PR #2, opened 2019-01-23
Add note about requiring "stack" user

derekhiggins:
configure_host puts a bunch of stuff in /home/stack
regardless of who ran the script for now lets just
use that user.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ from tripleo-quickstart here.
 
 - CentOS 7
 - ideally on a bare metal host
-- user with passwordless sudo access
+- run as user 'stack' with passwordless sudo access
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -11,7 +11,7 @@ from tripleo-quickstart here.
 
 - CentOS 7
 - ideally on a bare metal host
-- user with passwordless sudo access
+- run as user 'stack' with passwordless sudo access
```
hardys: This is fine, but you can run it as some other user if we add the provision/remote role to metalkube-setup-playbook.yml, or we set non_root_user to $USER before running ansible.

Whatever works at this point though :)

PR #7, opened 2019-01-29
01_install_requirements: add stack user

vrutkovs:
`02_configure_host.sh` has a playbook which expects `stack` user to be present, so it should be created in `01_install_requirements`.

Not sure if this is an effective way to check that the user exists though


hardys:
Yeah I think when you run quickstart there are roles which create the user, and we're missing those in the roles included here.

Long term it'd probably be better to remove all the stuff that expects a stack user, but adding this for now seems fine, thanks!

PR #9, opened 2019-01-29
Bump RHCOS to 47.284

vrutkovs:
This version contains newer kubelet (1.12), which has many issues
resolved so it should be the default


hardys:
Thanks, looks good but needs a rebase

PR #10, opened 2019-01-29
05_run_ocp: backup install config

vrutkovs:
`openshift-install create` would remove the config, so it should be
backed up as it might be useful

### DIFF
--- a/05_run_ocp.sh
+++ b/05_run_ocp.sh
@@ -45,7 +45,7 @@ sshKey: |
 EOF
 fi
 
-
+cp ocp/install-config{,.backup}.yaml 
 $GOPATH/src/github.com/openshift/installer/bin/openshift-install --dir ocp --log-level=debug create ignition-configs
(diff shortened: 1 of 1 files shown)


hardys:
Note that to get this script to re-run cleanly we also have to rm -fr ocp, otherwise the openshift-install tries to prompt for the missing data to re-create the install-config.yml, so I'm not sure storing any backup, if we create one, in that dir helps?

E.g my typical dev workflow is something like:

```
 ./ocp_cleanup.sh 
rm -fr ocp
./05_run_ocp.sh 
```

So if we're going to store the install-config, it should probably be somewhere else, logs perhaps?

But like I said I'm not sure what we gain in this case because it's inlined in the script.  I do find it annoying that it's removed when debugging interactive deploys though.

vrutkovs:
It doesn't seem to be used anywhere, so lets drop this for now

PR #11, opened 2019-01-29
Fix config handling

dhellmann:


.gitignore
```diff
@@ -1,4 +1,5 @@
 ocp/
 *.qcow2
 openshift-install.log
-logs/
\ No newline at end of file
+logs/
+config_*.sh
```
hardys: We shouldn't ignore the config_example.sh as that may change and is committed to the repo?

hardys: Hmm actually this should work OK provided that file is already added?  Sorry wasn't sure on first glance but provided that works then we can rebase and merge

dhellmann: Yeah, if the file is added explicitly the ignore rules are themselves ignored. Thanks for rebasing, too!

PR #13, opened 2019-01-30
Check for user stack in /etc/passwd

ukalifon:
Fix the bash syntax to test if the stack user needs to be created.
The old syntax failed with this error:

+ sudo '[' '!' cat /etc/passwd
+ grep stack ']'
grep: ]: No such file or directory
[: missing ']'

### DIFF
--- a/01_install_requirements.sh
+++ b/01_install_requirements.sh
@@ -55,6 +55,6 @@ if sudo [ ! -f /root/.ssh/id_rsa_virt_power ]; then
 fi
 
 # make sure stack user is created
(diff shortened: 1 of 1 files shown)


honza:
Yes, I hit this issue as well.

hardys:
Thanks, sorry but I fixed this is https://github.com/metalkube/dev-scripts/commit/579bc5fa80c6c941368ab628e57a49dc1e09f5eb - this PR was pushed around the same time so unfortunately I missed it

ukalifon:
This was fixed 10 minutes before my fix :)

PR #15, opened 2019-01-30
Add a newline after ssh key

honza:
Otherwise, the ssh key isn't added properly to the bootstrap VM, and you can't log in.

### DIFF
--- a/05_run_ocp.sh
+++ b/05_run_ocp.sh
@@ -42,6 +42,7 @@ pullSecret: |
   ${PULL_SECRET}
 sshKey: |
   ${SSH_PUB_KEY}
+
 EOF
 fi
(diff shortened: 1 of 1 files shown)


derekhiggins:
Is it possible you didn't have a ssh keypair when the bootstrap VM was created? I hit similar problems I think were caused by the key generation https://github.com/metalkube/dev-scripts/pull/18

honza:
I actually had to create the key manually because it was broken. So, yes, it's quite possible that #18 fixes it for me, too.

honza:
I pulled the new code, the new fix doesn't work for me, and this current PR doesn't work for me either. Thanks, Obama. :smile:

PR #20, opened 2019-01-31
add Makefile

elfosardo:
simplifying and condensing operations in a Makefile script


README.md
```diff
@@ -50,7 +56,6 @@ You can also see the status of the bootkube.sh script which is running via
 
 ## 2) Cleanup
 
-To clean up your environment you can run:
+To clean up your environment, run::
 
-- `./ocp_cleanup.sh`
-- `./libvirt_cleanup.sh`
+make clean
```
hardys: I'm fine with adding the make clean reference, but can we please leave the reference to the underlying scripts as well - most developers won't want to run make clean every time, and instead will iterate on something like:

```
./ocp_cleanup.sh
rm -fr ocp
./05_run_ocp.sh
```

So lets say something like 

- To clean up the ocp deployment run `./ocp_cleanup.sh`
- To clean up the dummy baremetal VMs and associated libvirt resources run `./libvirt_cleanup.sh`

Or, you can run `make clean` which will run all of the cleanup steps.

elfosardo: Hi! I'm still testing this and adding a dev-run option with the 3 steps will come next :)

README.md
```diff
@@ -70,8 +70,21 @@ And to deploy a node with the cirros image (image\_source above)
 You can also see the status of the bootkube.sh script which is running via
 `journalctl -b -f -u bootkube.service`.
 
-## 2) Cleanup
+## OCP cleanup & re-install
 
-To clean up your environment, run::
+To clean up the ocp deployment and re-install ocp, run::
+
+make dev-run
```
hardys: The idea of these scripts is developers can re-run whatever bits they need, IMHO there's not going to be a single target which suits all developers, so I'd rather we just referenced the individual cleanup scripts here

elfosardo: gotcha, I will remove this option, leave the clean command, but in the README I will reference the scripts as primary cleaning tools

hardys: Sounds good, when you're done can you also pls squash the commits?  Then we're good to merge I think, thanks!

PR #29, opened 2019-02-01
Setup a local DNS overlay so that BASE_DOMAIN queries would point to bootstrap

vrutkovs:
Configure NetworkManager to set nameserver to a local dnsmasq, this removes the need for `/etc/hosts` changes. This nameserver would be configured as default in bootstrap and other VMs.

dnsmasq would have `${CLUSTER_NAME}-api.${BASE_DOMAIN}` entry pointing to bootkube. Later on, once masters are stood up, more entries should be added.

### DIFF
--- a/05_run_ocp.sh
+++ b/05_run_ocp.sh
@@ -5,6 +5,11 @@ set -e
 source ocp_install_env.sh
 source common.sh
 
+# Switch NetworkManager to internal DNS
+sudo mkdir -p /etc/NetworkManager/conf.d/
+echo -e "[main]\ndns=dnsmasq" | sudo tee /etc/NetworkManager/conf.d/dnsmasq.conf
+systemctl restart NetworkManager
+
 # FIXME this is configuring for the libvirt backend which is dev-only ref
 # https://github.com/openshift/installer/blob/master/docs/dev/libvirt-howto.md
 # We may need some additional steps from that doc in 02* and also to make the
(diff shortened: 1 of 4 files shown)


libvirt_cleanup.sh
```diff
@@ -4,3 +4,4 @@ set -xe
 source common.sh
 
 ansible-playbook -b -vvv tripleo-quickstart-config/metalkube-teardown-playbook.yml -e @tripleo-quickstart-config/metalkube-nodes.yml -e local_working_dir=$HOME/.quickstart -e virthost=$HOSTNAME -e @config/environments/dev_privileged_libvirt.yml -i tripleo-quickstart-config/metalkube-inventory.ini
+sudo rm -rf /etc/NetworkManager/dnsmasq.d/openshift.conf /etc/NetworkManager/conf.d/dnsmasq.conf
```
hardys: We should probably remove openshift.conf in ocp_cleanup.sh, since that's owned by the ocp install script, vs the dnsmasq file which can persist (we could remove it here in libvirt cleanup tho I guess)

05_run_ocp.sh
```diff
@@ -5,6 +5,11 @@ set -e
 source ocp_install_env.sh
 source common.sh
 
+# Switch NetworkManager to internal DNS
+sudo mkdir -p /etc/NetworkManager/conf.d/
+echo -e "[main]\ndns=dnsmasq" | sudo tee /etc/NetworkManager/conf.d/dnsmasq.conf
+systemctl restart NetworkManager
```
hardys: This should include a test for the file as we don't need to do this every time we run the installer?

Relatedly this should probably go in 02_configure_host.sh where all the other host configuration happens?

hardys: Also we need sudo for the systemctl restart

hardys:
Ok looks like @vrutkovs is out today so I did a manual merge and pushed this via https://github.com/metalkube/dev-scripts/commit/12e7998ae0c0722a7732a80f232a4656711736d6

PR #30, opened 2019-02-01
Pass in ignition config to master nodes

derekhiggins:


ironic/Dockerfile
```diff
@@ -25,6 +25,7 @@ RUN crudini --set /etc/ironic/ironic.conf pxe ipxe_enabled true
 RUN crudini --set /etc/ironic/ironic.conf pxe pxe_config_template \$pybasedir/drivers/modules/ipxe_config.template
 
 RUN mkdir /tftpboot
+RUN echo '172.22.0.1 ostest-api.test.metalkube.org' >> /etc/hosts
```
vrutkovs: https://github.com/metalkube/dev-scripts/pull/29 would make this unnecessary

README.md
```diff
@@ -81,10 +81,18 @@ openstack baremetal node manage $NODE_UUID --wait
...
 ```
-openstack baremetal node deploy $NODE_UUID
+mkdir -p configdrive/openstack/latest
+cp ocp/master.ign configdrive/openstack/latest/user_data 
+openstack baremetal node deploy --config-drive configdrive $NODE_UUID
+```
+
+To ssh to the master nodes, you can route trafic through the bootstrap node
+```
+sudo ip route add 172.22.0.0/24 via $(getent hosts ostest-api.test.metalkube.org | awk '{ print $1 }')
```
hardys: We could also use sshuttle to tunnel to them I guess

README.md
```diff
@@ -81,10 +81,18 @@ openstack baremetal node manage $NODE_UUID --wait
 openstack baremetal node provide $NODE_UUID --wait
 ```
 
-And to deploy a node with the cirros image (image\_source above):
+And to deploy a master node with the coreos image (image\_source above) and passing in ignition config:
 
 ```
-openstack baremetal node deploy $NODE_UUID
+mkdir -p configdrive/openstack/latest
+cp ocp/master.ign configdrive/openstack/latest/user_data
```
vrutkovs:
```suggestion
curl -k https://ostest-api.test.metalkube.org:49500/config/master -o configdrive/openstack/latest/user_data 
```
This won't work without the network - `master.ign` is just a pointer to https://ostest-api.test.metalkube.org:49500/config/master. It should be fetched and placed there

ironic/Dockerfile
derekhiggins: Great because it the part of this patch that doesn't seem to work, when the containers is started /etc/hosts is rewritten ;-(

README.md
derekhiggins: Yup, any preference ?

README.md
hardys: But we expect to have the network, so the wrapper ignition file should be fine, right? Or is the curl a workaround for the issues you're seeing in your environment?

README.md
hardys: The route is fine for now, was just thinking aloud :)

README.md
vrutkovs: Neither helped in my case, so perhaps its fine to leave this as is

ironic/Dockerfile
derekhiggins: I've removed this line, PR #29 solves the problem

README.md
derekhiggins: ok

derekhiggins:
New version of patch tested on top of #29 (with a small fix to add a required sudo), working as expected.

PR #31, opened 2019-02-01
Image download updates

hardys:
Use an upstream location for the image download, and parameterize the docker build, which should make it easier to bump versions, and allows a fast local mirror for the 650MB qcow download :)


hardys:
Got an ack from @derekhiggins on slack so merging

PR #34, opened 2019-02-05
Remove reference to image checksum

derekhiggins:
We're using iscsi deploy, which doesn't check a
checksum (direct deploy does), lets not mislead people.


hardys:
Ah nice, I was thinking we'd have to calculate this

PR #37, opened 2019-02-05
Add ironic_client container to deploy the masters

hardys:
To automate the steps required to deploy the masters
we can run the same steps described in the README via
a container.

In future we might drive this sequence directly via some
code shared with the worker deployment, but for now
this helps us bootstrap a test environment.

### DIFF
--- a/05_run_ocp.sh
+++ b/05_run_ocp.sh
@@ -107,16 +107,27 @@ ssh -o "StrictHostKeyChecking=no" core@$IP sudo modprobe iscsi_tcp
 ssh -o "StrictHostKeyChecking=no" core@$IP sudo mkdir /run/httpd
 
 # Build and start the ironic container
-cat ironic/Dockerfile | ssh -o "StrictHostKeyChecking=no" core@$IP sudo dd of=Dockerfile
+"StrictHostKeyChecking=no" core@$IP sudo mkdir -p /home/core/ironic
+cat ironic/Dockerfile | ssh -o "StrictHostKeyChecking=no" core@$IP sudo dd of=ironic/Dockerfile
 ssh -o "StrictHostKeyChecking=no" core@$IP sudo podman build \
     --build-arg RHCOS_IMAGE_URL=${RHCOS_IMAGE_URL} \
     --build-arg RHCOS_IMAGE_VERSION=${RHCOS_IMAGE_VERSION} \
     --build-arg RHCOS_IMAGE_FILENAME_OPENSTACK=${RHCOS_IMAGE_FILENAME_OPENSTACK} \
-    -t ironic:latest .
+    -t ironic:latest ironic
 ssh -o "StrictHostKeyChecking=no" core@$IP sudo podman run \
     -d --net host --privileged --name ironic -v /run:/run:shared -v /dev:/dev localhost/ironic
 
 # Create a master_nodes.json file
 cat ~stack/ironic_nodes.json | jq '.nodes[0:3] |  {nodes: .}' | tee ocp/master_nodes.json
+
+# Build and start the ironic_client container
+ssh -o "StrictHostKeyChecking=no" core@$IP sudo mkdir -p /home/core/ironic_client
+ssh -o "StrictHostKeyChecking=no" core@$IP sudo mkdir -p /var/run/ironic_client_mnt
+cat ironic_client/Dockerfile | ssh -o "StrictHostKeyChecking=no" core@$IP sudo dd of=ironic_client/Dockerfile
+cat ocp/master.ign | ssh -o "StrictHostKeyChecking=no" core@$IP sudo dd of=/var/run/ironic_client_mnt/master.ign
+cat ocp/master_nodes.json | ssh -o "StrictHostKeyChecking=no" core@$IP sudo dd of=/var/run/ironic_client_mnt/master_nodes.json
+ssh -o "StrictHostKeyChecking=no" core@$IP sudo podman build -t ironic_client:latest ironic_client
+ssh -o "StrictHostKeyChecking=no" core@$IP sudo podman run \
+  -d --net host --privileged --name ironic_client -v /dev:/dev -v /var/run/ironic_client_mnt:/mnt localhost/ironic_client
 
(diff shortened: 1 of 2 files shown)


ironic_client/Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+RUN echo "openstack baremetal create /mnt/master_nodes.json" >> /bin/runironic_client
+RUN echo "mkdir -p configdrive/openstack/latest" >> /bin/runironic_client
+RUN echo "cp /mnt/master.ign configdrive/openstack/latest/user_data" >> /bin/runironic_client
+RUN echo "for node in \$(jq -r .nodes[].name /mnt/master_nodes.json); do" >> /bin/runironic_client
+RUN echo "  openstack baremetal node set \$node  --instance-info image_source=http://172.22.0.1/images/redhat-coreos-maipo-47.284-openstack.qcow2 --instance-info image_checksum=2a38fafe0b9465937955e4d054b8db3a --instance-info root_gb=25 --property root_device='{\"name\": \"/dev/vda\"}'" >> /bin/runironic_client
+RUN echo "  openstack baremetal node manage \$node --wait" >> /bin/runironic_client
+RUN echo "  openstack baremetal node provide \$node --wait" >> /bin/runironic_client
+RUN echo "  openstack baremetal node deploy --config-drive configdrive \$node" >> /bin/runironic_client
+RUN echo "done" >> /bin/runironic_client
+
```
hardys: TODO We'll probably want to wait for the masters to become active here (or fail)

ironic_client/Dockerfile
```diff
@@ -0,0 +1,25 @@
+FROM docker.io/centos:centos7
+
+RUN yum install -y python-requests
+RUN curl https://raw.githubusercontent.com/openstack/tripleo-repos/master/tripleo_repos/main.py | python - current-tripleo
+RUN yum install -y python-ironicclient python-openstackclient jq
+
+RUN echo "#!/usr/bin/bash" > /bin/runironic_client
```
vrutkovs: Why not just create the script here and `COPY` it during the build? This would create less layers

jamescassell: Or you could end each line with a backslash to avoid the multiple RUN statements.

hardys: Ack yeah good point, I just copied what we had already in ironic/Dockerfile, probably we can improve both

05_run_ocp.sh
```diff
@@ -107,16 +107,27 @@ ssh -o "StrictHostKeyChecking=no" core@$IP sudo modprobe iscsi_tcp
 ssh -o "StrictHostKeyChecking=no" core@$IP sudo mkdir /run/httpd
 
 # Build and start the ironic container
-cat ironic/Dockerfile | ssh -o "StrictHostKeyChecking=no" core@$IP sudo dd of=Dockerfile
+"StrictHostKeyChecking=no" core@$IP sudo mkdir -p /home/core/ironic
```
derekhiggins: Missing "ssh -o " here

ironic_client/Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+RUN echo "export OS_TOKEN=fake-token" >> /bin/runironic_client
+RUN echo "export OS_URL=http://localhost:6385/" >> /bin/runironic_client
+RUN echo "openstack baremetal create /mnt/master_nodes.json" >> /bin/runironic_client
+RUN echo "mkdir -p configdrive/openstack/latest" >> /bin/runironic_client
+RUN echo "cp /mnt/master.ign configdrive/openstack/latest/user_data" >> /bin/runironic_client
+RUN echo "for node in \$(jq -r .nodes[].name /mnt/master_nodes.json); do" >> /bin/runironic_client
+RUN echo "  openstack baremetal node set \$node  --instance-info image_source=http://172.22.0.1/images/redhat-coreos-maipo-47.284-openstack.qcow2 --instance-info image_checksum=2a38fafe0b9465937955e4d054b8db3a --instance-info root_gb=25 --property root_device='{\"name\": \"/dev/vda\"}'" >> /bin/runironic_client
+RUN echo "  openstack baremetal node manage \$node --wait" >> /bin/runironic_client
+RUN echo "  openstack baremetal node provide \$node --wait" >> /bin/runironic_client
+RUN echo "  openstack baremetal node deploy --config-drive configdrive \$node" >> /bin/runironic_client
```
derekhiggins: We'll no longer need to mention all of this in the README, and we wont need to install openstack clients their either. Actually wouldn't putting all of this in a container make it a little more difficult to retrace parts of the process while developing? So I guess my question is, why do we need to do this, is it to rely less on the virt host?

05_run_ocp.sh
hardys: whoops, thanks will fix - side-note, why do you use cat foo | ssh instead of scp here?

ironic_client/Dockerfile
hardys: Yeah we could run all this from the virt-host, I was trying to automate things, and yeah rely less on the virthost (prototype the pattern we expect for the final solution).

This could be premature though, maybe I should just run this script from 05* directoy for the time being?  No strong opinion but I'd like the master control-plane deployment to be automated one way or antother.

05_run_ocp.sh
derekhiggins: Not sure, it was either because I was originally going to put the file in ~root which is easier with sudo dd OR because I was too lazy to "man scp" to see if "-o StrictHostKeyChecking=no" worked with scp...., either way scp would be fine

ironic_client/Dockerfile
derekhiggins: I feel like it might help us iterate a bit faster if we kept this out of a container for the moment. So my preference would be for a script being called from 05 directly. But either would work so go with what ever your happier with.

hardys: Ok good point, I'll abandon this for now and move the script to the host, we can always package things in containers later when we're ready, thanks for the feedback

hardys:
Closing for now, will follow up with a version that runs on the host instead

PR #39, opened 2019-02-05
Using local user instead of user stack

### LINKED ISSUE #28: stack user required could cause issues if not checked
The scripts assume that the user 'stack' will be used to run the process but there's no check to verify and/or enforce that it's the actual user running the commands.
This could cause some issues during the process, e.g.
```
cat: /home/stack/ironic_nodes.json: Permission denied
```
due to the stack home dir being set by default at 700
A workaround exists for this issue, but others could arise.
In general, having the stack user hardcoded is not ideal.

elfosardo:
Using local user instead of user stack

This patch removes the need to create the user stack and just
uses a local user.
Instead of using the homedir, it creates a working-dir
/opt/dev-scripts to store temporary files.

Fixes #28

### DIFF
--- a/02_configure_host.sh
+++ b/02_configure_host.sh
@@ -9,7 +9,10 @@ source common.sh
 # Note we copy the playbook so the roles/modules from tripleo-quickstart
 # are found without a special ansible.cfg
 export ANSIBLE_LIBRARY=./library
+
 ANSIBLE_FORCE_COLOR=true ansible-playbook \
+    -e "non_root_user=$USER" \
+    -e "working_dir=$WORKING_DIR" \
     -e "roles_path=$PWD/roles" \
     -e @tripleo-quickstart-config/metalkube-nodes.yml \
     -e "local_working_dir=$HOME/.quickstart" \
@@ -18,7 +21,6 @@ ANSIBLE_FORCE_COLOR=true ansible-playbook \
     -i tripleo-quickstart-config/metalkube-inventory.ini \
     -b -vvv tripleo-quickstart-config/metalkube-setup-playbook.yml \
     | tee "ansible-$(date -u +%Y%m%d%H%M).log"
-
 
 # Allow local non-root-user access to libvirt ref
 # https://github.com/openshift/installer/blob/master/docs/dev/libvirt-howto.md#make-sure-you-have-permissions-for-qemusystem
--- a/common.sh
+++ b/common.sh
@@ -2,6 +2,7 @@
 
 SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 USER=`whoami`
+WORKING_DIR=${WORKING_DIR:-"/opt/dev-scripts"}
 
 if [ -z "$CONFIG" ]; then  
     # See if there's a config_$USER.sh in the SCRIPTDIR
@@ -21,6 +22,13 @@ if [ -z "$PULL_SECRET" ]; then
   echo "No valid PULL_SECRET set in ${CONFIG}"
   echo "Get a valid pull secret (json string) from https://cloud.openshift.com/clusters/install#pull-secret"
   exit 1
+fi
+
+if [ ! -d "$WORKING_DIR" ]; then
+    echo "Creating Working Dir"
+    sudo mkdir "$WORKING_DIR"
+    sudo chown "${USER}:${USER}" "$WORKING_DIR"
+    sudo chmod 755 "$WORKING_DIR"
 fi
 
 export RHCOS_IMAGE_URL=${RHCOS_IMAGE_URL:-"https://releases-rhcos.svc.ci.openshift.org/storage/releases/maipo/"}
--- a/config/environments/dev_privileged_libvirt.yml
+++ b/config/environments/dev_privileged_libvirt.yml
@@ -2,9 +2,7 @@
 # you can use these options.  This also makes VMs survive reboots \o/
 # This is particularly useful for local development environments
 non_root_chown: true
-non_root_user: stack
-undercloud_user: stack
-working_dir: "/home/{{ non_root_user }}"
+undercloud_user: "{{ non_root_user }}"
 ssh_user: root
 libvirt_uri: qemu:///system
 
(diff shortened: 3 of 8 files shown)


05_run_ocp.sh
```diff
@@ -117,6 +117,6 @@ ssh -o "StrictHostKeyChecking=no" core@$IP sudo podman run \
     -d --net host --privileged --name ironic -v /run:/run:shared -v /dev:/dev localhost/ironic
 
 # Create a master_nodes.json file
-cat ~stack/ironic_nodes.json | jq '.nodes[0:3] |  {nodes: .}' | tee ocp/master_nodes.json
+cat $WORKING_DIR/ironic_nodes.json | jq '.nodes[0:3] |  {nodes: .}' | tee ocp/master_nodes.json
```
derekhiggins: this will slightly break people with existing setups, but I think we can live with that at this stage

elfosardo: true, didn't consider that with all the buildup/teardown :/

common.sh
```diff
@@ -2,6 +2,7 @@
 
 SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 USER=`whoami`
+WORKING_DIR="/opt/dev-scripts"
```
hardys: We should make this overrideable e.g WORKING_DIR=${WORKING_DIR:-"/opt/dev-scripts"} - that means there's a workaround if anyone wants to temporarily use the /home/stack location by setting it in the per-user config files.

config/environments/dev_privileged_libvirt.yml
```diff
@@ -2,9 +2,9 @@
 # you can use these options.  This also makes VMs survive reboots \o/
 # This is particularly useful for local development environments
 non_root_chown: true
-non_root_user: stack
-undercloud_user: stack
-working_dir: "/home/{{ non_root_user }}"
+non_root_user: $USER
```
hardys: This won't work, will it?  Doesn't this need to be a j2 reference to an ansible variable?

E.g something like {{ansible_user_id}} ?

common.sh
elfosardo: yep, I will change that

config/environments/dev_privileged_libvirt.yml
elfosardo: I think this is a refuse from a wrong commit, I might have messed up something with my local repo.
I actually used {{ lookup('env','USER') }} since I think {{ansible_user_id}} will give root

hardys: You could just pass -e non_root_user=$USER here:

https://github.com/metalkube/dev-scripts/blob/master/02_configure_host.sh#L12

elfosardo: So remove all non_root_user definitions from the playbooks ?
We also have undercloud_user and ssh_user using the same variable, maybe reference the value of non_root_user instead ?

hardys: No I'm saying leave non_root_user alone and just pass the value you want to ansible-playbook :)

05_run_ocp.sh
celebdor:
```suggestion
jq '.nodes[0:3] |  {nodes: .}' "${WORKING_DIR}/ironic_nodes.json" | tee ocp/master_nodes.json
```

common.sh
```diff
@@ -23,6 +24,13 @@ if [ -z "$PULL_SECRET" ]; then
   exit 1
 fi
 
+if [ ! -d ${WORKING_DIR} ]; then
```
celebdor:
```suggestion
if [ ! -d "$WORKING_DIR" ]; then
```

common.sh
```diff
@@ -23,6 +24,13 @@ if [ -z "$PULL_SECRET" ]; then
...
+    echo "Creating Working Dir"
+    sudo mkdir ${WORKING_DIR}
```
celebdor:
```suggestion
    sudo mkdir "$WORKING_DIR"
```

common.sh
```diff
@@ -23,6 +24,13 @@ if [ -z "$PULL_SECRET" ]; then
...
+    sudo chown ${USER}:${USER} ${WORKING_DIR}
```
celebdor:
```suggestion
    sudo chown "${USER}:${USER}" "$WORKING_DIR"
```

common.sh
```diff
@@ -23,6 +24,13 @@ if [ -z "$PULL_SECRET" ]; then
...
+    sudo chmod 755 ${WORKING_DIR}
```
celebdor:
```suggestion
    sudo chmod 755 "$WORKING_DIR"
```

derekhiggins:
Patch tested on fresh centos and working with one edit
 - need to remove a reference to ~stack in 05_deploy_bootstrap_vm.sh (probably got added during a rebase)

02_configure_host.sh
```diff
@@ -9,7 +9,7 @@ source common.sh
 # Note we copy the playbook so the roles/modules from tripleo-quickstart
 # are found without a special ansible.cfg
 export ANSIBLE_LIBRARY=./library
-ansible-playbook -e roles_path=$PWD/roles -b -vvv tripleo-quickstart-config/metalkube-setup-playbook.yml -e @tripleo-quickstart-config/metalkube-nodes.yml -e local_working_dir=$HOME/.quickstart -e virthost=$HOSTNAME -e @config/environments/dev_privileged_libvirt.yml -i tripleo-quickstart-config/metalkube-inventory.ini
+ansible-playbook -e non_root_user=$USER -e roles_path=$PWD/roles -b -vvv tripleo-quickstart-config/metalkube-setup-playbook.yml -e @tripleo-quickstart-config/metalkube-nodes.yml -e local_working_dir=$HOME/.quickstart -e virthost=$HOSTNAME -e @config/environments/dev_privileged_libvirt.yml -i tripleo-quickstart-config/metalkube-inventory.ini
```
celebdor:
```suggestion
ansible-playbook -e "non_root_user=$USER" -e "roles_path=${PWD}/roles" -b -vvv tripleo-quickstart-config/metalkube-setup-playbook.yml -e @tripleo-quickstart-config/metalkube-nodes.yml -e "local_working_dir=${HOME}/.quickstart" -e "virthost=$HOSTNAME" -e @config/environments/dev_privileged_libvirt.yml -i tripleo-quickstart-config/metalkube-inventory.ini
```
Since you are touching the line, let's make it safer bash

elfosardo: yeah, I'm also going to remove the hardcoded working_dir in the ansible playbooks and udpate the command in libvirt_cleanup.sh

PR #40, opened 2019-02-05
Set necessary DNS in dnsmasq config

vrutkovs:
This ensures that:
* bootstrap's dnsmasq allocates defined IPs for masters
* "external" DNS resolves these IPs to `<cluster-name>-etcd-<index>` name
* sets etcd discovery DNS entry

### DIFF
--- a/05_run_ocp.sh
+++ b/05_run_ocp.sh
@@ -87,9 +87,11 @@ while ! sudo virsh domifaddr ${CLUSTER_NAME}-bootstrap | grep -q ${VM_MAC}; do
 done
 sudo virsh domifaddr ${CLUSTER_NAME}-bootstrap
 
+# NOTE: This hardcodes CLUSTER_NAME-api.BASE_DOMAIN to the bootstrap node.
+# TODO: Point instead to the DNS VIP when we have that. E.g.: server=/BASE_DOMAIN/DNS_VIP
 IP=$(sudo virsh domifaddr ostest-bootstrap | grep 122 | awk '{print $4}' | grep -o '^[^/]*')
-# TODO point to libvirt's DNS instead once correct hostname is set
-echo "address=/${CLUSTER_NAME}-api.${BASE_DOMAIN}/${IP}" | sudo tee /etc/NetworkManager/dnsmasq.d/openshift.conf
+echo "addn-hosts=/etc/hosts.openshift" | sudo tee /etc/NetworkManager/dnsmasq.d/openshift.conf
+echo "${IP} ${CLUSTER_NAME}-api.${BASE_DOMAIN}" | sudo tee /etc/hosts.openshift
 sudo systemctl restart NetworkManager
 
 # Wait for ssh to start
@@ -106,7 +108,23 @@ ssh -o "StrictHostKeyChecking=no" core@$IP sudo modprobe iscsi_tcp
 # doesn't exist on the host
 ssh -o "StrictHostKeyChecking=no" core@$IP sudo mkdir /run/httpd
 
+# Internal dnsmasq should reserve IP addresses for each master
+sudo cp -f ironic/dnsmasq.conf /tmp
+for i in 0 1 2; do 
+  NODE_MAC=$(cat ~stack/ironic_nodes.json | jq -r ".nodes[${i}].ports[0].address")
+  NODE_IP="172.22.0.2${i}"
+  HOSTNAME="${CLUSTER_NAME}-etcd-${i}.${BASE_DOMAIN}"
+  # Make sure internal dnsmasq would assign an expected IP
+  echo "dhcp-host=${NODE_MAC},${HOSTNAME},${NODE_IP}" >> /tmp/dnsmasq.conf
+  # Reconfigure "external" dnsmasq
+  echo "${NODE_IP} ${HOSTNAME} ${CLUSTER_NAME}-api.${BASE_DOMAIN}" | sudo tee -a /etc/hosts.openshift
+  echo "srv-host=_etcd-server-ssl._tcp.${CLUSTER_NAME}.${BASE_DOMAIN},${HOSTNAME},2380,0,0" | sudo tee -a /etc/NetworkManager/dnsmasq.d/openshift.conf
+done
+sudo systemctl reload NetworkManager
+cat /tmp/dnsmasq.conf | ssh -o "StrictHostKeyChecking=no" core@$IP sudo dd of=dnsmasq.conf
+
 # Build and start the ironic container
+cat ironic/runironic.sh | ssh -o "StrictHostKeyChecking=no" core@$IP sudo dd of=runironic.sh
 cat ironic/Dockerfile | ssh -o "StrictHostKeyChecking=no" core@$IP sudo dd of=Dockerfile
 ssh -o "StrictHostKeyChecking=no" core@$IP sudo podman build \
     --build-arg RHCOS_IMAGE_URL=${RHCOS_IMAGE_URL} \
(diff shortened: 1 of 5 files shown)


hardys:
@celebdor can you check this pls (for some reason I can't add you as a reviewer?)

05_run_ocp.sh
```diff
@@ -106,7 +106,23 @@ ssh -o "StrictHostKeyChecking=no" core@$IP sudo modprobe iscsi_tcp
...
+  NODE_MAC=$(cat ~stack/ironic_nodes.json | jq -r ".nodes[${i}].ports[0].address")
+  NODE_IP="172.22.0.2${i}"
+  HOSTNAME="${CLUSTER_NAME}-etcd-${i}.${BASE_DOMAIN}"
+  # Make sure internal dnsmasq would assign an expected IP
+  echo "dhcp-host=${NODE_MAC},${HOSTNAME},${NODE_IP}" >> ~stack/dnsmasq.conf
+  # Reconfigure "external" dnsmasq
+  echo "${NODE_IP} ${HOSTNAME} ${CLUSTER_NAME}-api.${BASE_DOMAIN}" | sudo tee -a /etc/hosts.openshift
+  echo "srv-host=_etcd-server-ssl._tcp.${CLUSTER_NAME}.${BASE_DOMAIN},${HOSTNAME},2380,0,0" | sudo tee -a /etc/NetworkManager/dnsmasq.d/openshift.conf
+done
+sudo systemctl restart NetworkManager
```
celebdor:
```suggestion
sudo systemctl reload NetworkManager
```

05_run_ocp.sh
```diff
@@ -88,8 +88,8 @@ done
 sudo virsh domifaddr ${CLUSTER_NAME}-bootstrap
 
 IP=$(sudo virsh domifaddr ostest-bootstrap | grep 122 | awk '{print $4}' | grep -o '^[^/]*')
-# TODO point to libvirt's DNS instead once correct hostname is set
-echo "address=/${CLUSTER_NAME}-api.${BASE_DOMAIN}/${IP}" | sudo tee /etc/NetworkManager/dnsmasq.d/openshift.conf
+echo "addn-hosts=/etc/hosts.openshift" | sudo tee /etc/NetworkManager/dnsmasq.d/openshift.conf
+echo "${IP} ${CLUSTER_NAME}-api.${BASE_DOMAIN}" | sudo tee /etc/hosts.openshift
```
celebdor: In https://github.com/metalkube/dev-scripts/pull/38 I'm going to move the api to be a VIP so that we can safely move it from the bootstrap to the cluster masters. Until then. I'm more than fine with going this way.

Maybe let's add a NOTE like:

```
# NOTE: This hardcodes CLUSTER_NAME-api.BASE_DOMAIN to the bootstrap node.
# TODO: Point instead to the DNS VIP when we have that. E.g.: server=/BASE_DOMAIN/DNS_VIP

File: README.md [hunk 3]
```diff
@@ -83,13 +83,13 @@ And to deploy a master node with the coreos image (image\_source above) and pass
 ```
 mkdir -p configdrive/openstack/latest
 cp ocp/master.ign configdrive/openstack/latest/user_data
-openstack baremetal node deploy --config-drive configdrive $NODE_UUID
+openstack baremetal node deploy --config-drive configdrive $NODE_UUID --wait
 ```
 
 To ssh to the master nodes, you can route trafic through the bootstrap node
 ```
 sudo ip route add 172.22.0.0/24 via $(getent hosts ostest-api.test.metalkube.org | awk '{ print $1 }')
```
derekhiggins: $ getent hosts ostest-api.test.metalkube.org 
Is now returning 4 entries, which breaks this command.

05_run_ocp.sh
```diff
@@ -106,7 +106,23 @@ ssh -o "StrictHostKeyChecking=no" core@$IP sudo modprobe iscsi_tcp
...
+cp -rf ironic/dnsmasq.conf ~stack/dnsmasq.conf
+for i in 0 1 2; do 
+  NODE_MAC=$(cat ~stack/ironic_nodes.json | jq -r ".nodes[${i}].ports[0].address")
+  NODE_IP="172.22.0.2${i}"
+  HOSTNAME="${CLUSTER_NAME}-etcd-${i}.${BASE_DOMAIN}"
+  # Make sure internal dnsmasq would assign an expected IP
+  echo "dhcp-host=${NODE_MAC},${HOSTNAME},${NODE_IP}" >> ~stack/dnsmasq.conf
+  # Reconfigure "external" dnsmasq
+  echo "${NODE_IP} ${HOSTNAME} ${CLUSTER_NAME}-api.${BASE_DOMAIN}" | sudo tee -a /etc/hosts.openshift
+  echo "srv-host=_etcd-server-ssl._tcp.${CLUSTER_NAME}.${BASE_DOMAIN},${HOSTNAME},2380,0,0" | sudo tee -a /etc/NetworkManager/dnsmasq.d/openshift.conf
```
celebdor: If I'm not mistaken, the etcd clustering resolution is only needed by the nodes themselves to cluster up. For this reason, I think it is better that you put DNS SRV records in the libvirt network definition like I do for A records in:
https://github.com/metalkube/dev-scripts/pull/38/files (network.xml.j2)

You can see an example of the SRV record in https://libvirt.org/formatnetwork.html#elementsAddress

It is for this reason that I am triyng to move the nodes from the 'default' libvirt node to a network I called 'baremetal'.

celebdor: In that libvirt section there's also the way how to make the host mac -> IP assignments you do here.

celebdor: The goal is for this baremetal network to eventually look and feel exactly like we expect the real baremetal network that will be used for the cluster networking.

05_run_ocp.sh
```diff
@@ -106,7 +106,23 @@ ssh -o "StrictHostKeyChecking=no" core@$IP sudo modprobe iscsi_tcp
...
+done
+sudo systemctl restart NetworkManager
+cat ~stack/dnsmasq.conf | ssh -o "StrictHostKeyChecking=no" core@$IP sudo dd of=dnsmasq.conf
```
celebdor: Does this mean that the dnsmasq that will be serving the deployment is from ironic? I would have expected ironic to only interact in the pxe booting and have the configuration be external. Is this something transitory?

vrutkovs: Yes, this is just for ironic containers to assign correct IP address. If libvirt xml can do that it should be the preferred way

README.md
vrutkovs: Fixed

05_run_ocp.sh
vrutkovs: Added the note.

DNSMasq hacks are fairly useful at this point, but eventually we should move to libvirt on this if possible

05_run_ocp.sh
```diff
@@ -106,7 +108,23 @@ ssh -o "StrictHostKeyChecking=no" core@$IP sudo modprobe iscsi_tcp
 # doesn't exist on the host
 ssh -o "StrictHostKeyChecking=no" core@$IP sudo mkdir /run/httpd
 
+# Internal dnsmasq should reserve IP addresses for each master
+cp -rf ironic/dnsmasq.conf ~stack/dnsmasq.conf
```
hardys: This doesn't work if you're not running this as the stack user - also doesn't need the recursive switch.

Also if we're going to copy to a working dir, can we make it a variable, as this will break when https://github.com/metalkube/dev-scripts/pull/39 gets fixed up I expect?

vrutkovs: Right, `/tmp` should be used here, so that original file won't be rewritten

vrutkovs: Fixed

PR #41, opened 2019-02-05
Switch to the "direct" deploy interface

derekhiggins:
For a single node this preforms a little better(about 2min45 vs
3min15 in a local env), but it should scale a lot better as it
offloads most of the work from the controller to the nodes being
deployed.


hardys:
Nice the performance is much better with this and the e1000 fix, thanks!

PR #43, opened 2019-02-06
Don't cat and grep when you can grep.

cmyster:


01_install_requirements.sh
```diff
@@ -55,7 +55,7 @@ if sudo [ ! -f /root/.ssh/id_rsa_virt_power ]; then
 fi
 
 # make sure stack user is created
-if ! cat /etc/passwd | grep -q "^stack:" ; then
+if ! grep -q "^stack:" /etc/passwd; then
```
celebdor:
```suggestion
if ! getent password stack > /dev/null; then
```

celebdor: adapted from the suggestion from @jamescassell

derekhiggins:
lgtm

PR #46, opened 2019-02-06
Remove sudo

derekhiggins:
Fix a last minute edit of the last patch. (Causes permission error further down the script)


derekhiggins:
merging

PR #47, opened 2019-02-06
Add 06_deploy_masters.sh

hardys:
This is a first-pass script which shows how to deploy the three
master nodes via Ironic.  Later this is likely to be automated
on the bootstrap VM (ref https://github.com/metalkube/dev-scripts/pull/37)
but for now having a script makes it easy to hack on.

Also rename 05_run_ocp.sh to 05_deploy_bootstrap_vm.sh
since that isn only doing the first step of the OCP deploy.


06_deploy_masters.sh
```diff
@@ -0,0 +1,22 @@
...
+export OS_URL=http://ostest-api.test.metalkube.org:6385/
+openstack baremetal create ocp/master_nodes.json
+mkdir -p configdrive/openstack/latest
+cp ocp/master.ign configdrive/openstack/latest/user_data
+for node in $(jq -r .nodes[].name ocp/master_nodes.json); do
+
+  # FIXME(shardy) we should parameterize the image and checksum (or calculate the latter)
+  openstack baremetal node set $node --instance-info image_source=http://172.22.0.1/images/redhat-coreos-maipo-47.284-openstack.qcow2 --instance-info image_checksum=2a38fafe0b9465937955e4d054b8db3a --instance-info root_gb=25 --property root_device='{"name": "/dev/vda"}'
+  openstack baremetal node manage $node --wait
+  openstack baremetal node provide $node --wait
```
dtantsur: If we ever enable cleaning, this will take time.

06_deploy_masters.sh
```diff
@@ -0,0 +1,22 @@
...
+  openstack baremetal node deploy --config-drive configdrive $node
+done
+
+# FIXME(shardy) we should wait for the node deploy to complete (or fail)
```
vrutkovs: Wait until we can ssh to masters?

PR #49, opened 2019-02-06
Pull RHCOS images earlier so that kubelet certs won't expire

vrutkovs:
Installer generates time-restricted certificates, so images should be downloaded earlier, so that installer assets won't expire

### DIFF
--- a/05_deploy_bootstrap_vm.sh
+++ b/05_deploy_bootstrap_vm.sh
@@ -4,6 +4,7 @@ set -e
 
 source ocp_install_env.sh
 source common.sh
+source get_rhcos_image.sh
 
 # FIXME this is configuring for the libvirt backend which is dev-only ref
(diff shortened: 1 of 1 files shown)


hardys:
Can you provide a bit more information on why this is needed?

It's not clear to me why pulling this at the start of the script is needed, and how that's related to the kublet cert expiry - this just downloads the qcow file right?

vrutkovs:
This downloads two images - qcow for bootstrap and gz for openstack images - before ignition files are generated. Downloading may take a while and installer can have time-sensitive assets - e.g certificates. AFAIK kubelet certs would now expire in 24h, but this might change in the future, so time-consuming actions should be done first

hardys:
Ok thanks for clarifying - perhaps it would be clearer if we just moved this into one of the earlier scripts e.g 02_configure_host or something, but for now this seems fine.

PR #50, opened 2019-02-06
Add hosts via Ironic during 'ocp_run'

vrutkovs:
This adds a new script to automatically provision nodes via Ironic and include in it `ocp_run` target


hardys:
I think this is basically the same as https://github.com/metalkube/dev-scripts/pull/47 ?

vrutkovs:
Right, closing it

PR #52, opened 2019-02-07
Fix permission denied for /tmp/dnsmasq.conf

bcrochet:
/tmp/dnsmasq.conf was copied earlier in the script with sudo, but then
attempted to write to without sudo.

### DIFF
--- a/05_deploy_bootstrap_vm.sh
+++ b/05_deploy_bootstrap_vm.sh
@@ -115,7 +115,7 @@ for i in 0 1 2; do
   NODE_IP="172.22.0.2${i}"
   HOSTNAME="${CLUSTER_NAME}-etcd-${i}.${BASE_DOMAIN}"
   # Make sure internal dnsmasq would assign an expected IP
-  echo "dhcp-host=${NODE_MAC},${HOSTNAME},${NODE_IP}" >> /tmp/dnsmasq.conf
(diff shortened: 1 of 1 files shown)


05_deploy_bootstrap_vm.sh
```diff
@@ -115,7 +115,7 @@ for i in 0 1 2; do
   NODE_IP="172.22.0.2${i}"
   HOSTNAME="${CLUSTER_NAME}-etcd-${i}.${BASE_DOMAIN}"
   # Make sure internal dnsmasq would assign an expected IP
-  echo "dhcp-host=${NODE_MAC},${HOSTNAME},${NODE_IP}" >> /tmp/dnsmasq.conf
+  echo "dhcp-host=${NODE_MAC},${HOSTNAME},${NODE_IP}" | sudo tee /tmp/dnsmasq.conf
```
celebdor:
```suggestion
  echo "dhcp-host=${NODE_MAC},${HOSTNAME},${NODE_IP}" | sudo tee -a /tmp/dnsmasq.conf
```

Without the `-a` you are not preserving the script semantics

hardys: Would it be better to just remove the sudo from the cp, since this file doesn't really need to be owned by root while in /tmp?

But otherwise yeah agree we need -a to append in this loop

derekhiggins:
thanks but a fix is already merged 
https://github.com/metalkube/dev-scripts/pull/46

PR #54, opened 2019-02-07
Clean previously deployed env

jtaleric:
Always rinse before you repeat.

### DIFF
--- a/06_deploy_masters.sh
+++ b/06_deploy_masters.sh
@@ -4,15 +4,26 @@ set -eux
 
 # Note This logic will likely run in a container (on the bootstrap VM)
 # for the final solution, but for now we'll prototype the workflow here
-
+instack="ocp/master_nodes.json"
 export OS_TOKEN=fake-token
 export OS_URL=http://ostest-api.test.metalkube.org:6385/
+
+# Clean previously env 
+nodes=$(openstack baremetal node list)
+for node in $(jq -r .nodes[].name ${instack}); do
+  if [[ $nodes =~ $node ]]; then
+    openstack baremetal node undeploy $node;
+    openstack baremetal node delete $node;
+  fi;
+done
+
 openstack baremetal create ocp/master_nodes.json
 mkdir -p configdrive/openstack/latest
 cp ocp/master.ign configdrive/openstack/latest/user_data
 for node in $(jq -r .nodes[].name ocp/master_nodes.json); do
 
   # FIXME(shardy) we should parameterize the image and checksum (or calculate the latter)
+  # Change to sda if using BM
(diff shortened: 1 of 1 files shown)


06_deploy_masters.sh
```diff
@@ -7,12 +7,18 @@ set -eux
 
 export OS_TOKEN=fake-token
 export OS_URL=http://ostest-api.test.metalkube.org:6385/
+
+# Clean previously env 
+for node in `openstack baremetal node list | grep master | awk '{print $2}'`; do  openstack baremetal node undeploy $node; done 
+for node in `openstack baremetal node list | grep master | awk '{print $2}'`; do  openstack baremetal node delete $node; done 
+
```
hardys: It'd be quicker to do this in a single loop with jq (see line 18 below) to avoid doing the baremetal node list twice?

jtaleric:
@hardys noted. Patch updated.

06_deploy_masters.sh
```diff
@@ -7,12 +7,20 @@ set -eux
 
 export OS_TOKEN=fake-token
 export OS_URL=http://ostest-api.test.metalkube.org:6385/
+
+# Clean previously env 
+for node in `openstack baremetal node list | grep master | awk '{print $2}'`; do
```
hardys: It might be better to use jq as previously mentioned, then we don't have any hard-coded dependencies on the name (it probably will be -master, but in theory folks could call those nodes whatever they want..

E.g use 

```
for node in $(jq -r .nodes[].name ocp/master_nodes.json); do
...
```

jtaleric: Oh, i see what you mean... sorry, I thought you just wanted to avoid calling baremetal node list twice.

06_deploy_masters.sh
```diff
@@ -4,15 +4,23 @@ set -eux
...
 # Note This logic will likely run in a container (on the bootstrap VM)
 # for the final solution, but for now we'll prototype the workflow here
-
+instack="ocp/master_nodes.json"
 export OS_TOKEN=fake-token
 export OS_URL=http://ostest-api.test.metalkube.org:6385/
+
+# Clean previously env 
+for node in $(jq -r .nodes[].name ${instack}); do
+  openstack baremetal node undeploy $node
```
derekhiggins: We need some kind of guard against running this when nodes haven't been deployed, (e.g. first run), otherwise the script just exits on error

+ for node in '$(jq -r .nodes[].name ${instack})'
+ openstack baremetal node undeploy openshift-master-0
Node openshift-master-0 could not be found. (HTTP 404)
[notstack@host11 dev-scripts]$

PR #55, opened 2019-02-07
Bare metal support

derekhiggins:
Allows us to boot the RHCOS image in a baremetal env (at least the one we've tested). The image boots but then fails to resolve ostest-api.test.metalkube.org. Which is better then we were.

Unfortunately the RDO IPA image has a bigger footprint, so needs a bump in RAM to fit the image download during deploy.

If merged this will require users to rerun the setup scripts from 02_...sh onwards


hardys:
This looks fine, but do we know why the RDO IPA ramdisk needs so much RAM?

Previously I've tested minimal VM tripleo deploys where the nodes only had the current 4096 so I'm a bit surprised that no longer works.

derekhiggins:
@hardys the API agent tries to download the RCHOS image to /tmp  which runs out of space. I only have the logs to go by but I can try and get onto a running API instance to see if I can get more details.

PR #56, opened 2019-02-08
Ignition patching system

celebdor:
This patch introduces a systematic way of patching ignition files and
provides a utility method for its usage.

All your patches to ignition files should follow jsonpatch rfc 6902.
You should place your patch json files in:

    json_patches/kind_of_machine

Then, in the code that starts the machine, you should run:

    apply_json_patches bootstrap "$IGN_FILE"

where bootstrap is the kind of machine from the possible:

- bootstrap
- masters
- workers

and IGN file is the ignition file that will be served to the machine via
PXE.

To run this, you should source utils.sh

### DIFF
--- a/01_install_requirements.sh
+++ b/01_install_requirements.sh
@@ -11,7 +11,7 @@ sudo yum -y update
 sudo yum -y install epel-release --enablerepo=extras
 sudo yum -y install curl vim-enhanced wget python-pip patch psmisc figlet golang dnsmasq NetworkManager crudini
 
-sudo pip install lolcat
+sudo pip install lolcat json-patch
 
 # for tripleo-repos install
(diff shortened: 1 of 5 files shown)


hardys:
Looks good, one nit - can we remove "served to the machine via PXE." from the commit message, as the ignition file is not passed via PXE, maybe just say "passed to the machine" as different methods are used for the bootstrap and master machines (qemu commandline vs config drive)

hardys:
Tested and this works great - one coment, would it be clearer if we named the directory ignition_patches or something?

celebdor:
> Tested and this works great - one comment, would it be clearer if we named the directory ignition_patches or something?

Very good point. I'll do that

PR #57, opened 2019-02-08
ocp_cleanup.sh - don't use openshift install to clean up

hardys:
We don't use "create cluster", so it doesn't really make sense to use the inverse openshift-install operation - it kinda works by luck because we name the bootstrap VM with the expected prefix, but we should probably just manually clean the things created via 05_deploy_bootstrap_vm.sh instead.

### DIFF
--- a/ocp_cleanup.sh
+++ b/ocp_cleanup.sh
@@ -4,8 +4,9 @@ set -e
 
 source ocp_install_env.sh
 
-$GOPATH/src/github.com/openshift/installer/bin/openshift-install --log-level=debug --dir ocp destroy cluster || true
-
+sudo virsh destroy "${CLUSTER_NAME}-bootstrap"
+sudo virsh undefine "${CLUSTER_NAME}-bootstrap" --remove-all-storage
+VOL_POOL=$(sudo virsh vol-pool "/var/lib/libvirt/images/${CLUSTER_NAME}-bootstrap.ign")
+sudo virsh vol-delete "${CLUSTER_NAME}-bootstrap.ign" --pool "${VOL_POOL}"
 rm -rf ocp
-
 sudo rm -rf /etc/NetworkManager/dnsmasq.d/openshift.conf


ocp_cleanup.sh
```diff
@@ -4,8 +4,10 @@ set -e
 
 source ocp_install_env.sh
 
-$GOPATH/src/github.com/openshift/installer/bin/openshift-install --log-level=debug --dir ocp destroy cluster || true
-
-rm -rf ocp
-
-sudo rm -rf /etc/NetworkManager/dnsmasq.d/openshift.conf
+if $GOPATH/src/github.com/openshift/installer/bin/openshift-install --log-level=debug --dir ocp destroy cluster; then
```
celebdor:
```suggestion
if "$GOPATH/src/github.com/openshift/installer/bin/openshift-install" --log-level=debug --dir ocp destroy cluster; then
```

ocp_cleanup.sh
```diff
@@ -4,8 +4,10 @@ set -e
...
+  rm -rf ocp
+  sudo rm -rf /etc/NetworkManager/dnsmasq.d/openshift.conf
+else
+  echo "Error running openshift-install destroy cluster, not cleaning ./ocp or /etc/NetworkManager/dnsmasq.d/openshift.conf"
```
celebdor:
```suggestion
 >&2 echo "Error running openshift-install destroy cluster, not cleaning ./ocp or /etc/NetworkManager/dnsmasq.d/openshift.conf"
```
Better print errors to stderr

ocp_cleanup.sh
hardys: I get quoting variables for string comparison in a test, but what's the logic behind quoting a command like this?

celebdor: GOPATH may contain spaces can't it?

Here's an example of the issue:

```
~ 🎩︎ echo $GOPATH
/home/celebdor/foo bar
~ 🎩︎ if $GOPATH/boo; then
echo "true"
else
echo "false"
fi
cd -- /home/celebdor/foo bar/boo
bash: cd: too many arguments
false
~ 🎩︎ if "$GOPATH/boo"; then
echo "true"
else
echo "false"
fi
true
```

vrutkovs:
`destroy cluster` won't ever succeed afaik, since we don't use `images` pool

ocp_cleanup.sh
hardys: Hmm okay - spaces in paths are an evil Windows thing IMHO so I don't get why anyone would ever do that on Linux - a bunch of stuff will break if the space isn't escaped I expect.

Anyway, thanks for clarifying!

ocp_cleanup.sh
```diff
@@ -4,8 +4,10 @@ set -e
 
 source ocp_install_env.sh
 
-$GOPATH/src/github.com/openshift/installer/bin/openshift-install --log-level=debug --dir ocp destroy cluster || true
-
-rm -rf ocp
-
-sudo rm -rf /etc/NetworkManager/dnsmasq.d/openshift.conf
+if "$GOPATH/src/github.com/openshift/installer/bin/openshift-install" --log-level=debug --dir ocp destroy cluster; then
```
derekhiggins: For this command to succeed for me I need to add a sudo, it can then delete the bootstrap node, with an error message about storage pools
FATAL Failed to destroy cluster: get storage pool "default": virError(Code=49, Domain=18, Message='Storage pool not found: no storage pool with matching name 'default'')

ocp_cleanup.sh
```diff
@@ -4,8 +4,8 @@ set -e
 
 source ocp_install_env.sh
 
-$GOPATH/src/github.com/openshift/installer/bin/openshift-install --log-level=debug --dir ocp destroy cluster || true
-
+sudo virsh destroy "${CLUSTER_NAME}-bootstrap"
+sudo virsh undefine "${CLUSTER_NAME}-bootstrap" --remove-all-storage
+sudo virsh vol-delete "${CLUSTER_NAME}-bootstrap.ign" --pool default
```
derekhiggins: I think s/default/images/ ?

hardys: Hmm, I think the bootstrap storage uses the default pool?

```
[shardy@tripleodev2 dev-scripts]$ sudo virsh pool-list
 Name                 State      Autostart 
-------------------------------------------
 default              active     yes       
 oooq_pool            active     yes       

[shardy@tripleodev2 dev-scripts]$ sudo virsh vol-list --pool default
 Name                 Path                                    
------------------------------------------------------------------------------
 ostest-bootstrap.ign /var/lib/libvirt/images/ostest-bootstrap.ign
 ostest-bootstrap.qcow2 /var/lib/libvirt/images/ostest-bootstrap.qcow2
```

Are you seeing something different in your environment?

hardys: I found that the --remove-all-storage removes the ostest-bootstrap.qcow2  but I had to manually clean the ign volume.

derekhiggins: Weird, my pool name is different

[root@host11 dev-scripts]# sudo virsh pool-list
 Name                 State      Autostart 
-------------------------------------------
 images               active     yes       
 oooq_pool            active     yes       

[root@host11 dev-scripts]# sudo virsh vol-list --pool default
error: failed to get pool 'default'
error: Storage pool not found: no storage pool with matching name 'default'

[root@host11 dev-scripts]# sudo virsh vol-list --pool images
 Name                 Path                                    
------------------------------------------------------------------------------
 ostest-bootstrap.ign /var/lib/libvirt/images/ostest-bootstrap.ign
 ostest-bootstrap.qcow2 /var/lib/libvirt/images/ostest-bootstrap.qcow2

I'll see if I can figure out why later

hardys: That's weird, I'll mark this WIP again until we figure out what's going on

ocp_cleanup.sh
```diff
@@ -4,8 +4,8 @@ set -e
...
 rm -rf ocp
```
vrutkovs:
```suggestion
rm -rf ocp
rm -rf configdrive```

hardys: Ah I forgot about the configdrive dir, maybe we should move that under the ocp dir but for now agree lets remove it, thanks!

hardys:
Ok updated to use virsh vol-pool - sorry @vrutkovs but I think I force-pushed over your changes?

PR #58, opened 2019-02-08
bootstrap vm deployment: support multiple nets

celebdor:
This patch gets us ready for when we add more nets to the deployment by
only retrieving the IP of one specific libvirt network. We had to drop
domifaddr because it misbehaved when there's multiple nets.

### DIFF
--- a/05_deploy_bootstrap_vm.sh
+++ b/05_deploy_bootstrap_vm.sh
@@ -85,16 +85,15 @@ sed -i "/<\/devices>/a <qemu:commandline>\n  <qemu:arg value='-fw_cfg'/>\n  <qem
 sudo virsh define ocp/bootstrap-vm.xml
 sudo virsh start ${CLUSTER_NAME}-bootstrap
 sleep 10
-VM_MAC=$(sudo virsh dumpxml ${CLUSTER_NAME}-bootstrap | grep "mac address" | head -n 1 | cut -d\' -f2)
-while ! sudo virsh domifaddr ${CLUSTER_NAME}-bootstrap | grep -q ${VM_MAC}; do
+
+while ! domain_net_ip ${CLUSTER_NAME}-bootstrap default; do
   echo "Waiting for ${CLUSTER_NAME}-bootstrap interface to become active.."
   sleep 10
 done
-sudo virsh domifaddr ${CLUSTER_NAME}-bootstrap
 
 # NOTE: This hardcodes CLUSTER_NAME-api.BASE_DOMAIN to the bootstrap node.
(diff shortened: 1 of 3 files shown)


utils.sh
```diff
@@ -21,3 +21,65 @@ for patch in "${PWD}/ignition_patches/${kind}/"*.json; do
 done
 sudo cp "$current" "$target"
 }
+
+function dom_mac_retrieve() {
```
hardys: Minor nit - would it be cleaner is we just made this a seperate script vs inlined?  We could create a uils directory, which could container a bash wrapper and any other tools (including python scripts etc)?

celebdor: I would like to keep it here for now and as soon as I land the PR for adding the second network I will move the scripts.

05_deploy_bootstrap_vm.sh
```diff
@@ -85,16 +85,15 @@ sed -i "/<\/devices>/a <qemu:commandline>\n  <qemu:arg value='-fw_cfg'/>\n  <qem
 sudo virsh define ocp/bootstrap-vm.xml
 sudo virsh start ${CLUSTER_NAME}-bootstrap
 sleep 10
-VM_MAC=$(sudo virsh dumpxml ${CLUSTER_NAME}-bootstrap | grep "mac address" | head -n 1 | cut -d\' -f2)
-while ! sudo virsh domifaddr ${CLUSTER_NAME}-bootstrap | grep -q ${VM_MAC}; do
+
+while ! domain_net_ip ${CLUSTER_NAME}-bootstrap default; do
```
hardys: This isn't working for me, it does `python /dev/fd/63 ostest-bootstrap default` which then fails with `No interface found for bridge default`

Should this really be looking for virbr0, or perhaps better we can retrieve it from the default virsh net XML?

```
[root@tripleodev2 ~]# virsh net-dumpxml default | grep "bridge name"
  <bridge name='virbr0' stp='on' delay='0'/>
```

utils.sh
```diff
@@ -21,3 +21,65 @@ for patch in "${PWD}/ignition_patches/${kind}/"*.json; do
...
+    fi
+}
+
+function net_iface_dhcp_ip() {
+local netname
+local hwaddr
+
+netname="$1"
+hwaddr="$2"
+sudo jq -r --arg HWADDR "$hwaddr" '.[] | select(."mac-address"==$HWADDR) | ."ip-address"' "/var/lib/libvirt/dnsmasq/${netname}.status"
```
hardys: Could we use `virsh net-dhcp-leases <network>` instead of parsing the status file directly?

hardys:
Latest revision is working well for me, thanks!

PR #60, opened 2019-02-11
WIP Run bootstrap service in a container

vrutkovs:
This removes the need for bootstrap VM and downloading an extra RHCOS image.
Instead of running RHCOS VM and provisioning it using bootstrap VM it does the following:

* Installs podman
* Determines MCD image name - pulls release image, uses `image machine-config-daemon` subcommand to get MCD pullspec
* Runs a CentOS systemd command in an unprivileged container with host network, mounting some directories from host and preparing container storage as a tmpfs
* Copies a bootstrap node preparation script (`bootstrap/bootstrap.sh`) and starts it as a systemd service (`bootstrap/prepare-bootstrap.service`)
* Bootstrap script would install `podman` and `systemd-journal-gateway` (required to setup RHCOS) inside CentOS container
* Boostrap script would pull MCD imagem run it and pass `boostrap.ign` file there.
* MCD would parse it and apply changes there, as if it was a RHCOS image and Ignition was passed on boot.
* After MCD is done, it would attempt to reboot the 'machine', fail and exit with error. `05_deploy_bootstrap_vm.sh` script would restart the container.
* Systemd inside bootstrap container would pick up new systemd services  - most notably bootkube - and run them.
* User machine should now be used as a bootstrap IP

This is, of course, very experimental and should not be merged yet.

TODO:
* [x] Template bootstrap.sh script to pass different MCD image
* [x] Avoid installing podman in a plain CentOS container - prepare a pre-installed version of it
* [ ] Teach MCD to avoid reboot attempt

### DIFF
--- a/01_install_requirements.sh
+++ b/01_install_requirements.sh
@@ -9,7 +9,7 @@ sudo sed -i "s/=enforcing/=permissive/g" /etc/selinux/config
 sudo yum -y update
 
 sudo yum -y install epel-release --enablerepo=extras
-sudo yum -y install curl vim-enhanced wget python-pip patch psmisc figlet golang dnsmasq NetworkManager crudini
+sudo yum -y install curl vim-enhanced wget python-pip patch psmisc figlet golang dnsmasq NetworkManager crudini podman
 
 sudo pip install lolcat json-patch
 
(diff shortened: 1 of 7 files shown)


vrutkovs:
This is potentially possible, but seems to be too complex:
* The image requires kubelet and CRI-O - so fresh versions of these have to be installed during install. These might not match RHCOS versions and cause a lot of weird issues
* Kubelet crashes currently when being run in container
* Container storage is being run in tmpfs, so it may run out of memory

Closing this for now, unless it would be necessary

PR #61, opened 2019-02-11
Reload NetworkManager instead of restarting

achuzhoy:
Restarting NetworkManager may cause network issues.
Better to reload instead.


02_configure_host.sh
```diff
@@ -40,4 +40,4 @@ sudo iptables -I INPUT -i virbr0 -p udp -m udp --dport 6230:6235 -j ACCEPT
 # Switch NetworkManager to internal DNS
 sudo mkdir -p /etc/NetworkManager/conf.d/
 sudo crudini --set /etc/NetworkManager/conf.d/dnsmasq.conf main dns dnsmasq
-sudo systemctl restart NetworkManager
+sudo systemctl reload NetworkManager
```
derekhiggins: I think we may need this to be restart in this script as it may not already be running, there is a restart in 05_deploy_bootstrap_vm.sh that probably could change to reload.

achuzhoy: ack, fixed

PR #62, opened 2019-02-11
Download ironic-inspector container and allow introspection to run

bfournie:
This gets the the ironic-inspector container and starts it. It also
changes the boot.ipxe file used for deployment to include
inspector.ipxe for introspection. If a deployment is not
being done, i.e. Ironic has not created a file in
/var/www/html/pxelinux.cfg/<MAC>, then the inspector.ipxe
file will be chainloaded to cause the IPA to run introspection
and return data to Ironic Inspector.

To access the Ironic Inspector API use:
export OS_TOKEN=fake-token
export OS_URL=http://ostest-api.test.metalkube.org:5050/

### DIFF
--- a/ironic/dualboot.ipxe
+++ b/ironic/dualboot.ipxe
@@ -1,0 +1,22 @@
+#!ipxe
+
+# NOTE(lucasagomes): Loop over all network devices and boot from
+# the first one capable of booting. For more information see:
+# https://bugs.launchpad.net/ironic/+bug/1504482
+set netid:int32 -1
+:loop
+inc netid 
+isset ${net${netid}/mac} || chain pxelinux.cfg/${mac:hexhyp} || goto inspector 
+echo Attempting to boot from MAC ${net${netid}/mac:hexhyp}
+chain pxelinux.cfg/${net${netid}/mac:hexhyp} || goto loop 
+
+# If no networks configured to boot then introspect first valid one
+:inspector
+chain inspector.ipxe || goto loop_done
+
+:loop_done
+echo PXE boot failed! No configuration found for any of the present NICs
+echo and could not find inspector.ipxe to use as fallback.
+echo Press any key to reboot...
+prompt --timeout 180
+reboot
--- a/ironic/inspector.ipxe
+++ b/ironic/inspector.ipxe
@@ -1,0 +1,9 @@
+#!ipxe
+
+
+:retry_boot
+echo In inspector.ipxe
+imgfree
+kernel --timeout 60000 http://172.22.0.1/images/ironic-python-agent.kernel ipa-inspection-callback-url=http://172.22.0.1:5050/v1/continue ipa-inspection-collectors=default,logs systemd.journald.forward_to_console=yes BOOTIF=${mac} ipa-debug=1 ipa-inspection-dhcp-all-interfaces=1 ipa-collect-lldp=1 initrd=ironic-python-agent.initramfs || goto retry_boot
+initrd --timeout 60000 http://172.22.0.1/images/ironic-python-agent.initramfs || goto retry_boot
+boot
--- a/01_install_requirements.sh
+++ b/01_install_requirements.sh
@@ -42,7 +42,7 @@ fi
 sudo yum -y update
 
 # make sure additional requirments are installed
-sudo yum install -y bind-utils ansible python-netaddr python-virtualbmc libvirt libvirt-devel libvirt-daemon-kvm qemu-kvm virt-install jq python-ironicclient python-openstackclient
+sudo yum install -y bind-utils ansible python-netaddr python-virtualbmc libvirt libvirt-devel libvirt-daemon-kvm qemu-kvm virt-install jq python-ironicclient python-ironic-inspector-client python-openstackclient
 
 if [ ! -f $HOME/.ssh/id_rsa.pub ]; then
     ssh-keygen -f ~/.ssh/id_rsa -P ""
--- a/05_deploy_bootstrap_vm.sh
+++ b/05_deploy_bootstrap_vm.sh
@@ -119,8 +119,10 @@ for i in 0 1 2; do
 done
 sudo systemctl reload NetworkManager
 cat /tmp/dnsmasq.conf | ssh -o "StrictHostKeyChecking=no" core@$IP sudo dd of=dnsmasq.conf
+cat ironic/dualboot.ipxe | ssh -o "StrictHostKeyChecking=no" core@$IP sudo dd of=dualboot.ipxe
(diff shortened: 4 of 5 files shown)


ironic/dualboot.ipxe
```diff
@@ -0,0 +1,28 @@
+#!ipxe
+
+# NOTE(lucasagomes): Loop over all network devices and boot from
+# the first one capable of booting. For more information see:
+# https://bugs.launchpad.net/ironic/+bug/1504482
+set netid:int32 -1
+:loop
+inc netid || chain pxelinux.cfg/${mac:hexhyp} || goto old_rom
```
dtantsur: Shouldn't this be `goto inspector`? Otherwise you'll never get to it. Check the bifrost's file: https://github.com/openstack/bifrost/blob/master/playbooks/roles/bifrost-ironic-install/files/boot.ipxe#L8

ironic/dualboot.ipxe
```diff
@@ -0,0 +1,28 @@
...
+isset ${net${netid}/mac} || goto inspector
```
dtantsur: I'm not sure what this does. I'm quite confident this `isset` will hold for most sane hardware.

ironic/dualboot.ipxe
```diff
@@ -0,0 +1,28 @@
...
+isset ${net${netid}/mac} || goto inspector 
+echo Attempting to boot from MAC ${net${netid}/mac:hexhyp}
+chain pxelinux.cfg/${net${netid}/mac:hexhyp} || goto loop 
+
+# If no networks configured to boot then introspect first valid one
+:inspector
+chain inspector.ipxe || goto loop_done
+
+:loop_done
+echo PXE boot failed! No configuration found for any of the present NICs.
```
dtantsur: Maybe also say something about inability to load inspector.ipxe?

ironic/dualboot.ipxe
```diff
@@ -0,0 +1,28 @@
...
+echo Press any key to reboot...
+prompt --timeout 180
+reboot
+
+:old_rom
```
dtantsur: This is redundant with loop_done.

ironic/inspector.ipxe
```diff
@@ -0,0 +1,9 @@
+#!ipxe
+
+
+:retry_boot
+echo In inspector.ipxe
+imgfree
+kernel --timeout 60000 http://172.22.0.1/images/ironic-python-agent.kernel ipa-inspection-callback-url=http://172.22.0.1:5050/v1/continue ipa-inspection-collectors=default,extra-hardware,numa-topology,logs systemd.journald.forward_to_console=yes BOOTIF=${mac} ipa-debug=1 ipa-inspection-dhcp-all-interfaces=1 ipa-collect-lldp=1 initrd=ironic-python-agent.initramfs || goto retry_boot
```
dtantsur: Let's remove `extra-hardware,numa-topology` at least for now. It can add a minute or two to introspection. Or do we actually need the extra hardware information from `python-hardware`?

ironic/dualboot.ipxe
dtantsur: Also `inc netid` cannot really fail, so the part after `||` is never executed?

ironic/inspector.ipxe
hardys: We can probably remove it initially - is there some docs reference where we can see exactly what data is collected by default, and what is added with extra-hardware?

I'm thinking in due course we'll want stuff like lldp data to validate networking, but initially probably make/model/cpu/ram/disks will be enough to match a basic profile of the expected platform.

ironic/dualboot.ipxe
derekhiggins: Ya, I guess we can probably just remove everything after the "||" and all of the old_rom section?

ironic/dualboot.ipxe
derekhiggins: To me this looks like execution keeps going to "loop:", incrementing  netid each time until this line fails (i.e. no more nics left to try), then it goes to inspector, without this line we're stuck in the loop?

dtantsur: Right, I think this should be
```
isset ${net${netid}/mac} || chain pxelinux.cfg/${mac:hexhyp} || goto inspector
```
(to keep fallback to `mac:hexhyp`, then fallback to inspector).

ironic/dualboot.ipxe
dtantsur: Yeah

derekhiggins:
> To access the Ironic Inspector API use:
> export OS_TOKEN=fake-token
> export OS_URL=http://ostest-api.test.metalkube.org:5050/
Do we also need to install "python2-ironic-inspector-client" on the virt host? that can go into 01_install_requirements.sh.

ironic/inspector.ipxe
dtantsur: Okay, just keep in mind that LLDP collection also makes inspection noticeably longer.

dtantsur:
Well, we can use inspection through ironic (and we probably should), but I guess it won't hurt installing it.

ironic/dualboot.ipxe
derekhiggins: That just happens on the next line, isn't it the same thing? 
> isset ${net${netid}/mac} || goto inspector 
> chain pxelinux.cfg/${net${netid}/mac:hexhyp} || goto loop 

if ${net${netid}/mac} is not set goto inspector
else chain....

dtantsur: No, because you need to also try `${mac:hexhyp}` in case of really old iPXE ROM (at least that's what ironic does).

ironic/dualboot.ipxe
bfournie: Yeah this is legacy from boot.ipxe in Ironic. Changed.

ironic/inspector.ipxe
bfournie: Removed extra-hardware,numa-topology

Makefile
```diff
@@ -15,7 +15,6 @@ build:
 
 ocp_run:
 	./05_deploy_bootstrap_vm.sh
-	./06_deploy_masters.sh
```
dtantsur: Is this intended?

bfournie: Nope, unintentional. Fixed.

ironic/dualboot.ipxe
derekhiggins: I don't see where ironic does that, all I see is
http://git.openstack.org/cgit/openstack/ironic/tree/ironic/drivers/modules/boot.ipxe#n9
> isset ${net${netid}/mac} || goto loop_done

bfournie: Yes ironic doesn't include the inspector support, just terminates if can't find the find in pxelinux.cfg

PR #64, opened 2019-02-13
Add wait_for_json to wait for ironic interaction

celebdor:


06_deploy_masters.sh
```diff
@@ -8,7 +8,12 @@ instack="ocp/master_nodes.json"
 export OS_TOKEN=fake-token
 export OS_URL=http://ostest-api.test.metalkube.org:6385/
 
-# Clean previously env 
+wait_for_json ironic \
```
hardys: This doesn't work because we don't source utils.sh in this script atm.

hardys:
We need to source the utils.sh, otherwise LGTM

06_deploy_masters.sh
celebdor: oops. Forgot that little step. I was trying it on the terminal by sourcing it
:-)

PR #65, opened 2019-02-13
Edit interface name in master's etcd discovery

vrutkovs:
etcd discovery container defaults to `eth0` interface to detect etcd IP and compare it with etcd SRV record.

This PR would pull MCS config, extract `etcd-member` k8s manifest, add `--if-name` argument to the container, pack it back and restart MCS to serve new ignition files to masters.

In the future this will be reverted and interface name would be a part of platform config, set in the installer. MCS would read this and automatically set `args` when assets are being generated

### DIFF
--- a/05_deploy_bootstrap_vm.sh
+++ b/05_deploy_bootstrap_vm.sh
@@ -143,4 +143,27 @@ ssh -o "StrictHostKeyChecking=no" core@$IP sudo podman run \
 # Create a master_nodes.json file
 jq '.nodes[0:3] | {nodes: .}' "${NODES_FILE}" | tee ocp/master_nodes.json
 
+MASTER_INTERFACE="eth1"
+
+# Fix etcd discovery on bootstrap
+rm -rf ocp/machineconfigs
+mkdir -p ocp/machineconfigs/temp
+# Find master machine config name
+while [ -z $(ssh -o StrictHostKeyChecking=no "core@$IP" sudo ls /etc/mcs/bootstrap/machine-configs/master*) ]; do sleep 5; done
+
+MASTER_CONFIG=$(ssh -o StrictHostKeyChecking=no "core@$IP" sudo ls /etc/mcs/bootstrap/machine-configs/master*)
+ssh -o "StrictHostKeyChecking=no" "core@$IP" sudo cat "${MASTER_CONFIG}" > ocp/machineconfigs/temp/master.yaml
+# Extract etcd-member.yaml part
+yq -r ".spec.config.storage.files[] | select(.path==\"/etc/kubernetes/manifests/etcd-member.yaml\") | .contents.source" ocp/machineconfigs/temp/master.yaml | sed 's;data:,;;' > ocp/machineconfigs/temp/etcd-member.urlencode
+# URL decode
+cat ocp/machineconfigs/temp/etcd-member.urlencode | urldecode > ocp/machineconfigs/etcd-member.yaml
+# Add a new param to args in discovery container
+sed -i "s;- \"run\";- \"run\"\\n    - \"--if-name=${MASTER_INTERFACE}\";g" ocp/machineconfigs/etcd-member.yaml
+# URL encode yaml
+cat ocp/machineconfigs/etcd-member.yaml | jq -sRr @uri > ocp/machineconfigs/temp/etcd-member.urlencode_updated
+# Replace etcd-member contents in the yaml
+sed "s;$(cat ocp/machineconfigs/temp/etcd-member.urlencode);$(cat ocp/machineconfigs/temp/etcd-member.urlencode_updated);g" ocp/machineconfigs/temp/master.yaml > ocp/machineconfigs/master.yaml
+# Copy the changed file back to bootstrap
+cat ocp/machineconfigs/master.yaml | ssh -o "StrictHostKeyChecking=no" "core@$IP" sudo dd of="${MASTER_CONFIG}"
+
 echo "You can now ssh to \"$IP\" as the core user"
--- a/utils.sh
+++ b/utils.sh
@@ -95,3 +95,7 @@ function network_ip() {
     fi
     echo "$ip"
 }
+
+function urldecode() {
+  echo -e "$(sed 's/+/ /g;s/%\(..\)/\\x\1/g;')"
+}
--- a/01_install_requirements.sh
+++ b/01_install_requirements.sh
@@ -11,7 +11,7 @@ sudo yum -y update
 sudo yum -y install epel-release --enablerepo=extras
 sudo yum -y install curl vim-enhanced wget python-pip patch psmisc figlet golang dnsmasq NetworkManager crudini
 
-sudo pip install lolcat json-patch
+sudo pip install lolcat json-patch yq
 
 # for tripleo-repos install
 sudo yum -y install python-setuptools python-requests


05_deploy_bootstrap_vm.sh
```diff
@@ -147,4 +147,26 @@ ssh -o "StrictHostKeyChecking=no" core@$IP sudo podman run \
 # Create a master_nodes.json file
 jq '.nodes[0:3] | {nodes: .}' "${WORKING_DIR}/ironic_nodes.json" | tee ocp/master_nodes.json
 
+MASTER_INTERFACE="eth1"
+
+# Fix etcd discovery on bootstrap
+mkdir -p ocp/machineconfigs/temp
+# Find master machine config name
+ssh -o "StrictHostKeyChecking=no" "core@$IP" sudo cat /etc/mcs/bootstrap/machine-configs/master* > ocp/machineconfigs/temp/master.yaml
```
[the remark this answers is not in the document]
vrutkovs: Right, its being stored in machine-config-operator assets. It should appear immediately after MCS is restarted on bootkube

derekhiggins:
Works for me
I0214 22:37:41.640184       1 run.go:57] ip addr is 192.168.111.44

But there is a lot there to follow, it might be ugly but how would you feel about editing the file with a sed, there would be a lot less going on to figure out for anybody looking at the code.

+ssh -o StrictHostKeyChecking=no core@$IP sudo sed -i.old \
+       -e "s/-%20%22--discovery-srv%3Dostest.test.metalkube.org%22/-%20%22--if-name=${MASTER_INTERFACE}%22%0A%20%20%20%20-%20%22--discovery-srv%3Dostest.test.metalkube.org%22/g" \
+       ${MASTER_CONFIG}

vrutkovs:
The code is ugly, the proper solution would be:
* A PR to installer, which adds a new `interface` setting for `platform: none`. It would store it in cluster config
* A PR to Machine Config Operator which would adjust container options after reading this setting from cluster config

hardys:
> The code is ugly, the proper solution would be:
> 
>     * A PR to installer, which adds a new `interface` setting for `platform: none`. It would store it in cluster config
> 
>     * A PR to Machine Config Operator which would adjust container options after reading this setting from cluster config

If we don't plan to do that now, can you please raise a Jira card to track these improvements please :)

vrutkovs:
@hardys done - https://jira.coreos.com/browse/KNIDEPLOY-107

05_deploy_bootstrap_vm.sh
```diff
@@ -143,4 +143,27 @@ ssh -o "StrictHostKeyChecking=no" core@$IP sudo podman run \
 # Create a master_nodes.json file
 jq '.nodes[0:3] | {nodes: .}' "${WORKING_DIR}/ironic_nodes.json" | tee ocp/master_nodes.json
 
+MASTER_INTERFACE="eth1"
+
+# Fix etcd discovery on bootstrap
+rm -rf ocp/machineconfigs
+mkdir -p ocp/machineconfigs/temp
+# Find master machine config name
+while [ -z $(ssh -o StrictHostKeyChecking=no "core@$IP" sudo ls /etc/mcs/bootstrap/machine-configs/master*) ]; do sleep 5; done
```
etingof: would it be a potentially infinite loop? would it make sense to terminate it eventually?

05_deploy_bootstrap_vm.sh
```diff
@@ -143,4 +143,27 @@ ssh -o "StrictHostKeyChecking=no" core@$IP sudo podman run \
...
+
+MASTER_CONFIG=$(ssh -o StrictHostKeyChecking=no "core@$IP" sudo ls /etc/mcs/bootstrap/machine-configs/master*)
```
etingof: observation: the `master*` glob could match more than one file, noting just in case.

05_deploy_bootstrap_vm.sh
```diff
@@ -143,4 +143,27 @@ ssh -o "StrictHostKeyChecking=no" core@$IP sudo podman run \
...
+ssh -o "StrictHostKeyChecking=no" "core@$IP" sudo cat "${MASTER_CONFIG}" > ocp/machineconfigs/temp/master.yaml
```
etingof: do we need two `ssh` calls (L154 & 155) to copy config(s)? could we do it  in one shot?

05_deploy_bootstrap_vm.sh
```diff
@@ -143,4 +143,27 @@ ssh -o "StrictHostKeyChecking=no" core@$IP sudo podman run \
...
+# Extract etcd-member.yaml part
+yq -r ".spec.config.storage.files[] | select(.path==\"/etc/kubernetes/manifests/etcd-member.yaml\") | .contents.source" ocp/machineconfigs/temp/master.yaml | sed 's;data:,;;' > ocp/machineconfigs/temp/etcd-member.urlencode
+# URL decode
+cat ocp/machineconfigs/temp/etcd-member.urlencode | urldecode > ocp/machineconfigs/etcd-member.yaml
+# Add a new param to args in discovery container
+sed -i "s;- \"run\";- \"run\"\\n    - \"--if-name=${MASTER_INTERFACE}\";g" ocp/machineconfigs/etcd-member.yaml
+# URL encode yaml
+cat ocp/machineconfigs/etcd-member.yaml | jq -sRr @uri > ocp/machineconfigs/temp/etcd-member.urlencode_updated
+# Replace etcd-member contents in the yaml
+sed "s;$(cat ocp/machineconfigs/temp/etcd-member.urlencode);$(cat ocp/machineconfigs/temp/etcd-member.urlencode_updated);g" ocp/machineconfigs/temp/master.yaml > ocp/machineconfigs/master.yaml
```
etingof: would it make sense to ensure that the desired substitution has taken place?

utils.sh
```diff
@@ -95,3 +95,7 @@ function network_ip() {
     fi
     echo "$ip"
 }
+
+function urldecode() {
+  echo -e "$(sed 's/+/ /g;s/%\(..\)/\\x\1/g;')"
```
etingof: nit: will this work for UTF8 URLs?

05_deploy_bootstrap_vm.sh
vrutkovs: Yes, it would finish as soon as MCS assets are generated. If there is some error with generating those masters won't be able to fetch ignition.

It usually takes ~20-30 secs to complete

05_deploy_bootstrap_vm.sh
vrutkovs: There are two predefined groups of assets installer generates - master and worker. These are hardcoded and won't change anytime soon, could be renamed to 'control_plane' though.

MCS adds a hash (possibly a checksum?) in the file name, so full file name can't be specified here.

05_deploy_bootstrap_vm.sh
vrutkovs: This is basically done for debugging simplicity - so fetching a master asset name is split from actual fetching

05_deploy_bootstrap_vm.sh
vrutkovs: `etcd-member.urlencode` is fetched by `yq` on L157, so it exists there. A better version would be using `yq` to replace the value, but the cmd line for that is terribly large

derekhiggins:
>> But there is a lot there to follow, it might be ugly but how would you feel about editing the file with a sed, there would be a lot less going on to figure out for anybody looking at the code.

> The code is ugly, the proper solution would be:

Just in case it wasn't clear, I was referring to my sed command that might be ugly. That said I still think it would be easier to follow then the 7 lines it would replace. Either way we need to merge something, So if your happy with the original patch would you mind adding a small detail in a comment about whats going on and I'll merge it.

vrutkovs:
>That said I still think it would be easier to follow then the 7 lines it would replace. 

A proper framework for patching resource would soon land in #79

>Either way we need to merge something, So if your happy with the original patch would you mind adding a small detail in a comment about whats going on and I'll merge it.

Updated commit message, PTAL

utils.sh
derekhiggins: Its a temporary fix that should be removed again soon, so going to merge as it works with what we currently have in the URL.

derekhiggins:
thanks.

PR #66, opened 2019-02-13
Set up metalkube/facet

honza:


03_ocp_repo_sync.sh
```diff
@@ -40,3 +41,12 @@ sync_go_repo_and_patch github.com/openshift/installer https://github.com/openshi
 
 # sync_go_repo_and_patch github.com/openshift/ci-operator https://github.com/openshift/ci-operator.git
 # sync_go_repo_and_patch github.com/sallyom/installer-e2e https://github.com/sallyom/installer-e2e.git
+
+sync_go_repo_and_patch github.com/metalkube/facet https://github.com/metalkube/facet.git
+
+# Build facet
+go get github.com/rakyll/statik
```
celebdor:
```suggestion
go get -v github.com/rakyll/statik
```
For better visibility into what is going on.

03_ocp_repo_sync.sh
```diff
@@ -40,3 +41,12 @@ sync_go_repo_and_patch github.com/openshift/installer https://github.com/openshi
...
+pushd $GOPATH/src/github.com/metalkube/facet
```
celebdor:
```suggestion
pushd "${GOPATH}/src/github.com/metalkube/facet"
```

README.md
```diff
@@ -31,6 +31,14 @@ This should result in some (stopped) VMs created by tripleo-quickstart on the
 local virthost and some other dependencies installed.
 
 - `./03_ocp_repo_sync.sh`
+
+After this step, you can run the [facet](https://github.com/metalkube/facet)
+server with:
+
+```
+$ go run $GOPATH/src/github.com/metalkube/facet/main.go server
```
celebdor:
```suggestion
$ go run "${GOPATH}/src/github.com/metalkube/facet/main.go" server
```

03_ocp_repo_sync.sh
```diff
@@ -5,6 +5,7 @@ source common.sh
 
 eval "$(go env)"
 echo "$GOPATH" | lolcat # should print $HOME/go or something like that
+export PATH="$PATH:$GOPATH/bin"
```
celebdor: Wouldn't it be better to just check that "${GOPATH}/bin" is in PATH and abort with a message if it is not, so users add it to the PATH properly in ~/.bashrc

honza: Given that we install golang in step 1, I assumed this was a clean machine with nothing set up. Why break the flow of running one step command after another just add something to your PATH?

README.md
hardys: This is fine for now, but I was thinking we'd probably have a wrapper script which can start the API, UI and any dependencies, that can be a follow-up though.

README.md
```diff
@@ -36,7 +36,7 @@ After this step, you can run the [facet](https://github.com/metalkube/facet)
 server with:
 
 ```
-$ go run $GOPATH/src/github.com/metalkube/facet/main.go server
+$ go run "${GOPATH}/src/github.com/metalkube/facet/main.go"" server
```
hardys: I think there's a spurious quote on the end of main.go here?

03_ocp_repo_sync.sh
hardys: I agree, we shouldn't require anything in the user .bashrc, this looks OK to me.

In the 05 script however we reference openshift-install via an explicit path though, which is quite nice as it's very clear where the command is running from (e.g the GOPATH vs some RPM or whatever), so we could do the same here.

hardys: This isn't working for me:

```
+ ./build.sh
I require the statik program but it's not installed.  Aborting.
You can install it with 'go get github.com/rakyll/statik'
```

However:

```
+ export PATH=/usr/local/bin:/usr/bin:/usr/local/sbin:/usr/sbin:/home/shardy/.local/bin:/home/shardy/bin:/home/shardy/go/bin
+ PATH=/usr/local/bin:/usr/bin:/usr/local/sbin:/usr/sbin:/home/shardy/.local/bin:/home/shardy/bin:/home/shardy/go/bin
+ which statik
/home/shardy/go/bin/statik
```

hardys: The issue is GOPATH isn't exported, so it's unset in the facet build.sh, so we need to export that, not the PATH

hardys: https://github.com/metalkube/facet/blob/master/build.sh#L3

README.md
honza: This does start everything. The UI is served at `/`, and the REST API is served at `/api`. The go server is both a static asset server, and a REST API server.

It can be also be started using the binary:

```
$ $GOPATH/src/github.com/metalkube/facet/bin/facet server
```

03_ocp_repo_sync.sh
honza: So, something like this?

```diff
-export PATH="$PATH:$GOPATH/bin"
+export GOPATH
```

honza: Or are you saying we should use `$GOPATH` in `build.sh`?

hardys: Yeah I'm saying we already do use GOPATH in build.sh, so we need to export it or, source the go env in build.sh - maybe we should have `eval "$(go env)"` in build.sh as it's needed when any other scripts call it?

README.md
hardys: Ack, Ok maybe later we have `05_run_metalkube_installer.sh` or something and we remove the current 05/06 scripts, the README seems fine for now :+1:

hardys:
Ok I think this is good now, but we need to land https://github.com/metalkube/facet/pull/24 and rebase/squash this branch before merging

PR #67, opened 2019-02-13
Allowing root_disk and instack.json to be set in advance.

achuzhoy:
Adding a valirable that if set is used as the root_device.
Making it possible to use a prepared instack.json. Previously
it was always trying to create one for libvirt based setup and
the MAC addresses in it didn't match the real ones on BM nodes.

### DIFF
--- a/06_deploy_masters.sh
+++ b/06_deploy_masters.sh
@@ -2,6 +2,7 @@
 
 set -eux
 source utils.sh
+source common.sh
 
 # Note This logic will likely run in a container (on the bootstrap VM)
 # for the final solution, but for now we'll prototype the workflow here
@@ -29,8 +30,7 @@ cp ocp/master.ign configdrive/openstack/latest/user_data
 for node in $(jq -r .nodes[].name $instack); do
 
   # FIXME(shardy) we should parameterize the image and checksum (or calculate the latter)
-  # Change to sda if using BM
-  openstack baremetal node set $node --instance-info image_source=http://172.22.0.1/images/redhat-coreos-maipo-47.284-openstack.qcow2 --instance-info image_checksum=2a38fafe0b9465937955e4d054b8db3a --instance-info root_gb=25 --property root_device='{"name": "/dev/vda"}'
+  openstack baremetal node set $node --instance-info image_source=http://172.22.0.1/images/redhat-coreos-maipo-47.284-openstack.qcow2 --instance-info image_checksum=2a38fafe0b9465937955e4d054b8db3a --instance-info root_gb=25  --property root_device="{\"name\": \"$ROOT_DISK\"}"
   openstack baremetal node manage $node --wait
   openstack baremetal node provide $node --wait
   openstack baremetal node deploy --config-drive configdrive $node
--- a/common.sh
+++ b/common.sh
@@ -11,8 +11,10 @@ EXT_IF=${EXT_IF:-}
 PRO_IF=${PRO_IF:-}
 # Internal interface, to bridge virbr0
 INT_IF=${INT_IF:-}
+#Root disk to deploy coreOS - use /dev/sda on BM
+ROOT_DISK=${ROOT_DISK:="/dev/vda"}
 
-if [ -z "$CONFIG" ]; then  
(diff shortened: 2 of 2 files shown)


06_deploy_masters.sh
```diff
@@ -29,8 +29,7 @@ cp ocp/master.ign configdrive/openstack/latest/user_data
 for node in $(jq -r .nodes[].name $instack); do
 
   # FIXME(shardy) we should parameterize the image and checksum (or calculate the latter)
-  # Change to sda if using BM
-  openstack baremetal node set $node --instance-info image_source=http://172.22.0.1/images/redhat-coreos-maipo-47.284-openstack.qcow2 --instance-info image_checksum=2a38fafe0b9465937955e4d054b8db3a --instance-info root_gb=25 --property root_device='{"name": "/dev/vda"}'
+  openstack baremetal node set $node --instance-info image_source=http://172.22.0.1/images/redhat-coreos-maipo-47.284-openstack.qcow2 --instance-info image_checksum=2a38fafe0b9465937955e4d054b8db3a --instance-info root_gb=25  --property root_device="{\"name\": $ROOT_DISK}"
```
derekhiggins: common.sh needs to be sourced at the top of this file

Also it looks like your missing double quotes around ROOT_DISK to make it valid json

tripleo-quickstart-config/roles/libvirt/setup/overcloud/tasks/main.yml
```diff
@@ -96,8 +96,9 @@
 # Generate the ironic node inventory files.  Note that this
 # task *must* occur after the above overcloud tasks, because if
 # `overcloud_nodes` is defined the template depends on the
-# `node_mac_map` variable.
+# `node_mac_map` variable. If the file exists - won't overwrite
 - name: Write ironic node json files
   template:
     src: ../templates/ironic_nodes.json.j2
     dest: "{{ working_dir }}/ironic_nodes.json"
+    force: no
```
hardys: Hmm, we need to be sure this is overwritten in the VM case (I think it's cleaned on teardown, but we've got be be sure it's not stale).

I think it would be better to leave this alone, and disable quickstart completely in the baremetal case?

Then we can just write $WORKING_DIR/ironic_nodes.json in the baremetal case (and add a test to ensure it exists)?

So add a conditional for baremetal to either skip the ansible run (or perhaps use tags if we still need the libvirt setup for the bootstrap VM?) here:

https://github.com/metalkube/dev-scripts/blob/master/02_configure_host.sh#L13


Then add a test that this file exists here, and it can be manually created:

https://github.com/metalkube/dev-scripts/blob/master/05_deploy_bootstrap_vm.sh#L111

I'm happy to make these changes in a seperate PR if you can drop the force: no from this one?

hardys: Also can you please drop the instack.json from the commit message (or if we leave something related to the nodes file here we need to say ironic_nodes.json as instack is a tripleo thing which we renamed here.

jtaleric: Ah, this reminds me @hardys ironic_nodes.json gets red to generate the DNS configuration -- mapping the mac address to the hostname and IP. 

We should have `common.sh` have 
```
NODES=${NODES:-}
```
If someone is doing a BM deployment, they add the node.json here, if it is blank we run through quickstart -- use this as the conditional to determine if we run through creating the VMBC?

hardys: Yeah it can either be a variable which accepts a json string, or a filename, doesn't really matter I guess, and yes then we can use it as a conditional to either skip quickstart completely or just run the bits we need to enable booting the bootstrap VM.

derekhiggins:
When I try this I get
++ ROOT_DISK=/dev/vda
common.sh: line 17: CONFIG: unbound variable

As 06_deploy_masters.sh has "set -u", we need update common.sh
-if [ -z "$CONFIG" ]; then  
+if [ -z "${CONFIG:-}" ]; then

PR #68, opened 2019-02-14
06_deploy_masters.sh wait for undeploy and tolerate failure

hardys:
We may fail in certain node states, and also we should wait for the
undeploy or the delete will fail.

### DIFF
--- a/06_deploy_masters.sh
+++ b/06_deploy_masters.sh
@@ -18,9 +18,9 @@ wait_for_json ironic \
 nodes=$(openstack baremetal node list)
 for node in $(jq -r .nodes[].name ${instack}); do
   if [[ $nodes =~ $node ]]; then
-    openstack baremetal node undeploy $node;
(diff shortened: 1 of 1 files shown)


06_deploy_masters.sh
```diff
@@ -18,9 +18,9 @@ wait_for_json ironic \
 nodes=$(openstack baremetal node list)
 for node in $(jq -r .nodes[].name ${instack}); do
   if [[ $nodes =~ $node ]]; then
-    openstack baremetal node undeploy $node;
-    openstack baremetal node delete $node;
-  fi;
+    openstack baremetal node undeploy --wait $node || true
```
derekhiggins: you need to add the "--wait" as the last param, otherwise it thinks the following param is the amount of seconds to wait for...

hardys: Oops thanks, now fixed

PR #74, opened 2019-02-15
WIP Masters should be part of DNS round-robin

vrutkovs:
`-api` should point to bootstrap node and masters, this address would be used for control plane. Initially it would start on bootstrap and then prepare 'real' masters on master/etcd nodes

TODO:
* [ ] Modify `baremetal.addnhosts` instead

### DIFF
--- a/tripleo-quickstart-config/roles/common/defaults/main.yml
+++ b/tripleo-quickstart-config/roles/common/defaults/main.yml
@@ -178,9 +178,6 @@ networks:
           port: 2380
           target: "{{ cluster_name }}-etcd-2.{{ base_domain }}"
       hosts:
-        - ip: "{{ baremetal_network_cidr | nthhost(5) }}"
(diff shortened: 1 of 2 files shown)


tripleo-quickstart-config/roles/common/defaults/main.yml
```diff
@@ -178,9 +178,6 @@ networks:
           port: 2380
           target: "{{ cluster_name }}-etcd-2.{{ base_domain }}"
       hosts:
-        - ip: "{{ baremetal_network_cidr | nthhost(5) }}"
-          hostnames:
-            - "{{ cluster_name }}-api"
```
celebdor: This makes the API not be a VIP anymore.

vrutkovs: Right, it breaks master setup, needs to be reworked

PR #76, opened 2019-02-15
Allow disabling master/worker VMs for baremetal testing

hardys:
When testing with real baremetal master/worker nodes we only
need the bootstrap VM, so we can disable some of the quickstart
roles by setting "NODES_PLATFORM" to something other than "libvirt"
which is the default.

NODES_FILE can also be overridden to point to your manually created
alternative to $WORKING_DIR/ironic_nodes.json (or you can just
copy it to that location).

This was created following conversation in https://github.com/metalkube/dev-scripts/pull/67

### DIFF
--- a/common.sh
+++ b/common.sh
@@ -2,7 +2,7 @@
 
 SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 USER=`whoami`
-WORKING_DIR=${WORKING_DIR:-"/opt/dev-scripts"}
+
 # Additional DNS
 ADDN_DNS=${ADDN_DNS:-}
 # External interface for routing traffic through the host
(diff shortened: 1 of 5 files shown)


common.sh
```diff
@@ -3,6 +3,8 @@
 SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 USER=`whoami`
 WORKING_DIR=${WORKING_DIR:-"/opt/dev-scripts"}
+NODES_FILE=${NODES_FILE:-"${WORKING_DIR}/ironic_nodes.json"}
+NODES_PLATFORM=${NODES_PLATFORM:-"libvirt"}
```
hardys: This isn't working as I expected, the WORKING_DIR is /opt/dev-scripts for NODES_FILE even if I override WORKING_DIR, I'll have to add a seperate conditional

derekhiggins: Maybe I don't get what you meant but It seems to work for me..

derekhiggins: Nevermind, I see what happened in your env now.

derekhiggins:
looks correct and doesn't appear to have broken anything ;-)

PR #77, opened 2019-02-15
Move master deploy to a different loop

vrutkovs:
This ensures all masters are started in a rapid succession, which helps 
etcd cluster to assemble


hardys:
Works for me, thanks!

PR #78, opened 2019-02-15
Rename instack to MASTER_NODES_FILE

hardys:
This should avoid confusion since instack is a legacy TripleO thing,
also make MASTER_NODES_FILE overridable so folks can do that
for baremetal testing as an alternative to NODES_FILE if they wish.

### DIFF
--- a/06_deploy_masters.sh
+++ b/06_deploy_masters.sh
@@ -6,7 +6,6 @@ source common.sh
 
 # Note This logic will likely run in a container (on the bootstrap VM)
 # for the final solution, but for now we'll prototype the workflow here
-instack="ocp/master_nodes.json"
 export OS_TOKEN=fake-token
 export OS_URL=http://ostest-api.test.metalkube.org:6385/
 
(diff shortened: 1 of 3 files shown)


celebdor:
Looks good to me. I'll need to try it.

06_deploy_masters.sh
```diff
@@ -6,7 +6,7 @@ source common.sh
 
 # Note This logic will likely run in a container (on the bootstrap VM)
 # for the final solution, but for now we'll prototype the workflow here
-instack="ocp/master_nodes.json"
+MASTER_NODES_FILE=${MASTER_NODES_FILE:-"ocp/master_nodes.json"}
```
derekhiggins: ocp/master_nodes.jsonis also mentioned in 05_deploy_bootstrap_vm.sh , should we move this line into common.sh and use it in the other file also?

hardys: Ah yeah good point, I'll update it

PR #79, opened 2019-02-18
Patch master node with passwd for virsh console

celebdor:
This patchset introduces a facility for patching master nodes just like
there is for bootstrap.

### DIFF
--- a/utils.sh
+++ b/utils.sh
@@ -14,12 +14,63 @@ wd=$(mktemp -d)
 target_name="$(basename "$target")"
 
 current="$target"
-for patch in "${PWD}/ignition_patches/${kind}/"*.json; do
+for patch in $(ls "${PWD}/ignition_patches/${kind}/"*.json); do
     current_patch_name="$(basename "$patch")"
     jsonpatch "$current" "$patch" > "${wd}/${target_name}_${current_patch_name}"
     current="${wd}/${target_name}_${current_patch_name}"
 done
-sudo cp "$current" "$target"
+
+if [[ "$current" != "$target" ]]; then
+    sudo cp "$current" "$target"
+fi
+}
+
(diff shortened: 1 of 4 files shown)


utils.sh
```diff
@@ -22,6 +22,54 @@ done
...
+function patch_node_ignition() {
+    local kind
+    local bootstrap
+    local wd
+    local config
+
+    kind="$1"
+    bootstrap="$2"
+    wd="$(mktemp -d)"
+    ssh_opts=("-o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null")
```
etingof: nit: do we need parenthesis here?

utils.sh
```diff
@@ -22,6 +22,54 @@ done
...
+
+    # Wait for the release config to be pulled
+    while
+        config=$(ssh "${ssh_opts[@]}" "core@$bootstrap" \
```
etingof: here and in other places where we `ssh` to the coming up host - would it make sense to tackle ssh `ConnectionAttempts` and/or `ConnectTimeout` options  in hope for more predicable timing?

utils.sh
celebdor: I'm making it into an array. That's why I put the parenthesis there. This way I can properly quote the var when I use it.

utils.sh
```diff
@@ -14,12 +14,63 @@ wd=$(mktemp -d)
...
+
+function patch_node_ignition() {
+    local kind
+    local bootstrap
+    local wd
+    local config
+
+    kind="$1"
+    bootstrap="$2"
+    wd="$(mktemp -d)"
```
derekhiggins: we appear to be leaking temp directories, but we can fix this up in another patch as there is another one to be fixed that isn't part of this patch.

PR #80, opened 2019-02-18
Also remove ocp/install-config.yaml during cleanup

dantrainor:
In situations where PULL_SECRET is either bad or broken, the
contents of this will still be written to install-config.yaml.
It's still assumed to be valid.  Since we source config_*.sh
anyway before running any of the top-level cleanups, we can
expect to just write this value to ocp/install-config anyway.

### DIFF
--- a/ocp_cleanup.sh
+++ b/ocp_cleanup.sh
@@ -10,3 +10,4 @@ VOL_POOL=$(sudo virsh vol-pool "/var/lib/libvirt/images/${CLUSTER_NAME}-bootstra
 sudo virsh vol-delete "${CLUSTER_NAME}-bootstrap.ign" --pool "${VOL_POOL}"
 rm -rf ocp
 sudo rm -rf /etc/NetworkManager/dnsmasq.d/openshift.conf
+rm -f ocp/install-config.yaml


ocp_cleanup.sh
```diff
@@ -10,3 +10,4 @@ VOL_POOL=$(sudo virsh vol-pool "/var/lib/libvirt/images/${CLUSTER_NAME}-bootstra
 sudo virsh vol-delete "${CLUSTER_NAME}-bootstrap.ign" --pool "${VOL_POOL}"
 rm -rf ocp
 sudo rm -rf /etc/NetworkManager/dnsmasq.d/openshift.conf
+rm -f ocp/install-config.yaml
```
derekhiggins: but we already remove the ocp directory 2 lines up, is that not working for some reason?

dantrainor: You're right, I don't know why I didn't pay closer attention to the line you mentioned.  I was clearly looking at one file and fixing another.  Thanks for the catch.

hardys: Ok so I think this PR can be closed, right?

hardys:
Ok I'l close this, feel free to reopen if needed.

PR #81, opened 2019-02-18
switch cluster domain to ostest.test.metalkube.org

celebdor:
Co-authored-by: Antoni Segura Puimedon <antoni@redhat.com>


hardys:
This is working well for me too so lets go ahead and merge

PR #82, opened 2019-02-19
Stop directing ansible logs to file

derekhiggins:
These are clogging up the top level directory and we
log all the scripts already (see common.sh), so we
don't need to do it a second time.


hardys:
Ah nice, yeah I noticed this last week as well

PR #84, opened 2019-02-19
Master avahi

celebdor:


celebdor:
It includes a previous PR. Once that one is merged, I'll rebase this.

utils.sh
```diff
@@ -56,6 +63,58 @@ function patch_node_ignition() {
...
+    local config_type
+    local files_template
+    local units_template
+    local octal_mode
+    local decimal_mode
+    local dest_path
+    local ifs
+    local simple_name
+
+    read -r -d '' files_template <<'EOF'
```
vrutkovs: bashism, this won't work when `sh ./ocp_run` is called

vrutkovs:
```suggestion
    files_template=$(cat <<EOF
```

utils.sh
```diff
@@ -56,6 +63,58 @@ function patch_node_ignition() {
...
+{"op": "add", "path": "${path}/-", "value": {"filesystem": "root", "path": "${dest_path}", "user": {"name": "root"}, "contents": {"source": "data:,${value}", "verification": {}}, "mode": ${decimal_mode}}}
+EOF
```
vrutkovs:
```suggestion
EOF
)
```

utils.sh
```diff
@@ -56,6 +63,58 @@ function patch_node_ignition() {
...
+    read -r -d '' units_template <<'EOF'
```
vrutkovs:
```suggestion
    units_template=$(cat <<EOF
```

utils.sh
```diff
@@ -56,6 +63,58 @@ function patch_node_ignition() {
...
+{"op": "add", "path": "/systemd/units/-", "value": {"contents": "${value}", "enabled": true, "name": "${simple_name}"}}
+EOF
```
vrutkovs:
```suggestion
EOF
)
```

utils.sh
celebdor: the file is headed with #!/bin/bash are we really not able to depend on bash?

celebdor: If that is the case, we lose `local` echo flags, etc

celebdor: without the single quoting of EOF, it will try to substitute the variables, which is not the intended behavior at this point (it is done later with envsubst)

utils.sh
```diff
@@ -56,6 +63,58 @@ function patch_node_ignition() {
...
+    local decimal_mode
+    local dest_path
+    local ifs
+    local simple_name
+
+    read -r -d '' files_template <<'EOF'
+{"op": "add", "path": "${path}/-", "value": {"filesystem": "root", "path": "${dest_path}", "user": {"name": "root"}, "contents": {"source": "data:,${value}", "verification": {}}, "mode": ${decimal_mode}}}
+EOF
+    read -r -d '' units_template <<'EOF'
+{"op": "add", "path": "/systemd/units/-", "value": {"contents": "${value}", "enabled": true, "name": "${simple_name}"}}
```
vrutkovs:
```suggestion
{"op": "add", "path": "/systemd/units/-", "value": {"contents": "\${value}", "enabled": true, "name": "\${simple_name}"}}
```

utils.sh
```diff
@@ -56,6 +63,58 @@ function patch_node_ignition() {
...
+    local files_template
+    local units_template
+    local octal_mode
+    local decimal_mode
+    local dest_path
+    local ifs
+    local simple_name
+
+    read -r -d '' files_template <<'EOF'
+{"op": "add", "path": "${path}/-", "value": {"filesystem": "root", "path": "${dest_path}", "user": {"name": "root"}, "contents": {"source": "data:,${value}", "verification": {}}, "mode": ${decimal_mode}}}
```
vrutkovs:
```suggestion
{"op": "add", "path": "\${path}/-", "value": {"filesystem": "root", "path": "\${dest_path}", "user": {"name": "root"}, "contents": {"source": "data:,\${value}", "verification": {}}, "mode": \${decimal_mode}}}
```
This should protect the values from being evaluated here

utils.sh
vrutkovs: hmm, weird, it seems its running as a bash but it still fails here

vrutkovs:
Patching works, but `avahi.sh` should be executable:
```
[core@master-0 ~]$ sudo systemctl status avahi
● avahi.service - mDNS announcing with Avahi
   Loaded: loaded (/etc/systemd/system/avahi.service; enabled; vendor preset: enabled)
   Active: activating (auto-restart) (Result: exit-code) since Wed 2019-02-20 16:43:33 UTC; 179ms ago
  Process: 6080 ExecStart=/usr/local/bin/avahi.sh (code=exited, status=203/EXEC)
 Main PID: 6080 (code=exited, status=203/EXEC)

Feb 20 16:43:33 master-0 systemd[1]: Unit avahi.service entered failed state.
Feb 20 16:43:33 master-0 systemd[1]: avahi.service failed.
[core@master-0 ~]$ /usr/local/bin/avahi.sh
-bash: /usr/local/bin/avahi.sh: Permission denied

```

machineconfig/master/spec/config/storage/files/avahi.sh L3Vzci9sb2NhbC9iaW4vYXZhaGkuc2gK
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
+set -e
+
+mkdir --parents /etc/avahi
+SHORTNAME="$(hostname -s)" envsubst < /etc/avahi/avahi-daemon.conf.template > /etc/avahi/avahi-daemon.conf
```
vrutkovs:
```suggestion
SHORT_HOSTNAME="$(hostname -s)" envsubst < /etc/avahi/avahi-daemon.conf.template > /etc/avahi/avahi-daemon.conf
```

PR #85, opened 2019-02-20
Don't generate clusterID

markmc:
Since 0.10.0, clusterID is no longer part of install-config.yaml

See openshift/installer#1057


hardys:
/lgtm

This reminds me that we also need to tag all created resouce with this ID like is done via terraform for AWS/OpenStack - we probably won't use the openshift-install destroy code yet, but it makes sense to align with the same tagging pattern for the future.

PR #75, opened 2019-02-15
Move ironic to the virt host

derekhiggins:

PR #86, opened 2019-02-20
Update install-config 

markmc:
Update to v1beta3 version. Remove everything except where we want to supply required information or override the defaults.

### DIFF
--- a/06_deploy_bootstrap_vm.sh
+++ b/06_deploy_bootstrap_vm.sh
@@ -13,30 +13,12 @@ source utils.sh
 if [ ! -d ocp ]; then
     mkdir -p ocp
     cat > ocp/install-config.yaml << EOF
-apiVersion: v1beta1
+apiVersion: v1beta3
 baseDomain: ${BASE_DOMAIN}
-machines:
-- name:     master
-  platform: {}
-  replicas: null
-- name:     worker
-  platform: {}
-  replicas: null
 metadata:
-  creationTimestamp: null
   name: ${CLUSTER_NAME}
-networking:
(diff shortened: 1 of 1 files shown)


markmc:
Sorry for all the pushes, should be done now.

hardys:
Thanks @markmc - sorry but can you rebase this please, I just merged #75 which touched the same script.

markmc:
No worries, rebased and squashed

06_deploy_bootstrap_vm.sh
```diff
@@ -13,30 +13,12 @@ source utils.sh
...
-    hostSubnetLength: 9
-  machineCIDR: 192.168.126.0/24
-  serviceCIDR: 172.30.0.0/16
-  type: OpenshiftSDN
 platform:
-  libvirt:
-    URI: qemu:///system
-    network:
-      if: tt0
+  libvirt: {}
```
hardys: Ah, yeah, good catch - the URI is there because I was originally trying to use the openshift-install libvirt backend to launch the bootstrap vm via "create cluster", but it turns out evern with replicas:null it still builds a single master node as well.

vrutkovs: While we're on this lets change this to `platform: none` - it should be the same as libvirt, but later on we change it for baremetal use cases

hardys: I tried this and you need to do:

```
platform:
  none: {}
```

Just setting platform to either null or the string none doesn't work, there's a json marshalling error from openshift-install

PR #87, opened 2019-02-20
Tolerate failure in *_cleanup.sh

hardys:
This means we still clean things up if e.g the bootstrap VM is still
defined but not running.

### DIFF
--- a/ocp_cleanup.sh
+++ b/ocp_cleanup.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-set -e
+set -x
 
 source ocp_install_env.sh
 
--- a/host_cleanup.sh
+++ b/host_cleanup.sh
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-set -xe
+set -x
 
 source common.sh
 


ocp_cleanup.sh
```diff
@@ -4,9 +4,9 @@ set -e
 
 source ocp_install_env.sh
 
-sudo virsh destroy "${CLUSTER_NAME}-bootstrap"
-sudo virsh undefine "${CLUSTER_NAME}-bootstrap" --remove-all-storage
-VOL_POOL=$(sudo virsh vol-pool "/var/lib/libvirt/images/${CLUSTER_NAME}-bootstrap.ign")
-sudo virsh vol-delete "${CLUSTER_NAME}-bootstrap.ign" --pool "${VOL_POOL}"
+sudo virsh destroy "${CLUSTER_NAME}-bootstrap" || true
+sudo virsh undefine "${CLUSTER_NAME}-bootstrap" --remove-all-storage || true
+VOL_POOL=$(sudo virsh vol-pool "/var/lib/libvirt/images/${CLUSTER_NAME}-bootstrap.ign")
```
derekhiggins: This could also fail I think, although if it does the vol-delete can't run, maybe we could just move the "rm" commands to the top and let this bit fail if it is going to.

hardys: Yeah, we could just remove the set -e instead and show the errors I guess

PR #88, opened 2019-02-20
Remove provisioning nic from the kernel command line

derekhiggins:
Having it there results in dracut attempting to dchp on boot.
This doesn't work and stalls the boot process once ironic is gone away.

### DIFF
--- a/ignition_patches/master/02_noneed_prov_nic.json
+++ b/ignition_patches/master/02_noneed_prov_nic.json
@@ -1,0 +1,27 @@
+[
+  {
+    "op": "add",
+    "path": "/spec/config/systemd/units/-",
+    "value": {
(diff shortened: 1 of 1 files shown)


hardys:
This raises the question of how we'll handle IPAM for existing nodes, I guess we'll have to pre-allocate the IPs for the master nodes, then somehow provide predictable IPs via the BareMetalHost abstraction in the operator-driven worker deployment case?

@dhellmann wondering if there are any established patterns for this we can reuse for the operator?

We'll want the provisioning network to remain accessible (with a discoverable IP that ideally doesn't change) after the node is provisioned I suspect, even in the case where Ironic is a short-lived appliance.

juliakreger:
@hardys I wonder if we should be instead calling it a back-end network that we also happen to be using for provisioning during the initial bootstrap. I think in that case, as a purely internal network, it should be fine to have some sort of static address assignment. I think the key is avoiding users having to decide, if that makes sense.

hardys:
Ok this works fine as-is, so lets merge it and figure out the best approach to address reservation later.

PR #90, opened 2019-02-21
Reduce the RHCOS disk size

derekhiggins:
The RHCOS image disk size is 16G so when IPA is writing
the qcow2 to disk during deployment, it has to write 16G (at least)
to the master disk. This happens 3 times, so on a virtual env we
are writting 50G during each deploy. The partitions on the RHCOS
image are only about 3.3G so we can safely reduce the disk size and
therfore reduce the amount of data ironic writes. When the master nodes
boot growfs runs and expands the filesyatem to the size available
on the master node, so we end up using all the space available anyways.

This patch in local tests takes 3 minutes off the deploy time for
the master nodes, so should speed up development time, in the
longer term we should look into changing the image upstream.

### DIFF
--- a/04_setup_ironic.sh
+++ b/04_setup_ironic.sh
@@ -43,7 +43,15 @@ fi
 RHCOS_IMAGE_FILENAME_RAW="${RHCOS_IMAGE_FILENAME_OPENSTACK}.raw"
 if [ ! -e "$IRONIC_DATA_DIR/html/images/$RHCOS_IMAGE_FILENAME_DUALDHCP" ] ; then
     pushd $IRONIC_DATA_DIR/html/images
-    qemu-img convert "$RHCOS_IMAGE_FILENAME_OPENSTACK" "${RHCOS_IMAGE_FILENAME_RAW}"
+
+    # Calculate the disksize required for the partitions on the image
+    # we do this to reduce the disk size so that ironic doesn't have to write as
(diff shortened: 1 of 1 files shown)


04_setup_ironic.sh
```diff
@@ -53,6 +53,11 @@ if [ ! -e "$IRONIC_DATA_DIR/html/images/$RHCOS_IMAGE_FILENAME_DUALDHCP" ] ; then
     sudo kpartx -d /dev/${LOOPBACK}
     sudo losetup -d /dev/${LOOPBACK}
     qemu-img convert -O qcow2 -c "$RHCOS_IMAGE_FILENAME_RAW" "$RHCOS_IMAGE_FILENAME_DUALDHCP"
+
+    # Reduce the disk size so that ironic doesn't have to write as much data during deploy
+    # This is a safe opperation only because the partitions on the disk are only about 3.3G
+    # TODO: add check for partition size to ensure we don't zap valid data if they change in future
+    qemu-img resize --shrink "$RHCOS_IMAGE_FILENAME_DUALDHCP" 4G
```
derekhiggins: I'm going to look into making this safer if you want to wait before merging.

hardys: Yeah maybe we could use virt-df or something to calculate the size?

This works fine though so happy to merge this as-is then you can push a followup to validate/calculate the size?

PR #91, opened 2019-02-21
Add firewall rule to allow IPA access to Ironic API

hardys:
This was failing for me, with a misleading "no route to host" error
from the IPA urllib code (the route was fine, but the firewall
blocked the port)

So, ensure we always open up this port for traffic over the brovc
bridge from the master/worker VMs to Ironic running on the host.

### DIFF
--- a/04_setup_ironic.sh
+++ b/04_setup_ironic.sh
@@ -39,6 +39,11 @@ if [ ! -e /etc/sysconfig/network-scripts/ifcfg-brovc ] ; then
     sudo ifup brovc
 fi
 
+# Add firewall rule to ensure the IPA ramdisk can reach the Ironic API on the host
+if ! sudo iptables -C INPUT -i brovc -p tcp -m tcp --dport 6385 -j ACCEPT; then
+    sudo iptables -I INPUT -i brovc -p tcp -m tcp --dport 6385 -j ACCEPT
+fi
+
 # Workaround so that the dracut network module does dhcp on eth0 & eth1
 RHCOS_IMAGE_FILENAME_RAW="${RHCOS_IMAGE_FILENAME_OPENSTACK}.raw"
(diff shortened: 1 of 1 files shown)


derekhiggins:
@hardys Looks good to me, I'm curious how port 80 worked, had you already a iptables rule for that, I'm wondering if we should also add it.

hardys:
> @hardys Looks good to me, I'm curious how port 80 worked, had you already a iptables rule for that, I'm wondering if we should also add it.

Hmm good point, I don't have a rule specific to brovc, but there's an IN_public_allow chain, I should probably re-test this on a clean host to be certain what we need to add, I guess adding an explicit brovc rule for port 80 would be fine though?

```
[shardy@tripleodev2 dev-scripts]$ sudo iptables -S| grep brovc
-A INPUT -i brovc -p tcp -m tcp --dport 6385 -j ACCEPT
-A INPUT -i brovc -p udp -m udp --dport 53 -j ACCEPT
-A INPUT -i brovc -p tcp -m tcp --dport 53 -j ACCEPT
-A INPUT -i brovc -p udp -m udp --dport 67 -j ACCEPT
-A INPUT -i brovc -p tcp -m tcp --dport 67 -j ACCEPT
-A FORWARD -i brovc -o brovc -j ACCEPT
-A FORWARD -o brovc -j REJECT --reject-with icmp-port-unreachable
-A FORWARD -i brovc -j REJECT --reject-with icmp-port-unreachable
-A OUTPUT -o brovc -p udp -m udp --dport 68 -j ACCEPT
[shardy@tripleodev2 dev-scripts]$ sudo iptables -S| grep "dport 80"
-A IN_public_allow -p tcp -m tcp --dport 80 -m conntrack --ctstate NEW -j ACCEPT
```

PR #94, opened 2019-02-22
Removing overcloud network as part of cleanup.

achuzhoy:
The patch takes care of removing the overcloud
network which had to be taken care of manually.

### DIFF
--- a/02_configure_host.sh
+++ b/02_configure_host.sh
@@ -40,6 +40,11 @@ if ! sudo iptables -C INPUT -i baremetal -p udp -m udp --dport 6230:6235 -j ACCE
     sudo iptables -I INPUT -i baremetal -p udp -m udp --dport 6230:6235 -j ACCEPT
 fi
 
+#Allow access to dualboot.ipxe
+if ! sudo iptables -C INPUT -p tcp --dport 80 -j ACCEPT ; then
+    sudo iptables -I INPUT -p tcp --dport 80 -j ACCEPT
+fi
+
 # Need to route traffic from the provisioning host.
 if [ "$EXT_IF" ]; then
   sudo iptables -t nat -A POSTROUTING --out-interface $EXT_IF -j MASQUERADE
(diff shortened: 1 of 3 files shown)


host_cleanup.sh
```diff
@@ -21,3 +21,5 @@ ANSIBLE_FORCE_COLOR=true ansible-playbook \
 sudo rm -rf /etc/NetworkManager/dnsmasq.d/openshift.conf /etc/NetworkManager/conf.d/dnsmasq.conf
 sudo virsh net-destroy baremetal
 sudo virsh net-undefine baremetal
+sudo virsh net-destroy overcloud
+sudo virsh net-undefine overcloud
```
juliakreger: Wouldn't these only be needed on a machine with a prior deployment via these scripts? tl;dr we could likely undo this in like a week?

achuzhoy: > Wouldn't these only be needed on a machine with a prior deployment via these scripts? tl;dr we could likely undo this in like a week?
Yes, if the respective virsh network doesn't exist - error will be printed and the script will continue executing.

02_configure_host.sh
```diff
@@ -40,6 +40,11 @@ if ! sudo iptables -C INPUT -i baremetal -p udp -m udp --dport 6230:6235 -j ACCE
     sudo iptables -I INPUT -i baremetal -p udp -m udp --dport 6230:6235 -j ACCEPT
 fi
 
+#Allow access to dualboot.ipxe
+if ! sudo iptables -C INPUT -p tcp --dport 80 -j ACCEPT ; then
+    sudo iptables -I INPUT -p tcp --dport 80 -j ACCEPT
```
celebdor:
```suggestion
    sudo iptables -I INPUT -p tcp --dport 80 -m comment --comment "Serving pxe files" -j ACCEPT
```

host_cleanup.sh
hardys: This is actually a bug in the quickstart roles I think, since the ./tripleo-quickstart-config/roles/environment/setup/tasks/main.yml file contains the task which creates this network, but it's not cleaned up in the destroy tasks.

Removing it here should be fine though, even though it's probably not strictly necessary unless the configuration of that network changes.

02_configure_host.sh
hardys: It's not only iPXE files, we also need it to download the image to the IPA ramdisk, so personally I'm fine with leaving out the comment, port 80 is ubiquitous enough most folks should figure out the kind of traffic it's used for :)

achuzhoy:
FWIW, 
took all the non-merged commits for a spin - worked fine. Ended up with deployed nodes.

derekhiggins:
lgtm

PR #95, opened 2019-02-25
Wait for masters to become active/booted

hardys:
This allows us to print some helpful output with the node IPs
and also will work better when we eventually want to test
subsequent things on the deployed cluster

### DIFF
--- a/07_deploy_masters.sh
+++ b/07_deploy_masters.sh
@@ -3,6 +3,7 @@
 set -eux
 source utils.sh
 source common.sh
+source ocp_install_env.sh
 
 # Note This logic will likely run in a container (on the bootstrap VM)
 # for the final solution, but for now we'll prototype the workflow here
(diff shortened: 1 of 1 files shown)


hardys:
I wonder if we should also check for the nodes to become Ready via oc here, e.g:

```
[shardy@tripleodev2 dev-scripts]$ oc --config ocp/auth/kubeconfig get nodes
NAME       STATUS    ROLES     AGE       VERSION
master-0   Ready     master    8m58s     v1.12.4+50c2f2340a
master-1   Ready     master    8m51s     v1.12.4+50c2f2340a
master-2   Ready     master    9m5s      v1.12.4+50c2f2340a
```

07_deploy_masters.sh
```diff
@@ -37,4 +37,41 @@ done
...
+
+echo "Master nodes up, you can ssh to the following IPs with core@<IP>"
+sudo virsh net-dhcp-leases baremetal
+
+while ! oc --config ocp/auth/kubeconfig get nodes | grep -q master; do
+  echo "Waiting for the master API to become ready..."
+  sleep 10
+done
+
+NODES_ACTIVE=$(oc --config ocp/auth/kubeconfig get nodes | grep master | grep Ready | wc -l)
```
derekhiggins: At this point you might also get nodes in a "NotReady" state
master-0   NotReady   master    26s       v1.12.4+50c2f2340a

After about a minute they seem to switch to "Ready", but the VIP still hasn't moved to a master node, this appears to happen (in local tests) about 4 minutes later, does this matter? If so we also should test for this. Something like
[[ $(ssh -o StrictHostKeyChecking=no  core@api.ostest.test.metalkube.org hostname) =~ master- ]] || echo Still Waiting

hardys: Ah thanks, I'd missed that, sure we can add the hostname check, or make the grep stricter I guess?

derekhiggins: I think I'd go for the hostname check as that seems to be the last thing that happens

07_deploy_masters.sh
```diff
@@ -37,4 +37,41 @@ done
 for node in $(jq -r .nodes[].name $MASTER_NODES_FILE); do
   openstack baremetal node deploy --config-drive configdrive $node
 done
-# FIXME(shardy) we should wait for the node deploy to complete (or fail)
+
+# Note we have to tolerate failure with the || echo "0" due to this issue
+# https://storyboard.openstack.org/#!/story/2005093
+NUM_ACTIVE=$(openstack baremetal node list --fields name --fields provision_state | grep master | grep active | wc -l || echo 0)
```
hardys: Hmm the workaround for the failing command isn't working, failing like:

```./07_deploy_masters.sh: line 45: [: 0
0: integer expression expected
```

So I guess I need to switch to a string comparison or ensure the echo 0 is cast to an int

hardys:
Ok lets hold on this until I address the feedback from Derek and figure out the error in the int comparison

derekhiggins:

hardys:
Ok pushed an update which I think resolves all issues, review feedback welcome :)

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/84/

derekhiggins:
LGTM

PR #96, opened 2019-02-25
Cleanup ssh known hosts in host_cleanup.sh

hardys:
This is helpful to avoid warnings when ssh is used manually
from the host to access the deployed nodes.

### DIFF
--- a/ocp_cleanup.sh
+++ b/ocp_cleanup.sh
@@ -10,3 +10,6 @@ VOL_POOL=$(sudo virsh vol-pool "/var/lib/libvirt/images/${CLUSTER_NAME}-bootstra
 sudo virsh vol-delete "${CLUSTER_NAME}-bootstrap.ign" --pool "${VOL_POOL}"
 rm -rf ocp
 sudo rm -rf /etc/NetworkManager/dnsmasq.d/openshift.conf
+
+# Cleanup ssh keys for baremetal network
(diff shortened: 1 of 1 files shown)


host_cleanup.sh
```diff
@@ -21,3 +21,6 @@ ANSIBLE_FORCE_COLOR=true ansible-playbook \
 sudo rm -rf /etc/NetworkManager/dnsmasq.d/openshift.conf /etc/NetworkManager/conf.d/dnsmasq.conf
 sudo virsh net-destroy baremetal
 sudo virsh net-undefine baremetal
+
+# Cleanup ssh keys for baremetal network
+sed -i "/^192.168.111/d" /home/$USER/.ssh/known_hosts
```
derekhiggins: I think I'd get more value from this if we put it into "ocp_cleanup.sh", as I regularly run it to bring up a new bootstrap node without running host_cleanup.sh, happy to merge it here either if you think its a better fit.

hardys: Yeah that could work too - particularly if we included stopping the openshift VMs in ocp_cleanup, e.g the undeploy part we do when re-running 07?

derekhiggins: Ya, we could add that too. I guess it would be cleaner that way. I usually just rerun 07_.. as the undeploy in there covers me anyways.

PR #103, opened 2019-02-27
Add go dep command to requirements.

imain:
Needed to build some go repositories.

### DIFF
--- a/01_install_requirements.sh
+++ b/01_install_requirements.sh
@@ -62,3 +62,10 @@ fi
 curl --silent --location https://dl.yarnpkg.com/rpm/yarn.repo | sudo tee /etc/yum.repos.d/yarn.repo
 curl --silent --location https://rpm.nodesource.com/setup_10.x | sudo bash -
 sudo yum install yarn nodejs -y
+
+# Install go dep command.  The packaged version is too old.  This is for development only atm.
+export GOPATH=$HOME/go
+mkdir -p $GOPATH/go/bin
+curl -o /tmp/install-go-dep.sh https://raw.githubusercontent.com/golang/dep/master/install.sh
+chmod 755 /tmp/install-go-dep.sh
+/tmp/install-go-dep.sh


celebdor:
which repo building fails without this?

imain:
None of these directly.  The operator-sdk needs it.

dhellmann:
The operator-sdk needs it, and operator-sdk is needed by metalkube/baremetal-operator

derekhiggins:

01_install_requirements.sh
```diff
@@ -62,3 +62,10 @@ fi
 curl --silent --location https://dl.yarnpkg.com/rpm/yarn.repo | sudo tee /etc/yum.repos.d/yarn.repo
 curl --silent --location https://rpm.nodesource.com/setup_10.x | sudo bash -
 sudo yum install yarn nodejs -y
+
+# Install go dep command.  The packaged version is too old.  This is for development only atm.
+export GOPATH=$HOME/go
```
hardys: Rather than hard-coding this, should we either do eval "$(go env)" or source ocp_install_env.sh which already does that?

russellb: +1 to `eval "$(go env)"`

01_install_requirements.sh
```diff
@@ -62,3 +62,10 @@ fi
...
+mkdir -p $GOPATH/go/bin
+curl -o /tmp/install-go-dep.sh https://raw.githubusercontent.com/golang/dep/master/install.sh
+chmod 755 /tmp/install-go-dep.sh
+/tmp/install-go-dep.sh
```
russellb: instead of the tmp file, how about

```
curl https://raw.githubusercontent.com/golang/dep/master/install.sh | sh
```

hardys: I think this has been obsoleted by #139 which did exactly that

hardys:
Obsoleted by #139

01_install_requirements.sh
sadasu: Yes, #139 takes care of installing dep and operator-sdk.

PR #104, opened 2019-02-27
fix OS_URL in README

dhellmann:
The OS_URL value in the readme points to the OpenShift API not Ironic.

Signed-off-by: Doug Hellmann <dhellmann@redhat.com>


celebdor:
This was working when ironic ran on the bootstrap node :wink:

PR #106, opened 2019-03-01
Check to make sure ironic containers are running

derekhiggins:
Now that these have simple healthchecks this should stall
the script if a problem is detected.


derekhiggins:

PR #110, opened 2019-03-04
Replace writing out of 80-libvirt.rules

derekhiggins:
This wasn't working but the libvirt-daemon package
provides a similar config in
/usr/share/polkit-1/rules.d/50-libvirt.rules

### DIFF
--- a/02_configure_host.sh
+++ b/02_configure_host.sh
@@ -23,17 +23,8 @@ ANSIBLE_FORCE_COLOR=true ansible-playbook \
     -i tripleo-quickstart-config/metalkube-inventory.ini \
     -b -vvv tripleo-quickstart-config/metalkube-setup-playbook.yml
 
-# Allow local non-root-user access to libvirt ref
-# https://github.com/openshift/installer/blob/master/docs/dev/libvirt-howto.md#make-sure-you-have-permissions-for-qemusystem
-if sudo test ! -f /etc/polkit-1/rules.d/80-libvirt.rules ; then
-  cat <<EOF | sudo dd of=/etc/polkit-1/rules.d/80-libvirt.rules
-polkit.addRule(function(action, subject) {
-  if (action.id == "org.libvirt.unix.manage" && subject.local && subject.active && subject.isInGroup("wheel")) {
-      return polkit.Result.YES;
-  }
-});
-EOF
-fi
+# Allow local non-root-user access to libvirt
+sudo usermod -a -G "libvirt" $USER
 
 # Allow ipmi to the virtual bmc processes that we just started
 if ! sudo iptables -C INPUT -i baremetal -p udp -m udp --dport 6230:6235 -j ACCEPT ; then
--- a/common.sh
+++ b/common.sh
@@ -28,6 +28,13 @@ fi
 source $CONFIG
 cat $CONFIG
 
+# Connect to system libvirt
+export LIBVIRT_DEFAULT_URI=qemu:///system
+if [ "$USER" != "root" -a "${XDG_RUNTIME_DIR:-}" == "/run/user/0" ] ; then
+    echo "Please use a non-root user, WITH a login shell (e.g. su - USER)"
+    exit 1
+fi
+
 WORKING_DIR=${WORKING_DIR:-"/opt/dev-scripts"}
 NODES_FILE=${NODES_FILE:-"${WORKING_DIR}/ironic_nodes.json"}
 NODES_PLATFORM=${NODES_PLATFORM:-"libvirt"}
--- a/06_deploy_bootstrap_vm.sh
+++ b/06_deploy_bootstrap_vm.sh
@@ -48,7 +48,7 @@ apply_ignition_patches bootstrap "$IGN_FILE"
 
 LATEST_IMAGE=$(ls -ltr redhat-coreos-maipo-*-qemu.qcow2 | tail -n1 | awk '{print $9}')
 sudo cp $LATEST_IMAGE /var/lib/libvirt/images/${CLUSTER_NAME}-bootstrap.qcow2
-sudo virt-install --connect qemu:///system \
+virt-install --connect qemu:///system \
              --import \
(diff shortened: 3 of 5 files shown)


markmc:
I haven't tested yet, but should work according to those libvirt-daemon policy:

```
polkit.addRule(function(action, subject) {
    if (action.id == "org.libvirt.unix.manage" &&
        subject.isInGroup("libvirt")) {
        return polkit.Result.YES;
    }
});
```

derekhiggins:
So adding the group works as expected, but looks like the patch doesn't go far enough, we also need to change the owner of some resources so a non root user can read/write them. Also our master(and worker) nodes are owned by root. So we're are left in a funny state where  virsh list for root shows them but not the bootstrap node, virsh list for the non root user is the other way around.

markmc:
I'd say we need:

```sh
export LIBVIRT_DEFAULT_URI=qemu:///system
```

otherwise all the `sudo virsh` you changed to `virsh` will be using `qemu:///session'

You can use 'virsh uri' to check

hardys:
I also noticed some problems with this if you ssh as root then su to a non-root account, in that case the $XDG_RUNTIME_DIR ends up pointing to the uid0 path so even with the correct policy applied you still can't access things as the non root user.

So maybe we can add some checks in the scripts to catch that as well?

markmc:
> I also noticed some problems with this if you ssh as root then su to a non-root account, in that case the $XDG_RUNTIME_DIR ends up pointing to the uid0 path so even with the correct policy applied you still can't access things as the non root user.
> 
> So maybe we can add some checks in the scripts to catch that as well?

Are you're sure it wasn't the "subject.local && subject.active" part of the policy in dev-scripts that was failing when you ssh?

hardys:
> > I also noticed some problems with this if you ssh as root then su to a non-root account, in that case the $XDG_RUNTIME_DIR ends up pointing to the uid0 path so even with the correct policy applied you still can't access things as the non root user.
> > So maybe we can add some checks in the scripts to catch that as well?
> 
> Are you're sure it wasn't the "subject.local && subject.active" part of the policy in dev-scripts that was failing when you ssh?

I don't think so, this was in an environment I debugged for Sandhya, and IIRC she logged in as root, then did su without "-" to reset the environment.  The result is an error like this which is independent of the polkit rules AFAICS:

```
[root@tripleodev2 ~]# sudo su shardy
[shardy@tripleodev2 root]$ echo $XDG_RUNTIME_DIR
/run/user/0
[shardy@tripleodev2 root]$ virsh list
error: failed to connect to the hypervisor
error: Cannot create user runtime directory '/run/user/0/libvirt': Permission denied
```

It's a corner case, but thought it worth mentioning since it was initially confusing and would be easy enough to validate in the scripts.

derekhiggins:
> It's a corner case, but thought it worth mentioning since it was initially confusing and would be easy enough to validate in the scripts.

Ya, looks like its a thing but once I set LIBVIRT_DEFAULT_URI(which I'm hoping to set in common.sh), its not a problem, maybe we should fail on it anyways so people don't hit it ouside of the scripts?

```
[root@host11 ~]# virsh list
 Id    Name                           State
----------------------------------------------------
 107   ostest-bootstrap               running
[root@host11 ~]# su notstack
[notstack@host11 root]$ virsh list 
error: failed to connect to the hypervisor
error: Cannot create user runtime directory '/run/user/0/libvirt': Permission denied

[notstack@host11 root]$ export LIBVIRT_DEFAULT_URI=qemu:///system
[notstack@host11 root]$ virsh list 
 Id    Name                           State
----------------------------------------------------
 107   ostest-bootstrap               running
[notstack@host11 root]$ 

```

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/85/

common.sh
```diff
@@ -28,6 +28,13 @@ fi
 source $CONFIG
 cat $CONFIG
 
+# Connect to system libvirt
+export LIBVIRT_DEFAULT_URI=qemu:///system
+if [ "$USER" != "root" -a "${XDG_RUNTIME_DIR:-}" == "/run/user/0" ] ; then
+    echo "Please use a non-root user, WITH a login sheel (e.g. su - USER)"
```
markmc: shell not sheel

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/86/

common.sh
hardys:
```suggestion
    echo "Please use a non-root user, WITH a login shell (e.g. su - USER)"
```

02_configure_host.sh
```diff
@@ -23,17 +23,8 @@ ANSIBLE_FORCE_COLOR=true ansible-playbook \
     -i tripleo-quickstart-config/metalkube-inventory.ini \
     -b -vvv tripleo-quickstart-config/metalkube-setup-playbook.yml
 
-# Allow local non-root-user access to libvirt ref
-# https://github.com/openshift/installer/blob/master/docs/dev/libvirt-howto.md#make-sure-you-have-permissions-for-qemusystem
```
hardys: For the avoidance of future confusion should we remove the old /etc/polkit-1/rules.d/80-libvirt.rules or just assume folks will be using a clean install?

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/89/

common.sh
```diff
@@ -28,6 +28,13 @@ fi
 source $CONFIG
 cat $CONFIG
 
+# Connect to system libvirt
+export LIBVIRT_DEFAULT_URI=qemu:///system
```
hardys: One problem is this isn't set by default in the environment, so you still need to either set it, or use sudo to see the VMs via virsh list outside of the scripts.  Can we perhaps set this globally somewhere?

hardys:
Ok other folks are hitting the XDG_RUNTIME_DIR thing so lets merge this and worry about the global setting of the URI as a follow-up

PR #111, opened 2019-03-04
Open ports for Facet backend and Yarn development server

dantrainor:
Open 3000/tcp and 8080/tcp on the host to allow access to Facet
backend and Yarn development server from remote locations, to aid
in remote development.

### DIFF
--- a/02_configure_host.sh
+++ b/02_configure_host.sh
@@ -51,6 +51,16 @@ if [ "$EXT_IF" ]; then
   sudo iptables -A FORWARD --in-interface baremetal -j ACCEPT
 fi
 
+# Add access to backend Facet server from remote locations
+if ! sudo iptables -C INPUT -p tcp --dport 8080 -j ACCEPT ; then
+  sudo iptables -I INPUT -p tcp --dport 8080 -j ACCEPT
+fi
+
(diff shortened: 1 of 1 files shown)


hardys:
I wonder if we should make these conditional, and disabled by default?

The original idea was that the development would happen on the host (either directly or via ssh), in which case for all developer virt setups we should only need firewall rules on the bridges needed on the VMs?

I'm not blocking on this as we do already have some unconditional rules for BMC and PXE/HTTP, but likewise those should perhaps be made conditional on NODES_PLATFORM=baremetal or something?

dantrainor:
I don't disagree with that.  A casual search showed we didn't rely on conditionals for other iptables rules so I didn't go that route, however.  These specific rules are used a little differently than other rules, so that might be the case.

hardys:
Yeah to be clear, we should probably make the existing pxe rules conditional as well (there are a few bridge related rules that already are though)

Are you planning to update this branch, or should we do a follow-up for all rules?

dantrainor:
Agreed.  For the sake of task completion I'd like to pursue a follow-up push that addresses all the rules.  I'd be ok doing this, but want to make sure I understand the conditionals we already have available in an effort to make sure it's done properly.

PR #113, opened 2019-03-05
Also cleanup api ssh known_hosts entry

hardys:
Some scripts access the cluster via the vip so we also need to clean
this to avoid ssh warnings when re-running the scripts

We may also want to clean this after the VIP fails over from the
bootstrap node, but I've not considered that in this patch


derekhiggins:
Build ABORTED, see build http://10.8.144.11:8080/job/dev-tools/82/

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/83/

derekhiggins:
So the first patch the CI reports on is one that ci didn't actually touch (we're only running make). Tested, it worked.

PR #114, opened 2019-03-05
manifests: Add CNV (kubevirt) manifests

fabiand:
This change adds the kubevirt manifests.

This is not all of the CNV manifests. But a start.

Signed-off-by: Fabian Deutsch <fabiand@fedoraproject.org>

### DIFF
--- a/manifests/README.cnv
+++ b/manifests/README.cnv
@@ -30,6 +30,11 @@ After creation you can wait for the application readiness with:
 ```
 
 
+### Running a VM
+
+Apply this config https://github.com/kubevirt/demo/blob/master/manifests/vm.yaml
+
+
 ## CDI
 
 TBD
--- a/manifests/README.cnv
+++ b/manifests/README.cnv
@@ -1,0 +1,30 @@
+# CNV Manifests
+
+## KubeVirt
+
+Done as described here:
+
+https://kubevirt.io/user-guide/docs/latest/administration/intro.html#2-alternative-flow-aka-operator-flow
+
+Created using:
+
+```
+VERSION=v0.15.0
+curl -Lo 10_cnv_kubevirt_op.yaml https://github.com/kubevirt/kubevirt/releases/download/$VERSION/kubevirt-operator.yaml
+curl -Lo 11_cnv_kubevirt_cr.yaml https://github.com/kubevirt/kubevirt/releases/download/$VERSION/kubevirt-cr.yaml
+```
+
+After creation you can wait for the application readiness with:
+
+```
+ kubectl wait kv kubevirt --for condition=Ready
+```
+
+
+## CDI
+
+TBD
+
+## Network
+
+TBD
--- a/manifests/README.cnv
+++ b/manifests/README.cnv
@@ -12,6 +12,15 @@ Created using:
 VERSION=v0.15.0
 curl -Lo 10_cnv_kubevirt_op.yaml https://github.com/kubevirt/kubevirt/releases/download/$VERSION/kubevirt-operator.yaml
 curl -Lo 11_cnv_kubevirt_cr.yaml https://github.com/kubevirt/kubevirt/releases/download/$VERSION/kubevirt-cr.yaml
+cat > 12_cnv_kubevirt_config.yaml <<EOY
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: kubevirt-config
+  namespace: kubevirt
+data:
+  debug.useEmulation: "true"
+EOY
 ```
 
 After creation you can wait for the application readiness with:
--- a/manifests/README.md
+++ b/manifests/README.md
@@ -1,0 +1,14 @@
+A number of manifests which will be deployed during installation.
+
+Assumption is that the list of sorted manifests is applicable using:
+
+```bash
+for MANIFEST in $(ls -1 | sort -h);
+do
+  kubectl apply -f $MANIFEST
+done
+```
+
+> **Important note:** We do not assume that all manifests can be applied in one
+> go using `kubectlapply -f .` as some manfiests will be depending on others,
+> i.e. a manifest introducing a CRD is needed before the CR can be created.
--- a/manifests/README.md
+++ b/manifests/README.md
@@ -10,5 +10,5 @@ done
 ```
 
 > **Important note:** We do not assume that all manifests can be applied in one
-> go using `kubectlapply -f .` as some manfiests will be depending on others,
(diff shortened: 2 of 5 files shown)


fabiand:
Please note that the initial PR uses manifests consuming upstream builds.
A follow up with downstream builds will follow.

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/88/

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/90/

manifests/README.md
```diff
@@ -0,0 +1,14 @@
...
+
+```bash
+for MANIFEST in $(ls -1 | sort -h);
+do
+  kubectl apply -f $MANIFEST
+done
+```
+
+> **Important note:** We do not assume that all manifests can be applied in one
+> go using `kubectlapply -f .` as some manfiests will be depending on others,
```
booxter: -> kubectl apply, -> manifests, -> will depend

manifests/README.cnv
```diff
@@ -0,0 +1,44 @@
...
+```
+
+After creation you can wait for the application readiness with:
+
+```
+ kubectl wait kv kubevirt --for condition=Ready
+```
+
+
+### RUnning a VM
```
booxter: -> Running

manifests/README.md
```diff
@@ -0,0 +1,14 @@
+A number of manifests which will be deployed during installation.
+
+Assumption is that the list of sorted manifests is applicable using:
+
+```bash
+for MANIFEST in $(ls -1 | sort -h);
```
booxter: nit: I am being told that ls already sorts by alphabetical order so there is probably no need for explicit sort

manifests/README.md
```diff
@@ -0,0 +1,14 @@
+A number of manifests which will be deployed during installation.
+
+Assumption is that the list of sorted manifests is applicable using:
```
booxter: Do you plan to hook in calling to ```oc apply```, or what's the reason for them to not be applied as part of ```make``` process?

manifests/README.cnv
```diff
@@ -0,0 +1,44 @@
...
+## KubeVirt
+
+Done as described here:
+
+https://kubevirt.io/user-guide/docs/latest/administration/intro.html#2-alternative-flow-aka-operator-flow
+
+Created using:
+
+```
+VERSION=v0.15.0
```
booxter: This is prone to become stale. (Uness we bump it each new release.) Should we entertain fetching the manifests dynamically as part of ```make``` process?

manifests/README.cnv
```diff
@@ -0,0 +1,44 @@
...
+curl -Lo 10_cnv_kubevirt_op.yaml https://github.com/kubevirt/kubevirt/releases/download/$VERSION/kubevirt-operator.yaml
+curl -Lo 11_cnv_kubevirt_cr.yaml https://github.com/kubevirt/kubevirt/releases/download/$VERSION/kubevirt-cr.yaml
+cat > 12_cnv_kubevirt_config.yaml <<EOY
```
booxter: (just curious) what does Y mean in ```EOY```?

booxter:
for the record, I also posted PR for additional network plugins here: https://github.com/metalkube/dev-scripts/pull/119 This PR assumes (but doesn't base on) this PR (incl. its README etc.)

manifests/README.md
fabiand: Done. My broken space bar...

manifests/README.cnv
fabiand: Done

manifests/README.md
fabiand: Well  - My main goal is to spell out what assumptions I made in the naming of these manifests.

But thanks.

manifests/README.md
fabiand: This PR is exclusively about providing the manifests.
@russellb and @hardys will consume them.

Thus I don#t personallyplan to do it, but the two of them I suppose.

manifests/README.cnv
fabiand: I do not plan to do this in this PR, this PR is about providing the initial stuff and seeing how it will be consumed.

but yes - anything we aim for on the long run needs to be automated.I'd even favor to get rid of this commited yamls in that case.

manifests/README.cnv
fabiand: end of yaml :)

fabiand:
Thanks for the review @booxter .

Again: The prupose of this PR is to provide initial manifests to be consumed by the installer.
However, this PR is not about integrating it into the installer flow.

@hardys @russellb are taking care to include it in the flow.

manifests/README.md
hardys: The current thinking is that we'll rely on the openshift-install logic that already exists to apply manifests - the "create manifests" stage generates some manifests, and we can add some extra files there, then we run the "create ignition-configs" step and those manifests are passed to the bootstrap VM and ultimately applied to the cluster.

We could manually apply them, but the idea is where possible to closely align with the existing openshift-install workflow and practices.

Do you see any barrier to trying that approach?  If not then I'll give it a try :)

manifests/README.cnv
hardys: The way this is handled in openshift-install is some code generates the yaml files, ref:

https://github.com/openshift/installer/tree/master/pkg/asset/manifests

And we have a fork of that repo for the kni work here:

https://github.com/metalkube/kni-installer/

So if we want to avoid the static yamls, they could be generated or fetched in that code?

manifests/README.cnv
```diff
@@ -30,6 +30,11 @@ After creation you can wait for the application readiness with:
 ```
 
 
+### RUnning a VM
```
hardys: Nit s/RU/Ru/

hardys:
Thanks for pushing this - I've not yet tested it, but seems like a good place to start, I noted one capitalization nit, and there's a typo in the 3rd commit message ("manifests: Reference to how to spanw a VM") so if you can fix those I'd say lets merge this, and I'll push a follow up showing how I expect this to integrate with the install flow.

Unless there are any objections I'll initially integrate by generating the manifests, adding these, then generating the ignition configs for deployment, and we can look at a potentially more direct integration via kni-installer in a future iteration

fabiand:
Thanks! Done.

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/110/

PR #118, opened 2019-03-05
Use a higher number pci bus for the provisioning nic

mcornea:
This change sets a higher pci bus number for the provisioning nic
device(default is 0x00) in order for the OS to detect this nic as
the 2nd network interface. This is so we match the networking
prerequisites[1] that we document for baremetal environments.

Note: we need to keep the provisioning nic first in the libvirt
domain definition to allow pxe booting via virtualbmc. Currently
virtualbmc only supports setting the domain's `network` boot device
which always sets the 1st nic as the boot device.

[1] https://docs.google.com/document/d/1qTBuV5uRTDd1rh7TByFFTbFfVPSEKD44sN6IKHojgYA/edit


mcornea:
Please note that with this change I am seeing the following error[1] when the master nodes boot the coreos image.

[1] https://drive.google.com/file/d/19shHja7occP606cTEYXyoCTkbaUzRNCp/view

derekhiggins:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/95/

mcornea:
> Please note that with this change I am seeing the following error[1] when the master nodes boot the coreos image.
> 
> [1] https://drive.google.com/file/d/19shHja7occP606cTEYXyoCTkbaUzRNCp/view

The latest change should address this ^ issue

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/97/

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/98/

derekhiggins:
The patch lgtm and it would be nice to get ride of the grub.cfg hack. I think the document you linked to describes a single users env, before approving do we know if this is what the target environment will look like?

hardys:
The google doc does describe the hardware used by @jtaleric but that doesn't represent the final target hardware AFAIK - I think we'll need more information to decide if this fits the final hardware but for now lgtm

tripleo-quickstart-config/roles/libvirt/setup/overcloud/templates/baremetalvm.xml.j2
```diff
@@ -31,6 +31,9 @@
       <mac address='{{ node_mac_map.get(item.name).get(network.name) }}'/>
       <source bridge='{{ network.bridge }}'/>
       <model type='{{ overcloud_libvirt_nic_model }}'/>
+{% if network.name == 'provisioning' %}
+      <address type='pci' domain='0x0000' bus='0x01' slot='0x03' function='0x0'/>
```
etingof: Considering bus numbers from zero up, would it be safer to use '0x00' here? I wonder if libvirt can take up the unused lower bus IDs?

Also, `domain` seems to be optional in the config, may be safer to drop it?

mcornea:
> The google doc does describe the hardware used by @jtaleric but that doesn't represent the final target hardware AFAIK - I think we'll need more information to decide if this fits the final hardware but for now lgtm

Sounds good, I'll keep this PR open in case we decide to go this route.

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/108/

PR #119, opened 2019-03-05
Install additional CNI / device network plugins

booxter:
These are network plugins installed by kubevirt-ansible.

OVS CNI pod needed adjustment in line with
https://github.com/kubevirt/ovs-cni/pull/70 to be able to create ovs-vsctl
executable on CoreOS.

### DIFF
--- a/manifests/120_cni_plugins.yaml
+++ b/manifests/120_cni_plugins.yaml
@@ -1,0 +1,49 @@
+---
+apiVersion: extensions/v1beta1
+kind: DaemonSet
+metadata:
+  name: kube-cni-plugins-amd64
+  namespace: kube-system
+  labels:
+    tier: node
+    app: cni-plugins
+spec:
+  template:
+    metadata:
+      labels:
+        tier: node
+        app: cni-plugins
+    spec:
+      hostNetwork: true
+      nodeSelector:
+        beta.kubernetes.io/arch: amd64
+      tolerations:
+      - key: node-role.kubernetes.io/master
+        operator: Exists
+        effect: NoSchedule
+      containers:
+      - name: cni-plugins
+        image: quay.io/kubevirt/cni-default-plugins:latest
+        command:
+        - /bin/bash
+        - -c
+        - |
+          cp -rf /usr/src/containernetworking/plugins/bin/* /opt/cni/bin/
+          echo "Entering sleep... (success)"
+          sleep infinity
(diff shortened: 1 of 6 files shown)


derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/96/

booxter:
/hold

Perhaps including a script to apply the manifests doesn't belong here if at all.

booxter:
@hardys please take a look, this is in addition to already merged kubevirt pieces. I believe you work on integration point for the ```make``` flow so I skip this part here.

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/118/

hardys:
@booxter thanks for this, please see https://github.com/metalkube/dev-scripts/pull/127 for my idea on how we should apply the manifests - it just adds them to those already generated for the openshift installer, so in theory we don't have to write any scripts to apply them.

Seems to work Ok but I've not tried this PR with it yet (only checked that the kubevirt CR exists) - can you check what I did with the numbering/indexes, as I think we may need to rename all of these additions, and I'm not sure if they should be applied before or after the kubevirt manifests?

booxter:
/hold

There is a discussion in upstream ```ovs-cni``` PR about what's the best place to put the ```ovs-vsctl``` binary to: whether it's ```/opt/cni/bin``` or ```/usr/local/bin```. Let's hold off this PR from merging until we get to agreement there.

booxter:
@hardys I've renamed all manifests to ```1NN_*.yaml``` format. Ordering wise, these additional CNI / device plugins should not care about the order in which they are posted in relation to KubeVirt operator. As long as users don't create VMIs before all components are up and running, it shouldn't matter. (Even then it's just a matter of failed VMIs/pods which probably should be expected when you, for some reason, post resources in the middle of installation.)

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/136/

booxter:
/hold cancel

OK, ```ovs-cni``` owner confirmed now that using ```/usr/local/bin/``` is the preferred way.

hardys:
Ok lets merge this, we don't yet have kubevirt working, and I think we'll want to move these to kni-installer but lets merge these and work those issues out via follow-up tasks.

PR #121, opened 2019-03-06
Adding check for failed string in status.

achuzhoy:
Sometimes the status can be "deploy failed" and this
patch adds the check to catch it too. Otherwise the
checking loop doesn't end.

### DIFF
--- a/07_deploy_masters.sh
+++ b/07_deploy_masters.sh
@@ -48,7 +48,7 @@ done
 # https://storyboard.openstack.org/#!/story/2005093
 NUM_ACTIVE=$(openstack baremetal node list --fields name --fields provision_state | grep master | grep active | wc -l || echo 0)
 while [ "$NUM_ACTIVE" != "3" ]; do
-  if openstack baremetal node list --fields name --fields provision_state | grep master | grep error; then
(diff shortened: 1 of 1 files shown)


07_deploy_masters.sh
```diff
@@ -48,7 +48,7 @@ done
 # https://storyboard.openstack.org/#!/story/2005093
 NUM_ACTIVE=$(openstack baremetal node list --fields name --fields provision_state | grep master | grep active | wc -l || echo 0)
 while [ "$NUM_ACTIVE" != "3" ]; do
-  if openstack baremetal node list --fields name --fields provision_state | grep master | grep error; then
+  if openstack baremetal node list --fields name --fields provision_state | grep master | grep error | grep failed; then
```
hardys: I'm not sure this will work, because the grep error will already have filtered entry with "failed" status?

Perhaps something like grep -e error -e failed, or a conditional regex would be better?

achuzhoy: You're absolutely right (as usual).
Corrected.

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/116/

PR #123, opened 2019-03-07
setup_ironic: Do not copy if src is dst

celebdor:
In case your WORKING DIR is the same directory that has the dev scripts
copying will fail complaining that the source and destination are the
same file. In this case, we should just not copy

### DIFF
--- a/04_setup_ironic.sh
+++ b/04_setup_ironic.sh
@@ -8,7 +8,9 @@ source common.sh
 source get_images.sh
 
 # ironic dnsmasq and ipxe config
-cp ironic/dnsmasq.conf $IRONIC_DATA_DIR/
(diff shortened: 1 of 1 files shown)


derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/122/

hardys:
We discussed this on slack, and I'm fine with it if folks think it's helpful - I'd assumed it would be better to keep the working files in a separate dir to avoid making a mess in the git checkout, untracked files etc, so if we do this perhaps we can add the expected things to .gitignore?

Or, we can just require that WORKING_DIR is some other location, and add a check for that?

celebdor:
Better to require WORKING_DIR to be some other location. Let's do a different PR for that.

PR #125, opened 2019-03-07
Add self-hosted OCP API Loadbalancer

yboaron:
This commit creates LB for OCP API, and load balances the
API traffic across all masters.

Comment#1: Because this LB runs on the master machine and OCP
API listen on API port (6443) on all interfaces, it's
required to allocate a new port that will be used as the
LB frontend (7443).

Comment#2: current implementation doesn't take into consideration the
VRRP status and LBs created on all the masters.

### DIFF
--- a/machineconfig/master/spec/config/storage/files/haproxy.cfg.template L2V0Yy9oYXByb3h5L2hhcHJveHkuY2ZnLnRlbXBsYXRlCg==
+++ b/machineconfig/master/spec/config/storage/files/haproxy.cfg.template L2V0Yy9oYXByb3h5L2hhcHJveHkuY2ZnLnRlbXBsYXRlCg==
@@ -1,0 +1,22 @@
+global
+    external-check
+defaults
+    mode                    tcp
+    log                     global
+    option                  dontlognull
+    retries                 3
+    timeout http-request    10s
+    timeout queue           1m
+    timeout connect 10s
+    timeout client 86400s
+    timeout server 86400s
+    timeout tunnel 86400s
+frontend  main
+  bind :7443
+  default_backend masters
+backend masters
+   option external-check
+   option log-health-checks
+   external-check path "/usr/bin:/usr/local/bin"
+   external-check command "/usr/local/etc/haproxy/members_healthcheck.sh"
+   balance     roundrobin
--- a/machineconfig/master/spec/config/storage/files/haproxy.sh L3Zhci91c3Jsb2NhbC9iaW4vaGFwcm94eS5zaAo=
+++ b/machineconfig/master/spec/config/storage/files/haproxy.sh L3Zhci91c3Jsb2NhbC9iaW4vaGFwcm94eS5zaAo=
@@ -1,0 +1,121 @@
+#!/usr/bin/env bash
+set -e
+
+
+function generate_haproxy_cfg {
+envsubst < $HOST_CFG_PATH/haproxy.cfg.template | sudo tee $HOST_CFG_PATH/haproxy.cfg
+
+etcd_members=$(dig +noall +answer -t SRV _etcd-server-ssl._tcp.$CLUSTER_DOMAIN | awk '{print $NF}')
+for item in $etcd_members
+do
+    ips=$(dig +noall +answer $item  | awk '{print $NF}')
+    if [[ "$?" -eq 0 && "${#ips[@]}" -ne 0 ]]; then
+        sudo echo "   server "$item" "$ips":"$OCP_API_PORT" weight 1 check inter 3s fall 3 rise 3" >>  $HOST_CFG_PATH/haproxy.cfg
+    fi
+done
+
+}
+
+function has_master_api_lb_topology_changed {
+etcd_members=$(dig +noall +answer -t SRV _etcd-server-ssl._tcp.$CLUSTER_DOMAIN   | awk '{print $NF}')
+for item in $etcd_members
+do
+    ips=$(dig +noall +answer $item | awk '{print $NF}')
+    haproxy_cfg_ip=($(grep $item $HOST_CFG_PATH/haproxy.cfg | awk '{print $3}' | sed 's/:.*//'))
+    if [[ "$?" -eq 0 && "${#ips[@]}" -ne 0 ]]; then
+       if [ "$haproxy_cfg_ip" != "$ips" ]; then
+           return 0
+       fi
+    else
+       if [[! -z "${haproxy_cfg_ip// }" ]];then
+          return 0
+       fi
+    fi
+done
+return 1
+}
+
+function start_haproxy {
+sudo mkdir --parents $HOST_CFG_PATH
+
+generate_haproxy_cfg
+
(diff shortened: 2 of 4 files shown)


hardys:
I wonder can we configure OpenShift so that the API only binds to the nic IP, then we can bind the normal port to the VIP with the LB?

I'm thinking using a non-standard port is liable to cause client confusion, and IIRC we achieved this same setup in TripleO by being careful with the service bind IPs (e.g use a specific IP not 0.0.0.0)

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/131/

yboaron:
@hardys 
I agree, if it's possible to configure Openshift to listen to port 6443 on specific IP (I'll send this question to openshift experts)  we should go with this approach.
Btw, with this approach, the LB should be bind to API VIP:6443, meaning we should create the LB only at the master machine that 'holds'  API VIP address.

hardys:
> Btw, with this approach, the LB should be bind to API VIP:6443, meaning we should create the LB only at the master machine that 'holds' API VIP address.

AFAIK we should still run haproxy on all masters, but have the frontend do a non-local bind to the VIP IP, that's managed by keepalived.

@celebdor may have other comments on this approach, but I was assuming we'd align pretty closely with how we already do this for OpenStack, which is as I described above.

celebdor:
I wonder if we could just do it with iptables load balancing the way kube-proxy does to avoid port problems.

yboaron:
Since current OCP installer code doesn't support configuring to which IP address/Network interface bind the API port (seems that this capability won't be added in the near future),
the OCP API LB should bind to port != OCP API port(6443).
    
Maybe we can use both Iptables and Haproxy (we"ll have member's health check), the suggested solution composed of :
    1. An IPTable rule that redirects API-VIP port(6443) to OCP API-LB port(7443)
    2. Haproxy LB

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/170/

machineconfig/master/spec/config/storage/files/haproxy.sh L3Zhci91c3Jsb2NhbC9iaW4vaGFwcm94eS5zaAo=
```diff
@@ -0,0 +1,110 @@
...
+do
+    ips=$(dig +noall +answer $item  | awk '{print $NF}')
+    if [[ "$?" -eq 0 && "${#ips[@]}" -ne 0 ]]; then
+        sudo echo "   server "$item" "$ips":"$OCP_API_PORT" weight 1 check inter 3s fall 3 rise 3" >>  $HOST_CFG_PATH/haproxy.cfg
+    fi
+done
+
+}
+
+function does_lb_topology_changed {
```
celebdor:
```suggestion
function has_master_api_lb_topology_changed {
```

machineconfig/master/spec/config/storage/files/haproxy.sh L3Zhci91c3Jsb2NhbC9iaW4vaGFwcm94eS5zaAo=
```diff
@@ -0,0 +1,110 @@
...
+etcd_members=$(dig +noall +answer -t SRV _etcd-server-ssl._tcp.$CLUSTER_DOMAIN   | awk '{print $NF}')
+for item in $etcd_members
+do
+    ips=$(dig +noall +answer $item | awk '{print $NF}')
+    haproxy_cfg_ip=($(grep $item $HOST_CFG_PATH/haproxy.cfg | awk '{print $3}' | sed 's/:.*//'))
+    if [[ "$?" -eq 0 && "${#ips[@]}" -ne 0 ]]; then
+       if [ "$haproxy_cfg_ip" != "$ips" ]; then
```
celebdor: shouldn't you be sorting before comparing? Otherwise if the answer gives the same answer but in a different order, it will give a false positive.

machineconfig/master/spec/config/storage/files/haproxy.sh L3Zhci91c3Jsb2NhbC9iaW4vaGFwcm94eS5zaAo=
```diff
@@ -0,0 +1,110 @@
...
+sudo podman run \
+            -d \
+            --name masters-haproxy \
+            --network=host \
+            --cap-add=NET_ADMIN \
+            -v $HOST_CFG_PATH:/usr/local/etc/haproxy:ro \
+            $HAPROXY_IMAGE
+
+# redirect traffic to OCP API LB
+sudo iptables -t nat -I PREROUTING --src 0/0 --dst $API_VIP -p tcp --dport $OCP_API_PORT -j REDIRECT --to-ports $LB_OCP_API_PORT -m comment --comment "OCP_API_LB_REDIRECT"
```
celebdor: Shouldn't you check the rule is already there and if that's the case, replace it?

celebdor: OH, I see you made a stop function

derekhiggins:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/204/

machineconfig/master/spec/config/storage/files/haproxy.sh L3Zhci91c3Jsb2NhbC9iaW4vaGFwcm94eS5zaAo=
yboaron: @celebdor 
I trigger reconfigure haproxy only if :: 
1. etcd-x IP (resolved by DNS) != etcd-x ip in haproxy cfg
Or
2. Can't resolve etcd-x IP in DNS but etcd-x IP appears in haproxy-cfg
I think that it should be fine, am I missing something?

derekhiggins:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/207/

machineconfig/master/spec/config/storage/files/haproxy.cfg.template L2V0Yy9oYXByb3h5L2hhcHJveHkuY2ZnLnRlbXBsYXRlCg==
```diff
@@ -0,0 +1,22 @@
+global
```
celebdor: I think it would be good to add stats:

```
listen haproxy_stats :50000
stats enable
stats uri /
no log
```

I'm fine with doing it in a separate PR

celebdor: Another conf example for stats https://www.datadoghq.com/blog/how-to-collect-haproxy-metrics/

PR #126, opened 2019-03-07
Allowing access to tftp server.

achuzhoy:
We need to have port 69 open for reaching tftp
server during pxebooting. It's not always open.

### DIFF
--- a/02_configure_host.sh
+++ b/02_configure_host.sh
@@ -51,6 +51,11 @@ if ! sudo iptables -C INPUT -p tcp --dport 80 -j ACCEPT ; then
     sudo iptables -I INPUT -p tcp --dport 80 -j ACCEPT
 fi
 
+#Allow access to tftp server for pxeboot
+if ! sudo iptables -C INPUT -p udp --dport 69 -j ACCEPT ; then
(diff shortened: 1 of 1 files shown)


hardys:
Looks good, thanks - can you confirm if this is only needed when testing with real baremetal nodes?

If so we can perhaps look at adding a conditional, since we probably don't want this (or the other pxe related ports) open on a box that's dedicated to VM testing (only open on the bridge from Ironic to the VMs)

achuzhoy:
It solely depends on the OS where we run the scripts on, so we can't predict.
I didn't see it on VM based setup - simply because the port was initially open (either by kickstart or initial automated configuration).
On a manually installed BM setup this port was closed and caused issues.

derekhiggins:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/132/

hardys:
My point is this unconditionally opens the port, including from outside the host, which is not needed in the VM case, we can just open it for the bridge device and not allow external access.

Anyway, I guess we can fix that in a follow-up, as this isn't the only place to do that.

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/134/

PR #130, opened 2019-03-07
Use `oc wait` to wait for condition instead of running oc commands

vrutkovs:
There is no need to repeat oc commands - it has `oc wait` subcommands which pauses until a particular `.status.condition` is set to True.

There is no `oc wait` for events, so this would run `oc get events` in a loop and look for `bootstrap-complete` event instead. Loop is limited to 60 attempts with 10 sec pause so bootstrapping is expected to complete in 10+ minutes.

If the cluster didn't start the script would output:
1. bootstrap.service logs
2. clusterversion operator status
3. operator status
4. List of failing pods

### DIFF
--- a/07_deploy_masters.sh
+++ b/07_deploy_masters.sh
@@ -71,10 +71,7 @@ for i in $(seq 0 2); do
   oc --config ocp/auth/kubeconfig wait nodes/master-$i --for condition=ready --timeout=600s
 done
 
-# Wait for bootstrap to complete
-while ! oc --config ocp/auth/kubeconfig get events -n kube-system --no-headers -o wide | grep -q bootstrap-complete; do
-  sleep 10
-done
+wait_for_bootstrap_event
 
 # disable NoSchedule taints for masters until we have workers deployed
 for num in 0 1 2; do
--- a/07_deploy_masters.sh
+++ b/07_deploy_masters.sh
@@ -58,15 +58,20 @@ done
 echo "Master nodes up, you can ssh to the following IPs with core@<IP>"
 sudo virsh net-dhcp-leases baremetal
 
-while [[ ! $(timeout -k 9 5 $SSH "core@api.${CLUSTER_NAME}.${BASE_DOMAIN}" hostname) =~ master- ]]; do
-  echo "Waiting for the master API to become ready..."
+# Wait for nodes to appear and become ready
+until oc --config ocp/auth/kubeconfig get nodes; do sleep 5; done
+NUM_NODES=$(oc --config ocp/auth/kubeconfig get nodes --no-headers | wc -l)
+while [ "$NUM_NODES" -ne 3 ]; do
   sleep 10
+  NUM_NODES=$(oc --config ocp/auth/kubeconfig get nodes --no-headers | wc -l)
+done
+for i in $(seq 0 2); do
+  oc --config ocp/auth/kubeconfig wait nodes/master-$i --for condition=ready --timeout=600s
 done
 
-NODES_ACTIVE=$(oc --config ocp/auth/kubeconfig get nodes | grep "master-[0-2] *Ready" | wc -l)
-while [ "$NODES_ACTIVE" -ne 3 ]; do
+# Wait for bootstrap to complete
+while ! oc --config ocp/auth/kubeconfig get events -n kube-system --no-headers -o wide | grep -q bootstrap-complete; do
   sleep 10
-  NODES_ACTIVE=$(oc --config ocp/auth/kubeconfig get nodes | grep "master-[0-2] *Ready" | wc -l)
 done
 
 # disable NoSchedule taints for masters until we have workers deployed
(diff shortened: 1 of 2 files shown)


derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/141/

07_deploy_masters.sh
```diff
@@ -69,9 +69,15 @@ done
 echo "Master nodes up, you can ssh to the following IPs with core@<IP>"
 sudo virsh net-dhcp-leases baremetal
 
+# Wait for nodes to become ready
+NUM_LEASES=$(oc --config ocp/auth/kubeconfig get nodes | wc -l)
```
hardys: We should probably rename the variable to NUM_NODES or something?

hardys:
Nice, thanks for this - I wasn't yet aware of the wait subcommand :)

LGTM but had one nit re variable naming which would be nice to fix.

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/143/

07_deploy_masters.sh
vrutkovs: Done

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/145/

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/147/

derekhiggins:
LGTM but can we squash it up a bit before merging, there are a bunch of patches fixing things from previous patches etc.. maybe thats why you kept it as a WIP but just making sure

vrutkovs:
Sure, the commits will be squashed, although it doesn't actually wait for bootstrap to complete now - it breaks when VIP switches to first master. Not sure how to add retries there

Perhaps I should leave just `oc wait` for nodes then?

hardys:
> Perhaps I should leave just `oc wait` for nodes then?

Sounds reasonable - leave the current logic until the check for nodes, as by that point we've waited for the VIP failover to happen, via the ssh/hostname check loop?

FWIW I don't see any problem with adding the CVO wait even if that does make the deploy take longer - as we've seen with the kubevirt stuff unexpected things happen when the cluster isn't fully up, so we may as well wait for it to be fully deployed before running any further scripts?

That said we probably do need to thing about the earliest point that workers can be deployed for the final implementation - I'm hoping we can figure out just adding manifests for all the additional pieces in due course, but for now serializing via scripts seems to be a reasonable way to enable further testing?

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/171/

07_deploy_masters.sh
```diff
@@ -89,6 +89,12 @@ for i in $(seq 0 2); do
...
 
-# Wait for clusterversion operator to report availability
-oc --config ocp/auth/kubeconfig wait clusterversion/version --for condition=available --timeout=600s
+# Wait for bootstrap to complete
+while read line; do
+  echo $line;
+  if [ "$(echo ${line} | awk '{print $4}')" == "bootstrap-complete" ]; then
+    break
+  fi
+done < <(oc --config ocp/auth/kubeconfig get events -n kube-system --no-headers -w -o wide)
```
hardys: You mentioned on slack needing to retry this, could you just omit the `-w` option and call the get events in a loop with grep?

e.g:

```
while ! oc --config ocp/auth/kubeconfig get events -n kube-system --no-headers -o wide | grep -q bootstrap-complete; do
  sleep 10
done

If we need to tolerate failure we can just || true in the conditional?
```

vrutkovs: Right, lets try that

vrutkovs: That worked. Squashed commits

derekhiggins:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/192/

derekhiggins:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/193/

derekhiggins:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/203/

vrutkovs:
@derekhiggins could the jenkins build timeout be extended? Now the script actually waits for bootstrap to complete, which is rarely less than an hour.

I'll merge that once CI runs would not be false-positive

hardys:
@derekhiggins I'm wondering what to do about the CI test failure - it does take a while in my environment to get the bootstrap-complete, but way less that 60mins what sort of env is CI running on?

derekhiggins:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/237/

derekhiggins:
Build ABORTED, see build http://10.8.144.11:8080/job/dev-tools/273/

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/282/

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/294/

PR #131, opened 2019-03-08
Fix ocp_cleanup to work for root user

vrutkovs:
Some home dirs may be outside of `/home` (for instance root), so lets use `$HOME` instead


derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/142/

PR #133, opened 2019-03-08
Define the storage pool requires sudo

e-minguez:


02_configure_host.sh
```diff
@@ -29,16 +29,16 @@ sudo usermod -a -G "libvirt" $USER
 # As per https://github.com/openshift/installer/blob/master/docs/dev/libvirt-howto.md#configure-default-libvirt-storage-pool
 # Usually virt-manager/virt-install creates this: https://www.redhat.com/archives/libvir-list/2008-August/msg00179.html
 if ! virsh pool-uuid default > /dev/null 2>&1 ; then
-    virsh pool-define /dev/stdin <<EOF
+    sudo virsh pool-define /dev/stdin <<EOF
```
hardys: Can you confirm you've tried this since #110 landed?  Also if it's a fresh install please ensure you logged out and back in again after the user was added to the libvirt group.

I'm not sure if the default pool is treated as special here though, cc @markmc @derekhiggins

e-minguez: I did steps 01 and 02 on Tuesday so I'm not sure TBH. Let me try again :)

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/146/

02_configure_host.sh
derekhiggins: Also there is some funny business going on here that I haven't quite but my finger on,  on some setups you might already have a "images" storage pool and need to remove it first, I also had to restart libvirtd when doing this.

e-minguez:
Reinstalled CentOS in the baremetal server and executed `make` succesfully. Closing this.
Thanks @hardys !

PR #140, opened 2019-03-08
Ensuring the file to edit exists.

achuzhoy:
This PR comes to apply a check prior to editing a file that may not exist,
which would cause the following error:
sed: can't read /home/stack/.ssh/known_hosts: No such file or directory


derekhiggins:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/159/

PR #143, opened 2019-03-08
Remove NoSchedule taints from masters

booxter:
Since we don't deploy workers yet, to do anything meaningful with the cluster
right now requires untainting master nodes for scheduling pods. This change
makes it part of automated deployment to save a manual step.

### DIFF
--- a/07_deploy_masters.sh
+++ b/07_deploy_masters.sh
@@ -79,5 +79,11 @@ while [ "$NODES_ACTIVE" -ne 3 ]; do
   sleep 10
   NODES_ACTIVE=$(oc --config ocp/auth/kubeconfig get nodes | grep "master-[0-2] *Ready" | wc -l)
 done
+
+# disable NoSchedule taints for masters until we have workers deployed
+for num in 0 1 2; do
+  oc --kubeconfig ocp/auth/kubeconfig adm taint nodes master-${num} node-role.kubernetes.io/master:NoSchedule-
+done
+
 oc --config ocp/auth/kubeconfig get nodes
 echo "Cluster up, you can interact with it via oc --config ocp/auth/kubeconfig <command>"


derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/161/

derekhiggins:
CI failed, see logs

```
+ for num in 0 1 2
*** PULL_SECRET ***
Error: unknown command "taint" for "oc"

Did you mean this?
	wait

Run 'oc --help' for usage.
```
I guess I need "set -o pipefail" in the script so it doesn't report SUCCESS 
Unfortunately, you can't see the command it ran, I've a PR up to fix that
https://github.com/metalkube/dev-scripts/pull/136

booxter:
Eh, that's me ASSUMING that `oc` has same commands as `kubectl` that I used here manually.

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/168/

07_deploy_masters.sh
```diff
@@ -79,5 +79,11 @@ while [ "$NODES_ACTIVE" -ne 3 ]; do
   sleep 10
   NODES_ACTIVE=$(oc --config ocp/auth/kubeconfig get nodes | grep "master-[0-2] *Ready" | wc -l)
 done
+
+# disable NoSchedule taints for masters until we have workers deployed
+for num in 0 1 2; do
+  oc --kubeconfig ocp/auth/kubeconfig adm taint nodes master-${num} node-role.kubernetes.io/master:NoSchedule-
+done
+
```
hardys: Will this be persistent, or will the machine config mean the taint gets reapplied?

Reason for asking is the openshift/installer docs describe doing this by adding a manifest with a modified MachineConfig that removes the taint:

https://github.com/openshift/installer/blob/master/docs/user/customization.md#control-plane-with-no-taints

Perhaps we should do that instead, based on the manifest changes proposed in https://github.com/metalkube/dev-scripts/pull/127 ?

russellb: While ideally we can follow docs, I got advice from @eparis that we should not use that approach.

I think the proposed approach is fine for now, until OpenShift defines another process for us that they recommend we adopt.

booxter: I know little about the right approach, but being lazy I will assume for now that Russell knows better. ;) Let me know if we indeed should switch to modified manifests.

@russellb do you have details on why we should not use manifests yet?

imain:
Yeah need this.  +1

07_deploy_masters.sh
eparis: editting the systemd unit file which is used to launch the kubelet to prevent it from setting the taint will come back and bite us at some indeterminate time in the future when that systemd unit file needs to be changed by OpenShift, but your custom version will still be there overwriting what we need to change. I really need to get that out of the docs.

The taint will only re-appear if the node object is deleted from the API server (and re-created by restarting the kubelet on that machine).

You can save yourself a for loop by doing:
`oc adm taint node -l node-role.kubernetes.io/master node-role.kubernetes.io/master:NoSchedule-`

derekhiggins: ok, sound like we're good to go with this @booxter wanna remove the forloop as suggested or merge as it?

booxter: I would remove but since it's merge already I don't think we should be bothered much.

PR #144, opened 2019-03-11
Add retry option to get_images script.

yprokule:
Add retry to curl calls to avoid sporadic network issues while
downloading images, for e.g.:
++ curl --insecure --compressed -L https://images.rdoproject.org/master/rdo_trunk/current-tripleo/ironic-python-agent.tar
++ tar -xf -
  0     0    0     0    0     0      0      0 --:--:--  0:02:07 --:--:--     0
  curl: (7) Failed connect to images.rdoproject.org:443; Connection timed out
  tar: This does not look like a tar archive
  tar: Exiting with failure status due to previous errors

### DIFF
--- a/get_images.sh
+++ b/get_images.sh
@@ -4,7 +4,7 @@ set -xe
 source common.sh
 
 if [ ! -f "$RHCOS_IMAGE_FILENAME" ]; then
-    curl --insecure --compressed -L -o "${RHCOS_IMAGE_FILENAME}" "${RHCOS_IMAGE_URL}/${RHCOS_IMAGE_VERSION}/${RHCOS_IMAGE_FILENAME}".gz
+    curl --insecure --compressed --retry 3 -L -o "${RHCOS_IMAGE_FILENAME}" "${RHCOS_IMAGE_URL}/${RHCOS_IMAGE_VERSION}/${RHCOS_IMAGE_FILENAME}".gz
 fi
 
 mkdir -p "$IRONIC_DATA_DIR/html/images"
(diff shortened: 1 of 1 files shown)


derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/172/

hardys:
Interested to know, does this happen spuriously, or was it due to a temporary RDO outage?

yprokule:
> Interested to know, does this happen spuriously, or was it due to a temporary RDO outage?

@hardys - difficult to say if RDO outage is the culprit. I hit it few times last week and re-running `make ironic` immediately after failure was passing.

hardys:
> > Interested to know, does this happen spuriously, or was it due to a temporary RDO outage?
> 
> @hardys - difficult to say if RDO outage is the culprit. I hit it few times last week and re-running `make ironic` immediately after failure was passing.

Ok weird, I'm not opposed to the retry but it'd be good to ensure we're not just working around a temporary outage - @derekhiggins maybe you can confirm if we ever did similar retries in tripleo-ci?  IIRC we didn't and just dealt with the occasional outages (and also mirrored things locally where possible).

derekhiggins:
> > > Interested to know, does this happen spuriously, or was it due to a temporary RDO outage?
> > 
> > 
> > @hardys - difficult to say if RDO outage is the culprit. I hit it few times last week and re-running `make ironic` immediately after failure was passing.
> 
> Ok weird, I'm not opposed to the retry but it'd be good to ensure we're not just working around a temporary outage - @derekhiggins maybe you can confirm if we ever did similar retries in tripleo-ci? IIRC we didn't and just dealt with the occasional outages (and also mirrored things locally where possible).

iirc we used to download this ipa image to a mirror server and each CI job got it from there. I forget what happened for users running tripleo-ci outside of CI(maybe they were building the image?) there was definitely times when the rdo servers were down. 

Having said that this isn't a problem any user should hit several times as the image should be downloaded once and then reused so maybe isn't a major problem? @yprokule were you deleting the cached file between run of dev-scripts? 
If we do decide to go with a retry then we should probably download to file and then untar as piping directly to tar is not going to work if it has already gotten data from a download that failed and then restarted.

yprokule:
> If we do decide to go with a retry then we should probably download to file and then untar as piping directly to tar is not going to work if it has already gotten data from a download that failed and then restarted.

@derekhiggins it's indeed reused when it's present. It happened when I run it on newly provisioned bm

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/184/

hardys:
@yprokule Not sure how to proceed with this - have you seen the need for retries again, or was it a one-off outage?

If the latter I'm inclined to close this without merging, as if RDO things are down other stuff will probably fail anyway e.g repos?

yprokule:
@hardys not this week. I'll reopen if see it's recurring. Thanks

PR #145, opened 2019-03-11
Revert "Switch to kube-system namespace for kubevirt manifests"

hardys:
Subsequent discussion indicates this isn't the right approach, and that
we instead need to fix the manifest ordering, and/or use a temporary
script that runs after the cluster is fully deployed.

This reverts commit a52d2b31ae239becee39f2c882601d55801efd19.

### DIFF
--- a/manifests/10_cnv_kubevirt_op.yaml
+++ b/manifests/10_cnv_kubevirt_op.yaml
@@ -1,4 +1,11 @@
 ---
+apiVersion: v1
+kind: Namespace
+metadata:
+  labels:
+    kubevirt.io: ""
+  name: kubevirt
+---
 apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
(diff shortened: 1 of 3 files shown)


derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/173/

booxter:
@hardys which ordering issue do you refer to? I think currently the main problem is that cluster deployment is abrupt by machine api operator baremetal mode not supported. Assuming this is fixed, which other (ordering?) issues do you envision?

hardys:
> @hardys which ordering issue do you refer to? I think currently the main problem is that cluster deployment is abrupt by machine api operator baremetal mode not supported. Assuming this is fixed, which other (ordering?) issues do you envision?

I just meant right now this doesn't work because we try to apply the manifests before the cluster is fully up - I wasn't 100% sure if by fixing the machine api operator baremetal type we can then be sure these manifests don't get applied until the cluster is fully deployed?

If we're confident that's the only fix needed then great, but I'm not super familiar with how the manifests are applied yet, so didn't want to assume it'd just work (again! ;)

booxter:
@hardys agreed, perhaps it's my naive expectation that installer won't apply additional manifests before cluster is ready for that. ;)

PR #152, opened 2019-03-12
Reordered 01 and added some comments

e-minguez:


01_install_requirements.sh
```diff
@@ -42,27 +47,56 @@ popd
 # Needed to get a recent python-virtualbmc package
 sudo tripleo-repos current-tripleo
 
-# Work around a conflict with a newer zeromq from epel
-if ! grep -q zeromq /etc/yum.repos.d/epel.repo; then
-  sudo sed -i '/enabled=1/a exclude=zeromq*' /etc/yum.repos.d/epel.repo
-fi
-sudo yum -y update
```
hardys: This yum update is significant - there are some packages which are newer in the tripleo repos, so we should do a yum update after configuring all the repos, AFAICS you just removed it, so we only yum update at the start of the script?

e-minguez: Added back the `yum update` after configuring the tripleo repos and a comment on why is needed.
Thanks!

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/185/

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/186/

PR #156, opened 2019-03-13
Test for existence of metalkube directories

derekhiggins:
These arn't files so the check never passed


derekhiggins:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/200/

PR #162, opened 2019-03-13
Removing the bootstrap VM as part of cleanup

achuzhoy:
The bootstap VM was remaining after 'make clean'.
This patch deals with its removal.

### DIFF
--- a/host_cleanup.sh
+++ b/host_cleanup.sh
@@ -24,6 +24,16 @@ ANSIBLE_FORCE_COLOR=true ansible-playbook \
     -b -vvv tripleo-quickstart-config/metalkube-teardown-playbook.yml
 
 sudo rm -rf /etc/NetworkManager/dnsmasq.d/openshift.conf /etc/NetworkManager/conf.d/dnsmasq.conf
+
+#Remove the bootstrap VM
+sudo virsh list --all --name|grep -q bootstrap && (
+    for i in $(sudo virsh list --all --name|grep bootstrap); do
+          sudo virsh destroy $i 2>/dev/null;
+          sudo virsh undefine $i 2>/dev/null;
+    done
+)
+
+#Remove the networks
 sudo virsh net-destroy baremetal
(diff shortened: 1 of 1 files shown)


host_cleanup.sh
```diff
@@ -24,6 +24,16 @@ ANSIBLE_FORCE_COLOR=true ansible-playbook \
     -b -vvv tripleo-quickstart-config/metalkube-teardown-playbook.yml
 
 sudo rm -rf /etc/NetworkManager/dnsmasq.d/openshift.conf /etc/NetworkManager/conf.d/dnsmasq.conf
+
+#Remove the bootstrap VM
+sudo virsh list --all --name|grep -q bootstrap && (
+    for i in $(sudo virsh list --all --name|grep bootstrap); do
+          sudo virsh destroy $i 2>/dev/null;
+          sudo virsh undefine $i 2>/dev/null;
```
jamescassell: Is there any storage to deal with here? Would you want `--remove-all-storage`?

derekhiggins:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/214/

markmc:
`kni-install destroy bootstrap` in ocp_cleanup.sh is supposed to take care of this. Would be good to know why you're not seeing that happening? Perhaps the logs show something?

hardys:
Yeah agreed this shouldn't be needed, so we should understand why ocp_cleanup.sh is not doing the job here.

hardys:
Ok so AFAICS this is failing because with a clean environment kni-installer isn't being built (since the repo move perhaps?), then ocp_cleanup.sh ignores that failure because we removed `set -e`, and it removes the ocp dir, so we can never retry :(

Will dig into it..

hardys:
Ok I think this was a temporary breakage related to the repo move, for some reason kni-installer didn't get built and we ignored the error trying to run it - @achuzhoy if you agree we can close this?

PR #165, opened 2019-03-14
 Remove if-name for etc discovery

markmc:
The if-name for etc discovery is no longer required because of:

https://github.com/openshift/machine-config-operator/pull/547

### DIFF
--- a/06_deploy_bootstrap_vm.sh
+++ b/06_deploy_bootstrap_vm.sh
@@ -41,9 +41,6 @@ $SSH -o ConnectionAttempts=500 core@$IP id
 # Create a master_nodes.json file
 jq '.nodes[0:3] | {nodes: .}' "${NODES_FILE}" | tee "${MASTER_NODES_FILE}"
 
-# Fix etcd discovery on bootstrap
-add_if_name_to_etcd_discovery "$IP" "eth1"
-
 # Generate "dynamic" ignition patches
 machineconfig_generate_patches "master"
 # Apply patches to masters
--- a/utils.sh
+++ b/utils.sh
@@ -147,36 +147,6 @@ function apply_yaml_patches() {
     yq -y '.' < "${wd}/${kind}.json" | sudo tee "$target" | sed -e 's/.*auth.*/***PULL_SECRET***/g'
 }
 
-# Add if-name param to etcd discovery container on masters
-function add_if_name_to_etcd_discovery() {
-    local ip
-    local if_name
-    local wd
-    local master_config
-
-    ip="$1"
-    if_name="$2"
-    wd=$(mktemp -d)
-
-    # Find master machine config name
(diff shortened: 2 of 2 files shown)


ocp_install_env.sh
```diff
@@ -10,7 +10,7 @@ export SSH_PUB_KEY="`cat $HOME/.ssh/id_rsa.pub`"
 export SSH_PRIV_KEY="$HOME/.ssh/id_rsa"
 
 # Temporary workaround pending merge of https://github.com/openshift/machine-api-operator/pull/246
-export OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE="quay.io/vrutkovs/kni-release-mar13:release"
+export OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE="registry.svc.ci.openshift.org/ci-op-r5vsk4k4/release:latest"
```
vrutkovs:
```suggestion
export OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE="quay.io/vrutkovs/kni-release:mar14"
```
The same release mirrored to quay, most likely its correct and these images won't be GC'ed

06_deploy_bootstrap_vm.sh
```diff
@@ -42,7 +42,7 @@ $SSH -o ConnectionAttempts=500 core@$IP id
 jq '.nodes[0:3] | {nodes: .}' "${NODES_FILE}" | tee "${MASTER_NODES_FILE}"
 
 # Fix etcd discovery on bootstrap
-add_if_name_to_etcd_discovery "$IP" "eth1"
+# add_if_name_to_etcd_discovery "$IP" "eth1"
```
hardys: Maybe we should just remove this and the associated function, or if we're leaving it commented add a reference to the mco patch?

hardys:
This works for me, not yet re-tested with the mirror from @vrutkovs

06_deploy_bootstrap_vm.sh
vrutkovs: Lets comment it first, we don't need to wait for MCO to lay files anymore afaik, so it not urgent

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/227/

ocp_install_env.sh
markmc: @hardys is going to submit a separate PR with that

06_deploy_bootstrap_vm.sh
markmc: There's still the patch_node_ignition() stuff

Update the patch to remove add_if_name_to_etcd_discovery() altogether. If we need it back, we can always get it from history

derekhiggins:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/229/

PR #169, opened 2019-03-14
ipmi_port is redundant on BM setup

achuzhoy:
This patch makes it possible to ignore ipmi_port,
that's not needed for BM nodes.

### DIFF
--- a/utils.sh
+++ b/utils.sh
@@ -270,7 +270,10 @@ function master_node_to_tf() {
     local_gb=$(master_node_val ${master_idx} "properties.local_gb")
     cpu_arch=$(master_node_val ${master_idx} "properties.cpu_arch")
 
-    ipmi_port=$(master_node_val ${master_idx} "driver_info.ipmi_port")
+    ipmi_port=$(master_node_val ${master_idx} "driver_info.ipmi_port // \"\"")
+    if [ -n "$ipmi_port" ]; then
+        ipmi_port="\"ipmi_port\"=      \"${ipmi_port}\""
+    fi
     ipmi_username=$(master_node_val ${master_idx} "driver_info.ipmi_username")
     ipmi_password=$(master_node_val ${master_idx} "driver_info.ipmi_password")
     ipmi_address=$(master_node_val ${master_idx} "driver_info.ipmi_address")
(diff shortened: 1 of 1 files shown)


utils.sh
```diff
@@ -270,7 +270,12 @@ function master_node_to_tf() {
     local_gb=$(master_node_val ${master_idx} "properties.local_gb")
     cpu_arch=$(master_node_val ${master_idx} "properties.cpu_arch")
 
-    ipmi_port=$(master_node_val ${master_idx} "driver_info.ipmi_port")
```
derekhiggins: Try this, if it works you can remove the other lines
`ipmi_port=$(master_node_val ${master_idx} "driver_info.ipmi_port // \"\"")`

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/239/

utils.sh
achuzhoy: This returns as empty value.
The thing is that later in this file we have:

  driver_info {
    "ipmi_port" = 

The value should be some valid port and not empty/null.

I replace the entire line with '${ipmi_port}'

derekhiggins: Sorry, I didn't read the rest of the code correctly, I was trying to avoid using $NODES_PLATFORM as we could have BM nodes with a port (although its unlikely) or libvirt nodes without a port. How about something like this

```
ipmi_port=$(master_node_val ${master_idx} "driver_info.ipmi_port // \"\"")
if [ -n "$ipmi_port" ]; then
    ipmi_port="\"ipmi_port\"=      \"${ipmi_port}\""
fi

```

derekhiggins:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/245/

hardys:
@derekhiggins Hey sounds like we need this to make progress on baremetal testing, did you have any further comments/changes here or should we go ahead and merge it?

derekhiggins:
> @derekhiggins Hey sounds like we need this to make progress on baremetal testing, did you have any further comments/changes here or should we go ahead and merge it?

lgtm, just wondering why the CI failed, but I guess its ok as masters would have been deployed at this stage, so lets merge.
> Mar 15 14:05:41 |  Unable to connect to the server: EOF

PR #174, opened 2019-03-15
Add script to deploy extra manifests

hardys:
We expect this to be handled via kni-installer but there were
issues with that ref #127 so lets just apply them via a script
for now, which is probably easier for test/dev workflows anyway.

### DIFF
--- a/09_deploy_manifests.sh
+++ b/09_deploy_manifests.sh
@@ -1,0 +1,13 @@
+#!/usr/bin/bash
+
+# FIXME - what are the prerequisites we need to wait for here, is waiting for 
+# bootstrap complete enough?
+
+# These will ultimately be deployed via kni-installer to the bootstrap node
+# where they will then be applied to the cluster, but for now we do it
+# manually
+cd manifests
+for manifest in $(ls -1 *.yaml | sort -h); do
+  oc --as system:admin --config ../ocp/auth/kubeconfig apply -f ${manifest}
+  echo "manifests/${manifest} applied"
+done


hardys:
cc @booxter this is basically just what's in the README but seems to work for me, any thoughts on the FIXME re what we should check/dependencies prior to running this?

hardys:
Also wondering what sanity checks we could add to the script to prove things worked as expected

booxter:
@hardys the last time I checked `dev-scripts` were not bringing up a cluster capable of applying `kubevirt` manifests creating resources inside `kubevirt` namespace. (But if I replaced the namespace to `kube-system` then it worked.) Have you checked that this script successfully applying manifests already in the directory?

hardys:
@booxter yes with the image override and various other fixes I believe they can be applied, but can you perhaps provide a list of sanity checks I can try to prove that and paste here (I did check and e.g the cni pods you added were running and the CR's created etc, but I'm not sure how to check if kubevirt is actually functional yet).

I waited a while after the `make` though until the CVO failed but most other stuff is up - what I'm not clear on is if the wait for bootstrap-success in #130 is sufficient or if we need to wait for other things to be up before applying these?

If you have a dev env perhaps you can also re-test and see how things look now?  Thanks! :)

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/247/

booxter:
@hardys (sorry for late reply)

Just make sure that:

1. services like `virt-handler` and `virt-api` and `virt-controller` are up.
2. you can `oc get vmis` and `oc get vms` without errors about unknown type. (Note that due to a bug in client-go it may not work on the very first call after deployment. Just call again.)

This should give enough of sanity check.

OVS-CNI, SR-IOV components are up because they use `kube-system` namespace. KubeVirt is different and uses `kubevirt` namespace, which had its issues due to deployment issues of KNI when I tried it.

hardys:
@booxter thanks for the reply, AFAICS it's working:

```
[shardy@tripleodev2 dev-scripts]$ oc --as system:admin --config ocp/auth/kubeconfig get pods --all-namespaces | grep -e kubevirt -e cni
kube-system                                             kube-cni-plugins-amd64-7glwv                                      1/1     Running            0          4m3s
kube-system                                             kube-cni-plugins-amd64-bflcr                                      0/1     Pending            0          4m3s
kube-system                                             kube-cni-plugins-amd64-txfnm                                      1/1     Running            0          4m3s
kube-system                                             kube-ovs-cni-amd64-fw7vs                                          0/2     Pending            0          3m59s
kube-system                                             kube-ovs-cni-amd64-jq77m                                          0/2     Pending            0          3m59s
kube-system                                             kube-ovs-cni-amd64-zppdm                                          0/2     Pending            0          3m59s
kube-system                                             kube-sriov-cni-amd64-c7mnt                                        0/1     Pending            0          4m1s
kube-system                                             kube-sriov-cni-amd64-dm52l                                        0/1     Pending            0          4m1s
kube-system                                             kube-sriov-cni-amd64-qhchd                                        0/1     Pending            0          4m1s
kubevirt                                                virt-api-7dcc9874d8-j88qb                                         1/1     Running            0          3m
kubevirt                                                virt-api-7dcc9874d8-r6sch                                         1/1     Running            0          3m
kubevirt                                                virt-controller-7c5994776b-h5cd6                                  1/1     Running            0          3m
kubevirt                                                virt-controller-7c5994776b-jnwvr                                  1/1     Running            0          3m
kubevirt                                                virt-handler-cjvvw                                                1/1     Running            0          3m
kubevirt                                                virt-handler-j6p56                                                1/1     Running            0          3m
kubevirt                                                virt-handler-mfb27                                                1/1     Running            0          3m
kubevirt                                                virt-operator-667649fb99-htdbp                                    1/1     Running            0          4m3s

```

Then launching a VM using `https://raw.githubusercontent.com/kubevirt/demo/master/manifests/vm.yaml` does something:

```
[shardy@tripleodev2 dev-scripts]$ oc --as system:admin --config ocp/auth/kubeconfig get vms
NAME     AGE   RUNNING   VOLUME
testvm   94s   false 
```

This is on an environment with VM masters though, so I'm not sure if it's possible to run the testvm?

Anyway, seems the files are getting applied (I did wait a while before running the script though), so probably we should merge this and iterate on any further debugging as a follow-up?

booxter:
@hardys yes this seems like it works.

To validate end-to-end, you would also need to:

1. enable software emulation: https://kubevirt.io/user-guide/docs/latest/administration/intro.html#software-emulation (before applying kubevirt manifests; OR just restart virt-*).
2. download virtctl: https://kubevirt.io/user-guide/docs/latest/administration/intro.html#client-side-virtctl-deployment
3. ./virtctl start testvm
4. ./virtctl console testvm

a virt-launcher pod should pop up into Running state.

But I don't think it's needed in this PR, it's clear from your output that kubevirt namespace is no longer affected since we can create resources there.

hardys:
Ok thanks @booxter lets rebase/merge this and I'll follow-up with some e2e testing (maybe as a utility script since we might not always want to run it)

hardys:
@booxter yeah my plan was to leave it out of the Makefile for now, since not everyone needs to test this atm, and for CI we ideally need some minimal e2e test if we add it to the default target.

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/281/

hardys:
Ok I think this is good to merge now, added @derekhiggins and @vrutkovs as they can perhaps review/merge then we'll iterate on the e2e test in a future patch

vrutkovs:
Merging this. CNV are not necessary for e2e test yet, so I guess it would be best to leave it out of CI for now

PR #175, opened 2019-03-15
Deploy baremetal-operator in the openshift-machine-api namespace.

sadasu:
Deploy baremetal-operator in the openshift-machine-api namespace/project.
Add cleanup logic for localized changes introduced by the above changes.
Add deploy of baremetal-operator as the last step of make and update
documentation.

### DIFF
--- a/08_deploy_bmo.sh
+++ b/08_deploy_bmo.sh
@@ -1,18 +1,18 @@
 #!/usr/bin/bash
 
+#source common.sh
 eval "$(go env)"
 
 # Get the latest bits for baremetal-operator
 export BMOPATH="$GOPATH/src/github.com/metalkube/baremetal-operator"
 
-source common.sh
-
-oc new-project bmo-project
+# Make a local copy of the baremetal-operator code to make changes
+cp -r $BMOPATH/deploy ocp/.
+sed -i 's/bmo-project/openshift-machine-api/g' ocp/deploy/role_binding.yaml
 
 # Start deploying on the new cluster
-
-oc apply -f $BMOPATH/deploy/service_account.yaml
-oc apply -f $BMOPATH/deploy/role.yaml
-oc apply -f $BMOPATH/deploy/role_binding.yaml
-oc apply -f $BMOPATH/deploy/crds/metalkube_v1alpha1_baremetalhost_crd.yaml
-oc apply -f $BMOPATH/deploy/operator.yaml
+oc --config ocp/auth/kubeconfig apply -f ocp/deploy/service_account.yaml --namespace=openshift-machine-api
+oc --config ocp/auth/kubeconfig apply -f ocp/deploy/role.yaml --namespace=openshift-machine-api
+oc --config ocp/auth/kubeconfig apply -f ocp/deploy/operator.yaml --namespace=openshift-machine-api
+oc --config ocp/auth/kubeconfig apply -f ocp/deploy/role_binding.yaml
+oc --config ocp/auth/kubeconfig apply -f ocp/deploy/crds/metalkube_v1alpha1_baremetalhost_crd.yaml
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
-.PHONY: default all requirements configure repo_sync ironic build ocp_run clean ocp_cleanup host_cleanup
-default: requirements configure repo_sync ironic build ocp_run
+.PHONY: default all requirements configure repo_sync ironic build ocp_run deploy_bmo clean ocp_cleanup host_cleanup
+default: requirements configure repo_sync ironic build ocp_run deploy_bmo
 
 all: default
 
@@ -21,6 +21,9 @@ build:
 ocp_run:
 	./06_deploy_bootstrap_vm.sh
 	./07_deploy_masters.sh
+
+deploy_bmo:
+	./08_deploy_bmo.sh
 
 clean: masters_cleanup ocp_cleanup host_cleanup
 
--- a/README.md
+++ b/README.md
@@ -83,6 +83,12 @@ generated in the previous step.
 After running `./07_deploy_masters.sh` note that it takes some time for the cluster to
 fully come up, many container images are downloaded before the k8s API is fully available.
 
+- `./08_deploy_bmo.sh`
+
+After running `./08_deploy_bmo.sh` the cluster that becomes active in the previous step
+is updated by deploying the baremetal-operator into the pre-existing "openshift-machine-api"
+project/namespace.
+
 ## Interacting with the deployed cluster
 
 When the master nodes are up and the cluster is active, you can interact with the API:


derekhiggins:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/248/

derekhiggins:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/249/

derekhiggins:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/255/

07_deploy_masters.sh
```diff
@@ -63,6 +63,9 @@ while [[ ! $(timeout -k 9 5 $SSH "core@api.${CLUSTER_NAME}.${BASE_DOMAIN}" hostn
   sleep 10
 done
 
+#Adding some delay so that all 3 masters come up
+sleep 10
```
russellb: Maybe it was just coincidence that this seemed to help?  Since the loop below is an infinite loop and has its own sleep, I don't imagine this changes things?  Unless calling "oc get nodes" 10 seconds later for the first time helps?  That would be pretty weird though ...

In any case, I think it would make sense in a separate PR

03_ocp_repo_sync.sh
```diff
@@ -60,3 +60,9 @@ popd
 
 # Install baremetal-operator
 sync_go_repo_and_patch github.com/metalkube/baremetal-operator https://github.com/metalkube/baremetal-operator.git
+
+# Update project name to openshift-machine-api
+pushd "${GOPATH}/src/github.com/metalkube/baremetal-operator/deploy"
+echo "  namespace: openshift-machine-api" >> service_account.yaml
```
russellb: What do you think of copying the upstream files into a dir under dev-scripts and modify them there during 08_deploy_bmo.sh?

host_cleanup.sh
```diff
@@ -28,3 +28,12 @@ sudo virsh net-destroy baremetal
...
 sudo virsh net-undefine provisioning
+
+# Remove local changes made to baremetal-operator code
+eval "$(go env)"
+pushd "$GOPATH/src/github.com/metalkube/baremetal-operator"
+# Another way to cleanup would be to "rm -rf baremetal-operator"
+git checkout master
+git checkout -- deploy/role_binding.yaml
+git checkout -- deploy/service_account.yaml
+popd
```
russellb: If you follow my suggestion of copying these files instead of modifying the contents of the git repo, you could drop these changes.

07_deploy_masters.sh
hardys: I've seen the same error, the issue is the set -e means the first failure causes the script to exit so we never retry.

So we either need to make the VIP healtcheck better so it doesn't fail-over the VIP until the API is actually ready, tolerate failure in the oc calls, or add some sleep/retry here.

Agree it would probably be better in a separate patch though.

03_ocp_repo_sync.sh
hardys: +1 that would probably be cleaner than modifying the git checkout

07_deploy_masters.sh
hardys: Actually this will probably be fixed via https://github.com/openshift-metalkube/dev-scripts/pull/130 if we can figure out how to land that and not break CI...

03_ocp_repo_sync.sh
sadasu: Yup, I can do that. I wasn't sure which approach would be preferable.

07_deploy_masters.sh
sadasu: Possible that it is coincidence. As soon as make failed at this point I was able to actually connect to the API at the command line. So, added this is a temporary workaround so that CI could actually get to my code in 08_deploy_bmo.sh. (And it did, in the run with this change in place.) I had patched in #130 during my tests and did get better results.

host_cleanup.sh
sadasu: Yes, making that changes that were suggested for 03_ocp_repo_sync.sh.

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/269/

08_deploy_bmo.sh
```diff
@@ -5,12 +5,16 @@ eval "$(go env)"
 # Get the latest bits for baremetal-operator
 export BMOPATH="$GOPATH/src/github.com/metalkube/baremetal-operator"
 
-oc --as system:admin --config ocp/auth/kubeconfig new-project bmo-project
+# Make a local copy of the baremetal-operator code to make changes
+cp -r $BMOPATH .
```
dhellmann: We really just need the deployment files, right? We don't need to copy the whole git repository? Should we put them in the ocp directory, since that's where we are putting all of our other temporary files like the master ignition files?

derekhiggins:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/270/

08_deploy_bmo.sh
sadasu: Ok. I was working on creating a deploy directory under dev-scripts instead of copying the whole repo. But, I can put it into the ocp directory too.

08_deploy_bmo.sh
```diff
@@ -5,12 +5,13 @@ eval "$(go env)"
...
-oc --as system:admin --config ocp/auth/kubeconfig apply -f $BMOPATH/deploy/service_account.yaml
-oc --as system:admin --config ocp/auth/kubeconfig apply -f $BMOPATH/deploy/role.yaml
-oc --as system:admin --config ocp/auth/kubeconfig apply -f $BMOPATH/deploy/role_binding.yaml
-oc --as system:admin --config ocp/auth/kubeconfig apply -f $BMOPATH/deploy/crds/metalkube_v1alpha1_baremetalhost_crd.yaml
-oc --as system:admin --config ocp/auth/kubeconfig apply -f $BMOPATH/deploy/operator.yaml
+oc --as system:admin --config ocp/auth/kubeconfig apply -f deploy/service_account.yaml
+oc --as system:admin --config ocp/auth/kubeconfig apply -f deploy/role.yaml
+oc --as system:admin --config ocp/auth/kubeconfig apply -f deploy/role_binding.yaml
+oc --as system:admin --config ocp/auth/kubeconfig apply -f deploy/crds/metalkube_v1alpha1_baremetalhost_crd.yaml
+oc --as system:admin --config ocp/auth/kubeconfig apply -f deploy/operator.yaml
```
dhellmann: We need to make sure this uses the openshift-machine-api namespace.

derekhiggins:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/272/

08_deploy_bmo.sh
sadasu: The change in line #10 is updating the namespace to openshift-machine-api in the role_binding.yaml file. Do you think it needs to be changed in other places? Other .yaml files do not explicitly have a namespace field. Do you want me to add the namespace field explicitly to the metadata section of the other (other than role_binding.yaml) .yaml files?

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/278/

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/284/

dhellmann:
This looks good to me now. @russellb you had changes requested, are you happy with the current version?

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/289/

hardys:
Note when this gets rebased that you need to pull latest master, then rebase the topic branch, not merge master to the topic branch (ref "Merge branch 'master' into 277-dev" commit above).

My workflow is like:

```
git checkout master
git pull
git checkout my_branch
git rebase -i origin # Here you can squash/fixup commits and deal with any conflicts
git push hardys my_branch -f
```

There are other approaches but I don't think we want merge commits on the feature branches, as I'm not sure if/how that will get resolved when we merge the PR?

hardys:
Sorry this needs another rebase because we merged the --config change

derekhiggins:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/328/

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/336/

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/339/

08_deploy_bmo.sh
```diff
@@ -1,18 +1,18 @@
...
 
 # Get the latest bits for baremetal-operator
 export BMOPATH="$GOPATH/src/github.com/metalkube/baremetal-operator"
 
-source common.sh
-
-oc new-project bmo-project
+# Make a local copy of the baremetal-operator code to make changes
+cp -r $BMOPATH/deploy ocp/.
+sed -i 's/bmo-project/openshift-machine-api/g' ocp/deploy/role_binding.yaml
```
dhellmann: This will pattern will need to change when https://github.com/metalkube/baremetal-operator/pull/88 is merged. We can do that separately if this PR merges first.

PR #176, opened 2019-03-18
Makefile: Add all target.

russellb:
"all" is a typical name for the default make target, so add it as an
alias for "default".  Then I can happily run "make clean all".


derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/267/

PR #178, opened 2019-03-18
Patches for BM

jtaleric:
Necessary patches for baremetal deployments
- ROOT_DISK was not being used in the TF
- Do not overwrite MASTER_NODES json if BM
- IPMI port with BM doesn't change


jtaleric:
Going to block this patch for now. There is work to fix ipmi issues already. I will submit a smaller patch addressing the ROOT_DISK issue.

derekhiggins:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/271/

PR #180, opened 2019-03-19
Use ROOT_DISK

jtaleric:
We provide the root disk via common.sh, use this instead of defining
a new var.


derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/280/

PR #183, opened 2019-03-19
add a script to approve the certs after they expire

dhellmann:
We shouldn't need this for very long, but to save everyone the effort
of keeping track of the command that needs to be run here's a script.


derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/292/

PR #184, opened 2019-03-19
WIP: Set up baremetal operator deployment with ironic.

imain:
Just doing a PR for visibility and for people to try it out if needed.

Co-Authored-By: Marius Cornea

### DIFF
--- a/08_deploy_bmo.sh
+++ b/08_deploy_bmo.sh
@@ -2,15 +2,39 @@
 
 eval "$(go env)"
 
+source common.sh
+
 # Get the latest bits for baremetal-operator
 export BMOPATH="$GOPATH/src/github.com/metalkube/baremetal-operator"
 
-oc --as system:admin --config ocp/auth/kubeconfig new-project bmo-project
+# Make a local copy of the baremetal-operator code to make changes
+cp -r $BMOPATH/deploy ocp/.
+sed -i 's/bmo-project/openshift-machine-api/g' ocp/deploy/role_binding.yaml
+
+# Kill the full ironic container on the host since it is performing DHCP and doesn't
+# allow our pod in openshift to take over.
+for name in ironic ironic-inspector httpd ; do 
+    sudo podman ps | grep -w "$name$" && sudo podman kill $name
+    sudo podman ps --all | grep -w "$name$" && sudo podman rm $name -f
+done
+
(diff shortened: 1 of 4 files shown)


derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/293/

08_deploy_bmo.sh
```diff
@@ -2,15 +2,39 @@
...
 
+source common.sh
+
 # Get the latest bits for baremetal-operator
 export BMOPATH="$GOPATH/src/github.com/metalkube/baremetal-operator"
 
-oc --as system:admin --config ocp/auth/kubeconfig new-project bmo-project
+# Make a local copy of the baremetal-operator code to make changes
+cp -r $BMOPATH/deploy ocp/.
+sed -i 's/bmo-project/openshift-machine-api/g' ocp/deploy/role_binding.yaml
```
hardys: I think this is supposed to be handled via https://github.com/openshift-metalkube/dev-scripts/pull/175 right?  So maybe we should rebase this patch on that one from @sadasu ?

08_deploy_bmo.sh
```diff
@@ -2,15 +2,39 @@
...
+
+# Kill the full ironic container on the host since it is performing DHCP and doesn't
+# allow our pod in openshift to take over.
```
bfournie: Looks like a good workaround, we'd probably be better off with a separate container for dnsmaq though so we can just kill that.  I'm working on a PR for separate containers and should have something shortly.  That will also fix the logging problems that Ian tracked in https://github.com/metalkube/metalkube-ironic/issues/13

imain: actually this breaks the cleanup.  Ironic loses info about the nodes its deployed.

Probably you can just ignore this whole thing :).

08_deploy_bmo.sh
imain: Yes you're right, didn't see that until now.  Also I don't think the copy is a good idea.. we need to fix up the baremetal-operator stuff

russellb:
This PR hasn't been updated in a while, so I"m going to close it out as stale.  Feel free to reopen or open a new PR when you're ready for folks to try an update to this.

PR #189, opened 2019-03-20
Removing leftover bootstrap VM as part of cleanup.

achuzhoy:
This patch aims to remove the leftover bootstrap
VM that remains on a hypervisor after clean.

### DIFF
--- a/08_deploy_bmo.sh
+++ b/08_deploy_bmo.sh
@@ -1,5 +1,7 @@
 #!/usr/bin/bash
 
+set -ex
+
 #source common.sh
 eval "$(go env)"
 
--- a/09_deploy_manifests.sh
+++ b/09_deploy_manifests.sh
@@ -1,5 +1,7 @@
 #!/usr/bin/bash
 
+set -ex
+
 # FIXME - what are the prerequisites we need to wait for here, is waiting for 
 # bootstrap complete enough?
 


vrutkovs:
I'd prefer kni-installer to do that if the installation succeeds

russellb:
This seems fine to me for now.  Indeed, `kni-installer` should do this eventually.  I imagine that could be done once kni-installer is updated to deploy the masters with terraform, as well.

hardys:
This should be done (indeed it is done) via ocp_cleanup.sh - we already discussed this via https://github.com/openshift-metalkube/dev-scripts/pull/162 and IMHO we should close this without merging, `make clean` already does the right thing?

achuzhoy:
This is what one may get now (several bootstrap vms)
[stack@rhos-compute-node-14 dev-scripts]$ sudo virsh list
 Id    Name                           State
----------------------------------------------------
 13    ostest-z4glr-bootstrap         running
 16    ostest-s5nvn-bootstrap         running

hardys:
@achuzhoy if this re-occurs please can you check the logs for the ocp_cleanup.sh script, I suspect something must have failed in the kni-installer cleanup for this to happen, so there could be an underlying bug there that we've not yet tracked.

PR #191, opened 2019-03-20
Provide initial rook artifacts

karmab:


manifests/131_rook_cm.yaml
```diff
@@ -0,0 +1,78 @@
+apiVersion: v1
+data:
+  csi-cephfsplugin-provisioner.yaml: "kind: StatefulSet\napiVersion: apps/v1beta2\nmetadata:\n
```
vrutkovs:
```suggestion
  csi-cephfsplugin-provisioner.yaml: |
    kind: StatefulSet
    apiVersion: apps/v1beta2
    metadata:
```
etc.

karmab:
@vrutkovs i didn't came up with the manifests, mostly used the existing ones from rook so i believe they actually should be let untouched

vrutkovs:
@karmab if these are upstream why not `curl` them? If these are expected to be changed these should be readable

karmab:
i m fine providing a 09 rook script, actually i prefer it
@hardys happy with that approach ?

karmab:
updated the PR to use a dedicated 10 script

derekhiggins:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/310/

karmab:
unreladed @derekhiggins ?

derekhiggins:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/314/

hardys:
I'm fine with the separate scripts, but this isn't enabling any extra disks for the VM setup - I guess we need those for the OSD's, so can you clarify how many and what size?

The ansible roles already have the templating to add the disks, so it should be pretty easy to add, even if we do it as a follow-up

10_deploy_rook.sh
```diff
@@ -0,0 +1,72 @@
+#!/usr/bin/bash
+
```
hardys: We need to source common.sh here so the KUBECONFIG variable is set, ref https://github.com/openshift-metalkube/dev-scripts/pull/177/

hardys: Also probably set -eux similar to other scripts?

10_deploy_rook.sh
```diff
@@ -0,0 +1,72 @@
...
+git clone https://github.com/rook/rook
+cd rook/cluster/examples/kubernetes/ceph
```
hardys: This git clone should go in `03_ocp_repo_sync.sh` I think, so it goes in the GOPATH and not in the dev-scripts directory?

10_deploy_rook.sh
```diff
@@ -0,0 +1,72 @@
...
+oc apply -f csi/rbac/cephfs
+sed -i '/ROOK_HOSTPATH_REQUIRES_PRIVILEGED/!b;n;c\          value: "true"' operator-with-csi.yaml
+oc create -f scc.yaml
+oc adm policy add-scc-to-user privileged -z rook-csi-rbd-provisioner-sa -n rook-ceph-system
+oc adm policy add-scc-to-user privileged -z rook-csi-rbd-attacher-sa -n rook-ceph-system
+oc adm policy add-scc-to-user privileged -z csi-rook-csi-rbd-plugin-sa -n rook-ceph-system
+oc adm policy add-scc-to-user privileged -z rook-csi-cephfs-plugin-sa -n rook-ceph-system
+oc adm policy add-scc-to-user privileged -z rook-csi-cephfs-provisioner-sa -n rook-ceph-system
+oc create -f operator-with-csi.yaml
+sleep 180
```
hardys: Can we poll or -w for some status/event instead of sleeping here?

10_deploy_rook.sh
```diff
@@ -0,0 +1,72 @@
...
+metadata:
+  name: rbd
+  namespace: rook-ceph
+spec:
+  failureDomain: osd
+  replicated:
+    size: 3
+EOF
+
+sleep 240
```
hardys: Same as above, for CI we want to wait for as little time as possible, and some poll/event check will be generally more robust?

See #130 for example where we started using oc wait to ensure a given condition or event happened?

10_deploy_rook.sh
```diff
@@ -0,0 +1,72 @@
+#!/usr/bin/bash
+
+git clone https://github.com/rook/rook
```
ktdreyer: We can do a shallow clone here to save time, right?

ktdreyer:
I work on RH's Ceph team, and we have some similar bash code for running Rook with OpenShift. Is there any meeting where your team is discussing this approach?

This example YAML approach is changing frequently (eg. https://github.com/rook/rook/pull/2820) so I want to be sure we're aligned.

hardys:
> I work on RH's Ceph team, and we have some similar bash code for running Rook with OpenShift. Is there any meeting where your team is discussing this approach?

Certainly happy to chat, but note the idea is that we consume whatever is the best-practice approach to deploy Ceph via Rook on OpenShift 4.x, not to reimplement anything that may already exist :)

These scripts are really just bootstrapping some temporary prototyping, and we expect/hope that ultimately we don't have any scripts and just some yaml manifests which are applied to the cluster via the machine-config-operator.

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/330/

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/331/

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/332/

10_deploy_rook.sh
karmab: done

derekhiggins:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/333/

hardys:
I tested this and the nodes didn't deploy correctly - is this working locally for you?

I'm wondering if we need to change the ironic root device hint, as it's perhaps writing the disk image to the wrong drive, or booting from the wrong one?

I'll investigate further if you think this should be working as-is

karmab:
@hardys i'm also hitting issues too with a secondary disk for the vms

derekhiggins:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/342/

karmab:
as a first pass, let s get the script available, we can investigate the additional disks failure separately

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/351/

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/355/

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/357/

hardys:
Ok we discussed and agreed to merge this as-is, and work on adding the extradisks support as a followup, since this can be tested now by manually adding the disks.

I'll take a look at why the extradisks stuff doesn't work, we had it enabled when this repo was originally created and it worked, so hopefully it's a simple fix to ensure the deploy/boot happens on the correct disk.

PR #192, opened 2019-03-20
WIP Automatically use latest RHCOS build.

russellb:
Instead of having to manually update the version number in this
script, always pull the latest RHCOS build.  That is what is done for
other platforms, and will help us catch any compatibility issues
sooner.

### DIFF
--- a/common.sh
+++ b/common.sh
@@ -57,8 +57,10 @@ if [ ! -d "$WORKING_DIR" ]; then
 fi
 
 export RHCOS_IMAGE_URL=${RHCOS_IMAGE_URL:-"https://releases-rhcos.svc.ci.openshift.org/storage/releases/maipo/"}
-export RHCOS_IMAGE_VERSION="${RHCOS_IMAGE_VERSION:-47.284}"
-export RHCOS_IMAGE_NAME="redhat-coreos-maipo-${RHCOS_IMAGE_VERSION}"
+export RHCOS_LATEST="$(curl ${RHCOS_IMAGE_URL}/builds.json | jq -r ".builds[0]")"
+export RHCOS_IMAGE_VERSION="${RHCOS_IMAGE_VERSION:-${RHCOS_LATEST}}"
+export RHCOS_IMAGE_FILENAME_OPENSTACK_GZ="$(curl ${RHCOS_IMAGE_URL}/${RHCOS_IMAGE_VERSION}/meta.json | jq -r '.images.openstack.path')"
+export RHCOS_IMAGE_NAME=$(echo $RHCOS_IMAGE_FILENAME_OPENSTACK_GZ | sed -e 's/-openstack.*//')
 # FIXME(shardy) note the -openstack image doesn't work for libvirt
 # as the qemu ignition config injection described in the docs at
 # https://coreos.com/os/docs/latest/booting-with-libvirt.html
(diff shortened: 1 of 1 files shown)


derekhiggins:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/315/

russellb:
While the patch works to grab the latest rhcos image, the cluster doesn't come up for me.  I believe the latest images are based on rhel 8, so we may have compatibility issues there to resolve first.

vrutkovs:
This won't boot masters yet - a new boot entry is created, but ostree won't find the deployment.

Also this needs to modify keepalived/haproxy to run only when pivot.service has finished provisioning

common.sh
```diff
@@ -68,7 +70,7 @@ export RHCOS_IMAGE_NAME="redhat-coreos-maipo-${RHCOS_IMAGE_VERSION}"
 export RHCOS_IMAGE_FILENAME="${RHCOS_IMAGE_NAME}-qemu.qcow2"
 export RHCOS_IMAGE_FILENAME_OPENSTACK="${RHCOS_IMAGE_NAME}-openstack.qcow2"
 export RHCOS_IMAGE_FILENAME_DUALDHCP="${RHCOS_IMAGE_NAME}-dualdhcp.qcow2"
-export RHCOS_IMAGE_FILENAME_LATEST="redhat-coreos-maipo-latest.qcow2"
+export RHCOS_IMAGE_FILENAME_LATEST="${RHCOS_IMAGE_NAME}-latest.qcow2"
```
derekhiggins: $RHCOS_IMAGE_FILENAME_LATEST  shouldn't contain the version number, we created it so that we would have a versionless link to point at.

russellb: It still doesn't have the version.  That variable contains the base of the name, without the version number.  I updated this to keep it consistent with other images.  The naming convention changed in the most recent images.

derekhiggins: I'm seeing it in the logs with a version number

```
++ RHCOS_IMAGE_NAME=rhcos-maipo-400.7.20190312.0
++ RHCOS_IMAGE_FILENAME_LATEST=rhcos-maipo-400.7.20190312.0-latest.qcow2
```

russellb: Ah, ooops :-)

russellb: I changed this back to a static filename, but using the latest naming convention

derekhiggins:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/334/

vrutkovs:
W/ #197 masters boot, but two services are failing to start:
```
Failed Units: 2
  rm-provision-nic.service
  systemd-udev-settle.service
```

rm-provision-nic fails with `Failed at step EXEC spawning /usr/local/bin/rm-provision-nic.sh: Permission denied`

cgwalters:
Right, we've got a mess currently due to the in-flight Ignition v3 switch.

If you want to test a known-to-work cluster, you should use the same version as the installer:
https://github.com/openshift/installer/blob/4354ddff20dff7b96d556e1dd9b1ceef1d869d42/hack/build.sh#L5

However newer RHCOS images have fixes like https://github.com/coreos/coreos-assembler/pull/442 that aren't in that version.  Hopefully we'll sort out Ignition v3 quickly.

hardys:
Closing this as it's been replaced by https://github.com/openshift-metalkube/dev-scripts/pull/259

PR #194, opened 2019-03-20
api-haproxy: change to docker library haproxy img

celebdor:
Now that we use internal haproxy httpchk, we do not need the custom
image we made.


derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/320/

PR #197, opened 2019-03-22
Fix the boot device hacks to follow symlinks

derekhiggins:
sed needs to be told to follow a simlink to prevent it
from converting a softlink to a regular file. Converting
it to a regular file meant when pivot created a new grub.cfg
we were no longer pointing at it and as a result we attempting
to switch roots into the old ostree.

Also make sure the baremetal network is the one mentioned in
/etc/default/grub so its correct if grub.cfg is regenerated.


derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/337/

vrutkovs:
This seems to unblock https://github.com/openshift-metalkube/dev-scripts/pull/192

PR #186, opened 2019-03-20
Added prerequisites script

e-minguez:

PR #198, opened 2019-03-22
Allow to use root user

vrutkovs:
tripleo scripts recommend using a non-root user, but dev-scripts work perfectly as a root. Also using root simplifies testing w/ Beaker machines

This reverts a part of #186

### DIFF
--- a/common.sh
+++ b/common.sh
@@ -68,12 +68,6 @@ if [ "$USER" != "root" -a "${XDG_RUNTIME_DIR:-}" == "/run/user/0" ] ; then
     exit 1
 fi
 
-# Check if root user
-if [[ $EUID -eq 0 ]]; then
-   echo "This script must be executed as regular user"
-   exit 1
-fi
-
 # Check if sudo privileges without password
 if ! sudo -n uptime &> /dev/null ; then
   echo "sudo without password is required"


common.sh
```diff
@@ -63,22 +63,6 @@ export SSH="ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -o C
...
-fi
-
-# Check if root user
-if [[ $EUID -eq 0 ]]; then
-   echo "This script must be executed as regular user"
-   exit 1
-fi
-
-# Check if sudo privileges without password
-if ! sudo -n uptime &> /dev/null ; then
```
hardys: We should probably leave the sudo check, as it's needed for non-root users, and it will still work fine when run as root?

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/338/

common.sh
vrutkovs: Right, sudo would work with root too

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/352/

common.sh
```diff
@@ -63,16 +63,6 @@ export SSH="ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -o C
 # Verify requisites/permissions
 # Connect to system libvirt
 export LIBVIRT_DEFAULT_URI=qemu:///system
-if [ "$USER" != "root" -a "${XDG_RUNTIME_DIR:-}" == "/run/user/0" ] ; then
```
hardys: Sorry we actually need this first check too - it will always pass when run as root, but it catches the case where someone has done `su` to a non-root user, e.g not `su -`, and in this case the XDG_RUNTIME_DIR still points at uid 0, consequently the scripts don't work as expected.

vrutkovs: Ah, makes sense. Reverted that part

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/376/

PR #202, opened 2019-03-22
set up the role binding before launching metalkube baremetal operator

dhellmann:
The role binding is needed to give the operator permission, so we
should wait to launch the operator after the role binding is there.


derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/345/

PR #203, opened 2019-03-22
be more resilient when changing the namespace for the bmo

dhellmann:
Change the sed pattern so we do not rely on what the current namespace
is, only on the fact that we are changing the namespace line. This
allows us to change the name we use when deploying on stock
kubernetes.


sadasu:
Yes, makes code more generic.

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/344/

PR #214, opened 2019-03-25
Configure etcd CNAME as a service

cybertron:
Instead of the semi-hard-coded CNAME generation in the coredns-mdns
plugin, use an mdns service record to create the etcd CNAMEs. This
is more flexible and will allow the plugin to be more generic since
it won't need to have OpenShift-specific logic in it.


derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/366/

cybertron:
No longer needed. This was included as part of #224

PR #218, opened 2019-03-26
dns-vip-prepender: Prevent race with crio

celebdor:
Cri-O and other service start after network-online.target.
dns-vip-prepender was being triggered by the NetworkManger dispatcher in
the "up" state, which happens right at network-online.target.
This means that if some service that ran after network-online target
needs CoreDNS mDNS resolution it may not get it depends on what happened
earlier. In order to address that, we move the dispatcher script to
"pre-up" which is strictly before NM desclares network-online.target as
achieved.


derekhiggins:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/382/

cybertron:
lgtm. This should also ensure that NM won't race with crio when it takes down the interface to apply changes.

PR #219, opened 2019-03-26
clean up /opt/dev-scripts

mlammon:
Signed-off-by: Mike Lammon <mike.lammon@gmail.com>

### DIFF
--- a/host_cleanup.sh
+++ b/host_cleanup.sh
@@ -28,3 +28,4 @@ sudo virsh net-destroy baremetal
 sudo virsh net-undefine baremetal
 sudo virsh net-destroy provisioning
 sudo virsh net-undefine provisioning
+sudo rm -rf /opt/dev-scripts/


host_cleanup.sh
```diff
@@ -28,3 +28,4 @@ sudo virsh net-destroy baremetal
 sudo virsh net-undefine baremetal
 sudo virsh net-destroy provisioning
 sudo virsh net-undefine provisioning
+sudo rm -rf /opt/dev-scripts/
```
elfosardo: this should be $WORKING_DIR

hardys: Also I'm not sure we should do this unconditionally, e.g in the typical dev workflow we do `make clean; make` or similar, and this will remove all the cached images thus increasing the test time and bandwidth usage.

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/383/

host_cleanup.sh
derekhiggins: Ya, I agree with shardy here, this would break a lot of peoples workflow, as we would loose cached images,

mlammon: OK, perhaps it should just be noted in the readme to start clean this needs to be removed.

PR #223, opened 2019-03-26
Bridges configuration adjustments for baremetal deployment

mcornea:
**Bridges configuration adjustments for baremetal deployment**

This change implements the following changes to closer resemble
the documented networking requirements[1]:

 - when NODES_PLATFORM is set to baremetal tripleo-quickstart-config
   sets up the baremetal virsh network as a simple bridge. Currently
   the baremetal network is set up as a nat network providing dhcp
   and dns services which are not required as those are expected to be
   external to the provisioning host.
 - configure as mac address of the baremetal bridge the same mac address
   as $INT_IF. This is so we keep using the same DHCP IP address and not
   lose network connectivity when adding the interface to the bridge.
 - configure sysconfig networks scripts for the physical interfaces and
   bridges and then apply them via restarting the network service. This
   should prevent losing network connectivity and it's required for
   automation purposes.

[1] https://docs.google.com/document/d/1YGbpi1IoUHzRwHdXOF9La8AJsovM4WnixwGl_tch1Sc/edit

### DIFF
--- a/02_configure_host.sh
+++ b/02_configure_host.sh
@@ -19,6 +19,8 @@ ANSIBLE_FORCE_COLOR=true ansible-playbook \
     -e "local_working_dir=$HOME/.quickstart" \
     -e "virthost=$HOSTNAME" \
     -e "platform=$NODES_PLATFORM" \
+    -e "baremetal_interface=$INT_IF" \
+    -e "provisioning_interface=$PRO_IF" \
     -e @config/environments/dev_privileged_libvirt.yml \
     -i tripleo-quickstart-config/metalkube-inventory.ini \
     -b -vvv tripleo-quickstart-config/metalkube-setup-playbook.yml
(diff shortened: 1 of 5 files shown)


karmab:
instead of using nmap to gather the ip, a better option would be to make sure bootstrap vm contains guest agent and uses that

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/386/

02_configure_host.sh
```diff
@@ -82,14 +99,27 @@ if [ "$INT_IF" ]; then
...
+    </network>
+EOF
+    sudo virsh net-define provisioning.xml
+    sudo virsh net-define baremetal.xml
+    sudo virsh net-start provisioning
+    sudo virsh net-start baremetal
+    echo -e "DEVICE=baremetal\nTYPE=Bridge\nHWADDR=$MAC\nONBOOT=yes\nNM_CONTROLLED=no\nBOOTPROTO=dhcp" | sudo dd of=/etc/sysconfig/network-scripts/ifcfg-baremetal
+    echo -e "DEVICE=eth0\nTYPE=Ethernet\nONBOOT=yes\nNM_CONTROLLED=no\nBOOTPROTO=none\nBRIDGE=baremetal" | sudo dd of=/etc/sysconfig/network-scripts/ifcfg-eth0
+    echo -e "DEVICE=eth1\nTYPE=Ethernet\nONBOOT=yes\nNM_CONTROLLED=no\nBOOTPROTO=none\nBRIDGE=provisioning" | sudo dd of=/etc/sysconfig/network-scripts/ifcfg-eth1
+    sudo systemctl restart network
```
hardys: We added NODES_PLATFORM to allow skipping the creation of the VMs, I wonder if you set that (to anything other than the default "libvirt", can we get the desired outcome, e.g quickstart will only configure the networks and skip the VM creation?

It seems kind of weird to skip VM setup on the host but still configure the networks etc though.
It seems

02_configure_host.sh
```diff
@@ -4,28 +4,45 @@ set -xe
...
+    else
+      sudo systemctl restart NetworkManager
+    fi
+fi    
 # Allow local non-root-user access to libvirt
 sudo usermod -a -G "libvirt" $USER
 
+# Restart libvirtd service to get the new group membership loaded
+sudo systemctl restart libvirtd
+
```
hardys: This could be a separate patch

07_deploy_masters.sh
```diff
@@ -50,14 +50,16 @@ popd
 
 echo "Master nodes active"
 
-NUM_LEASES=$(sudo virsh net-dhcp-leases baremetal | grep master | wc -l)
-while [ "$NUM_LEASES" -ne 3 ]; do
-  sleep 10
+if [ "$OOOQ_SETUP" == 'y' ]; then
```
hardys: I wonder if we can find a cleaner way to do this, maybe we ping the hosts by name instead which would work in both cases?

common.sh
```diff
@@ -127,3 +127,9 @@ LOGFILE="$LOGDIR/$(basename $0 .sh)-$(date +%F-%H%M%S).log"
 echo "Logging to $LOGFILE"
 # Set fd 1 and 2 to write to the log file
 exec 1> >( tee "${LOGFILE}" ) 2>&1
+
+# Set up virtual infrastructure with dev-scripts. Default: yes
+OOOQ_SETUP=${OOOQ_SETUP:-y}
```
hardys: If we can't reuse NODES_PLATFORM for this we should avoid the OOOQ name IMO - it makes sense to those of us who know where the ansible roles originated, but it's probably confusing to anyone that didn't previously work with OOOQ :)

02_configure_host.sh
mcornea: Great, I wasn't aware of the NODES_PLATFORM. I'm going to explore using it.

07_deploy_masters.sh
mcornea: Yes, we can leverage the DNS entries. I sent another PR to address it - https://github.com/openshift-metalkube/dev-scripts/pull/232

common.sh
mcornea: I agree, I'm going to refactor around using the NODES_PLATFORM option.

mcornea:
> Few comments added - it'd be easier to review if you broke up the patch into a series of commits, or perhaps several PRs - not blocking on that but just noting it for future patches :)

Thanks much for the review Steve! I broke it into several different PRs for ease of review:
https://github.com/openshift-metalkube/dev-scripts/pull/229
https://github.com/openshift-metalkube/dev-scripts/pull/232
https://github.com/openshift-metalkube/dev-scripts/pull/233
https://github.com/openshift-metalkube/dev-scripts/pull/235

I'm going to leave this one for exploring how to use the NODES_PLATFORM option to achieve what I wish.

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/405/

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/406/

mcornea:
No longer needed. Changes were addressed in https://github.com/openshift-metalkube/dev-scripts/commit/1d74bd97550b3c84fe69e0effd130b2adad7485d

PR #227, opened 2019-03-27
devtooling: Allow PR to be used when syncing repos

celebdor:
Sometimes it is nice to refer to some PR when building. This allows you
to do that by:

    sync_go_repo_and_patch github.com/openshift-metalkube/kni-installer https://github.com/openshift-metalkube/kni-installer.git 26

This means that instead of using kni origin/master, it will use the code
in PR #26

### DIFF
--- a/03_ocp_repo_sync.sh
+++ b/03_ocp_repo_sync.sh
@@ -7,12 +7,19 @@ eval "$(go env)"
 echo "$GOPATH" | lolcat # should print $HOME/go or something like that
 
 function sync_go_repo_and_patch {
-    DEST="$GOPATH/src/$1"
-    figlet "Syncing $1" | lolcat
+    local repo_gopath
+    local git_repo
+    local pr
+
+    repo_gopath="$1"
+    git_repo="$2"
+    pr="$3"
+    DEST="$GOPATH/src/$repo_gopath"
+    figlet "Syncing $repo" | lolcat
 
     if [ ! -d $DEST ]; then
         mkdir -p $DEST
-        git clone $2 $DEST
+        git clone "$git_repo" $DEST
     fi
 
     pushd $DEST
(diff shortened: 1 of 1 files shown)


derekhiggins:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/393/

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/395/

03_ocp_repo_sync.sh
```diff
@@ -21,7 +28,12 @@ function sync_go_repo_and_patch {
     git checkout master
     git pull --rebase origin master
     git branch -D we_dont_need_no_stinkin_patches || true
-    git checkout -b we_dont_need_no_stinkin_patches
+    if [[ -n "$pr" ]]; then
+        git fetch origin "pull/$pr/head:we_dont_need_no_stinkin_patches"
```
hardys: Sometimes PRs get stale, to ensure we always test the latest state (especially in CI) should we perhaps fetch the branch then rebase on master?

derekhiggins: +1 to rebasing, things are still changing quickly.

derekhiggins:
arn't we breaking compatibility with the existing patching functionality in this patch? Actually come to think of it we can already do what your trying to do

`sync_go_repo_and_patch github.com/openshift-metalkube/kni-installer https://github.com/openshift-metalkube/kni-installer.git https://patch-diff.githubusercontent.com/raw/openshift-metalkube/kni-installer/pull/13.patch`

and it take multiple PR's

hardys:
Yes this does break the existing patch-apply interface, personally I don't have a strong preference for one way or another - good point re the multiple PRs though @derekhiggins and also that solves the rebase problem because we apply them on master?

PR #229, opened 2019-03-27
Restart libvirtd service after adding user to libvirt group

mcornea:
When adding the user to the libvirt group for the first time we need
to restart the libvirtd service so it loads the new group membership
and allow the subsequent virsh pool commands to run without error.

### DIFF
--- a/02_configure_host.sh
+++ b/02_configure_host.sh
@@ -24,7 +24,11 @@ ANSIBLE_FORCE_COLOR=true ansible-playbook \
     -b -vvv tripleo-quickstart-config/metalkube-setup-playbook.yml
 
 # Allow local non-root-user access to libvirt
-sudo usermod -a -G "libvirt" $USER
+# Restart libvirtd service to get the new group membership loaded
+if ! id $USER | grep -q libvirt; then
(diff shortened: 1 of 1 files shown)


derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/398/

02_configure_host.sh
```diff
@@ -26,6 +26,9 @@ ANSIBLE_FORCE_COLOR=true ansible-playbook \
 # Allow local non-root-user access to libvirt
 sudo usermod -a -G "libvirt" $USER
 
+# Restart libvirtd service to get the new group membership loaded
```
hardys: Should we perhaps do this (and the usermod above) conditionally, so we don't restart libvirt unneccesarily?

E.g something like:

```
    sudo systemctl restart libvirtd
fi
```

derekhiggins:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/419/

hardys:
The CI failure appears unrelated, we get past this change and get a failure untainting the nodes after the OCP cluster is deployed

PR #236, opened 2019-03-28
Increase timeout for services that download content

yrobla:
When testing that with a slow connection, the services keep
restarting because they don't have enough time to download the
content. Increasing the timeout to, at least 10 minutes, ensures
that will work on more environments.


derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/407/

hardys:
Seems like a pretty long timeout to me, but unless @celebdor has any objections I'm fine with this merging

PR #243, opened 2019-03-29
Always source common.sh first

derekhiggins:
Users may have custom variables set in config_<user>.sh
If they have a custom CLUSTER_NAME or BASE_DOMAIN, we need
to set these first, before the defaults are used to set
CLUSTER_DOMAIN in ocp_install_env.sh.


derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/443/

PR #244, opened 2019-03-29
Remove ifcfg-provisioning during host_cleanup

cybertron:
There was a bug in this file previously, and it apparently doesn't
get updated properly. Remove the file during host cleanup so it will
be re-created correctly.


derekhiggins:
merging, ci isn't going to test this anyways.

derekhiggins:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/444/

PR #245, opened 2019-03-29
Add support for Ingress VIP

celebdor:
For the Openshift router support, the router pods will run on
masters-nodes (host networking).
This patch adds support for the Ingress VIP(192.168.111.4), one
of the master nodes will hold this IP (VRRP), creates wildcard
external DNS entry pointing to Ingress VIP and adds e2e verifictation test.

Co-authored-by: Antoni Segura Puimedon <antoni@redhat.com>

### DIFF
--- a/07_deploy_masters.sh
+++ b/07_deploy_masters.sh
@@ -73,4 +73,11 @@ for node in $(oc --config ocp/auth/kubeconfig get nodes --no-headers | sed -e 's
   oc label node $node node-role.kubernetes.io/worker=''
 done
 
+#Verify Ingress controller functionality
+echo "Verifying Ingress controller functionality, first we'll check that router pod responsive"
+until curl -o /dev/null -kLs --fail  tst.apps.${CLUSTER_DOMAIN}:1936/healthz; do sleep 20 && echo "Router pod not responding"; done
+echo "Router pod responding, checking connectivity to console via ingress controller"
+until wget --no-check-certificate https://console-openshift-console.apps.${CLUSTER_DOMAIN}; do sleep 10 && echo "rechecking"; done
+echo "Ingress controller is working!"
+
 echo "Cluster up, you can interact with it via oc --config ocp/auth/kubeconfig <command>"
--- a/06_deploy_bootstrap_vm.sh
+++ b/06_deploy_bootstrap_vm.sh
@@ -21,11 +21,14 @@ INFRA_ID=$(jq -r .infraID ocp/metadata.json)
 
 # NOTE: This is equivalent to the external API DNS record pointing the API to the API VIP
 if [ "$MANAGE_BR_BRIDGE" == "y" ] ; then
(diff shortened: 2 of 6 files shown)


celebdor:
Need to test the forwarders trick still

derekhiggins:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/446/

celebdor:
It worked on my env. Should we keep the Ingress controller test?

cybertron:
Ingress worked for me locally with this change.

I like the test. It's kind of like the ping test from tripleo as a basic smoke test of the cluster. The only concern I have about keeping it is that I did have a spurious failure while running it. I haven't run this patch enough to know whether that will be a common problem though. I suppose we could merge it as-is and back out the test if it ends up being a problem (although if it _is_ a problem we should probably address whatever is causing the underlying issue, so maybe that's an argument for keeping it too).

mcornea:
+1 worked for me as well on my environment:

```shell
[centos@provisionhost-0 dev-scripts]$ oc --config ocp/auth/kubeconfig  get routes
NAME      HOST/PORT         PATH   SERVICES   PORT   TERMINATION   WILDCARD
tst-svc   www.example.com          tst-svc    8080                 None
[centos@provisionhost-0 dev-scripts]$ oc --config ocp/auth/kubeconfig  get svc
NAME         TYPE           CLUSTER-IP      EXTERNAL-IP                            PORT(S)    AGE
kubernetes   ClusterIP      172.30.0.1      <none>                                 443/TCP    24m
openshift    ExternalName   <none>          kubernetes.default.svc.cluster.local   <none>     21m
tst-svc      ClusterIP      172.30.81.253   <none>                                 1234/TCP   10m
[centos@provisionhost-0 dev-scripts]$ curl  --header 'Host: www.example.com' whatever.apps.ostest.test.metalkube.org
tst-svc-1-h5ngb: HELLO! I AM ALIVE!!!
```

07_deploy_masters.sh
```diff
@@ -73,4 +73,19 @@ for node in $(oc --config ocp/auth/kubeconfig get nodes --no-headers | sed -e 's
   oc label node $node node-role.kubernetes.io/worker=''
 done
 
+#Verify Ingress controller functionality
+echo "Verifying Ingress controller functionality, first we'll check that router pod responsive"
+until curl -o /dev/null -kLs --fail  tst.apps.${CLUSTER_DOMAIN}:1936/healthz; do sleep 20 && echo "Router pod not responding"; done
+
+echo "Create service named tst-svc"
```
vrutkovs: There is already a lot of valid apps deployed in the cluster - like console - lets check that these are available?

if we want to keep the 'run arbitrary pod' scenario we'd have to take care of retries and error handling, which is terrible in bash

yboaron: @vrutkovs, done. 
Changed the routing test to check console service.

derekhiggins:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/456/

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/462/

PR #246, opened 2019-03-29
Restart network service instead of ifdown to avoid losing connectivity

mcornea:
To avoid losing network connectivity when bridging the physical interface
to the baremetal bridge we can use systemctl restart network. This is
useful for persisting the SSH connection when automating the deployment.

Also when the baremetal bridge is not managed by libvirt the provsioning host
uses the external DNS server for name resolution. This change sets the
condition so NetworkManager dnsmasq gets used only when the baremetal bridge
is managed by libvirt.

### DIFF
--- a/02_configure_host.sh
+++ b/02_configure_host.sh
@@ -123,13 +123,15 @@ if ! sudo iptables -C INPUT -p tcp --dport 3000 -j ACCEPT 2>/dev/null ; then
 fi
 
 # Switch NetworkManager to internal DNS
-sudo mkdir -p /etc/NetworkManager/conf.d/
-sudo crudini --set /etc/NetworkManager/conf.d/dnsmasq.conf main dns dnsmasq
-if [ "$ADDN_DNS" ] ; then
-  echo "server=$ADDN_DNS" | sudo tee /etc/NetworkManager/dnsmasq.d/upstream.conf
-fi
-if systemctl is-active --quiet NetworkManager; then
-  sudo systemctl reload NetworkManager
-else
-  sudo systemctl restart NetworkManager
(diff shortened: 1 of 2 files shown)


derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/447/

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/448/

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/449/

mcornea:
@derekhiggins Hi! Could you please check this change when you get a moment? Thanks!

02_configure_host.sh
```diff
@@ -73,14 +73,14 @@ sudo ifup baremetal
...
     echo -e "DEVICE=$INT_IF\nTYPE=Ethernet\nONBOOT=yes\nNM_CONTROLLED=no\nBRIDGE=baremetal" | sudo dd of=/etc/sysconfig/network-scripts/ifcfg-$INT_IF
-    sudo ifdown $INT_IF || true
-    sudo ifup $INT_IF
     if sudo nmap --script broadcast-dhcp-discover -e $INT_IF | grep "IP Offered" ; then
         echo -e "\nBOOTPROTO=dhcp\n" | sudo tee -a /etc/sysconfig/network-scripts/ifcfg-baremetal
-        sudo ifdown baremetal || true
-        sudo ifup baremetal
+        sudo systemctl restart network
+    else
+        sudo systemctl restart network
```
derekhiggins: nit: We have this line at the end of both the "if" and the "else", we could have it once, after the "if" statement

PR #248, opened 2019-03-29
root_disk shouldn't be hardcoded

achuzhoy:
Using the variable for root_disk that was
set by the user in config_${USER}.sh file.


derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/451/

hardys:
Oops, good catch, thanks!

PR #254, opened 2019-04-01
Exposed vars to customize master/worker specs

### LINKED ISSUE #253: default values for cpus and memory
currently, nodes are deployed with 2vcpus and 6Gb of ram, which causes pods to crashloop or not launch after initial bootstrapping.
we should increase those values to maybe 8 cpus and at least 8gb of ram

e-minguez:
This should fix #253 as will allow users to specify their own resources


derekhiggins:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/463/

hardys:
If we know the defaults are bad, shouldn't we just change the defaults?

My worry with this approach is everyone can have different specs, so we'll end up debugging weird problems where some folks allocate more/less resources - wouldn't it be better to ensure everyone uses the exact same spec, and we figure out what the minimum viable amount is for the testing we need to do?

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/480/

e-minguez:
I agree on having sane defaults is a must (for instance the monitoring stuff is not deployed due to the lack of cpu resources), but if we bump the cpu to 4 (4x3 + bootstrap) IDK if we all have enough horsepower to handle it.

hardys:
My point is really if we know the defaults are bad, it's best to be explicit about it, if folks don't have the resources to test, it's better to discover that at the start vs after hours of debugging OOM'd stuff on the nodes?

What if we increase the defaults, then allow overriding tripleo-quickstart-config/metalkube-nodes.yml via a variable, then we've got a simple way for "advanced" usage where folks can hack on that file, but it's clear what the default path is?

Also note it's not 4x3, pretty soon we'll want to deploy workers (at least 1) by default as well.

e-minguez:
> My point is really if we know the defaults are bad, it's best to be explicit about it, if folks don't have the resources to test, it's better to discover that at the start vs after hours of debugging OOM'd stuff on the nodes?
> 
+1
> What if we increase the defaults, then allow overriding tripleo-quickstart-config/metalkube-nodes.yml via a variable, then we've got a simple way for "advanced" usage where folks can hack on that file, but it's clear what the default path is?
> 
Done, I've increased vcpus to 4 and ram to 8gb. Also, I maintained the ability to customize it f needed.

> Also note it's not 4x3, pretty soon we'll want to deploy workers (at least 1) by default as well.

config_example.sh
```diff
@@ -9,3 +9,15 @@ set -x
...
+# Customize masters/workers specs
+# Memory in megabytes
+#export MASTER_MEMORY=6144
+# Num cpus
+#export MASTER_VCPU=2
+# Disk in gigabytes
+#export MASTER_DISK=50
+
+#export WORKER_MEMORY=6144
+#export WORKER_VCPU=8196
```
hardys: typo - 8196 VCPU! :)

02_configure_host.sh
```diff
@@ -15,6 +15,12 @@ ANSIBLE_FORCE_COLOR=true ansible-playbook \
     -e "non_root_user=$USER" \
     -e "working_dir=$WORKING_DIR" \
     -e "roles_path=$PWD/roles" \
+    -e "openshift_master_memory=${MASTER_MEMORY}" \
+    -e "openshift_master_vcpu=${MASTER_VCPU}" \
+    -e "openshift_master_disk=${MASTER_DISK}" \
+    -e "openshift_worker_memory=${WORKER_MEMORY}" \
+    -e "openshift_worker_vcpu=${WORKER_VCPU}" \
+    -e "openshift_worker_disk=${WORKER_DISK}" \
```
hardys: As I mentioned in my previous comment, would it perhaps be better to add a variable that overrides the  tripleo-quickstart-config/metalkube-nodes.yml file, and avoid putting these variables into the example config?

I'm just thinking if we want folks to use the defaults in most cases, we might want to avoid exposing the variables in the example config file, and if folks are changing this they may want to do stuff like modify the number/size of extra disks etc, which is all controlled via  tripleo-quickstart-config/metalkube-nodes.yml

config_example.sh
e-minguez: Not a typo... don't you have such a powerful server? :P

02_configure_host.sh
e-minguez: Ok, I'll fix it. Thanks!

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/485/

e-minguez:
@hardys PTAL, thanks!

derekhiggins:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/490/

tripleo-quickstart-config/metalkube-nodes.yml
```diff
@@ -29,13 +29,13 @@ node_count: 6
...
+    memory: '{{openshift_master_memory|default(default_memory, true)}}'
+    disk: '{{openshift_master_disk|default(default_disk, true)}}'
+    vcpu: '{{openshift_master_vcpu|default(default_vcpu, true)}}'
     extradisks: true
 
   openshift_worker:
-    memory: '{{openshift_memory|default(default_memory)}}'
-    disk: '{{openshift_disk|default(default_disk)}}'
-    vcpu: '{{openshift_vcpu|default(default_vcpu)}}'
+    memory: '{{openshift_worker_memory|default(default_memory, true)}}'
```
hardys: Is true a good default here?

FWIW I was expecting  us to leave the current variables as they are, and just expose a variable that enables easily passing a copy of this entire file, then folks can hack it however they want.

e-minguez: Sorry, I didn't removed the 'true' from the previous commit.

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/502/

README.md
```diff
@@ -152,3 +152,15 @@ at the command line, try:
...
+The default cpu/memory/disk resources when using virtual machines are provided
+by the [tripleo-quickstart-config/metalkube-nodes.yml](tripleo-quickstart-config/metalkube-nodes.yml) file:
+
+* 4 vCPU
+* 8 Gb RAM
+* 50 Gb HDD
+
+If required, the cpu/memory/disk can be customzied per host role
+(`openshift_master_memory`, `openshift_worker_memory`,...) by modifying that
+file prior to run the [04_setup_ironic.sh](04_setup_ironic.sh) script.
```
hardys: Modifying the script isn't a great interface for this - if we want users to be able to change the vars, we should add a variable to override  tripleo-quickstart-config/metalkube-nodes.yml as I mentioned a couple of times before, or alternatively allow for a user vars file via a variable.

Perhaps I can fix that as a follow-up if it's not clear, but given the commit title it seems like we should expose vars to customize, or change the commit message and readme and just change the defaults?

hardys:
Ok since the change from @derekhiggins landed to collect logs it's clear that some pods are getting OOMKilled in CI, so lets land this to improve that situation and I'll push a follow-up to fix the interface nits we've been discussing.

PR #222, opened 2019-03-26
Ingress controller - set number of pods to 3(number of masters)

yboaron:

PR #261, opened 2019-04-01
ingress: set HA type to UserDefined

vrutkovs:
This PR unblocks authorization and console operators.

Tested with #245

### DIFF
--- a/07_deploy_masters.sh
+++ b/07_deploy_masters.sh
@@ -67,6 +67,11 @@ done
 
 wait_for_bootstrap_event
 
+# Reconfigure ingress operator to use host ports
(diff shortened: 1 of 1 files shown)


derekhiggins:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/472/

celebdor:
@vrutkovs can you explain what difference the userDefined setting changes to fix this? The commit message doesn't go a long way to explain that :-)

vrutkovs:
userDefined would ensure ingress operator won't attempt to use cloud LB.

#222 is a proper way to fix it

PR #262, opened 2019-04-01
Set volumeSource to emptyDir in image-registry operator

vrutkovs:
This makes image-registry operator pass. In the prod version Rook should be deployed first to avoid using emptyDir for registry.

Once Rook code is finalized this config should be updated

### DIFF
--- a/assets/templates/99_registry.yaml
+++ b/assets/templates/99_registry.yaml
@@ -1,0 +1,20 @@
+# FIXME: Reconfigure this once Rook is deployed
+apiVersion: imageregistry.operator.openshift.io/v1
+kind: Config
+metadata:
+  generation: 1
+  name: cluster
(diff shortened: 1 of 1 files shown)


derekhiggins:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/473/

07_deploy_masters.sh
```diff
@@ -67,6 +67,11 @@ done
 
 wait_for_bootstrap_event
 
+# Image registry should temporarily use emptyDirs
```
hardys: Is it worth mentioning here that eventually rook will be deployed first so this won't be needed?

vrutkovs: Reworked this to use assets, added a comment in the template

PR #264, opened 2019-04-02
Make master node mutation be script safe

eparis:
oc get is for humans and is not script safe. Use a "-o" so we get something
script safe.

Drop all usage of --config since it was used inconsitenctly and the kubeconfig
was already set by common.sh

Stop doing bash loops for operations which are selectable across labels
by kubernetes.


derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/475/

celebdor:
Thanks!

hardys:
lgtm, thanks!

PR #265, opened 2019-04-02
README: describe the purpose and future of this repository

ktdreyer:
Explain this repository to new users, and show how we expect to promote the workflow concepts after they bake in:

  dev-scripts.git -> kni-installer.git -> openshift-installer.git

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,15 @@ This set of scripts configures some libvirt VMs and associated
 This is very similar to how we do TripleO testing so we reuse some roles
 from tripleo-quickstart here.
 
+We are using this repository as a work space while we figure out what the
+installer needs to do for bare metal provisioning. As that logic is ironed out,
+we are moving it into the [go-based
+kni-installer](https://github.com/openshift-metalkube/kni-installer).
+Eventually that fork of
+[openshift-installer](https://github.com/openshift/installer) will be merged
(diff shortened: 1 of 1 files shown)


derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/476/

README.md
```diff
@@ -7,6 +7,15 @@ This set of scripts configures some libvirt VMs and associated
 This is very similar to how we do TripleO testing so we reuse some roles
 from tripleo-quickstart here.
 
+We are using this repository as a work space while we figure out what the
+installer needs to do for bare metal provisioning. As that logic is ironed out,
+we are moving it into the [go-based
+kni-installer](https://github.com/openshift-metalkube/kni-installer).
```
hardys: I think the logic from the scripts will end up in either (kni|openshift)-install or the facet wrapper API:

https://github.com/openshift-metalkube/facet/tree/master/pkg/server

This will probably be where we put things that don't fit into the initial installer flow, e.g preparing node inventory via discovery so may be worth mentioning, also each of the scripts roughly maps to some API operation that the UI will eventually consume (even if the facet API just calls kni-install underneath)

ktdreyer: Added at https://github.com/openshift-metalkube/dev-scripts/pull/268

PR #270, opened 2019-04-02
add custom version of make-bm-worker

dhellmann:
Add a version of the tool for making a baremetal worker that knows
how to set the image properties to trigger provisioning immediately.

### DIFF
--- a/make-bm-worker/README.md
+++ b/make-bm-worker/README.md
@@ -1,0 +1,69 @@
+Registering Bare Metal Hosts
+============================
+
+The `make-bm-worker` tool may be a more convenient way of creating
+YAML definitions for workers than editing the files directly.
+
+```
+$ go run make-bm-worker/main.go -address ipmi://192.168.111.1:6233 -password password -user admin worker-99
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: worker-99-bmc-secret
+type: Opaque
+data:
+  username: YWRtaW4=
+  password: cGFzc3dvcmQ=
+
+---
+apiVersion: metalkube.org/v1alpha1
+kind: BareMetalHost
+metadata:
+  name: worker-99
+spec:
+  online: true
+  bmc:
+    address: ipmi://192.168.111.1:6233
+    credentialsName: worker-99-bmc-secret
+```
+
+The output can be passed directly to `oc apply` like this:
+
+```
+$ go run make-bm-worker/main.go -address ipmi://192.168.111.1:6233 -password password -user admin worker-99 | oc apply -f -
+```
+
+Include the `-image` option to include the image settings needed to
+trigger immediate provisioning:
+
+```
+$ go run make-bm-worker/main.go -address ipmi://192.168.111.1:6233 -password password -user admin -image worker-99
+---
+apiVersion: v1
+kind: Secret
+metadata:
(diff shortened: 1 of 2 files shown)


derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/494/

make-bm-worker/README.md
```diff
@@ -0,0 +1,68 @@
...
+  online: true
+  bmc:
+    address: 1.2.3.4
+    credentialsName: worker-99-bmc-secret
+
+  userData:
+    namespace: openshift-machine-api
+    name: worker-user-data
+  image:
+    url: "ERROR: template: yaml_out:26:13: executing "yaml_out" at <.ImageSourceURL>: can't evaluate field ImageSourceURL in type main.TemplateArgs
```
hardys: What should the url be when this isn't an error - do we pass the URL for the webserver on the host for now, until we figure out where to host the image on the cluster?

hardys: Tested and it works fine, so I guess we just need to fix the README?

```
    url: "http://172.22.0.1/images/redhat-coreos-maipo-latest.qcow2"
    checksum: "http://172.22.0.1/images/redhat-coreos-maipo-latest.qcow2.md5sum"
```

make-bm-worker/README.md
```diff
@@ -0,0 +1,68 @@
...
+  online: true
+  bmc:
+    address: 1.2.3.4
+    credentialsName: worker-99-bmc-secret
+```
+
+The output can be passed directly to `oc apply` like this:
+
+```
+$ go run make-bm-worker/main.go -address 1.2.3.4 -password password -user admin worker-99 | oc apply -f -
```
hardys: I wonder if we should make the examples more cut/paste friendly for the default virt environment?

E.g I've done:

```
$ go run make-bm-worker/main.go -address ipmi://192.168.111.1:6233 -password password -user admin -image worker-0 | oc --config ocp/auth/kubeconfig apply -f -
secret/worker-0-bmc-secret created
baremetalhost.metalkube.org/worker-0 created
```

That's not actually working yet with the demo operator running, but it'd be nice to have an out-of-the box example given the scope of this repo?

make-bm-worker/README.md
dhellmann: D'oh, I didn't notice that error message. I've updated it with a real value.

The actual URL to use will depend on the resolution to the question of how we're going to serve that image once the cluster is online. As long as the server on the provisioning host is available, the value there now will work.

make-bm-worker/README.md
dhellmann: When you say "demo operator" do you mean the baremetal operator with the -demo-mode flag? Because that iteration doesn't actually provision, it looks for hosts with specific names to move them all into the various states so the UX team can see what they look like.

hardys: Ah yeah I'd missed that, I was running `demo/launch_demo_bmo.sh` and expecting a worker to get built, I guess we can add a variable that allows that in a future PR, or copy/paste the necessary stuff into a `0N_deploy_workers.sh` pending fully integrated worker deployment

dhellmann:
Updated to include a valid address for a virt worker.

PR #273, opened 2019-04-02
Feature/deploy kafka

jparrill:
This PR achieves to deploy:

- Strimzi operator
- Kafka + Zookeeper cluster 
- Prometheus + Alert rules + Alertmanager
- Grafana + Prometheus Datasource and Dashboards to cover Kafka rules
- Kafka Producer/Consumer to send messages and recover those from Kafka cluster (by default 10 msg/sec/pod)
- All isolated into a namespace (By default: strimzi)

Configurable by `common.sh` and integrated with `./03_ocp_repo_sync.sh` script to pre-download repositories


derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/497/

11_deploy_kafka.sh
```diff
@@ -0,0 +1,75 @@
...
+
+# Recover Grafana Dashboard
+wget -q https://raw.githubusercontent.com/ppatierno/rh-osd-2018/master/grafana-dashboards/strimzi-kafka.json -O metrics/examples/grafana/strimzi-kafka.json
+wget -q https://raw.githubusercontent.com/ppatierno/rh-osd-2018/master/grafana-dashboards/strimzi-zookeeper.json -O metrics/examples/grafana/strimzi-zookeeper.json
+
+# Add Grafana Dashboards and Datasource
+GRAFANA_ROUTE=`oc get route grafana --template='{{ .spec.host }}'`
+PROMETHEUS_ROUTE=`oc get route prometheus --template='{{ .spec.host }}'`
+curl -X "POST" "http://${GRAFANA_ROUTE}/api/datasources" -H "Content-Type: application/json" --user admin:admin --data-binary '{ "name":"Prometheus","type":"prometheus","access":"proxy","url":"http://'${PROMETHEUS_ROUTE}'","basicAuth":false,"isDefault":true }'
+curl -X "POST" "http://grafana-kafka.apps.ostest.test.metalkube.org/api/dashboards/db" -H "Content-Type: application/json;charset=UTF-8"  --user admin:admin --data-binary @metrics/examples/grafana/strimzi-kafka.json
```
e-minguez: Use ${GRAFANA_ROUTE} instead

11_deploy_kafka.sh
```diff
@@ -0,0 +1,75 @@
...
+curl -X "POST" "http://grafana-kafka.apps.ostest.test.metalkube.org/api/dashboards/db" -H "Content-Type: application/json;charset=UTF-8"  --user admin:admin --data-binary @metrics/examples/grafana/strimzi-zookeeper.json
```
e-minguez: Use ${GRAFANA_ROUTE} instead

11_deploy_kafka.sh
```diff
@@ -0,0 +1,75 @@
...
+oc expose svc grafana -n ${KAFKA_NAMESPACE} || echo "Grafana route already exists" 
+
+# Recover Grafana Dashboard
+wget -q https://raw.githubusercontent.com/ppatierno/rh-osd-2018/master/grafana-dashboards/strimzi-kafka.json -O metrics/examples/grafana/strimzi-kafka.json
+wget -q https://raw.githubusercontent.com/ppatierno/rh-osd-2018/master/grafana-dashboards/strimzi-zookeeper.json -O metrics/examples/grafana/strimzi-zookeeper.json
+
+# Add Grafana Dashboards and Datasource
+GRAFANA_ROUTE=`oc get route grafana --template='{{ .spec.host }}'`
+PROMETHEUS_ROUTE=`oc get route prometheus --template='{{ .spec.host }}'`
+curl -X "POST" "http://${GRAFANA_ROUTE}/api/datasources" -H "Content-Type: application/json" --user admin:admin --data-binary '{ "name":"Prometheus","type":"prometheus","access":"proxy","url":"http://'${PROMETHEUS_ROUTE}'","basicAuth":false,"isDefault":true }'
```
e-minguez: To avoid traffic going out of the cluster, then going back, wouldn't be better to use the prometheus service instead?

jparrill: I also think that is the best way, but I've been suffering some fails regarding DNS issues and for now I think it's better to just use the external. I will fill an issue to change it to the SVC when it stays stable. Makes sense?

derekhiggins:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/501/

jparrill:
> Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/501/

Seems like is not creating the nodes ://

```
...
...
Apr 03 10:08:52 |  + oc get nodes
Apr 03 10:08:52 |  error: the server doesn't have a resource type "nodes"
Apr 03 10:08:52 |  + sleep 5
Apr 03 10:08:57 |  + oc Build timed out (after 90 minutes). Marking the build as failed.
```

hardys:
Do we have any info on the resource requirements for this?

The reason I ask is we're discussing an increase to the memory/cpu for VM test environments ref https://github.com/openshift-metalkube/dev-scripts/pull/254 (that may also help with some of the spurious CI failures) but it'd be good to understand if this will require much more resources to test successfully?

jparrill:
> Do we have any info on the resource requirements for this?
> 
> The reason I ask is we're discussing an increase to the memory/cpu for VM test environments ref #254 (that may also help with some of the spurious CI failures) but it'd be good to understand if this will require much more resources to test successfully?

But this fail happens before the script execution, the first time that build was finished, the execution were fine.

> Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/497/

This ^^

derekhiggins:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/510/

karmab:
@jparrill maybe it makes sense to move this script to an extras dir, as it's not really part of kni, but corresponds to an additional workflow that we want to implement at this point of time.
@hardys thoughts ?

jparrill:
> @jparrill maybe it makes sense to move this script to an extras dir, as it's not really part of kni, but corresponds to an additional workflow that we want to implement at this point of time.
> @hardys thoughts ?

Makes totally sense, @hardys if you agree I'll create an extra folder to add it into it.

jparrill:
@karmab @hardys it's done

hardys:
I'm fine with the extras dir, but I'd still like an answer to my question re resource usage - the previous reply that CI passed doesn't help because this stuff isn't tested in CI - I want to know how much memory we realistically need for folks to test this in the VM based setup locally.

hardys:
Also see my comments on https://github.com/openshift-metalkube/dev-scripts/pull/302 - we do need to start thinking about how this will work when using only kni-installer, e.g installer assets (or deploying layered products via some other operator) vs these dev-only scripts.

karmab:
the reason why we use an extras dir is because we believe that both kafka and knative (and other items) are not needed for kni (and as such shouldnt be moved to the kni installer).
we just need them for incoming demo purposes and should get rid of them afterward

jparrill:
@hardys anwering your question about resources:

for Kafka:
```
resources:
  requests:
    memory: 2Gi
    cpu: 500m
  limits:
    memory: 2Gi
    cpu: "1"
```

For Zookeeper:
```
resources:
  requests:
    memory: 1Gi
    cpu: "0.3"
  limits:
    memory: 1Gi
    cpu: "0.5"
```

Thi is what recommends Strimzi-Kafka folks for development environment that will support: 10-40 msgs/sec to the cluster

hardys:
Ok sounds good, perhaps we can follow-up with a README in the extras dir explaining the scripts are to demonstrate things on top of the deployed cluster, and that they may not work with the default VM sizes.

jparrill:
> Ok sounds good, perhaps we can follow-up with a README in the extras dir explaining the scripts are to demonstrate things on top of the deployed cluster, and that they may not work with the default VM sizes.

Yeah, will be usefull.

I ussually modify the node specs with those `sed` commands:

```
sed -i "s/default_vcpu: 4/default_vcpu: 8/" tripleo-quickstart-config/metalkube-nodes.yml
sed -i "s/default_memory: 8192/default_memory: 16384/" tripleo-quickstart-config/metalkube-nodes.yml
sed -i "s/extradisks_size: 8G/extradisks_size: 70G/"  tripleo-quickstart-config/roles/libvirt/defaults/main.yml
```

Issue #277, opened 2019-04-03
Version checks for RHEL fail in common.sh


sreichar:
Check for RHEL 7 fail with RHEL 7.6 errors - several errors - 1 example

++ sudo -n uptime
+++ awk -F= '/^ID=/ { print $2 }' /etc/os-release
+++ tr -d '"'
++ [[ ! rhel =~ ^(centos|rhel)$ ]]
+++ awk -F= '/^VERSION_ID=/ { print $2 }' /etc/os-release
+++ tr -d '"'
++ [[ 7.6 -ne 7 ]]
common.sh: line 84: [[: 7.6: syntax error: invalid arithmetic operator (error token is ".6")
+++ df / --output=fstype
+++ grep -v Type
++ FSTYPE=xfs

hardys:
Fixes landed so closing this, please reopen if any issues remain, or raise a follow-up issue

PR #281, opened 2019-04-03
Some fixes to help with issues seen with RHEL

sreichar:
Address #276 and #277

### DIFF
--- a/01_install_requirements.sh
+++ b/01_install_requirements.sh
@@ -40,7 +40,8 @@ sudo yum -y install \
   python-requests \
   python-setuptools \
   vim-enhanced \
-  wget
+  wget \
+  jq
 
 # We're reusing some tripleo pieces for this setup so clone them here
 cd
--- a/01_install_requirements.sh
+++ b/01_install_requirements.sh
@@ -1,7 +1,6 @@
 #!/usr/bin/env bash
 set -ex
 
-source common.sh
 
 # FIXME ocp-doit required this so leave permissive for now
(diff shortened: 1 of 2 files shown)


01_install_requirements.sh
```diff
@@ -43,6 +42,9 @@ sudo yum -y install \
   wget \
   jq
 
+# moved to after all called software is installed
+source common.sh
```
hardys: The problem with this is we set up logging in common.sh, so if anything fails before this, it's not logged (which will be bad, particularly in CI)

Perhaps we can break that into a separate file, e.g common_logging.sh and source that here, then either include that via common.sh, or update all the scripts that need it to source both files?

cc @stbenjam

derekhiggins:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/509/

01_install_requirements.sh
derekhiggins: > The problem with this is we set up logging in common.sh, so if anything fails before this, it's not logged (which will be bad, particularly in CI)

It shouldn't be a problem for CI as we also log the console which will get any errors in this script, it would be more of a problem for people trying to figure out what happened in their own env when scrollback is gone.

> 
> Perhaps we can break that into a separate file, e.g common_logging.sh and source that here, then either include that via common.sh, or update all the scripts that need it to source both files?
> 
> cc @stbenjam

sreichar:
jq fix not needed - actuall another issue

PR #282, opened 2019-04-03
create dedicated baremetal bridge on nodes

karmab:
This makes sure that both masters and workers get created with a brext bridge where the nic providing external access is attached.
The nic to pick is detected by checking which one has the default route
This allows to then create kubevirt vms which leverage this same bridge and as such can be directly accessed from the outside (without a need to expose services)

### DIFF
--- a/assets/files/etc/NetworkManager/dispatcher.d/98-brextscript
+++ b/assets/files/etc/NetworkManager/dispatcher.d/98-brextscript
@@ -1,0 +1,15 @@
+#!/usr/bin/env bash
+
+IF=$(ip r | grep default | grep -Po '(?<=dev )(\S+)')
+if [ ! -f /etc/sysconfig/network-scripts/ifcfg-$IF ] ; then
+echo brext acting on $IF
+cat <<EOF > /etc/sysconfig/network-scripts/ifcfg-$IF
+DEVICE=$IF
+BRIDGE=brext
+ONBOOT=yes
+NM_CONTROLLED=yes
+BOOTPROTO=none
+EOF
+systemctl restart NetworkManager
+nmcli conn down $IF
+fi
--- a/assets/files/etc/sysconfig/network-scripts/ifcfg-brext
+++ b/assets/files/etc/sysconfig/network-scripts/ifcfg-brext
@@ -1,0 +1,6 @@
+DEVICE=brext
+NAME=brext
+TYPE=Bridge
+ONBOOT=yes
+NM_CONTROLLED=yes
+BOOTPROTO=dhcp
--- a/assets/templates/98_master-brext.yaml
+++ b/assets/templates/98_master-brext.yaml
@@ -1,0 +1,23 @@
+apiVersion: machineconfiguration.openshift.io/v1
+kind: MachineConfig
+metadata:
+  generation: 1
+  labels:
+    machineconfiguration.openshift.io/role: master
+  name: 98-master-brext
+spec:
+  config:
+    ignition:
+      version: 2.2.0
+    storage:
+      files:
+      - contents:
+          verification: {}
+        filesystem: root
+        mode: 420
+        path: /etc/sysconfig/network-scripts/ifcfg-brext
+      - contents:
+          verification: {}
+        filesystem: root
+        mode: 493
+        path: /etc/NetworkManager/dispatcher.d/98-brextscript
--- a/assets/templates/98_worker-brext.yaml
+++ b/assets/templates/98_worker-brext.yaml
@@ -1,0 +1,23 @@
+apiVersion: machineconfiguration.openshift.io/v1
+kind: MachineConfig
+metadata:
+  generation: 1
+  labels:
+    machineconfiguration.openshift.io/role: worker
+  name: 98-worker-brext
+spec:
+  config:
+    ignition:
+      version: 2.2.0
+    storage:
+      files:
+      - contents:
+          verification: {}
+        filesystem: root
+        mode: 420
+        path: /etc/sysconfig/network-scripts/ifcfg-brext
+      - contents:
+          verification: {}
+        filesystem: root
+        mode: 493
+        path: /etc/NetworkManager/dispatcher.d/98-brextscript


assets/files/etc/sysconfig/network-scripts/ifcfg-eth1
```diff
@@ -0,0 +1,6 @@
+DEVICE=eth1
+NAME=eth1
```
hardys: Can you clarify which interface we expect to use here, the expected initial architecture I think is that we have only two, one for external/k8s cluster usage (which can include vlan isolation potentially), and another that's dedicated to provisioning but could potentially be repurposed after initial deployment.

It seems you're planning to repurpose the provisioning nic here?  I was thinking we might reuse it for e.g storage traffic on a dedicated subnet/vlan, but that we probably couldn't use that physical fabric for tenant traffic, as it's potentially going to have a route to BMC IPs etc?

karmab: no, eth1 is still for external cluster usage, in both physical and virtual env

hardys:
Added @celebdor and @derekhiggins so we can get this requirement aligned with the network/hardware team tasks - it seems like a reasonable requirement but I'm not yet familiar with how bridging the external nic will play with the OCP services...

assets/files/etc/NetworkManager/dispatcher.d/40-baremetalupdatescript
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
+ip link set baremetal up
+#ip link set baremetal type bridge vlan_filtering 1
```
dankenigsberg: Why I is commented out?

assets/files/etc/sysconfig/network-scripts/ifcfg-baremetal
```diff
@@ -0,0 +1,5 @@
+DEVICE=baremetal
```
dankenigsberg: In KNI context everything is bare metal, so this bridge name means very little. How about `brext`, making clear that it is a bridge that is connected to the external world?

derekhiggins:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/512/

derekhiggins:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/519/

assets/files/etc/sysconfig/network-scripts/ifcfg-baremetal
karmab: i m using baremetal for the bridge name,  since it s how it s named on the hypervisor side, when using libvirt setup.

dankenigsberg: Names should be set according to how customers see them, not according to what we used in our virtual test environment. What would a customer understand when she sees `baremetal` on her machine?

I personally think that the we should use `ext` in libvirt, and `brext` for the on-host bridge name.
@celebdor @derekhiggins  do you have an opinion about that?

assets/files/etc/NetworkManager/dispatcher.d/pre-up.d/z-baremetalupdatescript
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
+
+IF=ens4
+BRIDGE=baremetal
+if [ "$1" == "$IF" ] && [ -z "$( ip a | grep $IF | grep $BRIDGE)" ] ; then
+ cp -f /etc/resolv.conf /etc/resolv.conf.ori
+ DEFAULT_ROUTE=$( ip r | grep default | grep $IF )
+ for ip in $(ip -o addr |grep $IF | grep -v inet6 | awk '{print $4}') ; do
```
celebdor: filtering out ipv6 addressing that was put there by the dc dhcp seems a bit evil

assets/files/etc/NetworkManager/dispatcher.d/pre-up.d/z-baremetalupdatescript
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
+
+IF=ens4
```
celebdor: This should use the logic in /usr/local/bin/keepalived.sh to determine the right interface.

assets/files/etc/sysconfig/network-scripts/ifcfg-baremetal
celebdor: `ext` and `brext` feel like the right choice.

assets/files/etc/sysconfig/network-scripts/ifcfg-ens4
```diff
@@ -0,0 +1,5 @@
+DEVICE=ens4
```
celebdor: This file should be a template that is rendered by the dispatcher script once it learns which interface is the right one.

assets/files/etc/NetworkManager/dispatcher.d/pre-up.d/z-baremetalupdatescript
derekhiggins: Ya, on the dev2 baremetal env the interfaces are named differently

> 2: eno1: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc mq state UP mode DEFAULT group default qlen 1000\    link/ether e4:43:4b:52:79:54 brd ff:ff:ff:ff:ff:ff
> 3: eno2: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc mq state UP mode DEFAULT group default qlen 1000\    link/ether e4:43:4b:52:79:55 brd ff:ff:ff:ff:ff:ff

assets/files/etc/sysconfig/network-scripts/ifcfg-baremetal
karmab: switched to brext

assets/files/etc/NetworkManager/dispatcher.d/pre-up.d/z-baremetalupdatescript
karmab: changed the logic

assets/files/etc/NetworkManager/dispatcher.d/pre-up.d/z-baremetalupdatescript
karmab: changed the logic so that we rely on the interface which has default gw

assets/files/etc/sysconfig/network-scripts/ifcfg-ens4
karmab: switched to a template file created in the dispatcher script

assets/files/etc/NetworkManager/dispatcher.d/brextscript
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
```
dankenigsberg: from NetworkManager man page I learn that this script is executed quite often. I don't think it should ever generate another ifcfg file or modify its own con state. Wouldn't it create an endless mess?

        NetworkManager will execute scripts in the /etc/NetworkManager/dispatcher.d directory or subdirectories in alphabetical order in
       response to network events. Each script should be a regular executable file owned by root. Furthermore, it must not be writable by
       group or other, and not setuid.

       Each script receives two arguments, the first being the interface name of the device an operation just happened on, and second the
       action. For device actions, the interface is the name of the kernel interface suitable for IP configuration. Thus it is either
       VPN_IP_IFACE, DEVICE_IP_IFACE, or DEVICE_IFACE, as applicable. For the hostname and connectivity-change actions it is always "none".

karmab: changed so it will only generate needed file once

assets/files/etc/sysconfig/network-scripts/ifcfg-ens4
dankenigsberg: @celebdor @karimb could you give a top level review of what you attempt to do in this change?

If find it odd that you look for the default route (set by dhcp?) and attach a bridge to it, but without dhcp? Why restart NM? Why do it within the dispatcher script? And why without any care to the args passed by NM to it?

If expect us to know the nic name in advance, and simply drop here the ifcfg files for it and it's bridge, with dhcp enabled on the bridge.

russellb:
Can you add a description to this PR that explains why it is needed?  Thanks!

dankenigsberg:
Joining @russellb, I'd appreciate a top level explanation of this pr. Despite knowing its motivation, I don't understand how the host is expected to find which nic it should be used to connect virtual machines. I think that the name of the nic should be passed as an explicit argument of the install script.

karmab:
> Can you add a description to this PR that explains why it is needed? Thanks!

done

assets/files/etc/sysconfig/network-scripts/ifcfg-ens4
```diff
@@ -0,0 +1,5 @@
...
+BRIDGE=brext
```
hardys: We've resolved the hard-coded device name in 98-brextscript but I'm wondering how we can move away from this hard-coded file?

hardys:
I think we need to figure out how to avoid any hardcoded nic name, e.g this doesn't work in the default VM case:

```
[core@master-2 network-scripts]$ pwd
/etc/sysconfig/network-scripts

[core@master-2 network-scripts]$ ls
ifcfg-  ifcfg-brext  ifcfg-enp2s0  ifcfg-eth0  ifcfg-eth1

[core@master-2 network-scripts]$ grep BRIDGE ./*
./ifcfg-:BRIDGE=brext
./ifcfg-eth1:BRIDGE=brext

[core@master-2 network-scripts]$ cat ifcfg-
DEVICE=
BRIDGE=brext
ONBOOT=yes
NM_CONTROLLED=yes
BOOTPROTO=none
```

assets/files/etc/sysconfig/network-scripts/ifcfg-ens4
karmab: my bad, this is a left over, this file is no longer added to the nodes and we instead generate it on the file through the brext script

karmab:
> I think we need to figure out how to avoid any hardcoded nic name, e.g this doesn't work in the default VM case:
> 
> ```
> [core@master-2 network-scripts]$ pwd
> /etc/sysconfig/network-scripts
> 
> [core@master-2 network-scripts]$ ls
> ifcfg-  ifcfg-brext  ifcfg-enp2s0  ifcfg-eth0  ifcfg-eth1
> 
> [core@master-2 network-scripts]$ grep BRIDGE ./*
> ./ifcfg-:BRIDGE=brext
> ./ifcfg-eth1:BRIDGE=brext
> 
> [core@master-2 network-scripts]$ cat ifcfg-
> DEVICE=
> BRIDGE=brext
> ONBOOT=yes
> NM_CONTROLLED=yes
> BOOTPROTO=none
> ```

i didnt hit this issue and believe we re not using any harcoded name in this PR at this point

hardys:
The same problem I had before still exists for the virt platform, we try to add an interface with an empty string to the bridge, so it's not going to do anything:

Also brctl isn't available, @derekhiggins how did you make that work on baremetal, inside a container?

```
[root@master-0 ~]# brctl show
-bash: brctl: command not found
[root@master-0 ~]#  ip -o addr | grep brext
4: brext    inet 192.168.111.20/24 brd 192.168.111.255 scope global dynamic noprefixroute brext\       valid_lft 3265sec preferred_lft 3265sec
4: brext    inet6 fe80::2a4:20ff:fe1b:7fb0/64 scope link \       valid_lft forever preferred_lft forever
[root@master-0 ~]# cat /etc/sysconfig/network-scripts/ifcfg-
ifcfg-        ifcfg-brext   ifcfg-enp2s0  ifcfg-ens4    
[root@master-0 ~]# cat /etc/sysconfig/network-scripts/ifcfg-brext 
DEVICE=brext
NAME=brext
TYPE=Bridge
ONBOOT=yes
NM_CONTROLLED=yes
BOOTPROTO=dhcp
[root@master-0 ~]# cat /etc/sysconfig/network-scripts/ifcfg-
DEVICE=
BRIDGE=brext
ONBOOT=yes
NM_CONTROLLED=yes
BOOTPROTO=none
```

hardys:
Actually maybe this will work now as there's also ens4 added to the bridge which I didn't see before, so we just have to clean up the spurious `ifcfg-` file?

```
[root@master-0 ~]# cat /etc/sysconfig/network-scripts/ifcfg-ens4 
DEVICE=ens4 
BRIDGE=brext
ONBOOT=yes
NM_CONTROLLED=yes
BOOTPROTO=none
[root@master-0 ~]# ip a | grep ens
2: ens3: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UP group default qlen 1000
    inet 172.22.0.13/24 brd 172.22.0.255 scope global dynamic noprefixroute ens3
3: ens4: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc fq_codel master brext state UP group default qlen 1000
```

hardys:
Copying some discussion on this topic - @karmab did you see this, and any thoughts on if this is related to the spurious ifcfg file?

> Is the issue that 98-brextscript should look at the arguments to figure out when it's time to run, vs running every event?
> https://developer.gnome.org/NetworkManager/unstable/NetworkManager.html indicates we get a device/action and I don't see any conditional in the dispatcher script?
> and/or maybe this should be a pre-up.d that only triggers prior to brext coming up?

russellb:
Revisiting this ... I'm not sure this is something we should merge.  It was fine for demo purposes, but long term, I'd like to see host network state as something we can manage day 2.  I imagine an operator that can apply host network configuration changes based on CRDs.

Consider also that CNV and rook are components that could be deployed after the fact on a baremetal cluster, and they need to work.  This environment where we know they will be present from day 1 is not always the case.

To summarize, I'm -1 on merging this without a more comprehensive plan for host network management laid out.  I just spoke with @celebdor briefly about his team helping here.

hardys:
Ok so discussion indicates https://github.com/kubevirt/hyperconverged-cluster-operator should ultimately handle this as a day2 operation, since at the time of deploying the controlplane we don't actually know if kubevirt (or ceph, or any other layered application) will be deployed.

We have to move all day-1 hacks out of dev-scripts into operators (or the kni-installer and ultimately openshift/installer) so lets close this and focus on the final solution instead.

PR #285, opened 2019-04-03
 Address issue #277

sreichar:
RHEL's VERSION_ID is set to 7.X versus just 7 for CentOS,  cut after the '.' if it exist.


derekhiggins:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/517/

PR #302, opened 2019-04-05
knative install script

karmab:


hardys:
It'd be good to move away from the scripts soon, instead using installer assets, but otherwise lgtm

PR #305, opened 2019-04-08
Fixed repositories logic for RHEL vs CentOS

gklein:


01_install_requirements.sh
```diff
@@ -13,9 +13,9 @@ sudo yum -y update
 # Install EPEL required by some packages
 if [ ! -f /etc/yum.repos.d/epel.repo ] ; then
     if grep -q "Red Hat Enterprise Linux" /etc/redhat-release ; then
-        sudo yum -y install http://mirror.centos.org/centos/7/extras/x86_64/Packages/epel-release-7-11.noarch.rpm
-    else
         sudo yum -y install epel-release --enablerepo=extras
+    else
```
hardys: Can you provide more details on this - it switches the previous order and IME yum install epel-release on CentOS works fine, e.g we don't need to point to the RPM directly in that case, but we do in the RHEL case?

gklein: Not sure why it failed once, but it seems to work on both RHEL and CentOS now.

PR #307, opened 2019-04-08
Fix: error locating the common script on extra folder

jparrill:
Fixing a small error locating common script on new extra folder

### DIFF
--- a/extras/common.sh
+++ b/extras/common.sh
@@ -1,0 +1,1 @@
+../common.sh
--- a/extras/common.sh
+++ b/extras/common.sh
@@ -1,1 +1,0 @@
-../common.sh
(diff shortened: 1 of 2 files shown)


extras/common.sh
```diff
@@ -0,0 +1 @@
+../common.sh
```
hardys: This won't work if you run from some PWD that's not extras, instead you could use `$SCRIPTDIR` like:

https://github.com/openshift-metalkube/dev-scripts/blob/master/demo/launch_demo_bmo.sh#L11

jparrill: SCRIPTDIR will not xists on kafka.sh context I've just create a BASEDIR pointing to base repo dir that should do the same

hardys: Yeah I was suggesting you copy the same approach

PR #286, opened 2019-04-03
Set up the node -> machine mapping

bcrochet:
This manually updates the status of the Machine objects to contains:

1) nodeRef pointing back to the Node
2) addresses of the Node, copied to the Machine

PR #312, opened 2019-04-08
Run fix_certs script automatically

zaneb:
Because life is too short for broken certs.

### DIFF
--- a/06_create_cluster.sh
+++ b/06_create_cluster.sh
@@ -46,6 +46,10 @@ fi
 # Call kni-installer to deploy the bootstrap node and masters
 create_cluster ocp
 
+# Run the fix_certs.sh script periodically as a workaround for
+# https://github.com/openshift-metalkube/dev-scripts/issues/260
+sudo systemd-run --on-active=30s --on-unit-active=30m --unit=fix_certs.service $(dirname $0)/fix_certs.sh
+
 # Update kube-system ep/host-etcd used by cluster-kube-apiserver-operator to
 # generate storageConfig.urls
 patch_ep_host_etcd "$CLUSTER_DOMAIN"
--- a/fix_certs.sh
+++ b/fix_certs.sh
@@ -1,6 +1,8 @@
 #!/bin/bash
 
-# https://github.com/openshift-metalkube/dev-scripts/issues/141#issuecomment-474331659
+# https://github.com/openshift-metalkube/dev-scripts/issues/260
 
 export KUBECONFIG=$(dirname $0)/ocp/auth/kubeconfig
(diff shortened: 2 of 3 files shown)


06_create_cluster.sh
```diff
@@ -58,3 +58,5 @@ oc adm taint nodes -l node-role.kubernetes.io/master node-role.kubernetes.io/mas
 
 wait_for_cvo_finish ocp
 echo "Cluster up, you can interact with it via oc --config ${KUBECONFIG} <command>"
+
+sudo systemd-run --on-active=30s --on-unit-active=30m --unit=fix_certs.service $(dirname $0)/fix_certs.sh
```
hardys: I wonder if we should do this earlier, e.g right after the cluster create, or maybe when we see the masters are up?

Currently we're seeing timeouts waiting for the CVO to finish, but the cluster is mostly functional - in this case I guess we'd want the fix_certs workaround even if we timed out waiting on the CVO?

zaneb: That makes sense - you want it to be running even if it times out or you hit Ctrl-C on the script. I don't actually know at which point in the process it becomes possible to run the script successfully? (We don't want to start it too early, because then it won't try again for another 30 minutes.) Is it as soon as the master nodes are up?

hardys: I guess ideally we'd want to wait until the k8s API comes up and the csr API becomes available - probably we also need that for #286 as currently that assumes the API is ready as soon as the masters take DHCP leases, which is probably wrong.

Previously we waited for the bootstrap-complete, so we could do similar to the utils.sh wait_for_cvo_finish, but wait for bootstrap complete (e.g `kni-install upi bootstrap-complete`), then do some stuff, then wait for the CVO to complete (which can take a while).

@vrutkovs any thoughts on the best approach here?

hardys: Actually I think right after the create_cluster function should be OK, because kni-installer waits for the API to come up and bootstrap-complete to happen:

https://github.com/openshift-metalkube/kni-installer/blob/master/cmd/kni-install/create.go#L97

vrutkovs: Certs would exist right after bootstrap complete - these should be approved there. CVO may not finish without those

hardys: Ok sounds like we're in agreement to move the systemd-run to https://github.com/openshift-metalkube/dev-scripts/blob/master/06_create_cluster.sh#L48

derekhiggins:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/546/

PR #325, opened 2019-04-10
disable haproxy service until it is a static pod

celebdor:
Right now api-haproxy still is implemented as a systemd+podman piece of
infra. The problem is that when the podman pod storage breaks (as
happened with the other infra pods), the API ceases to be reachable.

Let's disable it until I reimplement it as a static pod.

### DIFF
--- a/assets/templates/99_master-api-haproxy.yaml
+++ b/assets/templates/99_master-api-haproxy.yaml
@@ -27,7 +27,7 @@ spec:
             [Install]
             WantedBy=multi-user.target
 
-          enabled: true
+          enabled: false
           name: api-haproxy.service
     storage:
       files:


dantrainor:
My deployment failed while testing this:

level=fatal msg="failed to initialize the cluster: Cluster operator machine-config is reporting a failure: Failed to resync 4.0.0-0.alpha-2019-04-10-154442 because: [timed out waiting for the condition during waitForControllerConfigToBeCompleted: controllerconfig is not completed: ControllerConfig has not completed: completed(false) running(false) failing(true), pool master has not progressed to latest configuration: configuration for pool master is empty, retrying]: timed out waiting for the condition"

Does this look to be related?

celebdor:
@dantrainor doesn't seem related to it

PR #329, opened 2019-04-10
Install centos libvirt repos on RHEL

vrutkovs:
This fixes the issue where default RHEL7 libvirt fails to start bootstrap node

### DIFF
--- a/01_install_requirements.sh
+++ b/01_install_requirements.sh
@@ -19,6 +19,26 @@ if [ ! -f /etc/yum.repos.d/epel.repo ] ; then
     fi
 fi
 
+# Install qemu-img-ev on RHEL
+if grep -q "Red Hat Enterprise Linux" /etc/redhat-release ; then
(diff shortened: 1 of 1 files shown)


01_install_requirements.sh
```diff
@@ -19,6 +19,26 @@ if [ ! -f /etc/yum.repos.d/epel.repo ] ; then
     fi
 fi
 
+# Install qemu-img-ev on RHEL
```
hardys: On my system this comes from rdo-qemu-ev, which is enabled via tripleo-repos, do we just need to move the install until after tripleo-repos has been run perhaps?

hardys:
Just to follow up on this, I guess we need to fix https://github.com/openshift-metalkube/dev-scripts/issues/278 then ensure the install happens after the repos are configured?

russellb:
closing as stale

PR #330, opened 2019-04-10
add a script to register hosts

dhellmann:
Add an optional step 11 to register hosts for the masters and workers.

Masters are registered with a machine reference, to prevent them from
showing up as available to be provisioned.

Workers are registered without a machine reference, so they are
available to be provisioned.

Add the missing machine and boot-mac arguments to make-bm-workers.

### DIFF
--- a/11_register_hosts.sh
+++ b/11_register_hosts.sh
@@ -1,0 +1,80 @@
+#!/usr/bin/bash
+
+set -ex
+
+source common.sh
+eval "$(go env)"
+
+# Get the latest bits for baremetal-operator
+export BMOPATH="$GOPATH/src/github.com/metalkube/baremetal-operator"
+
+function list_masters() {
+    cat $MASTER_NODES_FILE | \
+        jq '.nodes[] | {
+           name,
+           driver,
+           address:.driver_info.ipmi_address,
+           port:.driver_info.ipmi_port,
+           user:.driver_info.ipmi_username,
+           password:.driver_info.ipmi_password,
+           mac: .ports[0].address
+           } |
+           .name + " " +
+           .driver + "://" + .address + ":" + .port + " " +
+           .user + " " + .password + " " + .mac' \
+        | sed 's/"//g'
+}
+
+# Register the masters linked to their respective Machine objects.
+function make_bm_masters() {
+    while read name address user password mac; do
+        go run $SCRIPTDIR/make-bm-worker/main.go \
+           -address "$address" \
+           -password "$password" \
+           -user "$user" \
+           -machine-namespace openshift-machine-api \
+           -machine  "$(echo $name | sed 's/openshift/ostest/')" \
+           -boot-mac "$mac" \
+           "$name"
+    done
+}
+
+function list_workers() {
+    # Includes -machine and -machine-namespace
+    cat $NODES_FILE | \
+        jq '.nodes[] | select(.name | contains("worker")) | {
+           name,
+           driver,
+           address:.driver_info.ipmi_address,
+           port:.driver_info.ipmi_port,
+           user:.driver_info.ipmi_username,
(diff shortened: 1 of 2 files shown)


russellb:
Thanks @dhellmann.  I'm going to give this a try.

Can you also file an issue against kni-installer to track eventually moving this there?  kni-install creates the Machines, and it also has the data about the master hosts, so we should extend it to set all of this up as part of the deployment.

hardys:
Thanks this is working well for me using a bmo image that has https://github.com/metalkube/baremetal-operator/pull/128 applied (`quay.io/shardy/baremetal-operator:pr128_test`).

One question, the worker doesn't deploy but ends up powered on and in "manageable" state - is that expected atm?  I expected it to be claimed by the actuator then driven through the provisioning state, but instead it seems we power it on, introspection fails, then it does nothing because the state isn't provisioning, I know we have some missing pieces so I'll look at those next.

```
$ oc --config ocp/auth/kubeconfig get baremetalhosts -n openshift-machine-api
NAME                 STATUS   PROVISIONING STATUS   MACHINE USING THIS HOST   BMC                         ERROR
openshift-master-0   OK       ready                 ostest-master-0           ipmi://192.168.111.1:6230   
openshift-master-1   OK       ready                 ostest-master-1           ipmi://192.168.111.1:6231   
openshift-master-2   OK       ready                 ostest-master-2           ipmi://192.168.111.1:6232   
openshift-worker-0   OK       ready                                           ipmi://192.168.111.1:6233   
$ openstack baremetal node list
+--------------------------------------+--------------------+---------------+-------------+--------------------+-------------+
| UUID                                 | Name               | Instance UUID | Power State | Provisioning State | Maintenance |
+--------------------------------------+--------------------+---------------+-------------+--------------------+-------------+
| c638e487-6080-403d-a5dd-2d37080c95ee | openshift-master-1 | None          | power on    | active             | False       |
| 8a521eb4-b929-4bed-811a-af4f40e5bba7 | openshift-master-2 | None          | power on    | active             | False       |
| ca967ba2-0446-4b20-944a-189d2983db70 | openshift-master-0 | None          | power on    | active             | False       |
| 184af9d8-dd90-4f58-b447-6045caba0bc6 | openshift-worker-0 | None          | power on    | manageable         | False       |
+--------------------------------------+--------------------+---------------+-------------+--------------------+-------------+
$ oc --config ocp/auth/kubeconfig get nodes
NAME       STATUS     ROLES    AGE   VERSION
master-0   Ready      master   75m   v1.13.4+1ad602308
master-1   NotReady   master   76m   v1.13.4+1ad602308
master-2   Ready      master   75m   v1.13.4+1ad602308
$ oc --config ocp/auth/kubeconfig get machines -n openshift-machine-api
NAME                    INSTANCE   STATE   TYPE   REGION   ZONE   AGE
ostest-master-0                                                   75m
ostest-master-1                                                   75m
ostest-master-2                                                   75m
ostest-worker-0-8cswz                                             75m
ostest-worker-0-8plmk                                             75m
ostest-worker-0-rh4r7                                             75m
```

hardys:
I guess in future we could also add a variable to control the number of masters/workers (I know @yrobla was interested in that), then we could generate the right number of VMs, populate the json and set the MachineSet replicas according to the counts.

dhellmann:
@hardys, I expect the node in ironic to be in the manageable state and for the host object to be OK/Ready. Whether or not the actuator picks up the host depends on how many Machine objects exist asking for workers.
I'm not sure what the power state should be. I'll take a look at that.
I don't know why introspection would fail. The operator isn't driving that explicitly, so it's probably relying on an incompletely implemented discovery path.

dhellmann:
Cross referencing to the other issue: https://github.com/openshift-metalkube/kni-installer/issues/46

PR #334, opened 2019-04-11
Added updated Grafana dashboards

ppatierno:
This PR adds a new Kafka Grafana dashboard with some new rates metrics views.
@jparrill tbh I wasn't able to test it locally. Can you do that?


jparrill:
I've tested it and works fine.

LGTM

PR #335, opened 2019-04-11
Add sanity check when getting VIP for nameserver

yrobla:
Initially when the system is started, resolv.conf has some dummy resolv.conf
coming from the tree. When dns-vip-prepender is called the first time, it tries
to use that buggy nameserver and captures the error code of the dig command,
causing to set an incorrect nameserver. This patch changes dig for host, to
better capture error codes, and don't perform any action on resolv.conf if
nameserver cannot be resolved properly.


celebdor:
Could you rebase and squash?

yrobla:
opening a clean one

PR #336, opened 2019-04-11
move api haproxy to static pod

celebdor:
Just like we did for the rest of the kni network infra, move this from a
systemd+podman pod to a kubelet driven static pod


celebdor:
Needs to be tested still

derekhiggins:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/553/

russellb:
CI build worked other than the monitoring operator, which should be unrelated to this change.  I've also seen it work completely twice in a row in my env, so fine with me to merge.

PR #339, opened 2019-04-11
Make VM node specs configurable and bump defaults

hardys:
Add a VM_NODES_FILE variable, so that you can copy and modify
the metalkube-nodes.yml ansible vars file if desired (e.g to add
extra memory, or test with more/less nodes)

e.g

cp tripleo-quickstart-config/metalkube-nodes.yml ~
vim ~/metalkube-nodes.yml
echo "export VM_NODES_FILE=\"/home/$USER/metalkube-nodes.yml\"" >> config_$USER.sh

Proceed with caution if modifying this file, as the most well tested path
is using the defaults.

### DIFF
--- a/02_configure_host.sh
+++ b/02_configure_host.sh
@@ -11,12 +11,13 @@ source ocp_install_env.sh
 # Note we copy the playbook so the roles/modules from tripleo-quickstart
 # are found without a special ansible.cfg
 export ANSIBLE_LIBRARY=./library
+export VM_NODES_FILE=${VM_NODES_FILE:-tripleo-quickstart-config/metalkube-nodes.yml}
 
 ANSIBLE_FORCE_COLOR=true ansible-playbook \
     -e "non_root_user=$USER" \
     -e "working_dir=$WORKING_DIR" \
     -e "roles_path=$PWD/roles" \
-    -e @tripleo-quickstart-config/metalkube-nodes.yml \
+    -e @${VM_NODES_FILE} \
     -e "local_working_dir=$HOME/.quickstart" \
     -e "virthost=$HOSTNAME" \
     -e "platform=$NODES_PLATFORM" \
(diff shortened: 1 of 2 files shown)


hardys:
@karmab FYI I'm temporarily removing extradisks here so we can make progress on worker deployment, is that acceptable to you, it should be easy enough to use the new variable to pass a customized file which re-enables the extradisks?

derekhiggins:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/556/

02_configure_host.sh
```diff
@@ -11,12 +11,13 @@ source ocp_install_env.sh
 # Note we copy the playbook so the roles/modules from tripleo-quickstart
 # are found without a special ansible.cfg
 export ANSIBLE_LIBRARY=./library
+export VM_NODES_FILE=${VM_NODES_FILE=:-'tripleo-quickstart-config/metalkube-nodes.yml'}
```
derekhiggins: Typo here, message from CI
Apr 12 11:02:50 |  [0;31mCould not find or access '/home/notstack/dev-scripts/:-tripleo-quickstart-config/metalkube-nodes.yml' on the Ansible Controller.[0m

hardys: Oops, thanks!

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/557/

tripleo-quickstart-config/metalkube-nodes.yml
```diff
@@ -1,4 +1,4 @@
-default_memory: 8192
+default_memory: 16384
```
russellb: In general, I think we should match defaults used on other supported platforms.

That's 16 GB for masters, but 8 GB for workers.  We could save a bit by defaulting to 8 GB for the worker.

PR #340, opened 2019-04-11
mdns-publisher: remove master things

celebdor:
Without this change the masters would announce themselves as members of the etcd cluster as well


hardys:
Tested and this looks better, the worker starts the mdns-publisher, although the hostname still ends up as "localhost" and we're not pointing at the mDNS VIP in the resolv.conf

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/552/

PR #349, opened 2019-04-12
Pass -image option in make_bm_workers function

mcornea:
Without the image option the worker nodes seem to not end up in provisioning state and they boot of the disk instead of nic.

### DIFF
--- a/11_register_hosts.sh
+++ b/11_register_hosts.sh
@@ -67,6 +67,7 @@ function make_bm_workers() {
            -password "$password" \
            -user "$user" \
            -boot-mac "$mac" \
+           -image \
            "$name"
     done
 }


mcornea:
> We don't want to do this. We want the nodes registered and entering the Ready state, so that the actuator can find them and add the image settings. If we add the image settings ourselves, then the actuator will not connect the Host objects to the Machine objects properly, and then the rest of the cluster setup will not work.

@dhellmann is there any additional step that I need to do for this to happen? If I just run 11_register_hosts.sh script then the worker nodes stay in Ready state and do not get the image deployed.

dhellmann:
> > We don't want to do this. We want the nodes registered and entering the Ready state, so that the actuator can find them and add the image settings. If we add the image settings ourselves, then the actuator will not connect the Host objects to the Machine objects properly, and then the rest of the cluster setup will not work.
> 
> @dhellmann is there any additional step that I need to do for this to happen? If I just run 11_register_hosts.sh script then the worker nodes stay in Ready state and do not get the image deployed.

If the cluster-api-provider-baremetal (actuator) is running in the same namespace as the hosts, it should see them and start provisioning them. There was an issue which was just recently fixed that may have been affecting your system. See https://github.com/metalkube/cluster-api-provider-baremetal/pull/58

PR #353, opened 2019-04-15
Sleep for rook components to get up properly

karmab:


10_deploy_rook.sh
```diff
@@ -13,13 +13,15 @@ oc create -f common.yaml
 sed '/FLEXVOLUME_DIR_PATH/!b;n;c\          value: "\/etc/kubernetes\/kubelet-plugins\/volume\/exec"' operator-openshift.yaml > operator-openshift-modified.yaml
 sed -i 's/# - name: FLEXVOLUME_DIR_PATH/- name: FLEXVOLUME_DIR_PATH/' operator-openshift-modified.yaml
 oc create -f operator-openshift-modified.yaml
+sleep 120
 oc wait --for condition=ready  pod -l app=rook-ceph-operator -n rook-ceph --timeout=120s
```
hardys: I'm wondering why we need the sleep, when we already wait 120s for several conditions related to the operator/apps being ready below?

10_deploy_rook.sh
```diff
@@ -13,13 +13,15 @@ oc create -f common.yaml
...
 oc wait --for condition=ready  pod -l app=rook-ceph-agent -n rook-ceph --timeout=120s
 oc wait --for condition=ready  pod -l app=rook-discover -n rook-ceph --timeout=120s
 sed "s/useAllDevices: .*/useAllDevices: true/" cluster.yaml > cluster-modified.yaml
 sed -i 's/# port: 8443/port: 8444/' cluster-modified.yaml
 oc create -f cluster-modified.yaml
 oc create -f toolbox.yaml
+sleep 120
```
hardys: Can we do a wait here instead of an arbitrary sleep?

I'm worried about how we'll handle this in the final solution e.g via kni-install, shouldn't we have inter-dependent operators launched via manifests that will just wait until the needed things are ready?

I guess we could to imperative things in the cluster bootstrap code, but I was assuming we'd be able to throw the manifests at the cluster and just wait on the desired state, like most other components work?

hardys:
Ok lets do this a a temporary hack, but we should figure out the declarative answer to this asap, e.g how to stand everything up via an interface only of manifests.

travisn:
@karmab What issue are you seeing that requires a delay between creating the different manifests? It's expected that the operator would internally sequence the creation of the resources even after the manifests are all created at the same time. Please do open a rook issue if you're seeing some issue around the sequencing. Thanks!

PR #360, opened 2019-04-15
Only set hostname for single worker.

russellb:
We now create 1 worker VM instead of 3.


zaneb:
I endorse this message.

PR #361, opened 2019-04-15
Only create git branch if we are applying patches.

russellb:


03_ocp_repo_sync.sh
```diff
@@ -21,13 +21,15 @@ function sync_go_repo_and_patch {
...
-    git branch -D we_dont_need_no_stinkin_patches || true
-    git checkout -b we_dont_need_no_stinkin_patches
-
-    shift; shift;
-    for arg in "$@"; do
-        curl -L $arg | git am
-    done
+    if test "$#" -gt "2" ; then
+        git branch -D metalkube || true
+        git checkout -b metalkube
```
hardys: Looks good but can you also update the branch name in `./dev-setup.md` please?

russellb: done

PR #362, opened 2019-04-15
Specify master and worker counts in install-config.

russellb:
We now only create 1 worker VM, so specify the number of workers as 1
instead of the default of 3.


derekhiggins:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/562/

Issue #354, opened 2019-04-15
Nodes are getting NotReady status


AsherShoshan:
After 24 of a runnign cluster, nodes (which are tainted masters) getting into NotReady status.
SSH in to the node and systemctl status kubelet showing this:
kubelet.service - Kubernetes Kubelet
   Loaded: loaded (/etc/systemd/system/kubelet.service; enabled; vendor preset: enabled)
   Active: active (running) since Sun 2019-04-14 09:33:24 UTC; 22h ago
  Process: 1409 ExecStartPre=/bin/rm -f /var/lib/kubelet/cpu_manager_state (code=exited, status=0/SUCCESS)
  Process: 1407 ExecStartPre=/bin/mkdir --parents /etc/kubernetes/manifests (code=exited, status=0/SUCCESS)
 Main PID: 1411 (hyperkube)
    Tasks: 57 (limit: 26213)
   Memory: 246.6M
      CPU: 4h 37min 3.310s
   CGroup: /system.slice/kubelet.service
           └─1411 /usr/bin/hyperkube kubelet --config=/etc/kubernetes/kubelet.conf --bootstrap-kubeconfig=/etc/kubernetes/kubeconfig --rotate-certificates --kubeconfig=/var/lib/kubelet/kubeconfig --container>

Apr 15 08:18:24 master-1 hyperkube[1411]: E0415 08:18:24.206447    1411 reflector.go:125] k8s.io/kubernetes/pkg/kubelet/kubelet.go:444: Failed to list *v1.Service: Unauthorized
Apr 15 08:18:24 master-1 hyperkube[1411]: E0415 08:18:24.404757    1411 reflector.go:125] k8s.io/kubernetes/pkg/kubelet/config/apiserver.go:47: Failed to list *v1.Pod: Unauthorized
Apr 15 08:18:24 master-1 hyperkube[1411]: W0415 08:18:24.701927    1411 status_manager.go:485] Failed to get status for pod "openshift-authentication-operator-768788f4f7-t85x6_openshift-authentication-operat>
Apr 15 08:18:25 master-1 hyperkube[1411]: E0415 08:18:25.001594    1411 webhook.go:106] Failed to make webhook authenticator request: Unauthorized
Apr 15 08:18:25 master-1 hyperkube[1411]: E0415 08:18:25.001634    1411 server.go:245] Unable to authenticate the request due to an error: Unauthorized
[...]

PR #364, opened 2019-04-15
Config NTP server accessible on the red hat network.

russellb:
Related to issue #354

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,7 @@ logs/
 config_*.sh
 !config_example.sh
 assets/generated
+
+# Conditionally created for appropriate environments
+assets/templates/99_master-chronyd-redhat.yaml
+assets/templates/99_worker-chronyd-redhat.yaml
--- a/assets/files/etc/chrony.conf
+++ b/assets/files/etc/chrony.conf
@@ -1,0 +1,43 @@
+#
+# Configure an NTP server accessible on the redhat network.
+#
+
+# Use public servers from the pool.ntp.org project.
+# Please consider joining the pool (http://www.pool.ntp.org/join.html).
+#pool 2.rhel.pool.ntp.org iburst
+pool clock.redhat.com iburst
+
+# Record the rate at which the system clock gains/losses time.
+driftfile /var/lib/chrony/drift
+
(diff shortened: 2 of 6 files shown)


russellb:
still testing this ...

russellb:
verified that this put the right config in place in my test deployment

russellb:
Note this will break deployments not within the red hat network or on the VPN ...

derekhiggins:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/563/

hardys:
> Note this will break deployments not within the red hat network or on the VPN, but this is probably a net improvement, since that's probably the minority of these deployments. That's better than breaking the majority ...

Doesn't sound ideal - I assume that using an upstream NTP server in both cases won't work due to firewall rules or something?

Perhaps we could make it configurable?

russellb:
> > Note this will break deployments not within the red hat network or on the VPN, but this is probably a net improvement, since that's probably the minority of these deployments. That's better than breaking the majority ...
> 
> Doesn't sound ideal - I assume that using an upstream NTP server in both cases won't work due to firewall rules or something?
> 
> Perhaps we could make it configurable?

Done ... this will now only use `clock.redhat.com` if resolvable.  Indeed, other NTP servers won't work on the red hat network ...

derekhiggins:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/571/

AsherShoshan:
Still having the 'NotReady' after 15-19 hours.

hardys:
> Still having the 'NotReady' after 15-19 hours.

You'll have to provide more details and/or raise another issue please, others have reported success with this fix so you may be encountering some other problem.

bengland2:
ceph time sync issues are usually because the 3 monitors aren't all using a local time server IP.  For example, clock.redhat.com is a CNAME that will DNS-resolve to multiple time servers in different parts of the US and this will  break Ceph, which has a maximum time skew allowed of 50 msec or so. 

Does metal3 allow the user to specify a timeserver FQDN?  Defaulting to some interplanetary time server  doesn't work well for this.

PR #369, opened 2019-04-16
[wip] prevent extra routes to popup

karmab:
potential temporal workaround for extra routes appearing on the nodes over time and making go notready


russellb:
Thanks!  We merged #377 as an alternative workaround for this

Issue #332, opened 2019-04-11
"Make clean" script failed task  "TASK [teardown/user : Remove virt_power_key from remote authorized_keys]"


maveric-tellrv:
while executing 'make clean'  one of the task is failing 'TASK [teardown/user : Remove virt_power_key from remote authorized_keys]"  fatal: [localhost]: FAILED! => {
   "msg": "could not locate file in lookup: /home/bkr/.quickstart/id_rsa_virt_power.pub"
}

also I have observed that 
**'make clean'** does not delete the **baremetal** and **provisioning**  **bridge** 
 due to which when I execute 'make' it fails with msg": "error creating bridge interface baremetal: File exists"
[bkr@vm220 dev-scripts]$ brctl show 
bridge name	bridge id		STP enabled	interfaces
baremetal		8000.000000000000	no		
provisioning		8000.000000000000	no		
virbr0		8000.5254002e4699	yes		virbr0-nic

I need to manually remove the bridges to make run successful.

[bkr@vm220 dev-scripts]$ sudo ip link delete baremetal type bridge
[bkr@vm220 dev-scripts]$ sudo ip link delete provisioning type bridge

hardys:
Can you provide the *cleanup.sh logs?  This works fine in my environment so we'll need more information to help debug this I think.

maveric-tellrv:
[host_cleanup-2019-04-11-120833.log](https://github.com/openshift-metalkube/dev-scripts/files/3067461/host_cleanup-2019-04-11-120833.log)

hardys:
The id_rsa_virt_power.pub error can be ignored, I see the same in my logs and we set `ignore_errors: true` in `../tripleo-quickstart-config/roles/libvirt/teardown/user/tasks/main.yml` - I'll look at how we can clean that up though.

To debug the `make` error, again we'll need more info, e.g the logs for whatever script you see failing.

maveric-tellrv:
@hardy  attaching the logs  hope this will help
[Uploading 02_configure_host-2019-04-11-120526.log…]()

hardys:
I pushed a patch for the teardown error, lets track the removal of the bridge via https://github.com/openshift-metalkube/dev-scripts/issues/375 as it's unrelated to the title of this issue

PR #373, opened 2019-04-16
Remove ansible user teardown cleanup

hardys:
This only tries to remove the id_rsa_virt_power.pub from the
wrong directory, and since we create that via 01_install_requirements.sh
lets just drop this cleanup as it will always fail and give a
confusing error in the logs.

Fixes: #332


derekhiggins:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/572/

PR #377, opened 2019-04-17
Workaround to avoid the affect of the undeleted routes for API, DNS and INGRESS VIPs

yboaron:
This is a temporary solution to workaround the extra host route, 
It changes VIPs netmask to the master subnet netmask.

I set it as [WIP] since I didn't test it yet.

### DIFF
--- a/assets/files/etc/kubernetes/manifests/keepalived.yaml
+++ b/assets/files/etc/kubernetes/manifests/keepalived.yaml
@@ -52,6 +52,7 @@ spec:
       API_VIP="$(dig +noall +answer "api.${DOMAIN}" | awk '{print $NF}')"
       IFACE_CIDRS="$(ip addr show | grep -v "scope host" | grep -Po 'inet \K[\d.]+/[\d.]+' | xargs)"
       SUBNET_CIDR="$(/usr/local/bin/get_vip_subnet_cidr "$API_VIP" "$IFACE_CIDRS")"
+      NET_MASK="$(echo $SUBNET_CIDR | cut -d "/" -f 2)"
       INTERFACE="$(ip -o addr show to "$SUBNET_CIDR" | awk '{print $2}')"
       DNS_VIP="$(dig +noall +answer "ns1.${DOMAIN}" | awk '{print $NF}')"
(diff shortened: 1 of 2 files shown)


russellb:
Can you add a bit of an explanation about why this fixes the problem?

russellb:
also note another workaround proposed for the same issue here: https://github.com/openshift-metalkube/dev-scripts/pull/369

derekhiggins:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/579/

yboaron:
@russellb ,

This patch doesn't delete the redundant routes, created automatically by the kernel for each interface/IP address, (this should be fixed in the kernel), just set VIPs netmask to interface's main address netmask instead of 32, as a result of that the 'unneeded' routes created for the VIPs by the kernel will be identical to route for the interface main address, so we will not hit the 'connect: invalid argument' anymore.

Below you can see the routing table, IP a, and ping to API VIP  status (after keepalived set & move VIPs):

[core@master-1 ~]$ ip -4 a  | grep 192.168.111
    inet 192.168.111.21/24 brd 192.168.111.255 scope global dynamic noprefixroute ens4

[core@master-1 ~]$ ip r 
default via 192.168.111.1 dev ens4 proto dhcp metric 101 
10.128.0.0/14 dev tun0 scope link 
172.22.0.0/24 dev ens3 proto kernel scope link src 172.22.0.71 metric 100 
172.30.0.0/16 dev tun0 
192.168.111.0/24 dev ens4 proto kernel scope link src 192.168.111.21 metric 101 
192.168.111.0/24 dev ens4 proto kernel scope link src 192.168.111.2 metric 101 
192.168.111.0/24 dev ens4 proto kernel scope link src 192.168.111.5 metric 101 
[core@master-1 ~]$ ping 192.168.111.5
PING 192.168.111.5 (192.168.111.5) 56(84) bytes of data.
64 bytes from 192.168.111.5: icmp_seq=1 ttl=64 time=0.262 ms
64 bytes from 192.168.111.5: icmp_seq=2 ttl=64 time=0.148 ms

gklein:
I've applied this on a cluster for 24 hours. I can confirm it workaround the issue - all nodes are in ready state.

PR #383, opened 2019-04-17
Restart machine-controller after deploying baremetal-operator.

russellb:
This is a workaround for
https://github.com/metalkube/cluster-api-provider-baremetal/issues/57

The machine-controller currently requires that the BareMetalHost CRD
is defined before it starts up.  We'll fix this, but short term,
bounce the machine-controller after launching the baremetal-operator
and defining the BareMetalHost CRD in the cluster.


russellb:
@mhrivnak Thanks.  It'll be some time before we get the fix in a pinned release payload used here, though.

PR #390, opened 2019-04-18
Ceph config override using configMap

umangachapagain:
Reducing mon_osd_down_out_interval

Signed-off-by: Umanga Chapagain <chapagainumanga@gmail.com>

### DIFF
--- a/10_deploy_rook.sh
+++ b/10_deploy_rook.sh
@@ -31,7 +31,7 @@ data:
   config: |
     [global]
     osd_pool_default_pg_autoscale_mode = on
-    mon_osd_down_out_interval = "10s"
+    mon_osd_down_out_interval = "10"
 EOF
 
 sed 's/namespace: rook-ceph/namespace: openshift-storage/' operator-openshift.yaml > operator-openshift-modified.yaml
@@ -56,6 +56,9 @@ sed 's/namespace: rook-ceph/namespace: openshift-storage/' toolbox.yaml > toolbo
 sed -i "s@rook/ceph:master@rook/ceph:$ROOK_VERSION@" toolbox-modified.yaml
 oc create -f toolbox-modified.yaml
 
+oc wait --for condition=ready  pod -l app=rook-ceph-tools -n openshift-storage --timeout=120s
+oc -n openshift-storage exec $(oc -n openshift-storage get pod --show-all=false -l "app=rook-ceph-tools" -o jsonpath='{.items[0].metadata.name}') -- ceph mgr module enable pg_autoscaler --force
+
 cat <<EOF | oc create -f -
 apiVersion: ceph.rook.io/v1
 kind: CephBlockPool
@@ -68,6 +71,7 @@ spec:
     size: 2
 EOF
 
+sleep 100
 oc -n openshift-storage exec $(oc -n openshift-storage get pod --show-all=false -l "app=rook-ceph-tools" -o jsonpath='{.items[0].metadata.name}') --  ceph osd pool set rbd target_size_ratio .3
 
 cat <<EOF | oc create -f -
--- a/10_deploy_rook.sh
+++ b/10_deploy_rook.sh
@@ -21,6 +21,19 @@ oc create -f common-modified.yaml
 oc label namespace openshift-storage  "openshift.io/cluster-monitoring=true"
 oc policy add-role-to-user view system:serviceaccount:openshift-monitoring:prometheus-k8s -n openshift-storage
 
+cat <<EOF | oc create -f -
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: rook-config-override
+  namespace: openshift-storage
+data:
+  config: |
+    [global]
+    osd_pool_default_pg_autoscale_mode = on
+    mon_osd_down_out_interval = "10s"
+EOF
+
 sed 's/namespace: rook-ceph/namespace: openshift-storage/' operator-openshift.yaml > operator-openshift-modified.yaml
 sed -i 's/:rook-ceph:/:openshift-storage:/' operator-openshift-modified.yaml
 sed -i "s@rook/ceph:master@rook/ceph:$ROOK_VERSION@" operator-openshift-modified.yaml
@@ -54,6 +67,8 @@ spec:
   replicated:
     size: 2
 EOF
+
+oc -n openshift-storage exec $(oc -n openshift-storage get pod --show-all=false -l "app=rook-ceph-tools" -o jsonpath='{.items[0].metadata.name}') --  ceph osd pool set rbd target_size_ratio .3
 
 cat <<EOF | oc create -f -
 apiVersion: storage.k8s.io/v1
--- a/10_deploy_rook.sh
+++ b/10_deploy_rook.sh
@@ -30,8 +30,9 @@ metadata:
 data:
   config: |
     [global]
-    osd_pool_default_pg_autoscale_mode = on
-    mon_osd_down_out_interval = "10"
+    osd_pool_default_pg_num = 128
+    osd_pool_default_pgp_num = 128
+    mon_osd_down_out_interval = 10
 EOF
 
 sed 's/namespace: rook-ceph/namespace: openshift-storage/' operator-openshift.yaml > operator-openshift-modified.yaml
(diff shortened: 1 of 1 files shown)


umangachapagain:
@karmab Please take a look.

karmab:
tested it manually and saw:

image

umangachapagain:
will look into what is causing this.

ohadlevy:
@umangachapagain whats the status of the PR? thanks

umangachapagain:
> @umangachapagain whats the status of the PR? thanks

Tested it manually. Things are working. 
But looks like some tweaking is needed for automation because some ceph flags can not be set in runtime, apparently.
Will get it resolved asap.

10_deploy_rook.sh
```diff
@@ -43,6 +43,10 @@ sed 's/namespace: rook-ceph/namespace: openshift-storage/' toolbox.yaml > toolbo
 sed -i "s@rook/ceph:master@rook/ceph:$ROOK_VERSION@" toolbox-modified.yaml
 oc create -f toolbox-modified.yaml
 
+oc wait --for condition=ready  pod -l app=rook-ceph-tools -n openshift-storage --timeout=180s
+oc -n openshift-storage exec $(oc -n openshift-storage get pod --show-all=false -l "app=rook-ceph-tools" -o jsonpath='{.items[0].metadata.name}') -- ceph mgr module enable pg_autoscaler
+oc -n openshift-storage exec $(oc -n openshift-storage get pod --show-all=false -l "app=rook-ceph-tools" -o jsonpath='{.items[0].metadata.name}') -- ceph config set global osd_pool_default_pg_autoscale_mode on
```
liewegas: need to restart the mon daemon here since we're using the 14.2.0 build.  i'm not sure what the best/right way to do that is tbh, @leseb?

leseb: We should include that into Ceph config overrides, @umangachapagain please look into: https://rook.io/docs/rook/v0.9/advanced-configuration.html

liewegas: add something like this to the cluster.yaml:
```
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: rook-config-override
  namespace: openshift-storage
data:
  config: |
    [global]
    osd_pool_default_pg_autoscale_mode = on
```

umangachapagain: @liewegas Will this also enable the module which was previously done by `ceph mgr module enable pg_autoscaler` ?
@leseb Should I restart the mon daemon? If yes, how?

leseb: @umangachapagain you need **both** the `ceph mgr module enable pg_autoscaler` and the configmap.

You should inject the configmap before the injecting the cluster CR so you don't have to restart the mon.

10_deploy_rook.sh
```diff
@@ -21,6 +21,19 @@ oc create -f common-modified.yaml
...
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: rook-config-override
+  namespace: openshift-storage
+data:
+  config: |
+    [global]
+    osd_pool_default_pg_autoscale_mode = on
+    mon_osd_down_out_interval = "10s"
```
leseb:
```suggestion
    mon_osd_down_out_interval = "10"
```

10_deploy_rook.sh
umangachapagain: You mean,
Step 1. Inject configMap
Step 2. Deploy cluster
Step 3.  Deploy toolbox
Step 4. ` oc -n openshift-storage exec $(oc -n openshift-storage get pod --show-all=false -l "app=rook-ceph-tools" -o jsonpath='{.items[0].metadata.name}') -- ceph mgr module enable pg_autoscaler `

10_deploy_rook.sh
umangachapagain: ack

10_deploy_rook.sh
leseb: Correct.

umangachapagain: Step 4 can be done at runtime right? I remember some commands can't be `exec`ed on runtime.

leseb: Step 4 enables an mgr module, this can be done anytime the mgr is up and running, so no worries.

umangachapagain: Noted. Thanks.

umangachapagain:
@leseb Can we do `oc -n openshift-storage exec $(oc -n openshift-storage get pod --show-all=false -l "app=rook-ceph-tools" -o jsonpath='{.items[0].metadata.name}') --  ceph osd pool set rbd target_size_ratio .3` from configMap?

leseb:
@umangachapagain no, that's a pool property so you can only change it from a CLI call.

umangachapagain:
> @umangachapagain no, that's a pool property so you can only change it from a CLI call.

Noted. Thanks for all the help.

10_deploy_rook.sh
```diff
@@ -43,6 +56,9 @@ sed 's/namespace: rook-ceph/namespace: openshift-storage/' toolbox.yaml > toolbo
 sed -i "s@rook/ceph:master@rook/ceph:$ROOK_VERSION@" toolbox-modified.yaml
 oc create -f toolbox-modified.yaml
 
+oc wait --for condition=ready  pod -l app=rook-ceph-tools -n openshift-storage --timeout=120s
+oc -n openshift-storage exec $(oc -n openshift-storage get pod --show-all=false -l "app=rook-ceph-tools" -o jsonpath='{.items[0].metadata.name}') -- ceph mgr module enable pg_autoscaler
```
shtripat: This line gives the below error 

```
+ oc -n openshift-storage exec rook-ceph-tools-5b85446c4b-2fflw -- ceph mgr module enable pg_autoscaler
Error ENOENT: all mgr daemons do not support module 'pg_autoscaler', pass --force to force enablement
command terminated with exit code 2
```

@leseb ^^^

But if after this error we run the command from CLI it doesnt show any error and auto_scaler is enabled on ceph cluster.

liewegas:
Alternatively, to skip the autoscaler for now, drop the later hunks, and replace the configmap with
```
[global]
osd_pool_default_pg_num = 64
```
I *think* that'll be about the right number of PGs... double or halve as needed to make the warnings go away.

umangachapagain:
> 
> I _think_ that'll be about the right number of PGs... double or halve as needed to make the warnings go away.

Will try it out.
Would be better if pg_autoscaler could take care of it without any glitches.

liewegas:
> > I _think_ that'll be about the right number of PGs... double or halve as needed to make the warnings go away.
> 
> Will try it out.
> Would be better if pg_autoscaler could take care of it without any glitches.

I think it can with the (other) setting above, but let's worry about that after the summit demo is done.

karmab:
I tested the approach of creating the configmap with
`osd_pool_default_pg_num = 64`
but then the ceph pool didnt get created, because i was told by the operator that i missed the corresponding pgp setting.
I also tried adding `osd_pool_default_pgp_num = 64` but it didnt seem to get picked and i got the same error.

umangachapagain:
@karmab  Updated the PR to set both pg_num and pgp_num. It did work out for us as we just deployed 2 clusters with all these changes.

```
[root@bkr-hv02 ~]# oc get cm rook-config-override -oyaml
apiVersion: v1
data:
  config: |
    [global]
    osd_pool_default_pg_num = 128
    osd_pool_default_pgp_num = 128
    mon_osd_down_out_interval = 10
kind: ConfigMap
metadata:
  creationTimestamp: "2019-04-26T11:42:55Z"
  name: rook-config-override
  namespace: openshift-storage
  resourceVersion: "28805"
  selfLink: /api/v1/namespaces/openshift-storage/configmaps/rook-config-override
  uid: 6e87ca19-6818-11e9-8970-00cc68e859ab
[root@bkr-hv02 ~]# 

```

liewegas:
> @karmab Updated the PR to set both pg_num and pgp_num. It did work out for us as we just deployed 2 clusters with all these changes.
> 
> ```
> [root@bkr-hv02 ~]# oc get cm rook-config-override -oyaml
> apiVersion: v1
> data:
>   config: |
>     [global]
>     osd_pool_default_pg_num = 128
>     osd_pool_default_pgp_num = 128
>     mon_osd_down_out_interval = 10
> kind: ConfigMap
> metadata:
>   creationTimestamp: "2019-04-26T11:42:55Z"
>   name: rook-config-override
>   namespace: openshift-storage
>   resourceVersion: "28805"
>   selfLink: /api/v1/namespaces/openshift-storage/configmaps/rook-config-override
>   uid: 6e87ca19-6818-11e9-8970-00cc68e859ab
> [root@bkr-hv02 ~]# 
> ```

Do the pools have 128 or 100 pgs?  I think rook is explicitly setting the pg count and these default settings might not affect anything

liewegas:
Alternative approach: https://github.com/openshift-metal3/dev-scripts/pull/471

umangachapagain:
@karmab Updated PR in favor of #471 . This can also be merged.

liewegas:
lgtm.  this could also be done via a 'ceph config set global ...' command, but this is dev-scripts and it doesn't really matter :)

hardys:
We removed rook from dev-scripts so I'll close this now.

PR #392, opened 2019-04-19
Fix automatic deployment of libvirt workers.

russellb:
The following change recently went in to the baremetal-operator:

https://github.com/metalkube/baremetal-operator/commit/abed6b3e8381e410244b082fd54bf9569b66c3eb

With that change, I needed the baremetal-operator to use the "libvirt"
hardware profile for vbmc based workers.  This patch explicitly sets
that when using a virtual env.

The baremetal-operator also would have done this if the bmc address
had a "libvirt://" prefix, but we're registering them with an
"ipmi://" prefix here, even for virt, which works other than this.


russellb:
not needed after #393

PR #405, opened 2019-04-22
Only try to create workers if they exist

e-minguez:


11_register_hosts.sh
```diff
@@ -77,4 +77,4 @@ list_workers | make_bm_workers | tee $SCRIPTDIR/ocp/worker_crs.yaml
 
 oc --config ocp/auth/kubeconfig apply -f $SCRIPTDIR/ocp/master_crs.yaml --namespace=openshift-machine-api
 
-oc --config ocp/auth/kubeconfig apply -f $SCRIPTDIR/ocp/worker_crs.yaml --namespace=openshift-machine-api
+[ -f $SCRIPTDIR/ocp/worker_crs.yaml ] && oc --config ocp/auth/kubeconfig apply -f $SCRIPTDIR/ocp/worker_crs.yaml --namespace=openshift-machine-api
```
dhellmann: This file is created on line 76 above. What are the conditions you're seeing that lead to it not being created?

e-minguez: I don't have any workers in my `bm.json` (I'm thinking about creating the cluster with 3 masters, deploy rook, kubevirt, etc. then add workers eventually)

dhellmann: Does that somehow result in the file not being created? Maybe this needs to test if it's empty, instead of whether it exists?

e-minguez: You are right, the file is there but empty. I'll fix it, thanks!

PR #410, opened 2019-04-22
Increase the worker wait timeout

dantrainor:
The default worker wait timeout of 30s isn't sufficient while waiting for the node
to shut down, power cycle, and come back up before joining the cluster, particularly
in virtual environments where there are CPU, disk, or memory contensions.  Increase
the timeout to something more accommodating.


russellb:
Might need to be even higher

PR #412, opened 2019-04-22
Clean up disks through ironic and delete them

karmab:
helper script to clean up disks through ironic for baremetal envs

### DIFF
--- a/ironic_cleanup.sh
+++ b/ironic_cleanup.sh
@@ -1,0 +1,25 @@
+#!/bin/bash
+
+function wait_condition () {
+condition=0
+while [ "$condition" != "3" ] ; do
+  condition=$( openstack baremetal node list -c 'Provisioning State' | grep $1 | wc -l)
+  echo "waiting 5s for servers to be $1"
+  sleep 5
+done
+}
+
+export OS_TOKEN=fake-token
+export OS_URL=http://localhost:6385/
+
+SERVERS=`openstack baremetal node list -f value -c Name| xargs`
+if [ -z "$SERVERS" ] ; then
+    echo "nothing to do"
+    exit 0
+fi
+for server in $SERVERS ; do openstack baremetal node undeploy $server ; done
+wait_condition "available"
+for server in $SERVERS ; do openstack baremetal node manage $server ; done
+for server in $SERVERS ; do openstack baremetal node clean --clean-steps '[{"interface": "deploy", "step": "erase_devices_metadata"}]' $server ; done
+wait_condition "manageable"
+for server in $SERVERS ; do openstack baremetal node delete $server ; done


russellb:
I'd like to see some bug links in this script.  You can start with a bug against kni-installer, as the cluster destroy code there should ensure all of the right cleanup happens.

For the masters, kni-install should be using terraform to destroy the masters cleanly.  

The workers should be destroyed as well via this code path, as well.  It will need to remove all worker Machines, and then remove the BareMetalHosts, causing the baremetal-operator to deprovision them.

ironic_cleanup.sh
```diff
@@ -0,0 +1,25 @@
...
+  condition=$( openstack baremetal node list -c 'Provisioning State' | grep $1 | wc -l)
+  echo "waiting 5s for servers to be $1"
+  sleep 5
+done
+}
+
+export OS_TOKEN=fake-token
+export OS_URL=http://localhost:6385/
+
+SERVERS=`openstack baremetal node list -f value -c Name| xargs`
```
elfosardo: I'd rather see $(...) for command substitution in scripts

ironic_cleanup.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+
+function wait_condition () {
+condition=0
```
elfosardo: local condition

ironic_cleanup.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+
+function wait_condition () {
```
elfosardo: add some indentation to the function internals

ironic_cleanup.sh
```diff
@@ -0,0 +1,25 @@
...
+condition=0
+while [ "$condition" != "3" ] ; do
+  condition=$( openstack baremetal node list -c 'Provisioning State' | grep $1 | wc -l)
```
elfosardo: what is $1 in this case? maybe it's worth to use a local variable and assign $1 to it, for clarity

ironic_cleanup.sh
```diff
@@ -0,0 +1,25 @@
...
+
+export OS_TOKEN=fake-token
+export OS_URL=http://localhost:6385/
+
+SERVERS=`openstack baremetal node list -f value -c Name| xargs`
+if [ -z "$SERVERS" ] ; then
+    echo "nothing to do"
+    exit 0
+fi
+for server in $SERVERS ; do openstack baremetal node undeploy $server ; done
```
elfosardo: as far as I like one liner, maybe you can rewrite the for loops with a single function ?
unless we want to execute an operation at a time to all servers first

derekhiggins:
Ironic is now clean all nodes where they get registered, so we shouldn't normally need to run this script.
https://github.com/metalkube/metalkube-ironic/commit/e6081b9bb64bfefaac5e46f4d37b0ac606ded8f7

PR #414, opened 2019-04-23
Fix machine name in PATCH url

jtomasek:
$machine_name does not exist, this change fixes it to $machine

### DIFF
--- a/link-machine-and-node.sh
+++ b/link-machine-and-node.sh
@@ -17,6 +17,6 @@ node_name=$(echo $node | cut -f2 -d':')
 addresses=$(oc --config ocp/auth/kubeconfig get node ${node_name} -o json | jq -c '.status.addresses')
 
 curl -X PATCH \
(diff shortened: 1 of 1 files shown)


link-machine-and-node.sh
```diff
@@ -17,6 +17,6 @@ node_name=$(echo $node | cut -f2 -d':')
 addresses=$(oc --config ocp/auth/kubeconfig get node ${node_name} -o json | jq -c '.status.addresses')
 
 curl -X PATCH \
-     http://localhost:8001/apis/machine.openshift.io/v1beta1/namespaces/openshift-machine-api/machines/$machine_name/status \
+     http://localhost:8001/apis/machine.openshift.io/v1beta1/namespaces/openshift-machine-api/machines/$machine/status \
```
elfosardo: maybe wrap the variable in curly brackets ?

jtomasek: Done

PR #416, opened 2019-04-23
Remove unneeded iptables configuration

bfournie:
The iptables configuration is now done in the individual containers where the services are started, it can be removed from the script.


derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/619/

PR #426, opened 2019-04-23
Update Custom MAO bringup documentation to let CVO manage MAO again

sadasu:
While trying to bring up the custom MAO, the CVO was instructed to stop managing the MAO so that it doesn't immediately try to bring up the MAO that was stopped. When the new MAO is ready to be started, the CVO needs to be instructed to once again start managing the MAO for the new custom MAO to come up.

### DIFF
--- a/docs/custom-mao-and-capbm.md
+++ b/docs/custom-mao-and-capbm.md
@@ -94,6 +94,14 @@ Edit `custom-images.json` to have a modified image for the BareMetal case:
 
 ## 6) Now run the MAO
 
+Before running the MAO, we have to undo the change we did earlier by asking the
+cluster-version-operator to manage the machine-api-operator's deployment again.
+Without this, it will not scale the MAO back up.
+
+```sh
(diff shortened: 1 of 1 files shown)


docs/custom-mao-and-capbm.md
```diff
@@ -94,6 +94,14 @@ Edit `custom-images.json` to have a modified image for the BareMetal case:
 
 ## 6) Now run the MAO
 
+Before running the MAO, we have to undo the change we did earlier by asking the
+cluster-version-operator to manage the machine-api-operator's deployment again.
+Without this, it will not scale the MAO back up.
```
russellb: but we don't want it to scale the MAO back up, right?  You're running your own outside of the cluster, instead.  If it scales back up, it's not running your custom one.

sadasu: Ok. I misunderstood behavior to be that we were just replacing the image that the CVO was using to scale MAO back up.
I did not see the custom MAO come up either in the default or openshift-machine-api namespaces.

russellb: Yeah, you won't see it listed as a pod.  The `bin/machine-api-operator start ...` command below is what runs it.  It's running outside the cluster, so you won't see it listed anywhere.

Issue #421, opened 2019-04-23
11_register_hosts.sh fails with more than one worker nodes


mcornea:
**Describe the bug**
11_register_hosts.sh fails with more than one worker nodes:

```shell
2019-04-23 16:28:01 | + for node in '$(oc --config ocp/auth/kubeconfig get nodes -o template --template='\''{{range .items}}{{.metadata.uid}}:{{.metadata.name}}{{"\n"}}{{end}}'\'')'
2019-04-23 16:28:01 | ++ echo ab536217-65e4-11e9-a9d6-5254007cc020:worker-0
2019-04-23 16:28:01 | ++ cut -f2 -d:
2019-04-23 16:28:01 | + node_name=worker-0
2019-04-23 16:28:01 | + machine_name=ostest-worker-0
2019-04-23 16:28:01 | + [[ ostest-worker-0 == *\w\o\r\k\e\r* ]]
2019-04-23 16:28:01 | ++ oc --config ocp/auth/kubeconfig get machines -n openshift-machine-api
2019-04-23 16:28:01 | ++ grep worker-0
2019-04-23 16:28:01 | ++ cut -f1 '-d '
2019-04-23 16:28:02 | + machine_name='ostest-worker-0-4vv5f
2019-04-23 16:28:01 | + machine_name=ostest-worker-0
2019-04-23 16:28:01 | + [[ ostest-worker-0 == *\w\o\r\k\e\r* ]]
2019-04-23 16:28:01 | ++ oc --config ocp/auth/kubeconfig get machines -n openshift-machine-api
2019-04-23 16:28:01 | ++ grep worker-0
2019-04-23 16:28:01 | ++ cut -f1 '-d '
2019-04-23 16:28:02 | + machine_name='ostest-worker-0-4vv5f
2019-04-23 16:28:02 | ostest-worker-0-bdkrp
2019-04-23 16:28:02 | ostest-worker-0-nfw74'
2019-04-23 16:28:02 | + /home/centos/dev-scripts/link-machine-and-node.sh 'ostest-worker-0-4vv5f
2019-04-23 16:28:02 | ostest-worker-0-bdkrp
2019-04-23 16:28:02 | ostest-worker-0-nfw74' ab536217-65e4-11e9-a9d6-5254007cc020:worker-0
2019-04-23 16:28:02 | + set -e
2019-04-23 16:28:02 | + machine='ostest-worker-0-4vv5f
[...]

Issue #417, opened 2019-04-23
`worker-0` is hardcoded


e-minguez:
**Describe the bug**
When deploying workers with names different than `worker-0` the register script will never finish.

**To Reproduce**
Run a deployment with workers such as:

```
{
  "name": "kni1-worker-0",
  "driver": "ipmi",                                                                                                                               
...
```                                                             

**Expected/observed behavior**
The wait should be done for the node name instead

**Additional context**
https://github.com/openshift-metalkube/dev-scripts/blob/master/11_register_hosts.sh#L95

russellb:
OK - I think we should just skip this part if it's a baremetal deployment.  This was just an additional easy step to automate in the virt case where we can make more assumptions about what's configured.

e-minguez:
Closing this via #433
I'm assuming the procedure to add workers will be to use the following procedure https://github.com/metal3-io/baremetal-operator/blob/master/docs/dev-setup.md#using-bare-metal-hosts

russellb:
Yes, create `BareMetalHosts` for your workers, and then scale up the worker `MachineSet`

PR #433, opened 2019-04-24
Only wait for worker-0 in the default virt config.

russellb:
We automate waiting for a worker to come up and adding IPs to it for the
default virt configuration.  This is a helpful step for the common dev setup,
and it also runs in CI. For any other env, we just skip this, because we
can't automatically figure out the mapping between Machines and Nodes in
other cases, and must rely on running the link-machine-and-node.sh manually.

Related issues: #421 and #417


derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/598/

PR #436, opened 2019-04-24
Update to kubevirt 0.16.1

rmohr:
Update to kubevirt 0.16.1. This brings two fixes:

 * Ensure that VMI Pods are protected during the whole migration on drains
 * Allow disabling save-migration checks in libvirt. This allows to work around situations where libvirt does not properly recognise disks as migrateable, during development and testing.


fabiand:
@hardys @russellb @karmab please take a look

PR #437, opened 2019-04-24
oc proxy is required

e-minguez:


russellb:
This needs to be removed from `add-machine-ips.sh` too

e-minguez:
I agree on the `add-machine-ips.sh` script but in this case, the patch operation is being performed against `localhost:8081` and I'm not sure if `oc patch` can patch `status`

russellb:
I applied this to master along with the related fix

PR #445, opened 2019-04-24
Fix: chaging PVC for strimzi 0.11.2 version

jparrill:
The 0.11.2 version of Strimzi has 1G as default PVC size instead of 100 like master branch has. Changing it here.

### DIFF
--- a/extras/02_kafka.sh
+++ b/extras/02_kafka.sh
@@ -8,7 +8,7 @@ source ${BASEDIR}/common.sh
 # Kafka Strimzi configs
 KAFKA_NAMESPACE=${KAFKA_NAMESPACE:-strimzi}
 KAFKA_CLUSTERNAME=${KAFKA_CLUSTERNAME:-strimzi}
-KAFKA_PVC_SIZE=${KAFKA_PVC_SIZE:-10}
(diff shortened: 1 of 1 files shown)


extras/02_kafka.sh
```diff
@@ -39,7 +39,7 @@ oc wait --for condition=ready pod -l name=strimzi-cluster-operator -n ${KAFKA_NA
 
 # Modify Kafka cluster & Deploy
 sed -i "s/my-cluster/${KAFKA_CLUSTERNAME}/" metrics/examples/kafka/kafka-metrics.yaml
-sed -i "s/100Gi/${KAFKA_PVC_SIZE}Gi/" metrics/examples/kafka/kafka-metrics.yaml
+sed -i "s/1Gi/${KAFKA_PVC_SIZE}Gi/" metrics/examples/kafka/kafka-metrics.yaml
```
karmab: that should still be 100Gi for this sed

jparrill: it is, but applied by a variable set at the head of the file:

```
KAFKA_PVC_SIZE=${KAFKA_PVC_SIZE:-100}
￼```

PR #450, opened 2019-04-24
Add api-int DNS record to coredns

cybertron:
Add a static mdns record that points to the API VIP. This is to
support the new required DNS record from OpenShift.

### DIFF
--- a/assets/files/etc/kubernetes/manifests/coredns.yaml
+++ b/assets/files/etc/kubernetes/manifests/coredns.yaml
@@ -44,6 +44,9 @@ spec:
       source /etc/kubernetes/static-pod-resources/clusterrc
       export DOMAIN
       export NAME
+      # This should eventually be looked up from the config
+      API_VIP="$(dig +noall +answer "api.${DOMAIN}" | awk '{print $NF}')"
+      export API_VIP
       /usr/libexec/platform-python -c "from __future__ import print_function
       import os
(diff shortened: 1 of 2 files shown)


cybertron:
This is going to require changes in mdns-publisher as well. It also doesn't address the bootstrap node, which doesn't currently run mdns-publisher.

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/629/

markmc:
Please add a reference in the commit to the background context on the openshift change - e.g. https://github.com/openshift/installer/pull/1633 is a good reference

cybertron:
I updated my local commit message, so references should be there when I'm ready to drop the WIP from this. We need https://github.com/openshift-metal3/mdns-publisher/pull/7 before this will be effective though.

cybertron:
This version no longer has a dependency on mdns-publisher.

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/667/

hardys:
Hmm does this depend on other commits?  Since @cybertron mentioned it no longer depends on mdns-publisher I tried it locally and the masters can't resolve the api-int address to collect the ignition config.

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/675/

hardys:
Ignore my previous comment, I accidentally tested on a branch with #560 applied and CI confirms this as working so lgtm

PR #452, opened 2019-04-25
Workaround for make-bm-worker hardwareprofile

e-minguez:
https://github.com/metal3-io/baremetal-operator/issues/158

### DIFF
--- a/add-bm.sh
+++ b/add-bm.sh
@@ -1,0 +1,36 @@
+#!/bin/bash
+# Workaround for https://github.com/metal3-io/baremetal-operator/issues/158
+set -ex
+
+source common.sh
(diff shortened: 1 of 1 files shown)


e-minguez:
> Do we need this if we also have the change to the go tool?

This was just in case the other PR was so bad that you just close it :D

dhellmann:
> > Do we need this if we also have the change to the go tool?
> 
> This was just in case the other PR was so bad that you just close it :D

Aw, we wouldn't do that! Let's see about adding the conditional logic to the template like @russellb suggested and we can merge the other PR today. Ping me on slack if you want a hand with that.

e-minguez:
Fixed in the baremetal-operator

PR #461, opened 2019-04-25
Add 3 second delay before linking machines and nodes

dantrainor:
The preceeding steps leading up to running link-machine-and-node.sh
via add-machine-ips.sh might create a condition on slower systems
or systems with resource contention to not start the machine API
soon enough for the link-machine-and-node.sh steps to succesfully
interact with an API that hasn't yet come up.  The result of this
failure is "Failed connect to localhost:8001; Connection refused".

### DIFF
--- a/add-machine-ips.sh
+++ b/add-machine-ips.sh
@@ -13,5 +13,6 @@ for node in $(oc --config ocp/auth/kubeconfig get nodes -o template --template='
     if [[ "$machine_name" == *"worker"* ]]; then
         machine_name=$(oc --config ocp/auth/kubeconfig get machines -n openshift-machine-api | grep $node_name | cut -f1 -d' ')
     fi
+    sleep 3
     $SCRIPTDIR/link-machine-and-node.sh "$machine_name" "$node"
 done


add-machine-ips.sh
```diff
@@ -13,5 +13,6 @@ for node in $(oc --config ocp/auth/kubeconfig get nodes -o template --template='
     if [[ "$machine_name" == *"worker"* ]]; then
         machine_name=$(oc --config ocp/auth/kubeconfig get machines -n openshift-machine-api | grep $node_name | cut -f1 -d' ')
     fi
+    sleep 3
```
russellb: The error msg you posted would mean that `oc proxy` isn't up yet, and you have the sleep from before that gets run.  It happens inside link-machine-and-node.sh

russellb: I'd move your change into that script, but also make it do a programmatic check if the proxy is ready vs just a sleep

elfosardo: I agree with @russellb , the logic should be inside link-machine-and-node.sh
For example you could use curl in an until loop

hardys:
Ah sorry I missed this and just pushed https://github.com/openshift-metal3/dev-scripts/pull/467/ which shows how we can do this in a more programmatic way - feel free to merge that into this PR if you prefer

hardys:
I also raised https://github.com/openshift-metal3/dev-scripts/issues/466 to track this as I hit the same issue testing today

dantrainor:
I ran in to this again today:

+ addresses='[{"address":"192.168.111.20","type":"InternalIP"},{"address":"master-0","type":"Hostname"}]'
+ curl -X PATCH http://localhost:8001/apis/machine.openshift.io/v1beta1/namespaces/openshift-machine-api/machines/ostest-master-0/status -H 'Content-type: application/merge-patch+json' -d '{"status":{"addresses":[{"address":"192.168.111.20","type":"InternalIP"},{"address":"master-0","type":"Hostname"}],"nodeRef":{"kind":"Node","name":"master-0","uid":"c8ec7814-6b6e-11e9-b845-ee7e31e66e78"}}}'
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0curl: (7) Failed connect to localhost:8001; Connection refused

Just reading the PR notes from @russellb  and @hardys , I'll take the idea of a curl loop with an embedded delay as a means of mitigation.

dantrainor:
I see that @hardys has made some great changes for #467, I'll instead test with this.

dantrainor:
Closing in favor of #467

Issue #466, opened 2019-04-26
Connection refused reaching proxy on localhost:8001


hardys:
We seem to have a race where we try to access the proxy but it's not yet ready:

```
+ oc --config ocp/auth/kubeconfig proxy
++ oc --config ocp/auth/kubeconfig get node master-2 -o json
++ jq -c .status.addresses
+ addresses='[{"address":"192.168.111.22","type":"InternalIP"},{"address":"master-2","type":"Hostname"}]'
+ curl -X PATCH http://localhost:8001/apis/machine.openshift.io/v1beta1/namespaces/openshift-machine-api/machines/ostest-master-2/status -H 'Content-type: application/merge-patch+json' -d '{"status":{"addresses":[{"address":"192.168.111.22","type":"InternalIP"},{"address":"master-2","type":"Hostname"}],"nodeRef":{"kind":"Node","name":"master-2","uid":"766a7497-67fb-11e9-880f-12b410fc2262"}}}'
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
^M  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0curl: (7) Failed connect to localhost:8001; Connection refused
Starting to serve on 127.0.0.1:8001
```

**To Reproduce**

I saw this locally but I guess it's probably dependent on the time taken to start the proxy, probably the script should wait until the proxy port is accessible?

**Expected/observed behavior**

`link-machine-and-node.sh` probably needs some additional logic to check the proxy is accessible before trying to access it.

yprokule:
Has to be addressed by https://github.com/openshift-metal3/dev-scripts/pull/461

hardys:
> Has to be addressed by #461

Thanks, I'd missed that - IMHO a curl loop is better than the sleep approach, but we can work that out in the prs, #461 or #467

hardys:
Hmm it seems just waiting for the proxy to return json may not be enough:

```
Waiting for oc_proxy to respond++ date +%s
+ start_time=1556270394
+ curl -g -X GET http://localhost:8001/apis/machine.openshift.io/v1beta1/namespaces/openshift-machine-api/machines '-H Accept: application/json -H Content-Type: application/json'
+ jq .
[...]

PR #467, opened 2019-04-26
WIP Wait for proxy in link-machine-and-node.sh

hardys:
Otherwise we can race and the curl fails because the proxy port isn't
yet ready

Closes: #466

### DIFF
--- a/link-machine-and-node.sh
+++ b/link-machine-and-node.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+source utils.sh
+
 set -x
 set -e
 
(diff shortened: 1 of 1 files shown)


hardys:
Marking WIP as local testing still in-progress

russellb:
I've seen this problem locally sometimes, too.  This looks like an improvement to me, even if there's something even better to wait for.

dantrainor:
Still waiting on the rest of the deploy, but it had gotten past the point where it would have failed due to this condition.  I observed 3 sleeps of 5 seconds before link-machine-add-node.sh completed successfully.

LGTM.

PR #471, opened 2019-04-26
10_deploy_rook.sh: enable pg_autoscaler; set rbd pool target_ratio to .5

liewegas:
Signed-off-by: Sage Weil <sage@redhat.com>


10_deploy_rook.sh
```diff
@@ -70,6 +75,11 @@ parameters:
 reclaimPolicy: Retain
 EOF
 
+# tell ceph that the rbd pool will use ~50% of the cluster
+sleep 10
```
liewegas: not sure what a better way to wait for rook to create the above pool is.. or what a safe sleep time would be.  maybe just
```
while ! ceph osd pool ls | grep rbd ; do sleep 1 ; done
```
?

liewegas:
@karmab

PR #479, opened 2019-04-29
Allow custom NTP servers

e-minguez:


russellb:
Can we please retain "make this work automatically within red hat"?

The extra customization ability is nice, though.  Thanks.

utils.sh
```diff
@@ -26,10 +26,15 @@ function create_cluster() {
     $GOPATH/src/github.com/openshift-metalkube/kni-installer/bin/kni-install --dir "${assets_dir}" --log-level=debug create manifests
 
     # TODO - consider adding NTP server config to install-config.yaml instead
-    if host clock.redhat.com ; then
-        cp assets/templates/99_worker-chronyd-redhat.yaml.optional assets/templates/99_worker-chronyd-redhat.yaml
-        cp assets/templates/99_master-chronyd-redhat.yaml.optional assets/templates/99_master-chronyd-redhat.yaml
+    host clock.redhat.com > /dev/null && NTP_SERVERS="clock.redhat.com"
```
karmab: this means that if run within redhat network, you ll get this specific ntp server regarless of the option set for NT_SERVERS

e-minguez: Fixed I believe

russellb:
@e-minguez I'm good with adding the patch here, but it'll have to be rebased.  Sorry about that

e-minguez:
> @e-minguez I'm good with adding the patch here, but it'll have to be rebased. Sorry about that

Rebased!

extras/pre-00.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
```
russellb: I think we dropped the extras/ directory, so I'd say we can drop this file

e-minguez: Dropped

russellb:
thanks - just waiting for a CI run now

e-minguez:
> thanks - just waiting for a CI run now

:crossed_fingers:

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/661/

Issue #487, opened 2019-04-30
make clean does not remove any extra workers VMs 


gklein:
**Describe the bug**
When deploying a cluster with more than one worker, "make clean" does not remove any extra workers VM, just the first (default) one.

**To Reproduce**

*Deploy a cluster with 3 workers + 10gb extra disk for both masters and workers
*git rev-parse HEAD
2975d40
*export VM_NODES_FILE=~kni/metalkube-nodes.yml
```
20,29c20
<   - name: openshift_worker_1
<     flavor: openshift_worker
<     virtualbmc_port: 6234
<
<   - name: openshift_worker_2
<     flavor: openshift_worker
<     virtualbmc_port: 6235
<
<
< node_count: 6
---
> node_count: 4
36,37c27
<     extradisks_size: 10G
<     extradisks: true
---
>     extradisks: false
43,44c33
<     extradisks_size: 10G
<     extradisks: true
---
>     extradisks: false
```
* Run "make clean"

**Expected/observed behavior**
One workers VM is removed, but 2 others are still exist. 

```
$ sudo virsh list --all
 Id    Name                           State
----------------------------------------------------
 20    openshift_worker_2             running
 -     openshift_worker_1             shut off
```

All workers VMs that appears in $VM_NODES_FILE, should be removed with "make clean".

hardys:
Thanks, this is because we don't pass the VM_NODES_FILE to ansible in the `host_cleanup.sh`, which is easily fixed, but I'm working on a different approach which will remove VM_NODES_FILE and template the data dynamically, which should also solve this problem.

Issue #486, opened 2019-04-30
Extra workers deployed are failing to join the cluster (Virt)


gklein:
**Describe the bug**
Extra workers deployed on a  virt setup are failing to join the cluster

**To Reproduce**

* Deploy a cluster with 3 workers + 10gb extra disk for both masters and workers
   * git rev-parse HEAD
   2975d4035f38d6734fec2fb1be82fe1b6f4470c2
   * export VM_NODES_FILE=~kni/metalkube-nodes.yml
```diff ../metalkube-nodes.yml tripleo-quickstart-config/metalkube-nodes.yml
20,29c20
<   - name: openshift_worker_1
<     flavor: openshift_worker
<     virtualbmc_port: 6234
<
<   - name: openshift_worker_2
<     flavor: openshift_worker
<     virtualbmc_port: 6235
<
<
< node_count: 6
---
> node_count: 4
36,37c27
<     extradisks_size: 10G
<     extradisks: true
---
>     extradisks: false
43,44c33
<     extradisks_size: 10G
<     extradisks: true
---
>     extradisks: false
```

**Expected/observed behavior**
Workers should get a valid hostname/IP and connect to the cluster, but instead the hostname is set to localhost.locadomain so the workers are failing to join the cluster.

```Last login: Tue Apr 30 08:39:48 2019 from 192.168.111.1
[core@localhost ~]$ hostname
localhost.localdomain

Apr 30 07:25:11 localhost.localdomain hyperkube[1211]: E0430 07:25:11.324175    1211 kubelet.go:2273] node "localhost.localdomain" not found
Apr 30 07:25:11 localhost.localdomain hyperkube[1211]: I0430 07:25:11.325086    1211 kubelet_node_status.go:72] Attempting to register node localhost.localdomain
Apr 30 07:25:11 localhost.localdomain hyperkube[1211]: E0430 07:25:11.326267    1211 event.go:203] Server rejected event '&v1.Event{TypeMeta:v1.TypeMeta{Kind:"", APIVersion
:""}, ObjectMeta:v1.ObjectMeta{Name:"localhost.localdomain.159a2fbd4c61059f", GenerateName:"", Namespace:"default", SelfLink:"", UID:"", ResourceVersion:"", Generation:0, C
[...]

PR #492, opened 2019-04-30
Make number of workers configurable

hardys:
This removes the previous VM_NODES_FILE interface, so that we can dynamically
generate the data for master/worker nodes, and avoid issues where previous hard
coded things weren't updated via VM_NODES_FILE.

Instead, you can now pass NUM_MASTERS, NUM_WORKERS and VM_EXTRADISKS variables
to control the number of master/worker nodes, and if we create an extra disk
for ceph.

The previous cleanup in host_cleanup also didn't consider VM_NODES_FILE,
so that is updated to align with this new approach, fixing bug #487

Closes: #487
Related: #486


gklein:
I've tested this PR with:
```
export NUM_MASTERS="3"
export NUM_WORKERS="3"
export VM_EXTRADISKS="true"
````
Everything seems to work as expected, but it seems the workers only started a few minutes after the deployment was already done. 

I suggest we consider adding some logic to wait for the workers to appear before declaring that the deployment has completed.

hardys:
I looked at waiting for multiple workers, and it looks like it's difficult ref https://github.com/openshift-metal3/dev-scripts/issues/421 so I'd suggest we go ahead with this and figure out the solution as a follow-up?

derekhiggins:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/611/

russellb:
CI failed as the service we're grabbing RHCOS images from is down

russellb:
Tested this locally and it's working fine for me.  Thanks!

PR #497, opened 2019-05-01
Remove unused SSH_PRIV_KEY var.

russellb:


config_example.sh
```diff
@@ -9,3 +9,5 @@ set -x
 # Uncomment to build a copy of ironic or inspector locally
 #export IRONIC_INSPECTOR_IMAGE=https://github.com/metalkube/metalkube-ironic-inspector
 #export IRONIC_IMAGE=https://github.com/metalkube/metalkube-ironic
+
+export EXTERNAL_SUBNET="192.168.111.0/24"
```
hardys: If folks don't update their config, won't this break them on update because we're using it in 06, e.g:

`./06_create_cluster.sh:    INGRESS_VIP=$(python -c "from ansible.plugins.filter import ipaddr; print(ipaddr.nthhost('"$EXTERNAL_SUBNET"', 4))")`

Would it be safer perhaps to put the default into common.sh, then add it here commented to show how to override the default?

russellb: I figured they'd get a conflict in `ocp_install_env.sh` when updating if they had changed it from the default.  Leaving it here and noting it in config works too.

russellb: This one isn't a big deal.  I think i'm just going to drop this commit.

hardys: Ack, to clarify my concern wasn't for when someone had hacked on `ocp_install_env.sh`, it was what if they update dev-scripts and do nothing to their `config_$USER.sh`, AFAICT that will result in an error because EXTERNAL_SUBNET is undefined in `06_create_cluster.sh`?

So we'd have to specify a default in `common.sh`, then let the user optionally override in `config_$USER.sh`

russellb: Ah I see ... I didn't read close enough.  That makes sense :-)

Issue #481, opened 2019-04-29
10_deploy_rook.sh exits with error: no matching resources found


mcornea:
**Describe the bug**
10_deploy_rook.sh exits with error: no matching resources found after removing the sleeps in https://github.com/openshift-metal3/dev-scripts/pull/471

**To Reproduce**
Run 10_deploy_rook.sh on a virtual environment

**Expected/observed behavior**
Expected behavior: no failures

Observed behavior:
```shell
[centos@provisionhost-0 dev-scripts]$ cat logs/10_deploy_rook-2019-04-29-164344.log 
+ source common.sh
+++ go env
++ eval 'GOARCH="amd64"
GOBIN=""
GOCACHE="/home/centos/.cache/go-build"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GOOS="linux"
GOPATH="/home/centos/go"
GOPROXY=""
GORACE=""
GOROOT="/usr/lib/golang"
GOTMPDIR=""
GOTOOLDIR="/usr/lib/golang/pkg/tool/linux_amd64"
GCCGO="gccgo"
CC="gcc"
CXX="g++"
CGO_ENABLED="1"
GOMOD=""
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -fmessage-length=0 -fdebug-prefix-map=/tmp/go-build204779665=/tmp/go-build -gno-record-gcc-switches"'
+++ GOARCH=amd64
+++ GOBIN=
+++ GOCACHE=/home/centos/.cache/go-build
+++ GOEXE=
+++ GOFLAGS=
+++ GOHOSTARCH=amd64
+++ GOHOSTOS=linux
+++ GOOS=linux
+++ GOPATH=/home/centos/go
+++ GOPROXY=
+++ GORACE=
+++ GOROOT=/usr/lib/golang
+++ GOTMPDIR=
+++ GOTOOLDIR=/usr/lib/golang/pkg/tool/linux_amd64
+++ GCCGO=gccgo
+++ CC=gcc
+++ CXX=g++
+++ CGO_ENABLED=1
+++ GOMOD=
+++ CGO_CFLAGS='-g -O2'
[...]

PR #498, opened 2019-05-01
Increase rook sleeps to avoid error: no matching resources found

mcornea:
Fixes isue #481


karmab:
addressed in https://github.com/openshift-metal3/dev-scripts/commit/f6c98cb1678c8de4606324246865824e10e9a85e, (sorry for that)

PR #501, opened 2019-05-02
coredns-mdns: masters to always return SRV entries

celebdor:
Keepalived health for the coredns-mdns VIP is determined from being able
to give out Etcd records. The problem with that is that if some master
goes down, we are, by definition, not able to give out the total amount
of masters, which is the minimum that used to be set before we moved it
to 2. The reason why it used to be TOTAL_AMOUNT_OF_MASTERS was because
that is the right setting at the bootstrap stage and we were moving the
DNS VIP to the masters as soon as possible.

With this change and a companion one in kni-installer, we keep the
TOTAL_AMOUNT_OF_MASTERS, but only in the bootstrap keepalived
configuration. Once the cluster is formed and the bootstrap node is
taken down, the masters will run coredns-mdns without the
TOTAL_AMOUNT_OF_MASTERS restriction when responding to etcd SRV queries.

### DIFF
--- a/assets/files/etc/kubernetes/static-pod-resources/coredns/Corefile.template
+++ b/assets/files/etc/kubernetes/static-pod-resources/coredns/Corefile.template
@@ -1,7 +1,7 @@
 . {
     errors
     health
-    mdns $DOMAIN 2 $NAME
+    mdns $DOMAIN 0 $NAME
     forward . /etc/coredns/resolv.conf
     cache 30
     reload


celebdor:
Goes together with https://github.com/openshift-metal3/kni-installer/pull/71

hardys:
Tested locally with openshift-metal3/kni-installer#71 and this is working fine AFAICS

PR #503, opened 2019-05-02
Fix oc wait timeout in 10_deploy_rook.sh

dantrainor:
Most 'oc wait' commands have a timeout of 180s.  This may not be sufficient for
other wait conditions which have a timeout of 120s.  Make all the timeouts 180s.


dantrainor:
Changed typo to read "Make all the timeouts 180s".

karmab:
addressed in https://github.com/openshift-metal3/dev-scripts/commit/f6c98cb1678c8de4606324246865824e10e9a85e, (sorry for that)

PR #509, opened 2019-05-03
Increase oc wait timeout, again

dantrainor:
'oc create' commands may take more than 360s to complete.  Subsequent 'oc wait' commands
waiting for the condition of a pod being ready based on the 'oc create' command may
time out before completing, failing deployments.  The 'oc wait' command is a polling
operation, not a blocking operation.  Increasing this value should have no significant
impact on deployment time because as soon as the condition is met, the command returns.

### DIFF
--- a/10_deploy_rook.sh
+++ b/10_deploy_rook.sh
@@ -29,9 +29,9 @@ sed -i '/ROOK_MON_OUT_TIMEOUT/!b;n;c\          value: "40s"' operator-openshift-
 oc create -f operator-openshift-modified.yaml
 sleep 10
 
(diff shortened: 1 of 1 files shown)


dantrainor:
A timeout of 1200s resulted in a successful deployment, allowing 10_deploy_rook.sh to complete without error this time.

russellb:
seems harmless enough.  s/deploymenys/deployments/ in your commit message, though

dantrainor:
> seems harmless enough. s/deploymenys/deployments/ in your commit message, though

fixed, thx.

PR #510, opened 2019-05-06
Added 'imagePullPolicy: IfNotPresent' to some pods

e-minguez:
All the containers in the static pods have the `IfNotPresent` pull policy except those in this PR.

### DIFF
--- a/assets/files/etc/kubernetes/manifests/coredns.yaml
+++ b/assets/files/etc/kubernetes/manifests/coredns.yaml
@@ -32,6 +32,7 @@ spec:
       mountPath: "/etc/kubernetes/static-pod-resources"
     - name: kubeconfig
       mountPath: "/etc/kubernetes/kubeconfig"
+    imagePullPolicy: IfNotPresent
   - name: render-corefile
     image: quay.io/openshift/origin-node:latest
     command:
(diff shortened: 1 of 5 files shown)


hardys:
Can you add some information to the commit message explaining why this is needed please?

russellb:
Was this just for demo insurance, in case a cluster loses net access?  It makes sense for that env, but probably not in general

e-minguez:
> Was this just for demo insurance, in case a cluster loses net access? It makes sense for that env, but probably not in general

That was the main reason for this PR, but also to be consistent with the pull policies in the rest of the containers of those pods:

- https://github.com/openshift-metal3/dev-scripts/blob/master/assets/files/etc/kubernetes/manifests/coredns.yaml#L62
- https://github.com/openshift-metal3/dev-scripts/blob/master/assets/files/etc/kubernetes/manifests/coredns.yaml#L79
- https://github.com/openshift-metal3/dev-scripts/blob/master/assets/files/etc/kubernetes/manifests/keepalived.yaml#L92
- etc.

russellb:
@celebdor @cybertron please review this.

I'm inclined to remove the policy from other pods rather than merging this one.  Thoughts?

cybertron:
I think @celebdor added those manifests so he may have more context for why the policy was set that way, but reading https://kubernetes.io/docs/concepts/configuration/overview/#container-images makes me think we want to leave this blank and take the default behavior, which seems pretty sane.

bcrochet:
This is obsolete and should be closed. @e-minguez

e-minguez:
> This is obsolete and should be closed. @e-minguez

closing it

Issue #524, opened 2019-05-09
Running ./09_deploy_kubevirt.sh script places you in kubevirt-web-ui once it completes


rlopez133:
**Describe the bug**
When running the 09_deploy_kubevirt.sh from the dev-scripts repository, once it completes, it auto changes the existing namespace to the kubevirt-web-ui

**To Reproduce**
* ensure you have a running OCP cluster
* Within the dev-scripts repository, run the 09_deploy_kubevirt.sh script
* Once it successfully completes run `oc project` and it will state your in the kubevirt-web-ui

**Expected/observed behavior**
I expected that once the script completes it would continue to leave me in my existing project prior to running the script. 

**Additional context**
- I personally don't mind that it changed the project (avoids having to use the -n option) but it is inconsistent with how the 10_deploy_rook.sh script handles this. The 10_deploy_rook.sh **does not** auto switch you to the openshift-storage namespace.

- Ideally I'd like to see one of two things:
- Auto change the project but provide a message stating that the project was switched to X (where is the project name) once the script completes.

OR

- Let the user know that the script they just ran, that  the pods, resources, etc for that project are located in X where X is the name of the namespace and provide an example on how to see i.e. pods -- `oc get pods -n <namespace>`

PR #526, opened 2019-05-12
Making sure project scope is returned to the current project

gklein:
This should resolve https://github.com/openshift-metal3/dev-scripts/issues/524

### DIFF
--- a/09_deploy_kubevirt.sh
+++ b/09_deploy_kubevirt.sh
@@ -16,6 +16,7 @@ eval "$(go env)"
 # These will ultimately be deployed via kni-installer to the bootstrap node
 # where they will then be applied to the cluster, but for now we do it
 # manually
+export CURRENT_PROJECT=$(oc project -q)
 cd manifests
 for manifest in $(ls -1 *.yaml | sort -h); do
   oc --as system:admin --config ../ocp/auth/kubeconfig apply -f ${manifest}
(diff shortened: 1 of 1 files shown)


09_deploy_kubevirt.sh
```diff
@@ -36,3 +37,4 @@ oc apply -f operator.yaml
 sed "s/okdvirt/openshiftvirt/" crds/kubevirt_v1alpha1_kwebui_cr.yaml > crds/kubevirt_v1alpha1_kwebui_cr-modified.yaml
 sed -i "s/version: .*/version: \"$UIVERSION\"/" crds/kubevirt_v1alpha1_kwebui_cr-modified.yaml
 oc apply -f crds/kubevirt_v1alpha1_kwebui_cr-modified.yaml
+oc project ${CURRENT_PROJECT}
```
hardys: If anything fails in the script we'll fail to revert the project, I think we should run this via a trap, e.g something like:

  trap "oc project ${CURRENT_PROJECT}" EXIT

Set near the start of the script, e.g immediately after we read the current project, or immediately after we change it?

09_deploy_kubevirt.sh
```diff
@@ -16,6 +16,7 @@ eval "$(go env)"
 # These will ultimately be deployed via kni-installer to the bootstrap node
 # where they will then be applied to the cluster, but for now we do it
 # manually
+export CURRENT_PROJECT=$(oc project -q)
```
gklein:
```suggestion
export CURRENT_PROJECT=$(oc project -q)
trap "oc project ${CURRENT_PROJECT}" EXIT
```

09_deploy_kubevirt.sh
gklein: > If anything fails in the script we'll fail to revert the project, I think we should run this via a trap, e.g something like:
> 
> `trap "oc project ${CURRENT_PROJECT}" EXIT`
> 
> Set near the start of the script, e.g immediately after we read the current project, or immediately after we change it?
Yes, it would be safer this way.

PR #527, opened 2019-05-12
Haproxy API LB - increase iptables lock wait timeout

yboaron:
The haproxy static pod should verify the existence of an iptable
rule redirects traffic from OCP API to API LB frontend.
Sometimes Haproxy pod creates wrongly multiple instances of this rule
because it fails to read the current iptables rule list
(Another app is currently holding the xtables lock).

This patch increases the wait time for iptable lock to 10 seconds in
haproxy pod.


yboaron:
closes #528

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/631/

PR #530, opened 2019-05-13
Remove extras directory

karmab:
refer to https://github.com/rhdemo/2019-demo3


hardys:
Should we perhaps point to https://github.com/rhdemo/2019-demo3 in the commit message so folks can more easily find the new location?

karmab:
done

PR #533, opened 2019-05-13
Bring cdi back

karmab:
We removed cdi some time ago because there was a bug in the api resource discovery which prevents kubevirt to deploy.
it was then fixed on both sides so we are safe now to get cdi back with latest release of cdi ( and with any version of kubevirt)


hardys:
Is this dependent on the rebase to rc1?  Something in the commit message to remind us why it was removed, and why it's now being reinstated would probably be a good idea :)

russellb:
@karmab ping - see the question above

hardys:
We're now planning to remove this stopgap integration so the focus can switch the fully-integrated OLM operator driven approach, so lets close this ref https://github.com/openshift-metal3/dev-scripts/pull/603

PR #536, opened 2019-05-15
Add VIP DNS entries to vm_setup_vars.yml

hardys:
These have been removed from metal3-dev-env so we need to set
them here


hardys:
This will need to be merged together with https://github.com/metal3-io/metal3-dev-env/pull/9

PR #539, opened 2019-05-15
add kill proxy to host_cleanup.sh

dustinblack:
An error or interruption in `link-machine-and-node.sh` can leave a lingering `oc proxy` command. After a `make clean`, this proxy will still be running, and the `kill $proxy_pid` will error, aborting the deployment on the 06 script.

This simply adds a `pkill` to the `host_cleanup.sh` script to get that sucker out of there.


russellb:
good call ... it's annoying when this happens!

Issue #538, opened 2019-05-15
api-int should not be a required external DNS record


cybertron:
As part of the 4.1 rebase, we need to add support for the api-int DNS name. Unfortunately, because this is used during ignition (before we configure nodes to point at coredns) just adding a static entry in coredns is not sufficient. #523 is adding it to the libvirt dnsmasq, but this essentially makes it a required external record. Since it's actually for internal use only, we shouldn't be requiring that.

This issue is to track the tech debt associated with api-int and provide a reference for the work to remove the external DNS requirement. Currently the plan is to point ignition directly at the VIP and then switch to using api-int once we've configured the nodes to point at coredns.

cybertron:
For further information on api-int, see https://bugzilla.redhat.com/show_bug.cgi?id=1685704

cybertron:
This actually isn't completed until at least https://github.com/openshift-metal3/kni-installer/pull/82 merges. Preferrably after #560 merges too though to verify that the external record is no longer in use.

hardys:
> This actually isn't completed until at least [openshift-metal3/kni-installer#82](https://github.com/openshift-metal3/kni-installer/pull/82) merges. Preferrably after #560 merges too though to verify that the external record is no longer in use.

Ack thanks, I think it got auto-closed due to the commit message, I'm testing openshift-metal3/kni-installer#82  and #560 now so we can hopefully get those merged as well

cybertron:
Cool, thanks.

/me misses Related-Bug from OpenStack ;-)

PR #559, opened 2019-05-21
Add api_vip to install-config.yaml

cybertron:
This is needed for us to properly support the api-int DNS name on
baremetal. It is a prerequisite to the installer change:
https://github.com/openshift-metal3/kni-installer/pull/82

See #538 for further details.

### DIFF
--- a/06_create_cluster.sh
+++ b/06_create_cluster.sh
@@ -20,6 +20,18 @@ if [[ "${PULL_SECRET}" != *"cloud.openshift.com"* ]]; then
     exit 1
 fi
 
+# NOTE: This is equivalent to the external API DNS record pointing the API to the API VIP
+if [ "$MANAGE_BR_BRIDGE" == "y" ] ; then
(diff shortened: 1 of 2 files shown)


mcornea:
How do we set it for baremetal environments where the api-int record is different than 192.168.111.5 ?

hardys:
> How do we set it for baremetal environments where the api-int record is different than 192.168.111.5 ?

It should be set to whatever the api VIP DNS entry points to, because in `06_create_cluster.sh` we dig to resolve that entry to discover the API_VIP variable (which is the same IP we're setting here).

That probably means we should parameterize this, and allow for explicitly overriding API_VIP in the case where MANAGE_BR_BRIDGE is disabled, and probably move the dig so we can detect API_VIP before generating the install-config, thus avoiding this hard-coded entry?

cybertron:
Okay, it turns out that we were already looking up the API_VIP immediately after creating the config anyway, so I moved that block earlier and referenced API_VIP in the config. I think that should work for all environments.

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/669/

PR #571, opened 2019-05-24
Avoid messing the git repo

### LINKED ISSUE #566: assets/files/etc/chrony.conf modified during deploy
**Describe the bug**
Since we merged https://github.com/openshift-metal3/dev-scripts/pull/479 we leave the dev-scripts tree dirty after a deployment, because `utils.sh` does this:

```
echo "pool ${ntp} iburst" >> assets/files/etc/chrony.conf
```

After a deployment `assets/files/etc/chrony.conf` ends up in a modified state that needs a git reset - instead we should template this without modifying the git tracked files.

Also I think if you run multiple deployments without a git reset, we'll append the servers multiple times to the dirty `assets/files/etc/chrony.conf` ?

e-minguez:
Fixes #566

### DIFF
--- a/utils.sh
+++ b/utils.sh
@@ -14,6 +14,24 @@ function generate_assets() {
   done
 }
 
+function custom_ntp(){
+  # TODO - consider adding NTP server config to install-config.yaml instead
+  if [ -z ${NTP_SERVERS} ] && $(host clock.redhat.com > /dev/null); then
+    NTP_SERVERS="clock.redhat.com"
+  fi
+
+  if [ "$NTP_SERVERS" ]; then
+    cp assets/templates/99_worker-chronyd-custom.yaml.optional assets/generated/99_worker-chronyd-custom.yaml
+    cp assets/templates/99_master-chronyd-custom.yaml.optional assets/generated/99_master-chronyd-custom.yaml
+    NTPFILECONTENT=$(cat assets/files/etc/chrony.conf)
+    for ntp in $(echo $NTP_SERVERS | tr ";" "\n"); do
+      NTPFILECONTENT="${NTPFILECONTENT}"$'\n'"pool ${ntp} iburst"
+    done
+    NTPFILECONTENT=$(echo "${NTPFILECONTENT}" | base64 -w0)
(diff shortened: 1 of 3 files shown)


derekhiggins:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/690/

e-minguez:
Can we trigger the CI job again @derekhiggins ? I'm not sure if the failure is because the PR :$ Thanks!

utils.sh
```diff
@@ -44,6 +45,14 @@ function create_cluster() {
...
     cp ${assets_dir}/install-config.yaml{.tmp,}
     $OPENSHIFT_INSTALLER --dir "${assets_dir}" --log-level=debug create cluster
+
+    # Clean up
+    [ -f assets/files/etc/chrony.conf.tmp ] && \
+      mv assets/files/etc/chrony.conf{.tmp,}
+
+    for file in assets/templates/{99_worker-chronyd-custom.yaml,99_master-chronyd-custom.yaml}; do
+      [ -f ${file} ] && rm ${file}
+    done
```
hardys: These are already cleaned up in `ocp_cleanup.sh` but wouldn't it be easier/better to stop messing with the assets/files completey, and instead modify/copy things into assets/generated (e.g after generate_assets runs, which already cleans that directory, and it's already ignored by git)

hardys:
> Can we trigger the CI job again @derekhiggins ? I'm not sure if the failure is because the PR :$ Thanks!

You can re-trigger by copying the parameters into http://10.8.144.11:8080/job/dev-tools/build?delay=0sec but note I made some suggestions on how we might handle this more cleanly via the existing generated assets dir.

utils.sh
e-minguez: As `chrony.conf` needs to be modified after the yaml files are templated, I opted of doing it 'manually' instead.
Let me know about this solution.
Thanks @hardys !

derekhiggins:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/704/

e-minguez:
My bad. I believe I've fixed.

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/705/

hardys:
Thanks, this looks better, also tested locally although not with custom NTP servers.

PR #574, opened 2019-05-28
Also delete the interface

dantrainor:
When cleaning up the interface, remove the interface itself from the system.
The interface becomes unconfigured, though it won't delete the interface
from the system.  Ordinarily the next step would be to reboot the system,
at which time the interface would be deleted from the system.  However, not
rebooting the system could result in errors similar to: "error creating
bridge interface baremetal: File exists"


derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/689/

hardys:
I've run into this a couple of times, but AFAICT it's not because we need to reboot (I re-run dev-scripts multiple times each day without any problem).  So there seems to be some corner case where a failed deploy causes this.

It'd be good to further understand exactly when/why this is needed, but until then this seems like a reasonable additional cleanup - thanks!

PR #582, opened 2019-06-02
Use openshift/router-haproxy image for haproxy

yboaron:
This change updates haproxy static pod manifest to use
openshift/router-haproxy instead of docker/haproxy

### DIFF
--- a/assets/files/etc/kubernetes/manifests/haproxy.yaml
+++ b/assets/files/etc/kubernetes/manifests/haproxy.yaml
@@ -68,14 +68,43 @@ spec:
     imagePullPolicy: IfNotPresent
   containers:
   - name: haproxy
-    image: docker.io/library/haproxy:latest
-    args:
-    - "-W"
-    - "-db"
-    - "-S"
-    - "/var/run/haproxy/haproxy-master.sock,level,admin"
-    - "-f"
-    - "/etc/haproxy/haproxy.cfg"
+    image: quay.io/openshift/origin-haproxy-router:latest
+    command:
+    - "/bin/bash"
+    - "-c"
+    - |
+      #/bin/bash
+      reload_haproxy()
+      {
+        old_pids=$(pidof haproxy)
+        if [ -n "$old_pids" ]; then
+            /usr/sbin/haproxy -W -db -f /etc/haproxy/haproxy.cfg  -p /var/lib/haproxy/run/haproxy.pid -x /var/lib/haproxy/run/haproxy.sock -sf $old_pids &
+        else
+            /usr/sbin/haproxy -W -db -f /etc/haproxy/haproxy.cfg  -p /var/lib/haproxy/run/haproxy.pid &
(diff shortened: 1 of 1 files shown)


assets/files/etc/kubernetes/manifests/haproxy.yaml
```diff
@@ -68,15 +68,33 @@ spec:
...
+            /usr/sbin/haproxy -W -db -f /etc/haproxy/haproxy.cfg  -p /var/lib/haproxy/run/haproxy.pid &
+        fi
+      }
+      /usr/sbin/haproxy -W -db -f /etc/haproxy/haproxy.cfg  -p /var/lib/haproxy/run/haproxy.pid &
+      while true; do
+          sleep 10
+          if  grep -q RELOAD /haproxy_reload/msgq; then
+              reload_haproxy
+              echo "" > /haproxy_reload/msgq
+          fi
```
russellb: I think there’s a race condition here where a config change could get missed. A second config change could come in after a reload but before the msgq file is cleared. That change would be missed.

I know it seems incredibly unlikely. I can think of two options. 

1 - use locking on msgq as the shared resource. 

2 - treat msgq as a queue. Append a line for each reload request and only remove a line when reloading.

russellb:
And if you make it a real queue maybe it shouldn’t be a file and should be a Unix domain socket, something that has an interface closer to what’s needed?

yboaron:
@russellb , The race issue is a good point, I changed consumer (haproxy) poll interval to be smaller than half of producer (haproxy-monitor) poll interval, so we should be fine now with the race issue.
Performance wise it should be better to use something like unix domain socket instead of busy-wait polling.
Do u think it's acceptable if I'll send it in a follow-up patch?

russellb:
I think the race is still theoretically possible, even if you adjust the timeouts.  I think it's worth investigating other approaches before merging.

assets/files/etc/kubernetes/manifests/haproxy.yaml
```diff
@@ -108,8 +128,10 @@ spec:
       API_VIP="$(dig +noall +answer "api.${DOMAIN}" | awk '{print $NF}')"
 
       trap clean_prerouting_rules SIGINT SIGTERM
+      touch /haproxy_reload/msgq
+      chmod 777  /haproxy_reload/msgq
       while true; do
-          sleep 15
+          sleep 25
```
celebdor: How's this related to the PR? What's the motivation for increasing the sleep?

celebdor:
Agreed on locking or unix domain socket

bcrochet:
This is obsolete and should be closed. @yboaron

yboaron:
The haproxy static pod moved to MCO - we can delete this PR
10x @bcrochet

yboaron:
The haproxy static pod moved to MCO - we can close this PR
10x @bcrochet

Issue #581, opened 2019-06-02
01_install_requirements.sh failed  when installing python2-zmq


gklein:
**Describe the bug**
01_install_requirements.sh failed  when installing python2-zmq

**To Reproduce**
$ ./01_install_requirements.sh
Package ansible-2.8.0-2.el7.noarch already installed and latest version
Package 32:bind-utils-9.9.4-73.el7_6.x86_64 already installed and latest version
Package jq-1.5-10.el7.x86_64 already installed and latest version
Package 1:libguestfs-tools-1.38.2-12.100.lp.el7ev.noarch already installed and latest version
Package libvirt-4.5.0-10.el7_6.10.x86_64 already installed and latest version
Package libvirt-devel-4.5.0-10.el7_6.10.x86_64 already installed and latest version
Package libvirt-daemon-kvm-4.5.0-10.el7_6.10.x86_64 already installed and latest version
Package 2:nodejs-10.16.0-1nodesource.x86_64 already installed and latest version
Package podman-1.2-2.git3bd528e.el7.x86_64 already installed and latest version
Package python2-ironicclient-2.7.2-0.20190528160951.266a700.el7.noarch already installed and latest version
Package python2-ironic-inspector-client-3.5.0-0.20190311120303.9bb1150.el7.noarch already installed and latest version
Package python-lxml-3.2.1-4.el7.x86_64 already installed and latest version
Package python-netaddr-0.7.5-9.el7.noarch is obsoleted by python2-netaddr-0.7.19-5.el7.noarch which is already installed
Package python2-openstackclient-3.18.0-0.20190311121958.6868499.el7.noarch already installed and latest version
Package 10:qemu-kvm-1.5.3-160.el7_6.2.x86_64 is obsoleted by 10:qemu-kvm-ev-2.12.0-18.el7_6.5.1.x86_64 which is already installed
Package virt-install-1.5.0-1.el7.noarch already installed and latest version
Package unzip-6.0-19.el7.x86_64 already installed and latest version
Package yarn-1.16.0-1.noarch already installed and latest version
Resolving Dependencies
--> Running transaction check
---> Package python2-virtualbmc.noarch 0:1.5.0-0.20190401094255.bff0e6c.el7 will be installed
--> Processing Dependency: python-zmq >= 14.3.1 for package: python2-virtualbmc-1.5.0-0.20190401094255.bff0e6c.el7.noarch
--> Running transaction check
[...]

PR #583, opened 2019-06-03
Work around python-*zmq epel conflict

hardys:
This version now conflicts with the one from the RDO repos, so add
it to the exclude list for epel.

Fixes: #581


derekhiggins:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/694/

itzikb-redhat:
Works for me

PR #589, opened 2019-06-03
Updates to sync_repo_and_patch to support reproducible builds

p3ck:
For the CI workflow it is important to be able to reproduce builds.
This patch will do two things.  if a repos variable is defined with
repo/path/on/disk;SHA entries it will check out that specific
version instead master.  Second, if master is checked out this
will record the current HEAD commit.

Our CI workflow will cat the recorded_repos entry to config_$user.sh
which we then archive as an artifact.

This should allow us to re-run builds and get the same git repos.

### DIFF
--- a/utils.sh
+++ b/utils.sh
@@ -256,6 +256,40 @@ EOF
     oc -n openshift-etcd patch ep/host-etcd --patch "$patch"
 }
 
+function which_commit {
+    for repo_sha in $repos; do
+        IFS=\; read repo sha <<<"$repo_sha"
+        if [ "$1" = "$repo" ]; then
+            echo "$sha"
+            return
+        fi
+    done
+    echo "master"
+}
+
+function record_commit {
+    REPO_PATH=$1
+    COMMIT=$(git rev-parse HEAD)
+    if [ -e "$REPO_PATH/repos" ]; then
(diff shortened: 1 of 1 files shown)


russellb:
Can you please update the commit message with a more thorough explanation of the problem being solved, and how this patch addresses it?

p3ck:
yes - sorry about that.

p3ck:
@russellb I updated the commit and also the initial description here.  If you have any thoughts on reproducible builds I would be happy to listen.  It could be that we should simply archive all of these as artifacts in our build process.

russellb:
> For the CI workflow it is important to be able to reproduce builds.

I'm not sure why this is important, exactly.  I'd expect CI to be chasing the latest commits?

Stated a bit differently, is the problem statement: "For a given CI run, we need to know exactly which versions of all components were used" ?

p3ck:
@russellb CI will be testing the latest and the latest working combination will be passed onto QE and eventually released.  This matches the updated problem statement.

russellb:
> @russellb CI will be testing the latest and the latest working combination will be passed onto QE and eventually released. This matches the updated problem statement.

OK - there's no intention of using this repo in any sort of release pipeline.  We should discuss offline ...

PR #603, opened 2019-06-05
Drop rook and kubevirt scripts.

russellb:
These changes are to trim dev-scripts a bit further back to its
original intended scope, which is a development platform for bare
metal IPI support for the openshift installer.

These scripts made sense for demo purposes, but now that we're past
that, we should focus on evolving to the desired deployment flow.
We assume that the deployment of CNV (kubevirt++) or OCS (rook++) are
done as day 2 operations, presumably via an operator managed by OLM.

Kubevirt or rook developers may still want to manually deploy their
components on a dev-scripts cluster for development purposes,
bypassing OLM, or as a way to test things before the OLM path is
ready, but let's keep that separate from these scripts.  If desired,
those could be placed in a new repo that tells someone to deploy a
dev-scripts cluster as the first step.


derekhiggins:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/709/

russellb:
CI failed after timing out waiting for the worker node to appear.  I'm not sure how these changes would cause it.  I'll rebase and trigger another run.

derekhiggins:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/716/

karmab:
this makes sense

e-minguez:
What if instead removing the scripts/manifests are they moved into a contrib/ folder?

hardys:
> What if instead removing the scripts/manifests are they moved into a contrib/ folder?

We could, but the I think we need to make it very clear this is not the integration point for ceph/kubevirt stuff, so that we can focus on gaps in the operators that deploy those pieces.

So I'm in favor of removing these, since the eventual aim is to remove everything except the actual developer environment setup from these scripts.

derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/717/

Issue #590, opened 2019-06-03
10_deploy_rook.sh fails with error: timed out waiting for the condition while running 'oc wait --for condition=ready pod -l app=rook-ceph-tools -n openshift-storage --timeout=1200s'


mcornea:
**Describe the bug**
10_deploy_rook.sh fails with error: timed out waiting for the condition while running 'oc wait --for condition=ready pod -l app=rook-ceph-tools -n openshift-storage --timeout=1200s'

**To Reproduce**
Run make. Wait to finish.
Run 09_deploy_kubevirt.sh
Run 10_deploy_rook.sh

**Expected/observed behavior**

Observed behavior:

```shell
[centos@provisionhost-0 dev-scripts]$ ./10_deploy_rook.sh 
+ source logging.sh
+++ dirname ./10_deploy_rook.sh
++ LOGDIR=./logs
++ '[' '!' -d ./logs ']'
+++ basename ./10_deploy_rook.sh .sh
+++ date +%F-%H%M%S
++ LOGFILE=./logs/10_deploy_rook-2019-06-03-204530.log
++ echo 'Logging to ./logs/10_deploy_rook-2019-06-03-204530.log'
Logging to ./logs/10_deploy_rook-2019-06-03-204530.log
++ exec
+++ tee ./logs/10_deploy_rook-2019-06-03-204530.log
+ source common.sh
+++ go env
++ eval 'GOARCH="amd64"
GOBIN=""
GOCACHE="/home/centos/.cache/go-build"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GOOS="linux"
GOPATH="/home/centos/go"
GOPROXY=""
GORACE=""
GOROOT="/usr/lib/golang"
GOTMPDIR=""
GOTOOLDIR="/usr/lib/golang/pkg/tool/linux_amd64"
GCCGO="gccgo"
CC="gcc"
CXX="g++"
CGO_ENABLED="1"
GOMOD=""
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -fmessage-length=0 -fdebug-prefix-map=/tmp/go-build535695688=/tmp/go-build -gno-record-gcc-switches"'
[...]

Issue #570, opened 2019-05-24
After worker nodes boot it takes several minutes to get registered as nodes


mcornea:
**Describe the bug**
After worker nodes boot it takes several minutes to get registered as nodes.

Going through the kubelet log of a worker node we can see it first tries to register the node at 15:15:10:

```sh
[root@rhhi-node-3 core]# journalctl -l -u kubelet | grep 'register node' | head -1
May 24 15:15:10 rhhi-node-3 hyperkube[1240]: I0524 15:15:10.143247    1240 kubelet_node_status.go:72] Attempting to register node rhhi-node-3
```

but it only manages to register the node after 8 minutes.

```sh
[root@rhhi-node-3 core]# journalctl -l -u kubelet | grep 'registered node'
May 24 15:23:40 rhhi-node-3 hyperkube[1240]: I0524 15:23:40.111852    1240 kubelet_node_status.go:75] Successfully registered node rhhi-node-3
``` 

Based on my observations this time can vary, the longest that I've seen was around 15 minutes.

The kubelet log shows the following errors when the node doesn't get registered:

```sh
May 24 15:23:33 rhhi-node-3 hyperkube[1240]: E0524 15:23:33.094148    1240 kubelet_node_status.go:94] Unable to register node "rhhi-node-3" with API server: nodes is forbidden: User "system:anonymous" cannot create resource "nodes" in API group "" at the cluster scope
```

**To Reproduce**
Deploy worker nodes and watch the kubelet log on the worker node after it boots.

**Expected/observed behavior**

It takes several minutes for the worker node to get registered. I'd expect that the worker node would get registered right away when kubelet starts.

**Additional context**
kubelet.log - https://paste.fedoraproject.org/paste/FadFlN9S8JLD0m94YGM6qg

vrutkovs:
>I'd expect that the worker node would get registered right away when kubelet starts.

The node got registered when SDN container started, which requires a few images to be pulled, have mdns and coredns services to start.

mcornea:
> > I'd expect that the worker node would get registered right away when kubelet starts.
> 
> The node got registered when SDN container started, which requires a few images to be pulled, have mdns and coredns services to start.

It looks like the mdns and containers get started ~ 15:15 but the sdn container gets started ~ 15:23 which corresponds to the time when the node gets registered. So I guess the time to register varies based on the link parameters to quay.io where these images are hosted.

```sh
[...]

PR #604, opened 2019-06-05
Run fix_certs.sh every minute

mcornea:
It looks that running fix_certs.sh can help to avoid some issues such as #590 or #570.


hardys:
Thanks, I tested a similar workaround yesterday and it does seem to help - I'm not entirely sure why the ignition provided certs need immediately rotating, presumably something changed in the rebase to 4.2

PR #611, opened 2019-06-11
Move project cloning code into run_ci.sh

derekhiggins:
This was part of the jenkins job definition, moving it
here for more visibility, the code that clones dev-scripts
has to stay where it is though.

### DIFF
--- a/run_ci.sh
+++ b/run_ci.sh
@@ -82,6 +82,22 @@ sudo mount -o bind /opt/data/yumcache /var/cache/yum
 # Mount the kni-installer cache directory so we don't download a RHCOS image for each run
 sudo mount -o bind /opt/data/installer-cache /home/notstack/.cache/kni-install/libvirt
 
+# Clone the project being tested, "dev-scripts" will have been cloned in the jenkins
+# job definition, for all others we do it here
+ORG=openshift-metal3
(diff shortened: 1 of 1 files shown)


derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/726/

run_ci.sh
```diff
@@ -82,6 +82,22 @@ sudo mount -o bind /opt/data/yumcache /var/cache/yum
 # Mount the kni-installer cache directory so we don't download a RHCOS image for each run
 sudo mount -o bind /opt/data/installer-cache /home/notstack/.cache/kni-install/libvirt
 
+# Clone the project being tested, "dev-scripts" will have been cloned in the jenkins
+# job definition, for all others we do it here
+ORG=openshift-metal3
+if [[ "${REPO#*/}" =~ ^(baremetal-operator|metal3-dev-env|ironic-inspector-image|ironic-image|metal3-io.github.io|metal3-docs|base-image)$ ]] ; then
```
hardys: This is missing a couple of repos, but I'm not sure which are configured to run CI so we can fix in a follow-up - do we have a github API key on the host so we could retrieve the repos that way?

derekhiggins: Yup we have a key on the host but at the moment we only trigger CI on 4 repositories (below), if that expands we can look into getting list dynamically. 

repos
  - openshift-metal3/dev-scripts
  - openshift-metal3/kni-installer
  - metal3-io/ironic-image
  - metal3-io/ironic-inspector-image

Issue #512, opened 2019-05-07
Two IP addresses (DHCP)  allocated for Provisioning/External networks in the bootstrap node 


yboaron:
**Describe the bug**
Bootstrap node allocates two IP addresses (DHCP) instead of one for both Provisioning and External networks.

**To Reproduce**
Run deployment, ssh to bootstrap node before he's being deleted, and check 'ip -4 a'

**Expected/observed behavior**
a single IP address allocated for Baremetal/Provisioning networks.

**Additional context**
I 'tcpdumped' the  baremetal network at the host, and I can clearly see two DHCP requests for each network (with the same MAC address).
After looking deeply in bootstrap logs, It seems that the first DHCP request sent in initrd phase and the second one sent by NM.

yboaron:
issue #494 is a side effect of this bug.
After two IPs were created, the IP address that was allocated at initrd phase (192.168.111.43 in [1]) is removed after DHCP LEASE timeout.
But the corresponding routing entry (i.e: 192.168.111.0/24 dev ens3 proto kernel scope link src 192.168.111.43 )is not deleted, and a result of that we can't communicate with hosts in the external subnet until this route is deleted manually.

So, if it takes to the bootstrap more than DHCP lease timeout to finish his part, we may stuck in the status described in issue #494  

[1]
[core@localhost ~]$ ip route 
default via 192.168.111.1 dev ens3 
default via 192.168.111.1 dev ens3 proto dhcp metric 100 
10.88.0.0/16 dev cni0 proto kernel scope link src 10.88.0.1 linkdown 
172.22.0.0/24 dev ens4 proto kernel scope link src 172.22.0.26 
172.22.0.0/24 dev ens4 proto kernel scope link src 172.22.0.27 metric 101 
172.22.0.0/24 dev ens4 proto kernel scope link src 172.22.0.26 metric 101 
192.168.111.0/24 dev ens3 proto kernel scope link src 192.168.111.43 
192.168.111.0/24 dev ens3 proto kernel scope link src 192.168.111.44 metric 100 
192.168.111.0/24 dev ens3 proto kernel scope link src 192.168.111.43 metric 100 
192.168.111.0/24 dev ens3 proto kernel scope link src 192.168.111.2 metric 100 
[core@localhost ~]$

thom311:
I don't have now a logfile at hand to confirm, but last time I looked at this, the following appeared to happen:

[...]

PR #616, opened 2019-06-12
keepalived - extract the interface name only once

rgolangh:
In case the os has 2 interfaces, keepalived.conf would be rendered incorrectly because `INTERFACE` 
will contain 2 lines with the same name:
```
[root@dhcp-0-215 ~]# ip -o addr show to 10.35.0.0/23
2: ens3    inet 10.35.1.99/23 brd 10.35.1.255 scope global dynamic noprefixroute ens3\       valid_lft 86118sec preferred_lft 86118sec
2: ens3    inet 10.35.0.215/23 brd 10.35.1.255 scope global secondary dynamic noprefixroute ens3\       valid_lft 1700sec preferred_lft 1700sec
[root@dhcp-0-215 ~]# ip -o addr show to 10.35.0.0/23 | awk '{print $2}'
ens3
ens3
```

This fix would extract `ens3` even if we have 2 addresses on the same interface 

Review: @celebdor @yboaronn 

Related to:  https://github.com/openshift-metal3/dev-scripts/issues/512


derekhiggins:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/749/

Issue #617, opened 2019-06-13
Nodes deployment failed - cannot go from state 'deploy failed' to state 'manageable'


itzikb10:
**Describe the bug**
Using a virtualized setup (One hypervisor) installation fails 
```
level=error msg="Error: cannot go from state 'deploy failed' to state 'manageable'"
level=error
level=error msg="  on ../../../tmp/kni-install-626285310/masters/main.tf line 32, in resource \"ironic_deployment\" \"openshift-master-deployment\":"
level=error msg="  32: resource \"ironic_deployment\" \"openshift-master-deployment\" {"
level=error
```

Nodes are not deployed:
```
$ openstack baremetal node list
+--------------------------------------+--------------------+--------------------------------------+-------------+--------------------+-------------+
| UUID                                 | Name               | Instance UUID                        | Power State | Provisioning State | Maintenance |
+--------------------------------------+--------------------+--------------------------------------+-------------+--------------------+-------------+
| 3fb3633e-a9c2-46d4-a4a2-ed97a986a5e4 | openshift-master-1 | a9259fab-272a-40de-ba01-07ab8b7da0b1 | power on    | deploy failed      | False       |
| 844a883d-420f-47a3-a5bb-4aaac01be488 | openshift-master-0 | 1917e8dd-38a9-4e7d-aa86-fbac75c7d56a | power on    | deploy failed      | False       |
| 66db8847-1b5b-4a8d-8b04-f4c336c29f22 | openshift-master-2 | 6b64c177-9d20-4759-a45a-72ce2cd2eaf0 | power on    | deploy failed      | False       |
+--------------------------------------+--------------------+--------------------------------------+-------------+--------------------+-------------+
```

**From Ironic Logs**
```
/opt/dev-scripts/ironic/log/ironic/ironic-conductor.log:2019-06-13 07:33:54.231 74 ERROR concurrent.futures [req-cda4b2fc-f28a-402c-b128-594299ef5ea7 - - - - -] exception calling callback for <Future at 0x7f59fa9d0290 state=finished returned NoneType>: NodeNotLocked: Node 3fb3633e-a9c2-46d4-a4a2-ed97a986a5e4 found not to be locked on release
/opt/dev-scripts/ironic/log/ironic/ironic-conductor.log:2019-06-13 07:33:54.231 74 ERROR concurrent.futures Traceback (most recent call last):
[...]

PR #622, opened 2019-06-14
Extend Ironic conductor heartbeat timeout

cybertron:
In single node virtual environments, the system can get so overloaded
during deployment of the masters that the heartbeat times out, which
causes #617. This change overrides the heartbeat timeout to change
it from the default of 60 seconds to 120 seconds. In my experience,
this is sufficient to prevent the timeouts.

Note that this is making use of the environment driver[0] in oslo.config
for setting the value. I don't think we want to change the value
in the container config since it's primarily for virtual dev
environments.

0: https://docs.openstack.org/oslo.config/latest/reference/drivers.html#environment

### DIFF
--- a/04_setup_ironic.sh
+++ b/04_setup_ironic.sh
@@ -73,6 +73,7 @@ sudo podman run -d --net host --privileged --name mariadb --pod ironic-pod \
 
 sudo podman run -d --net host --privileged --name ironic --pod ironic-pod \
      --env MARIADB_PASSWORD=$mariadb_password \
+     --env OS_CONDUCTOR__HEARTBEAT_TIMEOUT=120 \
      -v $IRONIC_DATA_DIR:/shared ${IRONIC_IMAGE}
 
 # Start Ironic Inspector 


hardys:
Should we just change the default in the container?  I've been running into some similar issues running Ironic on the bootstrap VM (which is very CPU bound due to all the other k8s stuff running there)

cybertron:
I'm fine with that, but I don't know what, if any, negative implications raising the timeout would have. Maybe just that it would take longer for Ironic to notice if the conductor actually did go down?

dhellmann:
If the conductor goes down, kubernetes should notice faster than that timeout and restart it anyway.

dtantsur:
Given that we have only one conductor, it does not really matter :) But for a sane production environment 60 seconds should not be much (I'd expect people to make it lower not higher for a proper multi-node environment).

hardys:
Ok if this helps dev/test reliability lets merge the fix for now, but given that we expect to run Ironic in a single-node VM setup for the actual product, it'd be good to more clearly understand the bottlenecks here so we can try to optimize for that resource-contrained environment.

Issue #561, opened 2019-05-22
Remove master/worker assets from dev-scripts


hardys:
**Describe the bug**

We need to remove the temporary assets added while prototyping dns, vip and loadbalancing in dev-scripts, so that end-to-end baremetal deployments via only kni-installer (and ultimately openshift/installer) are possible without the development scripts.

The files affected are:

```
./files/etc/NetworkManager
./files/etc/NetworkManager/dispatcher.d
./files/etc/NetworkManager/dispatcher.d/pre-up.d
./files/etc/NetworkManager/dispatcher.d/pre-up.d/dns-vip-prepender
./files/etc/NetworkManager/dispatcher.d/pre-up.d/dns-vip-prepender-worker
./files/etc/chrony.conf
./files/etc/dhcp
./files/etc/dhcp/dhclient.conf.template
./files/etc/kubernetes
./files/etc/kubernetes/manifests
./files/etc/kubernetes/manifests/coredns.yaml
./files/etc/kubernetes/manifests/haproxy.yaml
./files/etc/kubernetes/manifests/keepalived.yaml
./files/etc/kubernetes/manifests/mdns-publisher-worker.yaml
./files/etc/kubernetes/manifests/mdns-publisher.yaml
./files/etc/kubernetes/static-pod-resources
./files/etc/kubernetes/static-pod-resources/coredns
./files/etc/kubernetes/static-pod-resources/coredns/Corefile.template
./files/etc/kubernetes/static-pod-resources/haproxy
./files/etc/kubernetes/static-pod-resources/haproxy/haproxy.cfg.template
./files/etc/kubernetes/static-pod-resources/haproxy/utils.sh
./files/etc/kubernetes/static-pod-resources/keepalived
./files/etc/kubernetes/static-pod-resources/keepalived/keepalived.conf.template
./files/etc/kubernetes/static-pod-resources/mdns
./files/etc/kubernetes/static-pod-resources/mdns/config.template
./files/etc/kubernetes/static-pod-resources/mdns/worker-config.template
./files/usr
./files/usr/local
./files/usr/local/bin
./files/usr/local/bin/clusterinfo
./files/usr/local/bin/fletcher8
./files/usr/local/bin/get_vip_subnet_cidr
./templates
./templates/99_ingress-controller.yaml
./templates/99_master-api-haproxy.yaml
./templates/99_master-chronyd-redhat.yaml.optional
./templates/99_master-core-password.yaml
[...]

PR #625, opened 2019-06-17
WIP: Use keepalived image from openshift

cybertron:


assets/files/etc/kubernetes/manifests/keepalived.yaml
```diff
@@ -94,7 +94,7 @@ spec:
   - name: keepalived
     securityContext:
       privileged: true
-    image: quay.io/celebdor/keepalived:latest
+    image: registry.svc.ci.openshift.org/ocp/4.2@sha256:daa9f390c43563b67546cd5b4cf3d8e351c3530f8091f523a73061fa441e8818
```
cybertron: Obviously I need to figure out how to reference this image properly.

04_setup_ironic.sh
```diff
@@ -73,6 +73,7 @@ sudo podman run -d --net host --privileged --name mariadb --pod ironic-pod \
 
 sudo podman run -d --net host --privileged --name ironic --pod ironic-pod \
      --env MARIADB_PASSWORD=$mariadb_password \
+     --env OS_CONDUCTOR__HEARTBEAT_TIMEOUT=120 \
```
cybertron: This is unrelated, but I need it for testing locally. Will remove before I push the final version of this change.

assets/files/etc/kubernetes/manifests/keepalived.yaml
russellb: You won't be able to do this "properly" in dev-scripts.  I expect these manifests will move to MCO, right?

There's a bit of magic here you need to hook into.  Assuming MCO is what has this manifest, then the image needs to be listed in this file as well: https://github.com/openshift/machine-config-operator/blob/master/install/image-references

By being listed in `image-references` for one of the cluster operators, the image will get included in the release payload.  There's also magic in CVO to convert image references in CVO deployed manifests to use the right image for that release payload.  It looks like MCO has this manifest that declares a config map of image references.  I would expect keepalived to be added here as well, so you can get the right image from your MCO code:  https://github.com/openshift/machine-config-operator/blob/master/install/0000_80_machine-config-operator_02_images.configmap.yaml

hardys: Yeah this will need to go into the MCO PR @bcrochet is working on, when that lands all this asset mangling will go away ref https://github.com/openshift-metal3/dev-scripts/pull/623

https://github.com/openshift-metal3/dev-scripts/issues/561 tracks the removal and the MCO PR is https://github.com/openshift/machine-config-operator/pull/795

russellb: In dev scripts I suppose you could use an “oc adm release” command to inspect the release image and pull out the pullspec for an image in that release payload. That would just be a very short term hack though.

assets/files/etc/kubernetes/static-pod-resources/keepalived/keepalived.conf.template
```diff
@@ -1,17 +1,22 @@
...
 vrrp_script chk_ocp {
-    script "curl -o /dev/null -kLs https://0:6443/readyz"
+    script "/usr/bin/curl -o /dev/null -kLs https://0:6443/readyz"
     interval 1
     weight 50
 }
 
 vrrp_script chk_dns {
-    script "host -t SRV _etcd-server-ssl._tcp.${DOMAIN} localhost"
+    script "/usr/bin/host -t SRV _etcd-server-ssl._tcp.${DOMAIN} localhost"
```
celebdor: This reminds me that we need to drop DNS VIP.

bcrochet:
This is obsolete and should be closed. @cybertron

bcrochet:
This is obsolete and can be closed. @cybertron

cybertron:
No longer needed.

PR #632, opened 2019-06-19
Add sanity check to vip-prepender

cybertron:
We recently had a deployment where the DNS VIP was not properly added
to resolv.conf. There isn't enough detail in the logs from the script
to determine what went wrong, so this change adds some more logging
and a sanity check that we actually did the thing we were trying to do.


hardys:
lgtm - I guess we'll want something similar to be added to https://github.com/openshift/machine-config-operator/pull/795 cc @bcrochet

metal3ci:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/776/

PR #657, opened 2019-07-09
Allow for forks with alt names

derekhiggins:
To allow for PR's from forks with names that don't match
the main repository name we've added an additional parameter
to the CI job, BASE_REPO=ORG/REPONAME, REPO hasn't changed.
An example of the values might be

+ REPO=derekhiggins/metalkube-ironic-test
+ BASE_REPO=metal3-io/ironic-image

### DIFF
--- a/run_ci.sh
+++ b/run_ci.sh
@@ -78,15 +78,11 @@ sudo mount -o bind /opt/data/installer-cache /home/notstack/.cache/kni-install/l
 
 # Clone the project being tested, "dev-scripts" will have been cloned in the jenkins
 # job definition, for all others we do it here
-ORG=openshift-metal3
-if [[ "${REPO#*/}" =~ ^(baremetal-operator|metal3-dev-env|ironic-inspector-image|ironic-image|metal3-io.github.io|metal3-docs|base-image)$ ]] ; then
-    ORG=metal3-io
-fi
 if [ -n "$REPO" -a -n "$BRANCH" ]  ; then
     pushd ~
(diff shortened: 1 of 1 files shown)


run_ci.sh
```diff
@@ -78,15 +78,11 @@ sudo mount -o bind /opt/data/installer-cache /home/notstack/.cache/kni-install/l
 
 # Clone the project being tested, "dev-scripts" will have been cloned in the jenkins
 # job definition, for all others we do it here
-ORG=openshift-metal3
```
dtantsur: I wonder if we can provide a fallback for BASE_REPO

hardys: Do we need to?  If it's a variable always set by jenkins then we shouldn't set a default?

dtantsur: At least when triggering a job manually you have to populate these things by hand. AFAIK there is no "recheck" here.

derekhiggins: Ya, we could leave this code in to set ORG if BASE_REPO isn't set, but it still wouldn't work if the fork name doesn't match the repository name.

metal3ci:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/859/

run_ci.sh
hardys: Yeah we have to manually trigger a job with the appropriate parameters, but I don't think defaulting BASE_REPO really helps much with that, since we need to populate other parameters anyway?

PR #660, opened 2019-07-09
make-bm-worker: Include consumer kind/version

zaneb:
The consumerRef should include not only the name of the consumer, but
also it's Kind and apiVersion.

Since in practice this script is used only for referencing
machine.openshift.io/v1beta1 Machine objects, just hard-code this and
revert the change of argument names from machine to consumer.

### DIFF
--- a/11_register_hosts.sh
+++ b/11_register_hosts.sh
@@ -32,15 +32,15 @@ function make_bm_masters() {
            -address "$address" \
            -password "$password" \
            -user "$user" \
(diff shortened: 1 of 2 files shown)


zaneb:
That extra apostrophe in the commit message was really bothering me :/

metal3ci:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/865/

PR #670, opened 2019-07-11
add externally provisioned flag when registering masters

dhellmann:
The host object API has changed to require a flag to explicitly mark hosts
as being externally provisioned. This change updates the command line tool
we use to register masters to add the flag.

Depends on https://github.com/metal3-io/baremetal-operator/pull/255


metal3ci:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/882/

PR #676, opened 2019-07-17
Gather correct data for ironic bmo configmap

karmab:


08_deploy_bmo.sh
```diff
@@ -16,6 +22,8 @@ sed -i 's/namespace: .*/namespace: openshift-machine-api/g' ocp/deploy/role_bind
 cp $SCRIPTDIR/operator_ironic.yaml ocp/deploy
 cp $SCRIPTDIR/ironic_bmo_configmap.yaml ocp/deploy
 sed -i "s#__RHCOS_IMAGE_URL__#${RHCOS_IMAGE_URL}#" ocp/deploy/ironic_bmo_configmap.yaml
+sed -i "s#provisioning_interface: \"ens3\"#provisioning_interface: \"${PRO_IF}\"#" ocp/deploy/ironic_bmo_configmap.yaml
+sed -i "s#cache_url: \"http://192.168.111.1/images\"#cache_url: \"http://${BAREMETAL_IP}/images\"#" ocp/deploy/ironic_bmo_configmap.yaml
```
hardys: This will leave the git-tracked ironic_bmo_configmap.yaml modified, we should probably sed into a new file, or rename the one tracked in git to .yaml.template or something?

We could also then put some easier placeholders in for sed to replace?

08_deploy_bmo.sh
```diff
@@ -6,6 +6,12 @@ source logging.sh
 source common.sh
 eval "$(go env)"
 
+# Set default value for provisioning interface
+PRO_IF=${PRO_IF:-ens3}
```
hardys: This hijacks an existing variable from common.sh, and they're not the same thing, the existing one is the provisioning interface on the host running dev-scripts, but this one is the interface used for provisioning on the target RHCOS nodes.

08_deploy_bmo.sh
karmab: no, it s modifying the generated file in ocp directory

08_deploy_bmo.sh
karmab: changed the name to PROV_IF to avoid overriding existing variables

08_deploy_bmo.sh
```diff
@@ -6,6 +6,12 @@ source logging.sh
 source common.sh
 eval "$(go env)"
 
+# Set default value for provisioning interface
+PROV_IF=${PRO_IF:-ens3}
```
hardys: Would this be clearer if we called it something like CLUSTER_PRO_IF?  Particularly since it's defaulting to the provisioning host PRO_IF, having a slightly different name for a different interface could be a bit confusing for folks that don't know the scripts well?

PR #680, opened 2019-07-22
Changes to use Ironic Stein

elfosardo:
This patch contains changes needed to be able to use Ironic Stein.

It has dependencies from other patches in other related repos
and they should all be merged at the same time, following the order below,
to guarantee a correct behavior.

Full patch set:
https://github.com/openshift-metal3/dev-scripts/pull/680
https://github.com/metal3-io/ironic-image/pull/79
https://github.com/metal3-io/ironic-inspector-image/pull/30
https://github.com/metal3-io/ironic-ipa-downloader/pull/3

### DIFF
--- a/ironic_cleanup.sh
+++ b/ironic_cleanup.sh
@@ -5,7 +5,7 @@ source logging.sh
 source common.sh
 
 # Kill and remove the running ironic containers
-for name in ironic ironic-api ironic-conductor ironic-exporter ironic-inspector dnsmasq httpd mariadb; do
+for name in ironic ironic-api ironic-conductor ironic-inspector dnsmasq httpd mariadb; do
     sudo podman ps | grep -w "$name$" && sudo podman kill $name
     sudo podman ps --all | grep -w "$name$" && sudo podman rm $name -f
 done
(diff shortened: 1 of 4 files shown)


elfosardo:
@iurygregory thanks and done all!

hardys:
The commit message says "It has dependencies from other patches in other related repos." but it's not actually dependent on the image changes, right?  Because it's just removing the exporter container?

I guess we can run CI to check the remaining functionality is still working, then we'll have a follow-up to switch to the new downstream images?

hardys:
> @hardys I think it's dependent since we are also removing things like mDNS etc, it's not only the ironic-exporter container and features.

Ack, my point is this PR doesn't actually switch to using those images yet, so it's not actually dependent on those changes?

@ elfosardo Should we be making that change here, landing the image PRs then running CI, or will that be a follow-up?

metal3ci:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/917/

ironic_cleanup.sh
```diff
@@ -5,7 +5,7 @@ source logging.sh
 source common.sh
 
 # Kill and remove the running ironic containers
-for name in ironic ironic-api ironic-conductor ironic-exporter ironic-inspector dnsmasq httpd mariadb; do
```
dtantsur: nit: would could keep ironic-exporter here for updates to existing environments

elfosardo:
@hardys the patch should be merged in order and this actually blocks all the others

PR #685, opened 2019-07-23
Makefile: Add "redeploy" target.

russellb:
When all goes well, "make redeploy" will result in a quicker
redeployment than "make clean all".

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,8 @@ default: requirements configure repo_sync ironic ocp_run deploy_bmo register_hos
 
 all: default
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -3,6 +3,8 @@ default: requirements configure repo_sync ironic ocp_run deploy_bmo register_hos
 
 all: default
 
+redeploy: ocp_cleanup ironic_cleanup ironic ocp_run deploy_bmo reigster_hosts bell
```
dtantsur: typo: register

russellb: oops, thanks

PR #689, opened 2019-07-24
Enable network service in RHEL8

### LINKED ISSUE #688: Rebooting the RHEL8 helper makes it unreachable
**Describe the bug**
After configuring the ifcfg-* files and rebooting the helper, it is unreachable because the network service is not enabled in RHEL8

**To Reproduce**
Run the 02 script and reboot the helper node.

Cannot log-in via network and `ip a` shows no IPs

**Additional context**

e-minguez:
It is disabled by default in RHEL8
Fixes #688

### DIFF
--- a/02_configure_host.sh
+++ b/02_configure_host.sh
@@ -112,6 +112,12 @@ if [ "$MANAGE_INT_BRIDGE" == "y" ]; then
     fi
 fi
 
+# If there were modifications to the /etc/sysconfig/network-scripts/ifcfg-*
+# files, it is required to enable the network service
+if [ "$MANAGE_INT_BRIDGE" == "y" ] || [ "$MANAGE_PRO_BRIDGE" == "y" ]; then
+  sudo systemctl enable network
+fi
+
 # restart the libvirt network so it applies an ip to the bridge
 if [ "$MANAGE_BR_BRIDGE" == "y" ] ; then
     sudo virsh net-destroy baremetal


02_configure_host.sh
```diff
@@ -111,6 +111,9 @@ if [ "$MANAGE_INT_BRIDGE" == "y" ]; then
         else
            sudo systemctl restart network
         fi
+        if [ "${RHEL8}" = "True" ] ; then
+          sudo systemctl enable network
+        fi
```
hardys: I'm wondering how this ends up disabled?  Was it a manual ISO install where no network was activated as part of the install?

Regardless this seems unrelated to the OS version, perhaps we should just unconditionally enable then restart it, to remove the restart duplication above?

hardys: Ah I see the restart de-duplication is handled in https://github.com/openshift-metal3/dev-scripts/pull/690 thanks!

e-minguez: I've added the enable only if the ifcfg-* files are modified

russellb: Is the `network` service only for handling `ifcfg` scripts?  In RHEL8 that's not installed by default.  We install it so we can still use `ifcfg` scripts.  That's probably why it has to be enabled in RHEL8 and we didn't see that before.

02_configure_host.sh
```diff
@@ -112,6 +112,12 @@ if [ "$MANAGE_INT_BRIDGE" == "y" ]; then
     fi
 fi
 
+# If there were modifications to the /etc/sysconfig/network-scripts/ifcfg-*
+# files, it is required to enable the network service
+if [ "$MANAGE_INT_BRIDGE" == "y" ] || [ "$MANAGE_PRO_BRIDGE" == "y" ]; then
+  sudo systemctl enable network
```
russellb: I suspect this was only needed for RHEL8 since we install `network-scripts` for `ifcfg` file support, which isn't there by default.  I suppose this is harmless elsewhere though.

hardys:
This is fine, but I can't help wondering if we should just move to nmcli for RHEL8 instead?

Here's an example of configuring an interface with nmcli:

https://github.com/hardys/installer/blob/issue_2060/data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template#L29

Not blocking this as a quick-fix, but wanted to have that discussion and now seems a good time to mention it :)

russellb:
+1 to moving to network manager for all of this!


On Jul 24, 2019, at 11:01 AM, Steven Hardy <notifications@github.com> wrote:

This is fine, but I can't help wondering if we should just move to nmcli
for RHEL8 instead?

Here's an example of configuring an interface with nmcli:

https://github.com/hardys/installer/blob/issue_2060/data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template#L29

Not blocking this as a quick-fix, but wanted to have that discussion and
now seems a good time to mention it :)

—
You are receiving this because you commented.
Reply to this email directly, view it on GitHub
<https://github.com/openshift-metal3/dev-scripts/pull/689?email_source=notifications&email_token=AACLQCQ3JUN7RHPEXDBYK7TQBBVCZA5CNFSM4IGNF4S2YY3PNVWWK3TUL52HS4DFVREXG43VMVBW63LNMVXHJKTDN5WW2ZLOORPWSZGOD2WTXVI#issuecomment-514669525>,
or mute the thread

e-minguez:
Can we have this merged then create an issue to track the NetworkManager modifications? Thanks!

hardys:
> Can we have this merged then create an issue to track the NetworkManager modifications? Thanks!
> 
> (also, TIL about https://github.com/nmstate/nmstate)

Sure, and yes maybe nmstate would be a nice thing to investigate as an alternative to shell hacking :)

PR #691, opened 2019-07-24
Switch to fully qualified RHCOS image URL

hardys:
This aligns with the format expected from openshift-install, and
means we can remove some backwards compatibility from the downloader
container ref https://github.com/openshift-metal3/rhcos-downloader/pull/6


metal3ci:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/927/

PR #693, opened 2019-07-24
Add validation for number of nodes in NODES_FILE

hardys:
Some folks have seen a confusing error like:

```
Terraform variables: Unknown BMC type 'null' for address null://null\""
```

It turns out this is because the NODES_FILE created outside
of dev-scripts contains only two nodes, so the jq slice to create
the 3 masters in the install-config returns null for the missing
node.

Adding some validation here should make it clearer, but probably we should
catch this via validation in the installer.

### DIFF
--- a/06_create_cluster.sh
+++ b/06_create_cluster.sh
@@ -44,6 +44,13 @@ if [ ! -d ocp ]; then
       build_installer
     fi
 
+    # Validate there are enough nodes to avoid confusing errors later..
+    NODES_LEN=$(jq '.nodes | length' ${NODES_FILE})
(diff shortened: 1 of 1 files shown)


metal3ci:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/928/

hardys:
Hmm CI failure looks real, testing locally

metal3ci:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/933/

metal3ci:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/934/

PR #699, opened 2019-07-26
WIP: assets: Drop ingress-controller manifest.

russellb:
The intention of this manifest was to change the node selector so that
the ingress controller would run on masters.


russellb:
Marked as WIP while CI runs ...

metal3ci:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/951/

metal3ci:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/954/

russellb:
The CI failure is real here.  This change is still needed, so more work is needed to figure out the right change to make to be able to drop this manifest.

russellb:
tracked via #705

PR #700, opened 2019-07-26
drop master host registration

dhellmann:
With https://github.com/openshift/installer/pull/2087 in place the
hosts for master nodes will be registered by the installer so we can
drop the step from dev-scripts.

### DIFF
--- a/11_register_hosts.sh
+++ b/11_register_hosts.sh
@@ -8,38 +8,6 @@ source logging.sh
 
 eval "$(go env)"
 
-function list_masters() {
-    cat $MASTER_NODES_FILE | \
-        jq '.nodes[] | {
-           name,
-           driver,
-           address:.driver_info.ipmi_address,
-           port:.driver_info.ipmi_port,
-           user:.driver_info.ipmi_username,
-           password:.driver_info.ipmi_password,
-           mac: .ports[0].address
-           } |
-           .name + " " +
-           .driver + "://" + .address + (if .port then ":" + .port else "" end) + " " +
-           .user + " " + .password + " " + .mac' \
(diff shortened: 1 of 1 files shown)


metal3ci:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/1128/

russellb:
Why not workers as well?

russellb:
This lgtm in any case. Removing the rest can be tested as a follow up.

hardys:
Agree it would be good to discuss the plan to remove worker registration and the 90_metal3_baremetalhost_crd.yaml definition (since that should now be handled by the MAO), but this seems like a step in the right direction and passed CI so lets merge it.

dhellmann:
The last time I looked the installer didn't have worker hardware details, so couldn't create the resources. I would prefer if we just leave that to a Day 2 operation, but I don't know if that's acceptable to the PMs.

There's a race condition in the bootstrapping code that means we need the host CRD registered earlier than it is if we wait for the MAO to come up. I'm working on that separately.

russellb:
> The last time I looked the installer didn't have worker hardware details, so couldn't create the resources. I would prefer if we just leave that to a Day 2 operation, but I don't know if that's acceptable to the PMs.

I thought the installer would create BareMetalHosts for every host we put in the install-config.  I see that dev-scripts only puts the first 3 hosts in the install-config, though.

hardys:
> I thought the installer would create BareMetalHosts for every host we put in the install-config. I see that dev-scripts only puts the first 3 hosts in the install-config, though.

Yeah this was my assumption too, but as you say we'll need to modify dev-scripts to populate the worker details in the install-config.

PR #701, opened 2019-07-26
Remove facet references.

russellb:
Facet development has been dormant for quite a while.  Some of the
relevant code was commented out, but we were still installing a few
unnecessary dependencies (yarn, nodejs).

Drop the facet related references from dev-scripts for now.  It can be
re-introduced when it's being actively used and we're sure the
additions are working properly.

When development of an installer UI resumes, it may take a different
form or be in a different location, so let's revisit dev-scripts
additions when that time comes.

### DIFF
--- a/01_install_requirements.sh
+++ b/01_install_requirements.sh
@@ -131,10 +131,6 @@ else
     # There are some packages which are newer in the tripleo repos
     sudo yum -y update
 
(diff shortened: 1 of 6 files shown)


russellb:
@honza @jtomasek you OK with this?

Issue #260, opened 2019-04-01
Certs not getting auto approved, stuck in pending state


russellb:
All of our deployments have a problem where certs aren't being approved automatically, and the following command is needed as a workaround:

```
oc get csr -o name | xargs -n 1 oc adm certificate approve
```
------------------------

Other PRs / issues tracked through this investigation:

- [x] https://github.com/openshift/machine-api-operator/pull/235
- [x] https://github.com/openshift/machine-api-operator/pull/268
- [x] https://github.com/openshift/cluster-api-provider-baremetal/pull/9
- [x] https://github.com/openshift/api/pull/274
- [x] https://github.com/openshift-metalkube/dev-scripts/pull/271
- [x] https://github.com/openshift-metalkube/kni-installer/issues/32
- [x] https://github.com/openshift/machine-api-operator/pull/271
- [x] https://github.com/metalkube/cluster-api-provider-baremetal/issues/49
- [x] https://github.com/openshift/machine-api-operator/pull/314 -- replaced by https://github.com/openshift/machine-api-operator/pull/321
- [x] https://github.com/metal3-io/cluster-api-provider-baremetal/pull/24
- [x] https://github.com/metal3-io/baremetal-operator/issues/190
 - [ ] introspection data needed on BareMetalHosts that represent externally provisioned masters - https://github.com/metal3-io/baremetal-operator/issues/242
- [ ] https://github.com/openshift-metal3/kni-installer/issues/46

russellb:
My machine-approver log has a bunch of:

```
I0321 18:09:39.113975       1 main.go:97] CSR csr-grgft added
I0321 18:09:39.133134       1 main.go:123] CSR csr-grgft not authorized: No target machine
I0321 18:14:32.453059       1 main.go:97] CSR csr-f66gv added
I0321 18:14:32.464451       1 main.go:123] CSR csr-f66gv not authorized: No target machine
I0321 18:18:11.088381       1 main.go:97] CSR csr-jmf46 added
I0321 18:18:11.102598       1 main.go:123] CSR csr-jmf46 not authorized: No target machine
I0321 18:22:27.112564       1 main.go:97] CSR csr-zxjkn added
I0321 18:22:27.127977       1 main.go:123] CSR csr-zxjkn not authorized: No target machine
I0321 18:27:20.454085       1 main.go:97] CSR csr-f85k9 added
```

The code of interest is authorizeCSR() in here:

[...]

PR #708, opened 2019-07-30
Fix certs even if no workers are deployed

e-minguez:


russellb:
Please hold on this. I need to understand why this isn’t working automatically for masters first. 

Related to #260

russellb:
I looked into it and now understand why this is required.

The docs were just updated last week to clarify that only the node client cert will be auto approved on an ongoing basis, and that some automation will always be required for the server certs.

https://bugzilla.redhat.com/show_bug.cgi?id=1720178

https://github.com/openshift/openshift-docs/commit/6912507d69af22bfb8bd49ac920ec84221b97afb

https://github.com/openshift/openshift-docs/pull/16060

11_register_hosts.sh
```diff
@@ -78,13 +78,13 @@ list_workers | make_bm_workers | tee $SCRIPTDIR/ocp/worker_crs.yaml
...
 
-# Check if file exists
-[ -s "$SCRIPTDIR/ocp/worker_crs.yaml" ] || exit 0
-
-oc --config ocp/auth/kubeconfig apply -f $SCRIPTDIR/ocp/worker_crs.yaml --namespace=openshift-machine-api
-
 # Run the fix_certs.sh script periodically as a workaround for
 # https://github.com/openshift-metalkube/dev-scripts/issues/260 This is only
 # required to approve certs for workers, as the master certs are approved
 # automatically during the bootstrap phase.
```
zaneb: Shouldn't we remove this last sentence from the comment again as well?

russellb: yes

PR #709, opened 2019-07-30
Fix provisioning IP in ironic-bmo-configmap for BM

### LINKED ISSUE #707: Wrong ironic_endpoint and ironic_inspector_endpoint on baremetal
**Describe the bug**
`ironic-bmo-configmap` has `ironic_endpoint: http://172.22.0.3:6385/v1/` and `ironic_inspector_endpoint: http://172.22.0.3:5050/v1/` but on baremetal deployments, I believe it should be the provisioning IP of the helper node instead.

**To Reproduce**
Run dev-scripts to completion and observe the `ironic-bmo-configmap`

The endpoints are calculated depending on the environment.

**Additional context**

e-minguez:
Fixes #707


russellb:
> I don't think we want this change. This is going to set the IP to refer to the IP of the provisioning host, but we want it to use the static IP managed by the containers running in the cluster so we can talk to the copy of Ironic running there.

That's right.

I imagine this needs to be changeable if a different subnet is used for the provisioning network, but by default this is the correct address, and it should definitely not be the provisioning host IP

PR #713, opened 2019-07-31
WIP: deploy solution within cluster to autosign certs

karmab:
As a temporary solution, we provide an artifact which will autosign certs each 5 minutes after initial deployment

### DIFF
--- a/06_create_cluster.sh
+++ b/06_create_cluster.sh
@@ -72,11 +72,11 @@ if [ $(sudo podman ps | grep -w -e "ironic-api$" -e "ironic-conductor$" -e "iron
     exit 1
 fi
 
-# Run the fix_certs.sh script periodically as a workaround for
-# https://github.com/openshift-metalkube/dev-scripts/issues/260
-sudo systemd-run --on-active=30s --on-unit-active=1m --unit=fix_certs.service $(dirname $0)/fix_certs.sh
-
 # Call openshift-installer to deploy the bootstrap node and masters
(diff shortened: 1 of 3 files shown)


russellb:
Take a look at this script - https://github.com/openshift/openshift-ansible/blob/release-3.11/roles/openshift_bootstrap_autoapprover/files/openshift-bootstrap-controller.yaml

what do you think about adapting that?

russellb:
also - if we switch to something in cluster, we can drop the cron job from the provisioning host

karmab:
> https://github.com/openshift/openshift-ansible/blob/release-3.11/roles/openshift_bootstrap_autoapprover/files/openshift-bootstrap-controller.yaml

actually, i had a look at this first, but it seems to be:
- specific to 3.11
- missing the image name
- too complex for a simple task

russellb:
> > https://github.com/openshift/openshift-ansible/blob/release-3.11/roles/openshift_bootstrap_autoapprover/files/openshift-bootstrap-controller.yaml
> 
> actually, i had a look at this first, but it seems to be:
> 
>     * specific to 3.11
> 
>     * missing the image name

I said adapt, not copy verbatim :-)

>     * too complex for a simple task

Well, I disagree.  I like the `oc observe` approach better, as it will react to any CSRs immediately instead of checking ever N minutes.

karmab:
@russellb provided an adaptation of the statefulset (untested though)

PR #716, opened 2019-07-31
add BareMetalHost CRD

dhellmann:
Until the machine-api-operator integration is finished we need the
host CRD to be defined during the installation process so the
installer can register host objects with the new cluster. This change
adds a copy of the CRD to the manifest templates so it is uploaded to
the cluster with the other manifests managed by the installer.


metal3ci:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/982/

russellb:
We could remove it from script #8 now, but it doesn't hurt anything for it to still be applied there too

PR #720, opened 2019-08-01
Wait for baremetal-operator pod to become ready

mcornea:
The baremetal-operator pod can take around 20 minutes to come up. Currently make exits successfully without the baremetal-operator being ready. This change waits for the pod to become ready before registering the baremetalhosts in 11_register_hosts.sh.


russellb:
There's no need to wait.  What's the motivation for the change?

mcornea:
> There's no need to wait. What's the motivation for the change?

I'd expect that when the installer finishes all the components are ready for use. Otherwise we need to check and wait post-deployment that the cluster is ready. It would also allow us to catch issues in case one of the baremetal-operator pods containers fail to start.

mcornea:
This is now covered by https://github.com/openshift-metal3/dev-scripts/blob/master/link-machine-and-node.sh#L92-L114

PR #722, opened 2019-08-05
Add MachineCIDR definition to install-config.yaml

bcrochet:
openshift/installer#2140 adds validation that the VIPs are a) in the
MachineCIDR and b) not in ClusterNetworks or ServiceNetworks.

This patch defines the MachineCIDR to EXTERNAL_SUBNET in order to
satisfy the first requirement.


cybertron:
We'll also need to add this to the example install-config in https://github.com/openshift/installer/blob/master/docs/user/metal/install_ipi.md

bcrochet:
/review

> We'll also need to add this to the example install-config in https://github.com/openshift/installer/blob/master/docs/user/metal/install_ipi.md

Thanks for the suggestion. I'll add it to my installer PR.

PR #728, opened 2019-08-07
rename the host CRD template to ensure it is installed earlier

dhellmann:
Asset manifests are installed in lexical order. The change in the
installer to register hosts uses a prefix of 95, so we need to use an
earlier value to ensure the CRD is registered before the host objects


metal3ci:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/1006/

PR #737, opened 2019-08-14
Change the order of the ironic containers

derekhiggins:
To ensure the IP address on the provisioning interface is set
We need static-ip-refresh to start before the others, observation
suggests the order in this file is the order they are started in.

### DIFF
--- a/operator_ironic.yaml
+++ b/operator_ironic.yaml
@@ -71,45 +71,24 @@ spec:
                   name: metal3-config
                   key: provisioning_interface
       containers:
-        - name: baremetal-operator
-          image: quay.io/metal3-io/baremetal-operator:master
-          ports:
-            - containerPort: 60000
-              name: metrics
+        - name: static-ip-refresh
+          image: quay.io/metal3-io/static-ip-manager:latest
           command:
-            - /baremetal-operator
+            - /refresh-static-ip
           imagePullPolicy: Always
+          securityContext:
+            privileged: true
           env:
-            - name: WATCH_NAMESPACE
-              valueFrom:
-                fieldRef:
-                  fieldPath: metadata.namespace
-            - name: POD_NAME
-              valueFrom:
-                fieldRef:
-                  fieldPath: metadata.name
-            - name: OPERATOR_NAME
-              value: "baremetal-operator"
-            - name: DEPLOY_KERNEL_URL
+            - name: PROVISIONING_IP
               valueFrom:
                 configMapKeyRef:
                   name: metal3-config
-                  key: deploy_kernel_url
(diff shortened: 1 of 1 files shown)


hardys:
Hmm we set this via static-ip-set in the initContainers section, but the lft there is 300seconds, I was expecting that to be enough given that it's at the end of the list after the *downloader containers?

https://github.com/openshift/ironic-static-ip-manager/blob/master/set-static-ip#L20
https://github.com/openshift-metal3/dev-scripts/blob/master/operator_ironic.yaml#L55

Note that when we have this figured out we need to ensure the corresponding change gets added to the MAO, the dev-scripts operator_ironic.yaml is going to be removed very soon cc @sadasu @imain

imain:
yeah it could be that time out.. :/  We should probably make it quite long for now.

imain:
If this happens again can you ssh into the masters, and check the provisioning interfaces?  One of them should have the IP specified (172.22.0.3 by default).  If that's not set anywhere then the timeout wasn't long enough.

derekhiggins:
> Hmm we set this via static-ip-set in the initContainers section, but the lft there is 300seconds, I was expecting that to be enough given that it's at the end of the list after the *downloader containers?

The problem wasn't the downloader containers, the problem was the time it takes to download the ironic, ironic-inspector etc.. container images, which happen after static-ip-set is finished,  these delay starting the other containers. Making sure we start static-ip-refresh first will hopefully ensure the IP is set when ironic starts.

imain:
static-ip-set is called as the last init container.  So the only wait is between when it runs and when the containers in the pod start.  I'm not sure that putting refresh first will totally solve the problem but it'd be worth checking/watching.

hardys:
Should we go ahead with this, or just get https://github.com/openshift/ironic-image/pull/15 in which makes the containers which depend on the provisioning interface wait until it's up?

russellb:
Changing the order is fine, but the fixes in the image seem like the real fix here.

derekhiggins:
> Should we go ahead with this, or just get [openshift/ironic-image#15](https://github.com/openshift/ironic-image/pull/15) in which makes the containers which depend on the provisioning interface wait until it's up?

Did you mean this comment here, I'm guessing this PR should now be closed as we no longer us operator_ironic.yaml ?

derekhiggins:
Closing, no longer relevant

hardys:
> Did you mean this comment here, I'm guessing this PR should now be closed as we no longer us operator_ironic.yaml ?

Yeah good point, I was thinking of the MAO ordering PR which already got closed, hopefully the image changes are sufficient.

PR #743, opened 2019-08-19
use REGISTRIES env variable

karmab:
- allow to set REGISTRIES env variable to create registries.conf with insecure registries

### DIFF
--- a/assets/files/etc/containers/registries.conf
+++ b/assets/files/etc/containers/registries.conf
@@ -1,0 +1,8 @@
+[registries.search]
+registries = [${REGISTRIES}, 'registry.access.redhat.com', 'docker.io']
+
+[registries.insecure]
+registries = [${REGISTRIES}]
+
+[registries.block]
+registries = []
--- a/assets/templates/99_master-registries-custom.yaml.optional
+++ b/assets/templates/99_master-registries-custom.yaml.optional
@@ -1,0 +1,17 @@
+apiVersion: machineconfiguration.openshift.io/v1
+kind: MachineConfig
+metadata:
+  labels:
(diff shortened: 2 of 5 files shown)


assets/files/etc/containers/registries.conf
```diff
@@ -0,0 +1,9 @@
+[root@u08-master-0 containers]# cat registries.conf
```
e-minguez: :rocket:

russellb:
Can you add some more explanation about this?  What is it?  Why is it needed?  Is there an openshift-kni/install-scripts issue that describes the need in more detail?

karmab:
the need is for people wanting to use their own insecure registries during the install

mcornea:
A good example where this is needed is for testing pre-release CNV 2.1 which uses images from an internal insecure registry.

russellb:
so this can be done post-install then?  and it's needed for CNV testing?  In that case, I wouldn't put it in dev-scripts.

There is an issue for this against install-scripts though: https://github.com/openshift-kni/install-scripts/issues/12

karmab:
> so this can be done post-install then? and it's needed for CNV testing? In that case, I wouldn't put it in dev-scripts.
> 
> There is an issue for this against install-scripts though: [openshift-kni/install-scripts#12](https://github.com/openshift-kni/install-scripts/issues/12)

so, it s useful for cnv but not only, it will probably be useful when we test ocs and would be useful for anyone wiling to deploy using a custom internal registry

russellb:
I still see this as out of scope for dev-scripts, which is aimed at a dev env for baremetal infrastructure platform support.  Feel free to discuss this further as an add-on for install-scripts.

PR #745, opened 2019-08-19
Ironic IP is wrong in docs

rdoxenham:
Now Ironic runs on the bootstrap VM it's no longer routable on localhost, updating

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,7 +126,7 @@ master-2   Ready     master    20m       v1.12.4+50c2f2340a
 
 ## Interacting with Ironic directly
 
-For manual debugging via openstackclient, you can use the following:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -130,7 +130,7 @@ For manual debugging via openstackclient, you can use the following:
 
 ```
 export OS_TOKEN=fake-token
-export OS_URL=http://localhost:6385/
+export OS_URL=http://172.22.0.2:6385/
 openstack baremetal node list
 ...
 ```
```
hardys: We should probably also reword the "Note that ..." comment below, something like "Note that the OS_URL above refers to the provisioning network IP of the bootstrap VM (started via openshift-install on the provisioning host).  After initial cluster bootstrapping is completed, this VM is destroyed, and Ironic services are then started on the deployed cluster, e.g one of the masters.  To access this instance ..."

PR #750, opened 2019-08-20
Fix virtualbmc service on RHEL-8

markmc:
We're installing virtualbmc into `/usr/loca`' and the systemd unit assumes it is in `/usr`. So the service is always in `failed` state.

virtualbmc is very cunning so when you run `vbmc add`, it goes ahead and forks the daemon:

```
 DEBUG VirtualBMC [-] Server at 50891 connection error: Server response timed out
 DEBUG VirtualBMC [-] Attempting to start `vbmcd` behind the scenes. Consider configuring your system to manage `vbmcd` via systemd. Automatic `vbmcd` start up will be removed in the future releases!
```

Things break badly only much later because cleanup with `systemctl stop` doesn't do what we expect.

Surely Ansible's systemd module would notice the service failing to start? Apparently not! https://github.com/ansible/ansible/issues/24450

### DIFF
--- a/01_install_requirements.sh
+++ b/01_install_requirements.sh
@@ -100,6 +100,7 @@ if [ "${RHEL8}" = "True" ] ; then
     git pull -r
     sudo pip3 install .
     curl 'https://review.rdoproject.org/r/gitweb?p=openstack/virtualbmc-distgit.git;a=blob_plain;f=virtualbmc.service;hb=HEAD' > virtualbmc.service
+    sed -i 's|/usr/bin/vbmcd|/usr/local/bin/vbmcd|' virtualbmc.service
     sudo mv virtualbmc.service /etc/systemd/system/.
     sudo systemctl daemon-reload
     popd ; popd


hardys:
Looks good, I suspect this is the same issue @russellb and I hit ref https://github.com/metal3-io/metal3-dev-env/issues/30

It'd be nice to de-duplicate this stuff (e.g by moving all the vbmc install into the metal3-dev-env ansible roles) but until we do, we'll need the same fix in metal3-dev-env I think.

markmc:
> Looks good, I suspect this is the same issue @russellb and I hit ref [metal3-io/metal3-dev-env#30](https://github.com/metal3-io/metal3-dev-env/issues/30)
> 
> It'd be nice to de-duplicate this stuff (e.g by moving all the vbmc install into the metal3-dev-env ansible roles) but until we do, we'll need the same fix in metal3-dev-env I think.

Looks like you were using dev-scripts install scripts, since metal3-dev-env doesn't have this RHEL-8 virtualbmc support

hardys:
> > Looks good, I suspect this is the same issue @russellb and I hit ref [metal3-io/metal3-dev-env#30](https://github.com/metal3-io/metal3-dev-env/issues/30)
> > It'd be nice to de-duplicate this stuff (e.g by moving all the vbmc install into the metal3-dev-env ansible roles) but until we do, we'll need the same fix in metal3-dev-env I think.
> 
> Looks like you were using dev-scripts install scripts, since metal3-dev-env doesn't have this RHEL-8 virtualbmc support

Ah good point, since the error was in the ansible teardown I incorrectly assumed we'd need a fix in the ansible role.

Still would be good to move all the virtbmc stuff into a common location e.g the ansible role in due course :)

russellb:
Thank you for tracking this down!

PR #751, opened 2019-08-20
add a script to add hardware data to a host

dhellmann:
Introspection is currently broken, and that prevents workers from
being useful. This script is a hack to add fake introspection data to
the host so it will appear to be complete and allow the worker to be
used.

### DIFF
--- a/skip-introspection-hack.sh
+++ b/skip-introspection-hack.sh
@@ -1,0 +1,67 @@
+#!/bin/bash
+
+source utils.sh
+
+set -x
+set -e
+
+host="$1"
+
+if [ -z "$host" ]; then
+    echo "Usage: $0 HOST"
+    exit 1
+fi
+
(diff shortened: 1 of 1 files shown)


hardys:
Should we merge this as an example of how to patch a host, or was it just a temporary PR to show @rdoxenham how to do it?

I tested with store_data = database locally and AFAICS it's now working, so probably we can reinstate introspection then this will no longer be required.

rdoxenham:
My vote - merge it... it can’t harm anything and we can always remove it later. Plus, as it stands now, until we fix the inspection store_data thing you can’t add any additional nodes post deployment as the BMO gets stuck waiting for inspection data even though Ironic states the node has finished (manageable). Therefore in the interim this patch will allow people to proceed with adding nodes, albeit with fake inspection data.

russellb:
we've merged the patch to re-enable the introspection data store

PR #755, opened 2019-08-21
Move to v1 install-config version

markmc:
As of openshift/installer#1589 the install-config format has been
locked in as the v1 version.


metal3ci:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/1075/

PR #764, opened 2019-08-23
Use provisioning bootstrap ip for debugging

karmab:
using the ip from provisioning network allows to use the same debugging process whether on virtual or physical environments

### DIFF
--- a/show_bootstrap_log.sh
+++ b/show_bootstrap_log.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
 
-BOOTSTRAP_VM_IP=$(sudo virsh net-dhcp-leases baremetal | grep -v master | grep ipv4 | tail -n1 | sed -e 's/.*\(192.*\)\/.*/\1/')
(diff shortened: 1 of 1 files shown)


show_bootstrap_log.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
 
-BOOTSTRAP_VM_IP=$(sudo virsh net-dhcp-leases baremetal | grep -v master | grep ipv4 | tail -n1 | sed -e 's/.*\(192.*\)\/.*/\1/')
-echo "Attempting to follow $1 on ${BOOTSTRAP_VM_IP} ..."
+BOOTSTRAP_VM_IP="172.22.0.2"
```
hardys: I don't think we should make this the default, because this IP relies on the static-ip-downloader starting via the ironic.service, it's something which can break (if we fail to pull the image for example, or if someone misconfigures the provisioning nic name), it'd be fine to mention this as an option in the docs though.

PR #765, opened 2019-08-23
document some of the variables used in config_example.sh

karmab:
Begin documenting some of the variables that can be set in config.sh env file


russellb:
ah yes, there were docs for these in README.md but not the config sample ...

PR #768, opened 2019-08-26
Don't set a pw for the core user on masters.

russellb:
Drop the manifest that sets a password for the core user on masters.
Most people probably don't even know this is here, and it shouldn't be
necessary since an ssh key is supported.  In any case, we definitely
should stop doing this by default.

Part of issue #561


metal3ci:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/1098/

PR #770, opened 2019-08-26
tools and instructions for running development version of baremetal-operator

dhellmann:


zaneb:
> The copy of the Deployment here is different from any of the others, because it doesn't include the baremetal-operator container at all.

Could we create it on the fly by transforming the one from OpenShift? That would eliminate the need to keep them in sync.

> The point of this is that I don't want to have to wait for container images to build every time I edit the operator code while I'm developing/testing changes.

I guess what I'm saying is that in addition to that there's another use case: building the cluster using a custom image as the 'real, full' version of metal3 that is needed for the cluster to come up, and see if the cluster does actually come up.

dhellmann:
> > The copy of the Deployment here is different from any of the others, because it doesn't include the baremetal-operator container at all.
> 
> Could we create it on the fly by transforming the one from OpenShift? That would eliminate the need to keep them in sync.

I thought about that, but decided writing a bunch of code to do it would be more work than necessary for now. If it turns into a source of friction, I'll be the first to jump on automating it.

> > The point of this is that I don't want to have to wait for container images to build every time I edit the operator code while I'm developing/testing changes.
> 
> I guess what I'm saying is that in addition to that there's another use case: building the cluster using a custom image as the 'real, full' version of metal3 that is needed for the cluster to come up, and see if the cluster does actually come up.

Now that the machine-api-operator manages the in-cluster deployment, testing a change in that deployment is a multi-step process that I wasn't trying to cover with this change. I agree it would be useful.

PR #772, opened 2019-08-27
[wip]: allow to use a preexisting openshift-install binary

karmab:
- allow to use openshift-install binary if found at root dir. this allows quickly confirming an issue is related to the installer code by relaunching an install with a known to work version


russellb:
agreed with @hardys on how to test a specific version

PR #776, opened 2019-08-28
Reinstate OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE

hardys:
This was removed in #758 but it turns out the default origin builds
still contain old ironic images, ref discussion on
https://github.com/openshift/ironic-inspector-image/pull/17

It also seems sane to ensure the release image set in dev-scripts is
the same as the installer, which currently is not the case, we're
defaulting to ocp builds and the installer to origin.

Closes: #775


hardys:
Thanks for the feedback @stbenjam - yes this is a source install, will rework so we only set the override in the build_installer function

PR #777, opened 2019-08-28
Stop recommending deprecated OS_URL

dtantsur:
It is being removed in https://review.opendev.org/#/c/677795/.
Use 'none' auth_type instead.


hardys:
+1 but would it be better to revive https://github.com/openshift-metal3/dev-scripts/pull/612 and generate a clouds.yaml instead?

Particularly since we now have ironic running in two different locations e.g bootstrap and on the cluster, having two differently named "clouds" to interact with might be nicer than documenting the override of the endpoint?

dtantsur:
This is an immediate fix to something that will break really soon. I'll get to the `clouds.yaml` idea soon, just too busy right now.

PR #785, opened 2019-08-30
add jira-sync integration scripts

dhellmann:
This patch adds the scripts used to pull data from github and bugzilla
into the team's jira instance. Placing them in a common place means
the team can modify them as needed, even if they still only run on my
development host.

After this PR is merged I will update the cron job to use the copy of
the scripts from this repo instead of my private copy.

### DIFF
--- a/jira-sync/jira_link_one.sh
+++ b/jira-sync/jira_link_one.sh
@@ -1,0 +1,22 @@
+#!/bin/bash
+
+set -e
+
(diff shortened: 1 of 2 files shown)


russellb:
I wonder if a new `jira-sync` repo would make sense?  This isn't really related to the dev env

dhellmann:
> I wonder if a new `jira-sync` repo would make sense? This isn't really related to the dev env

OK, I'll do that.

PR #796, opened 2019-09-06
Install python36-devel before doing pip install

hardys:
Some pip installs fail if we don't have the python-devel environment
installed prior to doing the install, in particular this seems to
break with the latest versions of python-libvirt on a minimal
install which lacks the devel libraries.


ahrechanychenko:
Tested locally, works for me

metal3ci:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/1127/

PR #719, opened 2019-08-01
Pull localy built images from local for masters

derekhiggins:
Add the ability for masters to pull localy built images
for development and ci jobs.

PR #801, opened 2019-09-06
WIP CI for networking components

cybertron:
This isn't working yet, but it provides a basic outline of what I think we need to be able to do in our CI runs. I'm pushing it as a WIP so I have something to discuss with the team and figure out how to move forward.

### DIFF
--- a/06_create_cluster.sh
+++ b/06_create_cluster.sh
@@ -44,7 +44,15 @@ if [ -n "$COREDNS_IMAGE" ]; then
     IMAGE_OVERRIDES="$IMAGE_OVERRIDES coredns=quay.io/cybertron/coredns:latest"
     popd
 fi
-# TODO: Build mdns-publisher and baremetal-runtimecfg images
+if [ -n "$BAREMETAL_RUNTIMECFG_IMAGE" ]; then
+    pushd "$BAREMETAL_RUNTIMECFG_IMAGE"
+    baremetal_runtimecfg_id=$(sudo buildah bud -f Dockerfile . | tail -n 1)
+    sudo podman tag $baremetal_runtimecfg_id quay.io/cybertron/baremetal-runtimecfg
+    sudo podman push quay.io/cybertron/baremetal-runtimecfg
+    IMAGE_OVERRIDES="$IMAGE_OVERRIDES baremetal-runtimecfg=quay.io/cybertron/baremetal-runtimecfg:latest"
+    popd
+fi
+# TODO: Build mdns-publisher
 
 # Build a new release, based on the configured release, that overrides the
 # appropriate images built above.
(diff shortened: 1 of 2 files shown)


06_create_cluster.sh
```diff
@@ -57,11 +60,16 @@ fi
...
 if [ -n "$IMAGE_OVERRIDES" ]; then
-    # This is consistently failing for me for some reason. It seems to be
-    # completing anyway though...
+    # This is what I would like to do, but it doesn't work
+#     oc adm release new -n ocp \
+#         --server https://api.ci.openshift.org \
+#         --from-release "$OPENSHIFT_RELEASE_IMAGE" \
+#         --to-image quay.io/cybertron/origin-release:v4.2 \
+#         $IMAGE_OVERRIDES || :
+    # This works in isolation but not when run as part of this script. :-/
```
hardys: cc @derekhiggins @stbenjam - can we use this as an alternative to https://github.com/openshift-metal3/dev-scripts/pull/759 and https://github.com/openshift-metal3/dev-scripts/pull/719 ?

It seems like a common approach here for all potentially overridden container images would be good, and using a custom release probably aligns better with integration into openshift CI?

imain: Yeah I think this would be super helpful for testing a lot of new stuff.

derekhiggins: I've closed 719, not sure how I ended up with two patch doing similar stuff...

I wonder if we should include some way of using a local registry (similar to https://github.com/openshift-metal3/dev-scripts/pull/759) to save bandwidth? I'm currently looking into integrating into the openshift build system which may eventually provide an alternative.

cybertron: Last time I checked, there wasn't really a way to use a local registry with the installer. There's no way to tell it to use --tls-verify=false when retrieving images, so it barfs on the "insecure" registry.

I would dearly love it if we could start using a local registry though. The amount of bandwidth and time that I burn from repeatedly pulling the same images is absurd. :-/

derekhiggins: > Last time I checked, there wasn't really a way to use a local registry with the installer. There's no way to tell it to use --tls-verify=false when retrieving images, so it barfs on the "insecure" registry.
> 
> I would dearly love it if we could start using a local registry though. The amount of bandwidth and time that I burn from repeatedly pulling the same images is absurd. :-/

I couldn't find a way to do it either without hacking the code, see the PR I linked to above
https://github.com/openshift-metal3/dev-scripts/pull/759/files#diff-13094614d976ac6f9d0cc2f87e41f332R36

cybertron:
This has been addressed elsewhere. No need to keep the PR open.

PR #810, opened 2019-09-18
WIP show how to deploy a 3rd master via BMO

hardys:
This illustrates the current steps needed to deploy a 3rd master
as a day-2 operation, e.g such that we could potentially deploy
in an environment with only 3 nodes by rebuilding the initial
provisioning host after the initial 2-nodes are deployed.
    
In-progress work on the etcd operator will make this easier/cleaner
in future, but for now it follows the recovery steps from the OCP 4.1
DR docs:
    
https://github.com/hexfusion/cluster-etcd-operator/pull/33
    
https://docs.openshift.com/container-platform/4.1/backup_and_restore/disaster_recovery/scenario-1-infra-recovery.htmlhttps://docs.ope
    
Note we also rename make-bm-worker to make-bm-node as this
now contains nothing specific to workers, so we can also use it
for master deployments.


hardys:
Note this is not to be merged, it's just to demonstrate the process - we might consider merging the changes to make-bm-worker later though, via a separate PR.

hardys:
This was just pushed as an example so I'll close the PR

PR #813, opened 2019-09-25
run_ci.sh copy logs for virtualbmc service

hardys:
Copy logs so we can ensure this is started and debug issues when ipmi commands fail

### DIFF
--- a/run_ci.sh
+++ b/run_ci.sh
@@ -14,6 +14,10 @@ function getlogs(){
     sudo podman logs coreos-downloader > $LOGDIR/coreos-downloader.log
     sudo podman logs ipa-downloader > $LOGDIR/ipa-downloader.log
 
+    sudo cp /var/log/virtualbmc/virtualbmc.log $LOGDIR/bootstrap-host-virtualbmc.log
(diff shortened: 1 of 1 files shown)


run_ci.sh
```diff
@@ -14,6 +14,10 @@ function getlogs(){
     sudo podman logs coreos-downloader > $LOGDIR/coreos-downloader.log
     sudo podman logs ipa-downloader > $LOGDIR/ipa-downloader.log
 
+    sudo cp /var/log/virtualbmc/virtualbmc.log $LOGDIR/bootstrap-host-virtualbmc.log
```
russellb: I think this log location just changed -- https://github.com/metal3-io/metal3-dev-env/pull/48/files

hardys: Hmm that may explain why vbmc is broken in our CI then ;)

hardys: I guess we need to start the containers as in 02_configure_host.sh from metal3-dev-env/pull/48 and I don't see any corresponding dev-scripts PR

I'll test locally and push one.

metal3ci:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/1169/

PR #815, opened 2019-09-25
Switch to containerized vbmc/sushy-tools

hardys:
This aligns with https://github.com/metal3-io/metal3-dev-env/pull/48
which landed recently and is required to fix metal3 CI.

PR #817, opened 2019-10-01
Only start vbmc containers for default NODES_PLATFORM

hardys:
This won't work and isn't needed in the case where
NODES_PLATFORM is set to e.g "baremetal", in which case
we don't configure the vbmc host directories and thus
the containers fail to start.

This was missed in #815 which switched over to the containerized
vbmc services.


acalhounRH:
Tested, and worked on bare-metal deployment.

metal3ci:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/1188/

PR #818, opened 2019-10-02
WIP: Add new items for configuring the BMO to the baremetal platform.

imain:
Add configuration information for the baremetal pod to the baremetal
platform config.  This can then be picked up in the installer and used
to set the platformStatus.

### DIFF
--- a/06_create_cluster.sh
+++ b/06_create_cluster.sh
@@ -5,8 +5,8 @@ set -e
 source logging.sh
 source utils.sh
 source common.sh
-source ocp_install_env.sh
 source rhcos.sh
+source ocp_install_env.sh
 
 # Do some PULL_SECRET sanity checking
 if [[ "${OPENSHIFT_RELEASE_IMAGE}" == *"registry.svc.ci.openshift.org"* ]]; then
--- a/ocp_install_env.sh
+++ b/ocp_install_env.sh
@@ -95,6 +95,11 @@ controlPlane:
 platform:
   baremetal:
     dnsVIP: ${DNS_VIP}
+    provisioningInterface: "ens3"
+    provisioningNetworkCIDR: "172.22.0.0/24"
+    provisioningIP: "172.22.0.3/24"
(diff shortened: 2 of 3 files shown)


imain:
You'll also need this:  https://github.com/openshift/installer/pull/2449

ocp_install_env.sh
```diff
@@ -95,6 +95,11 @@ controlPlane:
 platform:
   baremetal:
     dnsVIP: ${DNS_VIP}
+    provisioningInterface: "ens3"
+    provisioningNetworkCIDR: "172.22.0.0/24"
+    provisioningIP: "172.22.0.3/24"
+    provisioningDHCPRange: "172.22.0.10,172.22.0.100"
+    rhcosImageURL: "${RHCOS_IMAGE_URL}"
```
hardys: As mentioned on the installer PR we shouldn't add rhcosImageURL because it's data that's already known to the installer - we only care about RHCOS_IMAGE_URL in dev-scripts because we cache the image

ocp_install_env.sh
```diff
@@ -95,6 +95,11 @@ controlPlane:
 platform:
   baremetal:
     dnsVIP: ${DNS_VIP}
+    provisioningInterface: "ens3"
+    provisioningNetworkCIDR: "172.22.0.0/24"
+    provisioningIP: "172.22.0.3/24"
```
hardys: All of these should have defaults in the installer PR, e.g see `pkg/types/baremetal/defaults/platform.go`

06_create_cluster.sh
```diff
@@ -5,8 +5,8 @@ set -e
 source logging.sh
 source utils.sh
 source common.sh
-source ocp_install_env.sh
```
sadasu: The changes in this file seem to be a no-op.

imain:
This isn't needed any more.  The new version has sane defaults so you don't really need to configure it.

PR #819, opened 2019-10-03
Add option to use upstream images for ironic

derekhiggins:
Use upstream container images for the containers that
make up the ironic/bmo pod.

### DIFF
--- a/common.sh
+++ b/common.sh
@@ -39,6 +39,16 @@ source $CONFIG
 export OPENSHIFT_RELEASE_IMAGE="${OPENSHIFT_RELEASE_IMAGE:-registry.svc.ci.openshift.org/ocp/release:4.2}"
 export OPENSHIFT_INSTALL_PATH="$GOPATH/src/github.com/openshift/installer"
 
+# Switch Container Images to upstream, Installer defaults these to the openshift version
+if [ "${UPSTREAM_IRONIC:-false}" != "false" ] ; then
+    export IRONIC_LOCAL_IMAGE=${IRONIC_LOCAL_IMAGE:-"quay.io/metal3-io/ironic:master"}
+    export IRONIC_INSPECTOR_LOCAL_IMAGE=${IRONIC_INSPECTOR_LOCAL_IMAGE:-"quay.io/metal3-io/ironic-inspector:master"}
+    export IRONIC_IPA_DOWNLOADER_LOCAL_IMAGE=${IRONIC_IPA_DOWNLOADER_LOCAL_IMAGE:-"quay.io/metal3-io/ironic-ipa-downloader:master"}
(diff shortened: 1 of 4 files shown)


hardys:
Looks good, but I wonder if this should be optional, at least until we've got fully integrated into openshift CI?  

I was thinking in the case where metal3 CI votes on e.g openshift/installer we'll want to use the openshift images, but in the case where it's testing the metal3-io repos (and for much local development probably) we'll want to use these ones?

derekhiggins:
> Looks good, but I wonder if this should be optional, at least until we've got fully integrated into openshift CI?
> 
> I was thinking in the case where metal3 CI votes on e.g openshift/installer we'll want to use the openshift images, but in the case where it's testing the metal3-io repos (and for much local development probably) we'll want to use these ones?

Done, default behaviour remains the same but set "UPSTREAM_IRONIC=True" to switch to upstream images

config_example.sh
```diff
@@ -6,9 +6,10 @@ set +x
 export PULL_SECRET=''
 set -x
 
-# Uncomment to build a copy of ironic or inspector locally
-#export IRONIC_INSPECTOR_IMAGE=https://github.com/metal3-io/ironic-inspector
-#export IRONIC_IMAGE=https://github.com/metal3-io/ironic
+# Use <NAME>_LOCAL_IMAGE to build or use copy of container images locally e.g.
+#export IRONIC_INSPECTOR_LOCAL_IMAGE=https://github.com/metal3-io/ironic-inspector
+#export IRONIC_LOCAL_IMAGE=quay.io/username/ironic
+#export MACHINE_CONFIG_OPERATOR_LOCAL_IMAGE=https://github.com/openshift/machine-config-operator
```
dtantsur: Maybe add here:
```bash
# Switch to upstream metal3-io ironic images instead of openshift ones.
#export UPSTREAM_IRONIC=true
```

common.sh
```diff
@@ -39,6 +39,16 @@ source $CONFIG
 export OPENSHIFT_RELEASE_IMAGE="${OPENSHIFT_RELEASE_IMAGE:-registry.svc.ci.openshift.org/ocp/release:4.2}"
 export OPENSHIFT_INSTALL_PATH="$GOPATH/src/github.com/openshift/installer"
 
+# Switch Container Images to upstream, Installer defaults these to the openshift version
+if [ "$UPSTREAM_IRONIC" == "True" ] ; then
```
hardys: This variable is undefined by default, perhaps we should define it above with a default of "false", then check here for != "false" so we don't have to worry about case-sensitive "True"?

metal3ci:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/1200/

metal3ci:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/1203/

PR #835, opened 2019-10-25
Disable copy on write on the WORKING_DIR

derekhiggins:
For space reasons, We've moved WORKING_DIR onto the
BTRFS filesystem, this is chewing up diskspace so we
need to disable COW on this directory.


metal3ci:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/1269/

PR #844, opened 2019-11-12
Assert 'go' command is available before calling `go env`.

yprokule:
If `go` command is not available in default PATH, `go env` call fails,

``` bash
    +++ go env
    common.sh: line 3: go: command not found
    ++ eval ''
```

This change adds extra check to assert `go` binary is available in
default PATH, and if not prepend it with '/usr/local/go/bin'

### DIFF
--- a/common.sh
+++ b/common.sh
@@ -1,11 +1,13 @@
 #!/bin/bash
 
+export PATH="/usr/local/go/bin:$PATH"
+
 eval "$(go env)"
+
+export PATH="${GOPATH}/bin:$PATH"
 
 # Make dev-scripts work again temporarily
 export KNI_INSTALL_FROM_GIT=true
-
-export PATH="${GOPATH}/bin:/usr/local/go/bin:$PATH"
 
 # Ensure if a go program crashes we get a coredump
 #


yprokule:
/cc @hardys @stbenjam 

That's  a cosmetic change rather then fix for any issue/problem.

yprokule:
```
grep -R 'go: command not found' logs/
logs/01_install_requirements-2019-11-11-081902.log:common.sh: line 3: go: command not found
logs/01_install_requirements-2019-11-11-081902.log:lib/common.sh: line 5: go: command not found
logs/02_configure_host-2019-11-11-082554.log:common.sh: line 3: go: command not found
logs/03_build_installer-2019-11-11-082621.log:common.sh: line 3: go: command not found
logs/04_setup_ironic-2019-11-11-082648.log:common.sh: line 3: go: command not found
logs/06_create_cluster-2019-11-11-083120.log:common.sh: line 3: go: command not found
```

common.sh
```diff
@@ -1,8 +1,12 @@
 #!/bin/bash
 
+if [[ ! $(command -pv go) ]]; then
+    export PATH="/usr/local/go/bin:$PATH"
+fi
+
 eval "$(go env)"
 
-export PATH="${GOPATH}/bin:/usr/local/go/bin:$PATH"
+export PATH="${GOPATH}/bin:$PATH"
```
hardys: Do we need the conditional here, probably we should just always export PATH="/usr/local/go/bin:$PATH" before running go env?

Note that this will still fail on a clean install the very first time the 01 script gets run...

yprokule: At least it should not fail later when the *go* is installed

PR #846, opened 2019-11-18
Add web UI guidelines to README

jtomasek:
Add web UI guidelines to 'Interacting with the deployed cluster' section

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,6 +131,30 @@ master-1   Ready     master    20m       v1.12.4+50c2f2340a
 master-2   Ready     master    20m       v1.12.4+50c2f2340a
 ```
 
+Alternatively it is possible to manage the cluster using OpenShift Console web UI.
+The URL can be retrieved using
+```
+oc get routes --all-namespaces | grep console
+```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -131,6 +131,30 @@ master-1   Ready     master    20m       v1.12.4+50c2f2340a
 master-2   Ready     master    20m       v1.12.4+50c2f2340a
 ```
 
+Alternatively it is possible to manage the cluster using OpenShift Console web UI.
+The URL can be retrieved using
+```
+oc get routes --all-namespaces | grep console
+```
+By default, the URL is https://console-openshift-console.apps.ostest.test.metalkube.org
```
yprokule: nit: if it's highlighted - it's more readable, imho
```suggestion
By default, the URL is `https://console-openshift-console.apps.ostest.test.metalkube.org`
```

hardys: The URL gets highlighted as a link already so it's quite readable - see https://github.com/jtomasek/dev-scripts/tree/patch-1

PR #853, opened 2019-11-19
Make local registry address configurable

honza:
Or, in other words, let's not hardcode it anymore

### DIFF
--- a/04_setup_ironic.sh
+++ b/04_setup_ironic.sh
@@ -33,15 +33,15 @@ for IMAGE_VAR in $(env | grep "_LOCAL_IMAGE=" | grep -o "^[^=]*") ; do
         [ -e "$REPOPATH" ] || git clone $IMAGE $REPOPATH
         cd $REPOPATH
         export $IMAGE_VAR=${IMAGE##*/}:latest
-        export $IMAGE_VAR=192.168.111.1:5000/localimages/${!IMAGE_VAR}
+        export $IMAGE_VAR=$LOCAL_REGISTRY_ADDRESS/localimages/${!IMAGE_VAR}
         sudo podman build -t ${!IMAGE_VAR} .
         cd -
         sudo podman push --tls-verify=false ${!IMAGE_VAR} ${!IMAGE_VAR}
     fi
 
-    # Update the bootstrap and master nodes to treat 192.168.111.1:5000 as insecure
+    # Update the bootstrap and master nodes to treat LOCAL_REGISTRY_ADDRESS as insecure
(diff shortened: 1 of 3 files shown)


common.sh
```diff
@@ -77,8 +77,9 @@ if [ -z "$KNI_INSTALL_FROM_GIT" ]; then
 fi
 
 if env | grep -q "_LOCAL_IMAGE=" ; then
+    export LOCAL_REGISTRY_ADDRESS = ${LOCAL_REGISTRY_ADDRESS:-"192.168.111.1:5000"}
```
yprokule: Extra spaces
```suggestion
    export LOCAL_REGISTRY_ADDRESS=${LOCAL_REGISTRY_ADDRESS:-"192.168.111.1:5000"}
```

yprokule: IMHO `LOCAL_REGISTRY_ADDRESS` shouldn't be exported within condition. 
Being exported in outer block would allow images' mirroring(pr #850) not rely on presence of `*_LOCAL_IMAGE` variables, wdyt?

04_setup_ironic.sh
```diff
@@ -12,7 +12,7 @@ source ocp_install_env.sh
 # with "-" converted to "_" e.g. to use a custom ironic-inspector
 #export IRONIC_INSPECTOR_LOCAL_IMAGE=https://github.com/metal3-io/ironic-inspector-image
 #export IRONIC_RHCOS_DOWNLOADER_LOCAL_IMAGE=https://github.com/openshift-metal3/rhcos-downloader
-#export BAREMETAL_OPERATOR_LOCAL_IMAGE=192.168.111.1:5000/localimages/bmo:latest
+#export BAREMETAL_OPERATOR_LOCAL_IMAGE=$LOCAL_REGISTRY_ADDRESS/localimages/bmo:latest
```
hardys: This is just an example comment showing how to set this in config_$USER.sh, where common.sh isn't necessarily souurced, so maybe we should leave this one as-is?

common.sh
hardys: Yeah agreed, and I think the changes in 04_setup_ironic.sh will break without this unless *_LOCAL_IMAGE is specified

honza: Done

04_setup_ironic.sh
honza: Done

common.sh
honza: Done

metal3ci:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/1317/

Issue #600, opened 2019-06-05
Re-enable selinux on provisioning host


russellb:
01_install_requirements.sh disables selinux.  We need to get to where this is not required on the provisioning host.  This bug tracks removing this from dev-scripts.

ch-stark:
did a test today without selinux disabled 
commented out:
#if selinuxenabled ; then
    # FIXME ocp-doit required this so leave permissive for now
    #sudo setenforce permissive
    #sudo sed -i "s/=enforcing/=permissive/g" /etc/selinux/config
#fi

and did't find issues

russellb:
selinux is no longer disabled in dev-scripts

PR #858, opened 2019-12-03
remove duplicate selinuxenabled logic

ch-stark:
we will  rely on the settings in metal3-dev-env and maybe introduce some additional flag later

https://github.com/metal3-io/metal3-dev-env/blob/master/centos_install_requirements.sh#L9..L13

ref:
https://github.com/openshift-metal3/dev-scripts/issues/600

currently the function which gets disabled via this MR does not have effect

### DIFF
--- a/01_install_requirements.sh
+++ b/01_install_requirements.sh
@@ -9,14 +9,6 @@ if grep -q "Red Hat Enterprise Linux release 8" /etc/redhat-release 2>/dev/null 
     RHEL8="True"
 fi
 
-sudo yum install -y libselinux-utils
-if selinuxenabled ; then
-    # FIXME ocp-doit required this so leave permissive for now
-    sudo setenforce permissive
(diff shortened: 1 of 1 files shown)


01_install_requirements.sh
```diff
@@ -9,14 +9,6 @@ if grep -q "Red Hat Enterprise Linux release 8" /etc/redhat-release 2>/dev/null
...
 fi
 
-sudo yum install -y libselinux-utils
-if selinuxenabled ; then
-    # FIXME ocp-doit required this so leave permissive for now
-    sudo setenforce permissive
-    sudo sed -i "s/=enforcing/=permissive/g" /etc/selinux/config
-fi
-
-
```
hardys: For other reviewers, this is handled via https://github.com/metal3-io/metal3-dev-env/blob/master/centos_install_requirements.sh#L9..L13 which is run below L26

metal3ci:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/1343/

PR #861, opened 2019-12-11
WIP prototype running two clusters with shared provisioning network

hardys:
Until the installer supports configuring the provisioning subnet, we
may be able to deploy two clusters on the same virthost with only
different external networks, provided the metal3 pod on the first
cluster is disabled after any workers are deployed.

This is pretty hacky atm, but aims to prove if this will be possible

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 ocp/
+ocp2/
 *.qcow2
 openshift-install.log
 logs/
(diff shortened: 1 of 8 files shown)


hardys:
FYI @dhellmann @stbenjam this isn't yet working as something is broken with the vbmc for the additional nodes

hardys:
@dhellmann @stbenjam FYI this now works to to the point where the master nodes for the second cluster are deployed, not yet sure if it will complete bootstrapping though.

It depends on metal3-io/metal3-dev-env#151 and I've been testing with `export METAL3_DEV_ENV="/home/shardy/metal3-dev-env"` to include that.

The process to test is run `make` as normal, then run `13_optional_configure_host_cluster2.sh` which creates the additional VMs and baremetal2 bridge/network, then run `14_optional_create_cluster2.sh` which deploys the same as the normal 06 script, but using a different config in the `ocp2` dir

hardys:
This has been superseded by thte work @dhellmann is doing to enable multi-cluster with isolated networks

PR #865, opened 2019-12-17
Do not curl for release image if we provide it

yrobla:
When running on a disconnected install, the host does not
have access to curl for the image version. But we can provide
it on config file. So avoid looking for that release if
we already are providing it.


hardys:
lgtm, thanks!

metal3ci:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/1375/

PR #878, opened 2020-01-07
Add conditions for Centos 8

yrobla:
Centos 8 behaves in a similar way as RHEL 8. So we
need to add the check for Centos 8 and include it on
the needed conditionals.


01_install_requirements.sh
```diff
@@ -9,6 +9,10 @@ if grep -q "Red Hat Enterprise Linux release 8" /etc/redhat-release 2>/dev/null
     RHEL8="True"
 fi
 
+if grep -q "CentOS Linux release 8" /etc/redhat-release 2>/dev/null; then
+    CENTOS8="True"
+fi
+
```
hardys: This doesn't appear to be used anywhere?

common.sh
```diff
@@ -171,6 +171,10 @@ if grep -q "Red Hat Enterprise Linux release 8" /etc/redhat-release 2>/dev/null
     export RHEL8="True"
 fi
 
+if grep -q "CentOS Linux release 8" /etc/redhat-release 2>/dev/null; then
+    CENTOS8="True"
```
dtantsur: 1. Do you need `export` here? RHEL8 uses it.

2. Cannot we reuse the same RHEL8 variable? Do you expect any differences?

hardys: yes this should be export

It could use the same variable but it's probably OK to add a new one - currently the only differences are the repos enabled in metal3-dev-env (which figures out the OS version without this variable), but I suppose we might find something which works different on centos.

The goal is to remove all these conditionals though, when we remove support for centos7, at that point we'd only special case things which really were different between rhel and centos (ideally nothing).

metal3ci:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/1405/

metal3ci:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/1406/

hardys:
lgtm but we need to land https://github.com/metal3-io/metal3-dev-env/pull/157 first - I'll try to re-test that soon on RHEL8 so we can remove the WIP and get it merged

hardys:
@yrobla could you rebase this when you get a moment please?  Hopefully https://github.com/metal3-io/metal3-dev-env/pull/223 will land soon, then we can add this for centos8 dev-scripts support

metal3ci:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/1522/

metal3ci:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/1541/

hardys:
This was merged via https://github.com/openshift-metal3/dev-scripts/pull/957

Issue #880, opened 2020-01-07
Building the installer locally doesn't work with MIRROR_IMAGES


cybertron:
**Describe the bug**
When MIRROR_IMAGES is in use and KNI_INSTALL_FROM_GIT=true is set, the installer ends up referencing a release that it can't access. This is because https://github.com/openshift-metal3/dev-scripts/blob/1536fdeaeac964ecc27e876ec3e509e09563b93d/common.sh#L82 explicitly sets the release image to the upstream release, but with mirroring we remove the upstream pull secret and only provide the one for the local registry.

**To Reproduce**
Set MIRROR_IMAGES=true and KNI_INSTALL_FROM_GIT=true before running dev-scripts. Do not include any LOCAL_IMAGE overrides as that further changes the behavior.

**Expected/observed behavior**
I expect the custom installer to use the local registry with the mirrored images. What is actually happening is that the installer tries to pull an upstream release image that it doesn't have credentials for.

cybertron:
#881 unblocked me because with that fix and a LOCAL_IMAGE override my deployment succeeds. However, I think that may be masking the problem. With LOCAL_IMAGE set it's ignoring the local mirror and deploying from the ci registry anyway. Possibly the LOCAL_IMAGE logic is overwriting the mirrored release that gets created? Something is still definitely wrong though.

hardys:
@cybertron thanks for the report, I can confirm this is broken, with both MIRROR_IMAGES and a LOCAL_IMAGE set I'm seeing the bootstrap VM use the upstream registry if I also set KNI_INSTALL_FROM_GIT.

PR #881, opened 2020-01-07
Always use local release with MIRROR_IMAGES

cybertron:
Previously [0] would set the release override to the ci image instead
of the local one that we need to use when mirroring. This change moves
the logic for setting it into a common location where the image
override will be set correctly regardless of what combination of env
vars is in use.

0: https://github.com/openshift-metal3/dev-scripts/blob/1536fdeaeac964ecc27e876ec3e509e09563b93d/common.sh#L82

### DIFF
--- a/common.sh
+++ b/common.sh
@@ -83,9 +83,12 @@ if [ -z "$KNI_INSTALL_FROM_GIT" ]; then
 fi
 
 if env | grep -q "_LOCAL_IMAGE=" ; then
+    export MIRROR_IMAGES=true
+fi
+
+if [ -n "$MIRROR_IMAGES" ]; then
     # We're going to be using a locally modified release image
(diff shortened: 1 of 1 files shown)


cybertron:
This is a partial fix for #880. My local installer tests are still failing because my local release image is referencing ci images that the installer doesn't have access to. I haven't figured out if that's because I messed something up in my local environment or if more changes are needed to get this working. I'm pretty certain this change is correct though so I figured I'd push it up for discussion.

hardys:
lgtm but added derek for review

derekhiggins:
/lgtm
thanks

metal3ci:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/1407/

PR #882, opened 2020-01-09
Define default for KNI_INSTALL_FROM_GIT

hardys:
This is undefined as exposed by setting
set -euxo pipefail in the 04 script.

Setting that is also desirable because
currently we ignore any failure in the
oc adm release mirror because the results
are piped to tee.


metal3ci:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/1408/

PR #883, opened 2020-01-09
Only extract installer when not building from source

hardys:
We don't need to extract the installer if we know we'll build it from source.


metal3ci:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/1409/

PR #884, opened 2020-01-09
handle unset variables

dhellmann:
Move the initialization of WORKING_DIR earlier in common.sh so it
comes before the first use to avoid errors due to the variable not
being defined.

### DIFF
--- a/common.sh
+++ b/common.sh
@@ -39,6 +39,8 @@ source $CONFIG
 # mirror images for installation in restricted network
 export MIRROR_IMAGES=${MIRROR_IMAGES:-}
 
+WORKING_DIR=${WORKING_DIR:-"/opt/dev-scripts"}
+
 # variables for local registry configuration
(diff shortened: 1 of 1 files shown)


common.sh
```diff
@@ -56,9 +56,18 @@ export KNI_INSTALL_FROM_GIT=${KNI_INSTALL_FROM_GIT:-}
 # See https://openshift-release.svc.ci.openshift.org for release details
 #
 # if we provide OPENSHIFT_RELEASE_IMAGE, do not curl. This is needed for offline installs
+if echo $SHELLOPT | grep -q nounset; then
+    _reset_nounset=true
+else
+    _reset_nounset=false
+fi
+set +u
 if [ -z "${OPENSHIFT_RELEASE_IMAGE}" ]; then
```
zaneb: Wouldn't it be simpler to change this to:

?

dhellmann: Where's the fun in that?

I'll wait for my cluster to build and push the change.

PR #887, opened 2020-01-14
Always build a local release image if mirroring

derekhiggins:
Both MIRROR_IMAGES=true and *_LOCAL_IMAGE rely on this
image being build so we should build it in both cases.
We had only been building it if *_LOCAL_IMAGE was set.

Fixes: #880


metal3ci:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/1420/

PR #892, opened 2020-01-20
Fix locale issue when checking for XFS compatibility

andfasano:
Filesystem type check failed when locale was not English. Thanks @hardys for the suggested solution.

### DIFF
--- a/common.sh
+++ b/common.sh
@@ -180,14 +180,14 @@ if grep -q "Red Hat Enterprise Linux release 8" /etc/redhat-release 2>/dev/null 
 fi
 
 # Check d_type support
-FSTYPE=$(df ${FILESYSTEM} --output=fstype | grep -v Type)
+FSTYPE=$(df "${FILESYSTEM}" --output=fstype | tail -n 1)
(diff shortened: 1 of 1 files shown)


hardys:
lgtm - one question, did the same problem in metal3-dev-env (ref https://github.com/metal3-io/metal3-dev-env/pull/188) cause the install to fail in exactly the same way?

I'm guessing so, failing at https://github.com/openshift-metal3/dev-scripts/blob/master/01_install_requirements.sh#L18 - so I wonder if we should remove the duplicate checks from dev-scripts.

Not blocking on that for this PR, just a possible follow-up, initially metal3-dev-env was a copy of some code from dev-scripts, but since then we've changed things to call metal3-dev-env from dev-scripts, and it seems there's still some duplication we can potentially remove.

andfasano:
Yes, both fixes are currently needed until the duplication will be removed

metal3ci:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/1433/

PR #893, opened 2020-01-20
ocp_cleanup: Remove any stale .ign volumes

hardys:
These can get left as the --remove-all-storage doesn't remove them
(they're passed via -fw-cfg not strictly owned by the domain).


metal3ci:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/1434/

PR #898, opened 2020-01-23
Fix some issues calculated the openshift version

hardys:
Modify the way we calculate the actual OC version - the previous string manipulation doesn't work so we ended up downloading the client every time.  Also make a similar change to improve the regex used to extract OPENSHIFT_VERSION from the release pullspec.


metal3ci:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/1440/

russellb:
CI failure is unrelated to this change.  It downloaded `oc` fine

Issue #895, opened 2020-01-22
Add validation of WORKING_DIR


hardys:
WORKING_DIR defaults to /opt/dev-scripts, which isn't ideal because if you install e.g RHEL8 with the default partition scheme, you only get 50G on the root LV.

A common solution to this is to override WORKING_DIR, but you must not put the directory inside an existing user home directory (because it's not world readable), if you do then the vbmc power control will fail with an error like:

```
Stderr: 'Running Get PICMG Properties my_addr 0x20, transit 0, target 0x20\nError response 0xc1 from Get PICMG Properities\nRunning Get VSO Capabilities my_addr 0x20, transit 0, target 0x20\nInvalid completion code received: Invalid command\nDiscovered IPMB address 0x0\nSet Chassis Power Control to Up/On failed: Node busy\n': oslo_concurrency.processutils.ProcessExecutionError: Unexpected error while running command.
```

We should add validation to check that $WORKING_DIR is world readable and/or isn't in any existing user homedir.

Probably some validation of the space available would also be good, either by figuring out a sensible limit or perhaps better by checking for free space in common.sh and bailing out if the volume backing $WORKING_DIR is 90% full or something?

andfasano:
Validations check about the fact that the $WORKING_DIR is world readable and that enough space is available looks reasonable to me and very likely do not add any breaking change and should be sufficient. I would avoid checking if it's in the homedir.

hardys:
Closed via #900

PR #900, opened 2020-01-24
Add sanity checks to verify the working dir

andfasano:
This PR addresses issue #895.

A couple of sanity checks have been added to verify that:
1) The WORKING_DIR is world readable
2) Enough free space is available to successfully setup dev-scripts. Since this value can be difficult to estimate precisely, it can be configured via the MIN_SPACE_REQUIRED setting, by default set to 80Gb

To minimize the impact on the existing code, these checks have been placed in a separate script invoked just after commons.sh

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,7 +158,7 @@ Run sshuttle:
 sshuttle -r <user>@<virthost> 192.168.111.0/24
 ```
 
-
+To access the web Console use the `kubeadmin` user, and password generated in the `ocp/auth/kubeadmin-password` file.
 
 
 ## Interacting with Ironic directly
--- a/common.sh
+++ b/common.sh
@@ -65,7 +65,11 @@ export MIRROR_IP=${MIRROR_IP:-$PROVISIONING_HOST_IP}
 # mirror images for installation in restricted network
 export MIRROR_IMAGES=${MIRROR_IMAGES:-}
 
+# The dev-scripts working directory
 WORKING_DIR=${WORKING_DIR:-"/opt/dev-scripts"}
+
+# The minimum amount of space required for a default installation, expressed in GB
+MIN_SPACE_REQUIRED=${MIN_SPACE_REQUIRED:=80}
 
 # variables for local registry configuration
 export LOCAL_REGISTRY_PORT=${LOCAL_REGISTRY_PORT:-"5000"}
--- a/sanitychecks.sh
+++ b/sanitychecks.sh
@@ -1,0 +1,32 @@
+#!/bin/bash
+
+function verifyWorkingDirIsWorldReadable {
+  if [ ! -d $WORKING_DIR ]; then
+    echo "WORKING_DIR ${WORKING_DIR} is not a directory"
+    exit 1
+  fi
+
+  if [ ! -r $WORKING_DIR ]; then
+    echo "Unable to access WORKING_DIR ${WORKING_DIR}"
+    exit 1
+  fi
+
+  if find "${WORKING_DIR}" -maxdepth 0 ! -perm -o=r | grep . ; then
+    echo "The WORKING_DIR ${WORKING_DIR} is not world-readable!"
+    exit 1 
+  fi
+}
+
+function verifyFreeSpace {
+  AVAIL=$(df -h --output=avail -B 1G $WORKING_DIR | tail -n 1)
+
+  if (( $AVAIL < $MIN_SPACE_REQUIRED )); then 
+    echo "Not enough free space, at least MIN_SPACE_REQUIRED=${MIN_SPACE_REQUIRED} GB required"
+    exit 1
+  fi
+}
+
+verifyFreeSpace
+verifyWorkingDirIsWorldReadable
+
+echo "Sanity checks passed"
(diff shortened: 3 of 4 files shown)


common.sh
```diff
@@ -39,8 +39,12 @@ source $CONFIG
 # mirror images for installation in restricted network
 export MIRROR_IMAGES=${MIRROR_IMAGES:-}
 
+# The dev-scripts working directory
 WORKING_DIR=${WORKING_DIR:-"/opt/dev-scripts"}
 
+# The minimum amount of space required for a default installation, expressed in GB
+MIN_SPACE_REQUIRED=${MIN_SPACE_REQUIRED:=80}
```
hardys: If this variable is only needed in sanitychecks.sh, perhaps just define it in there?

We have a large number of variables in common.sh already, so it would be good to keep only those variables which are really common defined here I think.

andfasano: Initially I defined in the `sanitychecks.sh` file, but then I noticed that all the settings were centralized in `commons.sh` and did not want to increase the level of configuration scattering, but agree that very likely it's better to keep only stuff that is in common.
Then remains open the problem about how to document it, since in the README file there's only  a generic mention to the settings contained in the `commons.sh`. Maybe we can a small additional configuration paragraph in the README for that.

sanitychecks.sh
```diff
@@ -0,0 +1,46 @@
...
+    echo "Unable to access WORKING_DIR"
+    exit 1
+  fi
+
+  ATTRIBS=$(stat -L -c "0%a" "$WORKING_DIR")
+
+  if (( ($ATTRIBS & 05) != 0 )); then
+
+   WD_PATH=$(dirname $WORKING_DIR)
+   while [ "$WD_PATH" != "/" ] ; do
```
hardys: I think this function can be simplified, how about something like:

```
if find "${WORKING_DIR}" -maxdepth 1 ! -perm -o=r | grep -q . ; then
    echo "The WORKING_DIR ${WORKING_DIR}" is not world-readable!"
    exit 1
fi
```

common.sh
hardys: Personally I'd move the variable to `sanitychecks.sh` and mention the variable name in the error message, then it becomes self-documenting if anyone hits the validation error

sanitychecks.sh
andfasano: Seems a good idea, I tested it but I had some troubles with the `maxdepth` option, since it fails whenever the target dir already contains (for any reason) a sub directory not readable by others. Instead with `-maxdepth 0` works fine also in such cases.

README.md
```diff
@@ -125,6 +125,14 @@ Then you can interact with the k8s API on the bootstrap VM e.g
 You can also see the status of the bootkube.sh script which is running via
 `journalctl -b -f -u bootkube.service`.
 
+## Login credentials
+
+Once succesfully deployed, you can access the OpenShift instance by using the following credentials:
+
+ Username | Password
+ -------- | ---------
+ kubeadmin  | See the content of the secret in `ocp/auth/kubeadmin-password`
```
hardys: we shouldn't document interacting with the username/password IMO or there's a risk they end up in the shell history, see https://github.com/openshift-metal3/dev-scripts/blob/master/README.md#interacting-with-the-deployed-cluster where we already document using the --config option to point to the kubeconfig file instead.

Note you can make this persistent in your environment by adding e.g `export KUBECONFIG=$HOME/dev-scripts/ocp/auth/kubeconfig` to your bashrc.

andfasano: Ok for the shell, but the idea here was to document how to access via the web interface, as I did not find any reference in the existing dev-scripts documentation, thus I thought it could have been a useful piece of information especially for first-users. I think at least a reference to something could be useful (ie https://docs.openshift.com/container-platform/4.1/web-console/web-console.html)

sanitychecks.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
andfasano: What would be the benefit in this case?

README.md
hardys: Ah good point, perhaps we should have a section explaining how to access the web console, as there's currently some info about that in https://github.com/openshift-metal3/dev-scripts#interacting-with-the-deployed-cluster ?

andfasano: We could add a  "Accessing the web console" section next to the mentioned section and move the instructions there

metal3ci:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/1470/

metal3ci:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/1475/

metal3ci:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/1476/

sanitychecks.sh
andfasano: Hi @hardys, I've retested the approach and found a case where it seems to not work properly when WORKING_DIR is placed within the _current user_ home directory (ie, user `andrea` and `WORKING_DIR=/home/andrea/wd") and `wd` is flagged as world-readable - while `/home/andrea' is not. Not sure how to cover it case with `find`, previous approach was fine also in such case.

metal3ci:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/1495/

PR #914, opened 2020-01-29
Add podman_cleanup step

honza:
This stops and removes all containers, and removes all images.

### DIFF
--- a/podman_cleanup.sh
+++ b/podman_cleanup.sh
@@ -1,0 +1,6 @@
+#!/usr/bin/env bash
+set -x
+
+sudo podman stop --all
+sudo podman rm --all --force
+sudo podman image rm --all --force
--- a/Makefile
+++ b/Makefile
@@ -31,5 +31,8 @@ ironic_cleanup:
 host_cleanup:
 	./host_cleanup.sh
 
+podman_cleanup:
+	./podman_cleanup.sh
+
 bell:
 	@echo "Done!" $$'\a'


podman_cleanup.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+set -x
+
+sudo podman stop --all
+sudo podman rm --all --force
+sudo podman image rm --all --force
```
andfasano: Removing everything on the host where dev-scripts have been executed could be pretty strong in case other containers/images have been used

podman_cleanup.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+set -x
+
+sudo podman stop --all
+sudo podman rm --all --force
```
andfasano: I would also `-v` to avoid dangling volumes, if any

podman_cleanup.sh
dhellmann: Indeed. I do sometimes run other things on my dev host. The cleanup logic already looks for things it knows were started by dev-scripts. Is it missing the names of any pods?

Maybe we want to add the script, but not included it in "make clean" so that it's available but not required?

honza: > Maybe we want to add the script, but not included it in "make clean" so that it's available but not required?

Agreed

hardys: We already remove the ironic pod, what is this removing that the current cleanup didn't remove?

Also I don't think we should remove any images by default - otherwise we'll end up re-pulling the images every single test run.

I do occasionally have to prune old images in my test environment, but IMHO that's preferable to wasting time and bandwidth downloading the same container over and over?

hardys: If we want to have a podman_cleanup target, I think it would be safer to do `sudo podman image prune --all` in stead of unconditionally removing all-the-images (which as previously noted could easily remove things unrelated to the dev-scripts setup).

hardys:
@honza can you provide some more context about the problem this aims to solve?  At first glance this seems like a very brute force approach, and it's probably unsuitable for most developer workflows - I guess is related to CI usage?

honza:
I'm fine to abandon this if you don't think it's useful.

When I was working on the mirror releases, and found myself doing something similar often as the existing cleaning mechanism didn't remove enough.

hardys:
> When I was working on the mirror releases, and found myself doing something similar often as the existing cleaning mechanism didn't remove enough.

Ok so the requirement is probably to purge all unused local images, remove the registry container and erase any state e.g $WORKING_DIR/registry I think?

I'm wondering if perhaps something like an images_clean or registry_clean would be helpful, as we do sometimes want to remove all locally cached images or reconfigure the registry, but not necessarily by removing absolutely everything that happens to be running via podman?

hardys:
This was superseded by https://github.com/openshift-metal3/dev-scripts/pull/984 which includes a target to prune images

PR #924, opened 2020-02-06
Version local registry cert files

cybertron:
We only want to generate these cert files when necessary, either on
a new deployment of the registry or when the cert files change.
The previous check for existence meant the latter case wasn't handled
correctly. Even if we fixed that somehow, the md5 check later wasn't
effective either because the directory is bind mounted into the
container so we're essentially checking that the md5 of the file
matches itself.

To address these issues, I've added a version number to the cert and
key files for the registry. If/when we make further changes to the
cert configuration for the registry we just need to bump that version
to force an update of the registry. The registry restart is also
explicitly requested any time the certs are re-generated so we don't
have to try to figure out which cert is in use.

It's possible we could try to inspect the existing cert for differences
with the desired cert, but that would be more complicated and I'm not
sure it would be less error-prone than this approach.

### DIFF
--- a/common.sh
+++ b/common.sh
@@ -73,6 +73,7 @@ export REGISTRY_USER=${REGISTRY_USER:-ocp-user}
 export REGISTRY_PASS=${REGISTRY_PASS:-ocp-pass}
 export REGISTRY_DIR=${REGISTRY_DIR:-$WORKING_DIR/registry}
 export REGISTRY_CREDS=${REGISTRY_CREDS:-$HOME/private-mirror.json}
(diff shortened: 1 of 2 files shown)


hardys:
Thanks for fixing this, it's something I missed in https://github.com/openshift-metal3/dev-scripts/pull/902

I was thinking we'd inspect the existing certs, but this approach seems fine as an alternative - I'll test locally but lgtm!

metal3ci:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/1494/

PR #925, opened 2020-02-07
Don't always install operator-sdk

hardys:
This is only required when running the BMO locally so
lets only install it when needed - this saves us several
minutes in every CI run :)


hardys:
We also always download and install the oc tarball but I'll try to figure out how to cache that in a followup

metal3ci:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/1496/

PR #926, opened 2020-02-07
Fix world-readable check in sanitychecks.sh

hardys:
This issue was mentioned in #900 but I missed it when
merging - we need to ensure world-readable access to
the entire path, not only the $WORKING_DIR which may
be a subdirectory.

Also made the script executable and moved a variable to
enable easier local testing, and reversed the order of
the size and access test (the size test will always fail
if the directory is not readable, so we should check for
access first).


andfasano:
Reviewed and tested and looks fine to me. Fully agree on the reversed order check

metal3ci:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/1497/

PR #927, opened 2020-02-07
Add 03_build_installer.sh to README.md

slagle:
The 03_build_installer.sh script is run when "make" is executed, however
it was not mentioned in README.md. Manually running the scripts as shown
in README.md prior to this patch would result in an error as the ocp
installer was never extracted.


hardys:
Thanks!

PR #928, opened 2020-02-07
Use a better value for $PS4

slagle:
Sets a value for $PS4 which will log the script name and line number,
and function name (if applicable).

As most of the scripts already run -x, this doesn't really add any extra
verbosity and is very helpful when reviewing logs and failures.

### DIFF
--- a/common.sh
+++ b/common.sh
@@ -2,6 +2,9 @@
 
 export PATH="/usr/local/go/bin:$PATH"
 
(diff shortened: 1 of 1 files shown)


hardys:
lgtm - it will make the logs easier to follow.

There is still some possibility of confusion when we log a script name from a different directory, e.g when we cd to metal3-dev-env, but logging the $PWD as well will probably make the line prefix too long.

PR #941, opened 2020-02-21
Adding test to verify workingdir is writeable

andfasano:


sanitychecks.sh
```diff
@@ -20,6 +20,11 @@ function verifyWorkingDirIsWorldReadable {
     echo "The WORKING_DIR ${WORKING_DIR} is not world-readable!"
     exit 1 
   fi
+
+  if ! sudo -u nobody test -w ${WORKING_DIR}; then
+    echo "The WORKING_DIR ${WORKING_DIR} is not world-writable!"
```
hardys: The working dir shouldn't be world writeable - it needs to be world readable and writeable by the user (and root)?

andfasano: Ok let me check since I had a failure some days ago due a lack of permission

andfasano: The problem I faced was that I created as WORKING_DIR `/home/dev-scripts` via sudo but without changing the owner, so it was writable for root but not for the local user

hardys: Ok a check that it's writeable by $USER would be reasonable, although that should not happen if you let dev-scripts create the directory so it's something of a corner case.

andfasano: Agree I'll change the code in the direction discussed then

metal3ci:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/1537/

PR #942, opened 2020-02-21
Minor tip for sshuttle usage

andfasano:
Minor tip to improve the usage of sshuttle to connect to the web console

### DIFF
--- a/README.md
+++ b/README.md
@@ -161,6 +161,12 @@ Run sshuttle:
 sshuttle -r <user>@<virthost> 192.168.111.0/24
 ```
 
+*Tip:* to specify a different ssh key for `<virthost>`, use the following command:
+
+```
+sshuttle -e "ssh -i ~/.ssh/mycustomkey_rsa" -r <user>@<virthost> 192.168.111.0/24
+```
+
 To access the web Console use the `kubeadmin` user, and password generated in the `ocp/auth/kubeadmin-password` file.
 
 


README.md
```diff
@@ -161,6 +161,12 @@ Run sshuttle:
 sshuttle -r <user>@<virthost> 192.168.111.0/24
 ```
 
+*Tip:* to specify a different ssh key for `<virthost>`, use the following command:
+
+```
+sshuttle -e "ssh -i ~/.ssh/mycustomkey_rsa" -r <user>@<virthost> 192.168.111.0/24
+```
```
hardys: I'm interested to understand when/why this would happen in the default dev-scripts environment?

We create a key in the default location ref https://github.com/openshift-metal3/dev-scripts/blob/master/02_configure_host.sh#L11 so when would this alternative key get generated in the documented environment?

andfasano: I got this case while working on the CI on Packet.net: the CI job takes care of provisioning a baremetal instance for a given account on Packet.net with a given ssh key, and after that checkout and build dev-scripts on the instance

hardys: Ack, to keep the instructions as simple as possible, perhaps we should start a separate readme file for CI specific steps?

andfasano: Well, originally I figured out it as a more generic and handy tip for sshuttle, as it could not be too much obvious. The CI in this case was just the triggering context (building a different instance before running dev-script), as AFAIK we do not access the web console in CI tests.

hardys: My concern is introducing variables which aren't applicable in the default case - this is an advanced case where you've got to tunnel to a remote host and are using a customized SSH_PUB_KEY, so it may be confusing for new users to see this and wonder where e.g mycustomkey_rsa is supposed to come from?

andfasano: If for "default case" you mean running dev-scripts on the same local machine where the web browser is executed then I agree that it is not needed. Indeed I was pointing out to another scenario (the advanced one), where dev-scripts is executed on a remote machine and the web browser is launched from the local user machine (another similar example is when I run dev-scripts on a lab machine and I want to connect to the cluster from my laptop). Could make sense to rephrase it under this point of view (trying to separate different use cases) or do you think it could be still confusing?

hardys: Yes rephrasing this to describe any additional steps needed when connecting to a remote machine could make sense - do we typically need a custom SSH key in that case though?

By "default case" I just meen the typical non-advanced case which we document in the README, my concern is just that we keep that concise and not confusing.

andfasano: Well, a custom key is required when provisioning upfront the server, such as in the case of packet.net

hardys: Right, so probably it would be best to either drop this PR, or convert it into one which explains how to run dev-scripts in the CI environment?

andfasano: Probably for sshuttle it's ok to use also the generated key (but it must be copied to the user machine before)

andfasano: Ok let's drop it

PR #943, opened 2020-02-21
Add support for coredns-mdns to run_ci.sh

cybertron:
coredns-mdns is somewhat unique in that it doesn't become its own
image, it gets included in the coredns image as a vendored plugin.
Because of this, we need a bit more logic around building proposed
changes than just exporting the appropriate local image env var.

This change adds the ability to override the Dockerfile name for the
image build in dev-scripts (because coredns is also unusual in that
way) and adds the necessary steps to clone and update the coredns
code so it can be built with the proposed change.

### DIFF
--- a/04_setup_ironic.sh
+++ b/04_setup_ironic.sh
@@ -38,7 +38,10 @@ for IMAGE_VAR in $(env | grep "_LOCAL_IMAGE=" | grep -o "^[^=]*") ; do
         cd $REPOPATH
         export $IMAGE_VAR=${IMAGE##*/}:latest
         export $IMAGE_VAR=$LOCAL_REGISTRY_DNS_NAME:$LOCAL_REGISTRY_PORT/localimages/${!IMAGE_VAR}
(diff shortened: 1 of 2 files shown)


metal3ci:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/1536/

cybertron:
It looks like the failure was downloading an image from quay. Is there a way to trigger a retest of metal3ci?

hardys:
> It looks like the failure was downloading an image from quay. Is there a way to trigger a retest of metal3ci?

You can take the parameters from the failing build e.g http://10.8.144.11:8080/job/dev-tools/1536/parameters/ and copy/paste them into a new build "Run with Parameters" in the Jenkins UI, e.g http://10.8.144.11:8080/job/dev-tools/build?delay=0sec

I've just done that resulting in http://10.8.144.11:8080/job/dev-tools/1538/

metal3ci:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/1538/

cybertron:
Good to know, thanks!

PR #945, opened 2020-02-25
Add possibility to pass boot_mode

yrobla:
Create a new property boot_mode in the properties
dictionary. It can accept two values (UEFI, legacy) and
will default to UEFI. This will be passed to install-config
and will be picked by the installer.


metal3ci:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/1660/

PR #947, opened 2020-02-26
Add option to create more VMs than OCP workers

luis5tb:
This allows scaling number of workers with machineset obj in a
more easy way as the libvirt VMs are already created and registered
in ironic


02_configure_host.sh
```diff
@@ -29,15 +29,15 @@ fi
 export VM_NODES_FILE=${VM_NODES_FILE:-}
 if [ ! -z "${VM_NODES_FILE}" ]; then
   echo "VM_NODES_FILE is no longer supported"
-  echo "Please use NUM_MASTERS, NUM_WORKERS and VM_EXTRADISKS variables instead"
+  echo "Please use NUM_MASTERS, NUM_VM_WORKERS and VM_EXTRADISKS variables instead"
```
hardys: This should still document the NUM_WORKERS interface I think?

luis5tb: As in the ansible-playbook right below the information used is NUM_VM_WORKERS, I assumed that was what was needed there.

hardys: The existing option is NUM_WORKERS, and unless folks are doing advanced testing e.g scale out there is no need to mention NUM_VM_WORKERS here IMO.

NUM_WORKERS is still the primary interface that determines the number of deployed workers.

hardys:
lgtm, thanks!

metal3ci:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/1543/

hardys:
@luis5tb would you mind rebasing this please?

metal3ci:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/1587/

metal3ci:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/1602/

metal3ci:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/1620/

ocp_install_env.sh
```diff
@@ -117,7 +117,7 @@ $(network_configuration)
     dnsVIP: ${DNS_VIP}
     hosts:
 $(node_map_to_install_config_hosts $NUM_MASTERS 0 master)
-$(node_map_to_install_config_hosts $NUM_WORKERS $NUM_MASTERS worker)
+$(node_map_to_install_config_hosts $NUM_VM_WORKERS $NUM_MASTERS worker)
```
hardys: I tested and found one problem with this - all of the workers get registered via the install-config, even though the replicas are set to NUM_WORKERS - this can cause problems because if NUM_VM_WORKERS is large (I tried 20) all of the VMs try to start enrollment then introspection at the same time, and in my environement this causes issues with VBMC that leads to failures.

I think perhaps we should leave this set to NUM_WORKERS, and generate the BMH yaml so that can be applied after initial deployment?

hardys: Note we did have code that did the BMH generation prior to https://github.com/openshift-metal3/dev-scripts/pull/897 so we could perhaps restore some of that to do what's described above.

hardys:
I noticed another problem - we need to account for the extra workers during cleanup here:

https://github.com/openshift-metal3/dev-scripts/blob/master/host_cleanup.sh#L19

config_example.sh
```diff
@@ -37,7 +41,7 @@ set -x
 # Provisioning interface within the cluster
 #export CLUSTER_PRO_IF="eno1"
 
-# Which disk to deploy 
+# Which disk to deploy
```
elfosardo: nit: unrelated change

config_example.sh
```diff
@@ -28,6 +28,10 @@ set -x
 # Indicate number of workers to deploy
 #export NUM_WORKERS=0
 
+# Indicate number of VMs to be created (as possible workers). Must be greater
+# than NUM_WORKERS
+#export NUM_VM_WORKERS=0
```
elfosardo: perhaps a good idea adding a check in the code to verify this condition ?

luis5tb:
> I noticed another problem - we need to account for the extra workers during cleanup here:
> 
> https://github.com/openshift-metal3/dev-scripts/blob/master/host_cleanup.sh#L19
This is already included (changed to NUM_VM_WORKERS

ocp_install_env.sh
luis5tb: how is that different from setting NUM_WORKERS to 20?

hardys: > how is that different from setting NUM_WORKERS to 20?

The difference is you could scale out in smaller increments and the hosts aren't all registered via the installer (otherwise you'd just use NUM_WORKERS, right?)

hardys:
> This is already included (changed to NUM_VM_WORKERS

Hmm yeah I see - it didn't work for me locally so will re-test, thanks!

hardys:
Sorry @luis5tb - seems like this PR got stale and https://github.com/openshift-metal3/dev-scripts/pull/1033 has now landed which I think addresses the same use-case, are you OK if we close this one now?

hardys:
Ok closing this as obsolete since #1033 covers the same use-case

luis5tb:
> Ok closing this as obsolete since #1033 covers the same use-case

Sorry for the delay! Yes, perfectly fine to close it! Nice that it was done on #1033!

PR #955, opened 2020-03-04
Fix some hardcoded network references

hardys:
These will break when we move the baremetal network to ipv6 ref #951 and also we missed a couple of references to the "baremetal" network in #948


metal3ci:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/1564/

PR #958, opened 2020-03-04
Re-add tripleo repo cleanup

cybertron:
This accidentally got dropped in [0]. It's still needed to prevent
dependency issues on re-deploy.

0: https://github.com/openshift-metal3/dev-scripts/commit/f64ebda3671a6d6e7620898181e0648b40b5a9f3

### DIFF
--- a/host_cleanup.sh
+++ b/host_cleanup.sh
@@ -23,7 +23,7 @@ ANSIBLE_FORCE_COLOR=true ansible-playbook \
     -i ${VM_SETUP_PATH}/inventory.ini \
     -b -vvv ${VM_SETUP_PATH}/teardown-playbook.yml
 
-sudo rm -rf /etc/NetworkManager/dnsmasq.d/openshift-${CLUSTER_NAME}.conf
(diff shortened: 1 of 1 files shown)


hardys:
@cybertron while it looks like the removal was a rebase mistake, I think this should no longer be needed since https://github.com/metal3-io/metal3-dev-env/pull/220 landed?

cybertron:
I still hit it yesterday in dev-scripts, and I don't think 220 fixes this. It removes the tripleo-repos tool, but not the repos that it installed, which are not part of the package.

hardys:
> I still hit it yesterday in dev-scripts, and I don't think 220 fixes this. It removes the tripleo-repos tool, but not the repos that it installed, which are not part of the package.

Ah yes you're right - and I missed that when testing metal3-io/metal3-dev-env#220 because a previous version of the PR did remove the repos.

Probably we should fix that in metal3-dev-env but for now I'll go ahead and merge this - thanks!

cybertron:
Yeah, I originally did it here because I don't think dev-scripts calls the metal3-dev-env cleanup scripts, but if we do it early in the run like your patch did instead of on cleanup then maybe that's not a problem anymore. I guess there's an argument that we shouldn't leave the repos lying around anyway, but I'm not too concerned about that.

PR #968, opened 2020-03-11
Forward mode of baremetal bridge needs to be flexible

yrobla:
If we choose not to manage our bridges, we assume that those
are already created on the host. In that case we cannot use nat,
because it will conflict with the existing bridge. We need to
use bridge mode (shared physical device).


metal3ci:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/1608/

metal3ci:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/1613/

metal3ci:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/1617/

PR #975, opened 2020-03-16
Remove RHCOS image hack for ipv6

hardys:
I tested on 4.4.0-0.ci-2020-03-14-105727 which contains the bump
from https://github.com/openshift/installer/pull/3271 and confirmed
that we no longer need this.


metal3ci:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/1631/

PR #977, opened 2020-03-18
Add -g to curl commands that might get ipv6 addresses

### LINKED ISSUE #976: curl fails with ipv6 addresses
Tried my first ipv6 deployment with dev-scripts and it hung trying to download images. The error appears to be:

```
+(./04_setup_ironic.sh:150): curl --fail 'http://[fd00:1101::1]/images/rhcos-44.81.202003062006-0-openstack.x86_64.qcow2.gz.sha256sum'
curl: (3) [globbing] error: bad range specification after pos 9
```

It looks like we need to add -g to the curl commands so it doesn't interpret the brackets as glob expressions. I'm testing a patch locally that does so.

cybertron:
Without this, attempts to curl from ipv6 fail with:

  curl: (3) [globbing] error: bad range specification after pos 9

-g tells curl to disable globbing so it will just use the url as
provided.

Closes #976

### DIFF
--- a/04_setup_ironic.sh
+++ b/04_setup_ironic.sh
@@ -147,8 +147,8 @@ fi
 sudo podman wait -i 1000 ipa-downloader
 
 # Wait for images to be downloaded/ready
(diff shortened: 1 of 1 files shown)


russellb:
not a problem with all curl versions, but `-g` should be fine always

PR #980, opened 2020-03-18
update show_bootstrap_log.sh for ipv6

dhellmann:


hardys:
I suspect we can simplify the logic to get the IP but this looks fine in the context of making it work for ipv6

PR #981, opened 2020-03-18
fix bootstrap vm volume cleanup

dhellmann:
The volume list output starts with a blank space so the grep pattern
looking for the cluster name at the start of the line was not
matching. Fix this by switching the awk to extract the volume name to
the before the grep to filter the volumes.


hardys:
Ah good catch, I'd not yet spotted why these needed periodic manual cleanup :)

PR #982, opened 2020-03-18
improve vbmc and sushy-tools management

dhellmann:
* Do not stop them when we start a cluster, in case we have multiple
  clusters.

* Do not try to start them if they are already running from a previous
  cluster.

* Remove the vbmc PID file when starting it because it frequently
  leaves the file behind.


metal3ci:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/1650/

metal3ci:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/1653/

PR #984, opened 2020-03-19
add more cleanup scripts and Makefile target

dhellmann:
Use `make realclean` to run all of the new cleaning steps, in addition
to the old ones.

cache_cleanup.sh removes user cache files related to the installer.

registry_cleanup.sh removes the local registry service and its files.

workingdir_cleanup.sh removes the entire working directory.

### DIFF
--- a/cache_cleanup.sh
+++ b/cache_cleanup.sh
@@ -1,0 +1,7 @@
+#!/usr/bin/env bash
+set -x
+
+source logging.sh
+source common.sh
+
+rm -rf $HOME/.cache/openshift-installer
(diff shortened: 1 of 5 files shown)


cache_cleanup.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+set -x
+
+source logging.sh
+source common.sh
+source utils.sh
```
hardys: I don't think utils.sh is used in any of these?

hardys:
Looks good modulo utils.sh nit

Another thing we might want to clean is stale/old local podman images, e.g:

```
    sudo podman image prune --all
```

I find myself doing this periodically to reclaim space where old images build up.

cache_cleanup.sh
dhellmann: Fixed.

dhellmann:
> Looks good modulo utils.sh nit
> 
> Another thing we might want to clean is stale/old local podman images, e.g:
> 
> ```
>     sudo podman image prune --all
> ```
> 
> I find myself doing this periodically to reclaim space where old images build up.

Good idea. I added that to the latest revision.

PR #986, opened 2020-03-21
Remove add-machine-ips script

markmc:
Not used/referenced since commits be6cfd76 and 5c381840.


russellb:
The scripts are not used directly, but are still used by some people.

They are required to get Machine and Node resources linked together.  AFAIK the only thing that requires it is the host management UI, which expects to find links between Node-Machine-BareMetalHost.  Those links exist for workers, but not masters.

Related bz: https://bugzilla.redhat.com/show_bug.cgi?id=1801238

I suppose if we keep them in the repo, they could be moved to a dir with a README that explains why they still exist, and with a link to the related bug?  I can do that if you agree.

hardys:
Yeah we should probably leave this until we land a fix to associate the master node/machines.

https://github.com/openshift-metal3/dev-scripts/issues/917 was raised quite recently as this impacts UX things - discussions are in progress on https://github.com/openshift/enhancements/pull/229 to enable continuous introspection which would be one way to fix this.

Another would be to do introspection in the installer terraform workflow and add the data to the BMH during the install workflow (which isn't necessarily mutually exclusive to the continuous introspection approach, but we could save some walltime by skipping this step during deployment).  @stbenjam has some code which does this, and it may be we go that way if the continuous-introspection work is not accepted for 4.5

add-machine-ips.sh
```diff
@@ -1,18 +0,0 @@
-#!/bin/bash
-set -x
-set -e
-
-source logging.sh
-source utils.sh
-source common.sh
-source ocp_install_env.sh
-
-for node in $(oc --config ocp/auth/kubeconfig get nodes -o template --template='{{range .items}}{{.metadata.uid}}:{{.metadata.name}}{{"\n"}}{{end}}'); do
```
markmc: `ocp/auth/kubeconfig` is wrong

markmc:
Ok, if this is still needed, it's broken because it refers to `ocp/auth/kubeconfig` which should now be `ocp/auth/$CLUSTER_NAME/kubeconfig`

markmc:
Why did commit 5c38184 remove csr_hack from `Makefile` ?

russellb:
> Why did commit [5c38184](https://github.com/openshift-metal3/dev-scripts/commit/5c38184019ff5dcba45c337acd34af4a6df39de2) remove csr_hack from `Makefile` ?

The scripts *used* to serve a purpose related to CSR approval, but they are no longer needed for that purpose.  We never completed getting Machines and Nodes linked together automatically for masters, and the urgency to do so really dropped off when it was no longer required for master CSR approval.

The CSR approval workflow for masters works without this because first the CSRs are automatically approved during bootstrapping:

* https://github.com/openshift/installer/blob/master/data/data/bootstrap/systemd/units/approve-csr.service
* https://github.com/openshift/installer/blob/master/data/data/bootstrap/files/usr/local/bin/approve-csr.sh

and then the cluster-machine-approver was fixed to automatically approve certificate renewals for a Node that already has a valid certificate.  

* https://bugzilla.redhat.com/show_bug.cgi?id=1737611
* https://github.com/openshift/cluster-machine-approver/pull/38

The host management UI was built on environments using these scripts, and makes this Node-Machine link assumption, and is now the only thing broken by it not being automated.

The scripts could be re-added to the makefile under a different name to make it clear it's not related to the CSR workflow anymore.

PR #988, opened 2020-03-21
Remove broken KUBECONFIG env var

markmc:
common.sh still points `KUBECONFIG` at `ocp/auth` instead of
`ocp/$CLUSTER_NAME/auth` but it moved in commit f64ebda

It looks like we don't actually need `KUBECONFIG` set anywhere these
days.


dhellmann:
I have a patch in #959 to fix the location. We do run some oc commands, so I would expect those to need the variable set.

markmc:
Ok, closing

PR #990, opened 2020-03-26
Add ASSETS_EXTRA_FOLDER for additional manifests

yrobla:
There is the need to deploy clusters with customized settings,
that will need to include additional manifests at deploy time.
Add the possibility to include an ASSETS_EXTRA_FOLDER, that will
allow to put additional manifests there and will be copied to
the manifests before install.

### DIFF
--- a/utils.sh
+++ b/utils.sh
@@ -70,11 +70,15 @@ function create_cluster() {
       cp assets/metal3-cbo-deployment.yaml ${assets_dir}/openshift/.
     fi
 
+    if [ ! -z "${ASSETS_EXTRA_FOLDER:-}" ]; then
+      cp -rf ${ASSETS_EXTRA_FOLDER}/*.yaml ${assets_dir}/openshift/
+    fi
+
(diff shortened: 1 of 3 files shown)


utils.sh
```diff
@@ -49,6 +49,10 @@ function create_cluster() {
     mkdir -p ${assets_dir}/openshift
     cp -rf assets/generated/*.yaml ${assets_dir}/openshift
 
+    if [ ! -z "${ASSETS_EXTRA_FOLDER:-}" ]; then
+      cp -rf ${ASSETS_EXTRA_FOLDER}/* ${assets_dir}/openshift
```
karmab: you want to make sure only yaml files are copied to the openshift directory or bootstrapping will fail (in a non obvious way)

hardys:
lgtm other than the comment from @karmab which looks like an easy fix and would be good to address before merging?

metal3ci:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/1713/

PR #992, opened 2020-03-26
multi-cluster: do not tie the registry hostname to the cluster name

dhellmann:
Use a hostname that isn't tied to the cluster for the registry to
allow it to be used for multiple clusters.


metal3ci:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/1671/

dhellmann:
Abandoning this approach

Issue #932, opened 2020-02-11
Re-running `02_configure_host.sh` can cause unexpected results


hardys:
An issue noted by @dtantsur in https://github.com/openshift-metal3/dev-scripts/pull/919 regarding mismatched mac addresses in the generated `ironic_nodes.json` has been debugged and I'm raising this to keep track of the issue.

The problem is it's possible to re-run `02_configure_host.sh` without first running `host_cleanup.sh` which tears down the existing libvirt resources.

In this situation, the ansible libvirt modules don't modify the existing resources, but we still generate new macs in the ansible roles - thus we write out an `ironic_nodes.json` which doesn't match reality.

We probably have two options, either fail fast and don't allow 02 to be run if there are existing libvirt things, or we can make the metal3-dev-env roles smarter and ensure they don't overwrite existing `ironic_nodes.json` in the case where no new resources have been created.

PR #998, opened 2020-03-31
Pass NODES_FILE into teardown

derekhiggins:
So that it can be cleaned up.

Fixes: https://github.com/openshift-metal3/dev-scripts/issues/932

### DIFF
--- a/host_cleanup.sh
+++ b/host_cleanup.sh
@@ -20,6 +20,7 @@ ANSIBLE_FORCE_COLOR=true ansible-playbook \
     -e "extradisks=$VM_EXTRADISKS" \
     -e "virthost=$HOSTNAME" \
     -e "manage_baremetal=$MANAGE_BR_BRIDGE" \
+    -e "nodes_file=$NODES_FILE" \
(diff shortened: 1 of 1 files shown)


hardys:
I'm not sure this is the entire fix to #932 - I think @dtantsur was expecting 02 to be idempotent, and I suspect we need changes to metal3-dev-env to make that possible (it would be helpful for example to be able to add nodes after a deploy, as an alternative to the approach proposed in https://github.com/openshift-metal3/dev-scripts/pull/947)

This looks fine though, makes sense to clean it up on teardown.

derekhiggins:
> I'm not sure this is the entire fix to #932 - I think @dtantsur was expecting 02 to be idempotent, and I suspect we need changes to metal3-dev-env to make that possible (it would be helpful for example to be able to add nodes after a deploy, as an alternative to the approach proposed in #947)

There is an associated metal3-dev-env patch, I mentioned it in the issue #932 but never mentioned it here
https://github.com/metal3-io/metal3-dev-env/pull/279

PR #1005, opened 2020-04-03
add variable to control use of remote libvirturi in install-config

dhellmann:
We only want to use the remote libvirt feature when configuring a
cluster for hive, so use a flag to toggle that behavior.


metal3ci:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/1702/

PR #1006, opened 2020-04-03
move the mirror log file to working dir with cluster-specific name

dhellmann:
We want the file to have the cluster name in it so we get different
mirror instructions for each cluster. We want the file to be in a
place that we clean up when we clean the registry so we don't
use stale info.

### DIFF
--- a/common.sh
+++ b/common.sh
@@ -231,6 +231,10 @@ if [[ "$OPENSHIFT_VERSION" != "4.3" ]]; then
   export BMC_DRIVER=${BMC_DRIVER:-redfish}
 fi
 
+# Both utils.sh and 04_setup_ironic.sh use this log file, so set the
+# name one time. Users should not override this.
+export MIRROR_LOG_FILE=${REGISTRY_DIR}/${CLUSTER_NAME}-image_mirror-${OPENSHIFT_RELEASE_TAG}.log
+
 # Switch Container Images to upstream, Installer defaults these to the openshift version
 if [ "${UPSTREAM_IRONIC:-false}" != "false" ] ; then
     export IRONIC_LOCAL_IMAGE=${IRONIC_LOCAL_IMAGE:-"quay.io/metal3-io/ironic:master"}
--- a/04_setup_ironic.sh
+++ b/04_setup_ironic.sh
@@ -61,9 +61,8 @@ if [ ! -z "${MIRROR_IMAGES}" ]; then
     # pull from one registry and push to local one
     # hence credentials are different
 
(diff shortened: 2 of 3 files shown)


metal3ci:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/1701/

common.sh
```diff
@@ -124,6 +124,10 @@ fi
 
 export OPENSHIFT_RELEASE_TAG=$(echo $OPENSHIFT_RELEASE_IMAGE | sed -E 's/[[:alnum:]\/.-]*release.*://')
 
+# Both utils.sh and 04_setup_ironic.sh use this log file, so set the
+# name one time. Users should not override this.
+export MIRROR_LOG_FILE=${REGISTRY_DIR}/${CLUSTER_NAME}-image_mirror-${OPENSHIFT_RELEASE_TAG}.log
```
hardys, line 129: The PR description says we want this to be cleaned between runs, but in the default `make clean all` case the registry dir is not cleaned (deliberately, so we can cache container images between runs).

Would it be better to put this in either `$OCP_DIR` (which is already cleaned between runs), or perhaps `$WORKING_DIR/$CLUSTER_NAME` (which is created in common.sh but AFAICS isn't cleaned between runs, but we could add cleaning it?)

dhellmann: The phrasing on that message was poor. We only want the file removed when we do clean the registry, and we always want it removed when we clean the registry. I've updated the text.

metal3ci:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/1716/

metal3ci:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/1718/

metal3ci:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/1720/

common.sh
hardys: But we could switch to a different mirrored release between test runs, even if we don't clean the registry?

I guess we'll overwrite the mirror log file in that case but I just want to make sure only cleaning this on registry clean is what we want?

dhellmann: The version is included in the filename via $OPENSHIFT_RELEASE_TAG so we get a new file for each version and for each cluster. That lets you install different versions into different clusters and mirror them all from the same registry instance.

metal3ci:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/1725/

metal3ci:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/1743/

PR #1009, opened 2020-04-07
multi-cluster: add script to install hive with required configuration

dhellmann:
Include a variable for managing which hive image to use.  This makes it
easier to deploy a private build consistently.

### DIFF
--- a/README.md
+++ b/README.md
@@ -301,4 +301,3 @@ The following adds an additional file `/etc/test` as an example:
 ```
 export IGNITION_EXTRA="ignition/file_example.ign"
 ``` 
-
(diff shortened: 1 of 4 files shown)


metal3ci:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/1709/

hardys:
I wonder if we should calculate KUBECONFIG in the case where it's not already set by the user, and set it in common.sh (currently we do it in `run_ci.sh` and `run-custom-mao.sh` but I think this PR assumes the user env already set it?)

Anyway not a blocker since this passed CI, perhaps we can discuss it and address that as a followup.

PR #1010, opened 2020-04-07
multi-cluster: support a log file prefix for logs that "go together"

dhellmann:


logging.sh
```diff
@@ -1,9 +1,13 @@
 # Log output automatically
 LOGDIR="$(dirname $0)/logs"
-if [ ! -d "$LOGDIR" ]; then
-    mkdir -p "$LOGDIR"
+if [[ -z "${LOGPREFIX:-}" ]]; then
```
russellb, line 3: want to document LOGPREFIX somewhere?

dhellmann: I'm trying to break #959 up to make it easier to review, but maybe I should have left this one in that series.

I don't know that anyone would actually use this variable, but the new multicluster.sh script uses it so that all of the logs for a given batch job end up in the same subdirectory.

russellb: OK, not a big deal

metal3ci:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/1711/

PR #1017, opened 2020-04-16
Add dual stack IPv4+IPv6 support

russellb:
This commit allows configuring a dual-stack IPv4+IPv6 environment.  It
changes a bunch of configuration to reflect that IPv4 and IPv6
settings are required at the same time, creates a dual-stack
"baremetal" network environment, and configures the OpenShift cluster
with both IPv4 and IPv6 subnets.

This isn't actually expected to result in a working OpenShift cluster
yet, but this provides a development and test environment for
dual-stack OpenShift clusters.

### DIFF
--- a/01_install_requirements.sh
+++ b/01_install_requirements.sh
@@ -15,8 +15,10 @@ if [ -z "${METAL3_DEV_ENV}" ]; then
 fi
 
 pushd ${METAL3_DEV_ENV}
-# FIXME: Pin to aafbb89df8df88bc3ecbfe59592149c929432515
-git reset aafbb89df8df88bc3ecbfe59592149c929432515 --hard
+# Pin to a specific metal3-dev-env commit to ensure we catch breaking
+# changes before they're used by everyone and CI.
+# TODO -- come up with a plan for continuously updating this
+git reset b19da74e06062d0c21ade7dea1c6e8d09f7f4e48 --hard
 ./centos_install_requirements.sh
 ansible-galaxy install -r vm-setup/requirements.yml
 ANSIBLE_FORCE_COLOR=true ansible-playbook \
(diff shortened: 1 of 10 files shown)


metal3ci:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/1731/

russellb:
@stbenjam @hardys CI passed!  ready for a look when you have a chance

russellb:
For reference, this makes use of https://github.com/metal3-io/metal3-dev-env/pull/280

metal3ci:
Build SUCCESS, see build http://10.8.144.11:8080/job/dev-tools/1732/

01_install_requirements.sh
```diff
@@ -15,8 +15,10 @@ if [ -z "${METAL3_DEV_ENV}" ]; then
 fi
 
 pushd ${METAL3_DEV_ENV}
-# FIXME: Pin to aafbb89df8df88bc3ecbfe59592149c929432515
-git reset aafbb89df8df88bc3ecbfe59592149c929432515 --hard
+# Pin to a specific metal3-dev-env commit to ensure we catch breaking
+# changes before they're used by everyone and CI.
+# TODO -- come up with a plan for continuously updating this
+git reset b19da74e06062d0c21ade7dea1c6e8d09f7f4e48 --hard
```
hardys, line 21: We should do this in the conditional L11..L14 - we only want to reset in the case where METAL3_DEV_ENV isn't set by a developer for testing a local checkout IMO.

hardys: That can be a followup if you prefer but I noticed it trying to locally test with a newer metal3-dev-env so perhaps we can just move it in this PR?

hardys:
I noticed one issue which I'd like to fix, but I can push a PR to resolve it as a followup, otherwise lgtm and it passed CI

PR #1022, opened 2020-04-21
New OpenShift CI Docker image

andfasano:
Refactor Dockerfile.ci to be used with OpenShift CI pipeline:src, required for OpenShift CI integration (see https://github.com/openshift/release/pull/8420)

### DIFF
--- a/images/Dockerfile.ci
+++ b/images/Dockerfile.ci
@@ -4,7 +4,7 @@
 FROM centos:7
 RUN yum -y install python3 git && \
     yum clean all && \
-    pip3 install ansible packet-python 
-RUN mkdir /root/src
+    pip3 install ansible packet-python && \
+    mkdir /root/src
 WORKDIR /root/src
 
--- a/images/Dockerfile.ci
+++ b/images/Dockerfile.ci
@@ -1,9 +1,10 @@
 # This Dockerfile builds the image used by the e2e-metal-ipi test steps in the OpenShift CI.
 # For more details about the test see https://steps.svc.ci.openshift.org/job/openshift-baremetal-operator-master-e2e-metal-ipi
 
-FROM registry.access.redhat.com/ubi8/ubi
-RUN dnf -y install python3 && \
-    dnf clean all && \
-    pip3 install ansible && \
-    pip3 install packet-python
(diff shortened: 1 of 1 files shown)


images/Dockerfile.ci
```diff
@@ -1,9 +1,10 @@
 # This Dockerfile builds the image used by the e2e-metal-ipi test steps in the OpenShift CI.
 # For more details about the test see https://steps.svc.ci.openshift.org/job/openshift-baremetal-operator-master-e2e-metal-ipi
 
-FROM registry.access.redhat.com/ubi8/ubi
-RUN dnf -y install python3 && \
-    dnf clean all && \
-    pip3 install ansible && \
-    pip3 install packet-python
+FROM centos:7
```
hardys, line 4: Is this consistent with other openshift CI?  We've been aligning with centos8 generally so I'd expected this to go from ubi8 to centos8?  /cc @elfosardo

andfasano: This topic was discussed with the DPTP team, and they don't see a requirement for a test only image to inherit from ubi8.
In this case the requirement is dictated by the `build_root` image (see https://github.com/openshift/release/blob/1f8a505632558b432594390dff38cb7886488dd6/ci-operator/config/openshift-metal3/dev-scripts/openshift-metal3-dev-scripts-master__ocp-4.5.yaml#L6), because the Dockerfile.ci will be based on that image. 
The image in the `build_root` configuration has the benefit to be already prepared to be a valid build root image, and has already go available in case it will be needed for further tests on devscripts.

elfosardo: @hardys unfortunately the src images are based on rhel7, I would also push for anything with an 8 at the end, but at the moment it seems not possible at least for this scope :/

images/Dockerfile.ci
```diff
@@ -1,9 +1,10 @@
...
+RUN yum -y install python3 git && \
+    yum clean all && \
+    pip3 install ansible packet-python 
+RUN mkdir /root/src
```
elfosardo, line 8: you can aggregate this with the precedent RUN and save one layer! :)

PR #1029, opened 2020-04-29
Drop CentOS 7 support completely

russellb:
CentOS 7 was marked deprecated in #1000.  Everyone really needs to be
on RHEL 8 or CentOS 8 at this point.  Drop CentOS 7 support completely
now that a warning has been in place for a bit.

RHEL 8 (or CentOS 8) is required because that's the environment we
would expect customers to use to run the installer.  Any issues with
CentOS 7 are not worth our focus anymore.

### DIFF
--- a/02_configure_host.sh
+++ b/02_configure_host.sh
@@ -88,11 +88,8 @@ EOF
     virsh pool-autostart default
 fi
 
-if [ "${RHEL8}" = "True" ] || [ "${CENTOS8}" = "True" ] ; then
-    ZONE="\nZONE=libvirt"
-
-    sudo systemctl enable --now firewalld
-fi
+ZONE="\nZONE=libvirt"
+sudo systemctl enable --now firewalld
 
 if [ "$MANAGE_PRO_BRIDGE" == "y" ]; then
     # Adding an IP address in the libvirt definition for this network results in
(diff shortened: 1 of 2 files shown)


hardys:
+1 but /cc @andfasano who was asking about the status of CentOS7 support for IBMCloud CI investigations - did you get any indication when RHEL/CentOS8 may be an option there yet?

russellb:
> +1 but /cc @andfasano who was asking about the status of CentOS7 support for IBMCloud CI investigations - did you get any indication when RHEL/CentOS8 may be an option there yet?

It's not an option there, but I just consider that a blocker to using that platform.  I don't think it's a good idea to move our testing off the platform that's closest to what production environments would be using.

hardys:
> It's not an option there, but I just consider that a blocker to using that platform. I don't think it's a good idea to move our testing off the platform that's closest to what production environments would be using.

Yeah I agree but wanted to raise visibility of the issue :)

hardys:
Looks good but I think there is some dead code we can remove:

https://github.com/openshift-metal3/dev-scripts/blob/master/02_configure_host.sh#L91

https://github.com/openshift-metal3/dev-scripts/blob/master/common.sh#L388

I guess we could remove the USE_FIREWALLD conditionals too since thats always used now?

russellb:
> Looks good but I think there is some dead code we can remove:
> 
> https://github.com/openshift-metal3/dev-scripts/blob/master/02_configure_host.sh#L91
> 
> https://github.com/openshift-metal3/dev-scripts/blob/master/common.sh#L388
> 
> I guess we could remove the USE_FIREWALLD conditionals too since thats always used now?

Oh, right.  Thanks for the reminder.  I'll see what I can remove and update this PR.

PR #1033, opened 2020-05-05
support workers not included in initial deployment

dhellmann:
Add NUM_EXTRA_WORKERS variable to manage creating more workers than
are deployed initially. A manifest file for registering these
additional hosts is created in the ocp directory for the cluster.

### DIFF
--- a/02_configure_host.sh
+++ b/02_configure_host.sh
@@ -45,8 +45,11 @@ if [[ "${IP_STACK}" == "v6" ]]; then
   done
 fi
 
-ANSIBLE_FORCE_COLOR=true ansible-playbook \
+export ANSIBLE_FORCE_COLOR=true
+
+ansible-playbook \
     -e @vm_setup_vars.yml \
+    -e "ironic_prefix=${CLUSTER_NAME}_" \
     -e "cluster_name=${CLUSTER_NAME}" \
     -e "provisioning_network_name=${PROVISIONING_NETWORK_NAME}" \
     -e "baremetal_network_name=${BAREMETAL_NETWORK_NAME}" \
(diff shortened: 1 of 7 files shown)


02_configure_host.sh
```diff
@@ -66,6 +69,31 @@ ANSIBLE_FORCE_COLOR=true ansible-playbook \
...
+if [ ${NUM_EXTRA_WORKERS} -ne 0 ]; then
+    ansible-playbook \
+        -e @vm_setup_vars.yml \
+        -e "ironic_prefix=${CLUSTER_NAME}_extra_" \
+        -e "cluster_name=${CLUSTER_NAME}" \
+        -e "provisioning_network_name=${PROVISIONING_NETWORK_NAME}" \
+        -e "baremetal_network_name=${BAREMETAL_NETWORK_NAME}" \
+        -e "working_dir=$WORKING_DIR" \
+        -e "num_masters=0" \
+        -e "num_workers=$NUM_EXTRA_WORKERS" \
```
derekhiggins, line 82: Instead of duplicating this ansible-playbook command is this possible in the first call with  
-e "num_workers=$(( $NUM_WORKERS + $NUM_EXTRA_WORKERS ))"

dhellmann: The `nodes_file` value is different to, because we don't want the extra workers written to the list of nodes that are given to the installer to start.

derekhiggins: > The `nodes_file` value is different too, because we don't want the extra workers written to the list of nodes that are given to the installer to start.

Ack, I've been doing something similar in a local env (using vm-setup/setup-playbook.yml to set up a virt env for baremetal-deploy/ansible-ipi-install), I'm passing a high number of workers into the playbook to create lots of workers but then using a smaller number for the compute replicas in install-config. The result is a X deployed workers and X + Extra workers registered and inspected by ironic. This mightn't suit your use case though.

dhellmann: Either approach may work for me. I just need an extra node and don't want to wait for a 5 node cluster to deploy because it seems to take waaay longer than 3 nodes.

PR #1040, opened 2020-05-27
don't hardcode provisioning ip

honza:
Instead of hardcoding it, we look it up from the install state file. We
also update the operator-sdk command to conform to the latest usage
pattern.

### DIFF
--- a/metal3-dev/run.sh
+++ b/metal3-dev/run.sh
@@ -62,10 +62,10 @@ oc apply -f $OUTDIR/deployment-dev.yaml -n openshift-machine-api
 
 # Set some variables the operator expects to have in order to work
 export OPERATOR_NAME=baremetal-operator
-export DEPLOY_KERNEL_URL=http://172.22.0.3:6180/images/ironic-python-agent.kernel
-export DEPLOY_RAMDISK_URL=http://172.22.0.3:6180/images/ironic-python-agent.initramfs
-export IRONIC_ENDPOINT=http://172.22.0.3:6385/v1/
-export IRONIC_INSPECTOR_ENDPOINT=http://172.22.0.3:5050/v1/
+
+for var in IRONIC_ENDPOINT IRONIC_INSPECTOR_ENDPOINT DEPLOY_KERNEL_URL DEPLOY_RAMDISK_URL; do
+    export "$var"=$(cat $OUTDIR/deployment-full.yaml | yq -r ".spec.template.spec.containers[0].env | map(select( .name == \""${var}"\"))[0].value")
+done
 
 # Wait for the ironic service to be available
 wait_for_json ironic "$IRONIC_ENDPOINT" 300 \


metal3-dev/run.sh
```diff
@@ -13,12 +15,23 @@ fi
...
 source $(dirname $(dirname $0))/utils.sh
 
+if [ ! -f "ocp/ostest/.openshift_install_state.json" ]; then
+    echo "openshift_install_state does not exist, is the cluster up?"
+    exit 1
+fi
+
+PROVISIONING_HOST_IP=$(jq  -r '."*installconfig.InstallConfig".config.platform.baremetal.provisioningHostIP' < ocp/ostest/.openshift_install_state.json)
+
+if [ $IP_STACK == "v6" ]; then
```
hardys, line 25: You could use the wrap_if_ipv6 function from utils.sh for this?

metal3-dev/run.sh
```diff
@@ -1,5 +1,7 @@
 #!/bin/bash -xe
 
+source $(dirname $(dirname $0))/common.sh
```
hardys, line 3: This doesn't work if you cd into metal3-dev then run the script, I think you need something similar to SCRIPTDIR in common.sh e.g `source $(dirname $(dirname $0))/common.sh`

metal3-dev/run.sh
```diff
@@ -13,12 +15,23 @@ fi
 
 source $(dirname $(dirname $0))/utils.sh
 
+if [ ! -f "ocp/ostest/.openshift_install_state.json" ]; then
```
hardys, line 18: Again this won't work if you cd into the directory first, so it should use SCRIPTDIR (either calculated here or in common.sh)

metal3-dev/run.sh
```diff
@@ -13,12 +15,23 @@ fi
...
+    echo "openshift_install_state does not exist, is the cluster up?"
+    exit 1
+fi
+
+PROVISIONING_HOST_IP=$(jq  -r '."*installconfig.InstallConfig".config.platform.baremetal.provisioningHostIP' < ocp/ostest/.openshift_install_state.json)
```
dhellmann, line 23: The path to the state file depends on the cluster name, which isn't always "ostest".

honza:
This probably needs to be rebased on top of #1034

hardys:
> This probably needs to be rebased on top of #1034

Now merged so this can be rebased

metal3-dev/run.sh
```diff
@@ -62,10 +62,10 @@ oc apply -f $OUTDIR/deployment-dev.yaml -n openshift-machine-api
 
 # Set some variables the operator expects to have in order to work
 export OPERATOR_NAME=baremetal-operator
-export DEPLOY_KERNEL_URL=http://172.22.0.3:6180/images/ironic-python-agent.kernel
-export DEPLOY_RAMDISK_URL=http://172.22.0.3:6180/images/ironic-python-agent.initramfs
-export IRONIC_ENDPOINT=http://172.22.0.3:6385/v1/
-export IRONIC_INSPECTOR_ENDPOINT=http://172.22.0.3:5050/v1/
+
+for var in IRONIC_ENDPOINT IRONIC_INSPECTOR_ENDPOINT DEPLOY_KERNEL_URL DEPLOY_RAMDISK_URL; do
+    export "$var"=$(cat deployment-full.yaml | yq -r ".spec.template.spec.containers[0].env | map(select( .name == \""${var}"\"))[0].value")
```
andfasano, line 67: I think it's `$OUTDIR/deployment-full.yaml`

honza: Good catch

PR #1050, opened 2020-06-16
ipv6: Enable dual-stack feature set

russellb:
Dual-stack IPv6 is not yet enabled by default in OpenShift since it
doesn't fully work yet.  We have a feature set that allows you to turn
it on for testing purposes.  When dev-scripts is configured to bring
up a dual-stack test environment, this change adds a manifest to
enable dual-stack in the cluster.

Now that a lot of prerequisite work has started merging, we're ready
to push the dual-stack CI jobs further and this change will ensure
that dual-stack is enabled in those jobs.

### DIFF
--- a/utils.sh
+++ b/utils.sh
@@ -49,6 +49,14 @@ function create_cluster() {
     mkdir -p ${assets_dir}/openshift
     cp -rf assets/generated/*.yaml ${assets_dir}/openshift
 
+    if [[ "${IP_STACK}" == "v4v6" ]]; then
+        # The IPv6DualStack feature is not on by default, because it doesn't
+        # fully work yet.  If we're running a dual-stack test environment,
+        # we will turn on the IPv6DualStackNoUpgrade feature set for testing
+        # purposes.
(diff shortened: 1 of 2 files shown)


russellb:
I tested this locally and the manifest was applied properly.  e2e-metal-ipi will verify that this doesn't break single-stack ipv6 installs

utils.sh
```diff
@@ -49,6 +49,14 @@ function create_cluster() {
     mkdir -p ${assets_dir}/openshift
     cp -rf assets/generated/*.yaml ${assets_dir}/openshift
 
+    if [[ "${IP_STACK}" == "v4v6" ]]; then
+        # The IPv6DualStack feature is not on by default, because it doesn't
+        # fully work yet.  If we're running a dual-stack test environment,
+        # we will turn on the IPv6DualStackNoUpgrade feature set for testing
+        # purposes.
+        cp assets/ipv6-dual-stack-no-upgrade.yaml ${assets_dir}/openshift/.
```
dhellmann, line 57: The trailing `.` looks odd, but doesn't seem to cause any issues.

PR #1056, opened 2020-06-30
add a script to reset all hosts

dhellmann:
Sometimes when working on the BMO it is useful to be able to reset the
status of the hosts in the cluster.

### DIFF
--- a/metal3-dev/reset-hosts.sh
+++ b/metal3-dev/reset-hosts.sh
@@ -1,0 +1,30 @@
+#!/bin/bash
+
+set -ex
+
+SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )/.." && pwd )"
+
+bmo_path=$GOPATH/src/github.com/metal3-io/baremetal-operator
+if [ ! -d $bmo_path ]; then
+    echo "Did not find $bmo_path" 1>&2
+    exit 1
+fi
+
+hosts=$(oc get baremetalhost -n openshift-machine-api -o name | cut -f2 -d/)
+
+(pushd $bmo_path &&
+     go run ./cmd/reset-host/main.go -n openshift-machine-api $hosts)
+
+# The openstack command depends on having OS_CLOUD set. Set it to a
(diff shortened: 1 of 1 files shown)


metal3-dev/reset-hosts.sh
```diff
@@ -0,0 +1,20 @@
...
+    echo "Did not find $bmo_path" 1>&2
+    exit 1
+fi
+
+hosts=$(oc get baremetalhost -n openshift-machine-api -o name | cut -f2 -d/)
+
+(pushd $bmo_path &&
+     go run ./cmd/reset-host/main.go -n openshift-machine-api $hosts)
+
+# NOTE: assumes OS_CLOUD is set correctly
```
hardys, line 16: We could just set it since this will always require OS_CLOUD=metal3 in a dev-scripts environment?

metal3-dev/reset-hosts.sh
```diff
@@ -0,0 +1,20 @@
...
+for host in $hosts; do
+    openstack baremetal node maintenance set $host || continue
+    openstack baremetal node delete $host
```
hardys, line 19: This will only work when the script is run from the dev-scripts root directory, we could make it also work when folks `cd metal3-dev` if we define `SCRIPTDIR` like in `run.sh` and pushd/cd there prior to running the /usr/local/bin/openstack wrapper script.

Also I think we can save some time by moving the node delete out of the loop, since that supports passing multiple nodes (doesn't look like node maintenance set does though so we still need the loop)

metal3-dev/reset-hosts.sh
dhellmann: I'm not always in a dev-scripts environment these days, but I can set the variable if it isn't already set.

metal3-dev/reset-hosts.sh
dhellmann: done

dhellmann:
/hold

we need to merge https://github.com/metal3-io/baremetal-operator/pull/568 before this one

PR #1061, opened 2020-07-10
Enable the ability to block multicast traffic for testing

bcrochet:
In order to properly test unicast VRRP with keepalived, it is desirable to
block multicast on the provisioning host.

This patch does so with ebtables. Additional config will be required in
vm_setup_vars.yml to create external SRV records, since coredns-multicast
will no longer be working.

### DIFF
--- a/02_configure_host.sh
+++ b/02_configure_host.sh
@@ -235,3 +235,19 @@ OPENSTACKCLIENT_PATH="${OPENSTACKCLIENT_PATH:-/usr/local/bin/openstack}"
 if ! command -v openstack | grep -v "${OPENSTACKCLIENT_PATH}"; then
 	  sudo ln -sf "${METAL3_DEV_ENV_PATH}/openstackclient.sh" "${OPENSTACKCLIENT_PATH}"
 fi
+
+# Block Multicast with ebtables
+if [ "$DISABLE_MULTICAST" == "true" ]; then
+    for dst in 224.0.0.251 224.0.0.18; do
+        sudo ebtables -A INPUT --pkttype-type multicast -p ip4 --ip-dst ${dst} -j DROP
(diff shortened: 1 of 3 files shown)


02_configure_host.sh
```diff
@@ -207,3 +207,21 @@ OPENSTACKCLIENT_PATH="${OPENSTACKCLIENT_PATH:-/usr/local/bin/openstack}"
...
+        sudo ebtables -A OUTPUT --pkttype-type multicast -p ip4 --ip-dst ${dst} -j DROP
+    done
+
+    # for dst in ff02::fb ff02::12; do
+    for dst in ff02::fb; do
+        sudo ebtables -A INPUT --pkttype-type multicast -p ip6 --ip6-dst ${dst} -j DROP
+        sudo ebtables -A FORWARD --pkttype-type multicast -p ip6 --ip6-dst ${dst} -j DROP
+        sudo ebtables -A OUTPUT --pkttype-type multicast -p ip6 --ip6-dst ${dst} -j DROP
+    done
+fi
```
hardys, line 226: Should we handle the case where someone tests with DISABLE_MULTICAST=true, then (without rebooting) wants to test again with it enabled?

bcrochet: Sorry... I thought I had the removal in a clean script. Let me add that.

bcrochet: Never mind... host_cleanup.sh has an ebtable flush. That should do the trick.

yboaron:
/lgtm
Tested it with unicast Keepalived deployment - looks OK.

hardys:
Looks good - I wonder if we should have specific chains instead of the unconditional flush?  Perhaps that would be safer and could be done as a followup?

PR #1064, opened 2020-07-13
Further single-stack IPv6 vs dual-stack updates

danwinship:
Followup to #1062. There are a bunch of other places that assumed that if an IPv6 subnet is defined, then we prefer IPv6. Also, two places used `${IP_STACK}` in a way that didn't take the possibility of `"v4v6"` into account.

I'm not sure if [this code](https://github.com/openshift-metal3/dev-scripts/blob/b1e7c89fb5412b4da001c25fce5a5a394bebb635/02_configure_host.sh#L159) is single-stack IPv6-specific or if it works with dual-stack?

Not tested because I don't know how to test it...

/assign @russellb

### DIFF
--- a/02_configure_host.sh
+++ b/02_configure_host.sh
@@ -33,7 +33,7 @@ if [ ! -z "${VM_NODES_FILE}" ]; then
   exit 1
 fi
 
-if [[ "${IP_STACK}" == "v6" ]]; then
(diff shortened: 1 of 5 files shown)


russellb:
these changes lgtm as long as `e2e-metal-ipi` passes.  Thanks!

danwinship:
e2e-metal-ipi passed

PR #1066, opened 2020-07-17
04_setup_ironic: check that checksums exist and always verify them

dtantsur:
Currently we only download and verify checksums only if the image
files do not exist. This poses two problems:
1) Incomplete downloads are not detected on a re-run
2) It's possible to end up with an image but no checksum

### DIFF
--- a/04_setup_ironic.sh
+++ b/04_setup_ironic.sh
@@ -111,18 +111,18 @@ for IMAGE in ${IRONIC_IMAGE} ${VBMC_IMAGE} ${SUSHY_TOOLS_IMAGE} ; do
 done
 
 CACHED_MACHINE_OS_IMAGE="${IRONIC_DATA_DIR}/html/images/${MACHINE_OS_IMAGE_NAME}"
-if [ ! -f "${CACHED_MACHINE_OS_IMAGE}" ]; then
+if [ ! -f "${CACHED_MACHINE_OS_IMAGE}" -o ! -f "${CACHED_MACHINE_OS_IMAGE}.sha256sum" ]; then
   curl -g --insecure -L -o "${CACHED_MACHINE_OS_IMAGE}" "${MACHINE_OS_IMAGE_URL}"
   echo "${MACHINE_OS_IMAGE_SHA256} ${CACHED_MACHINE_OS_IMAGE}" | tee ${CACHED_MACHINE_OS_IMAGE}.sha256sum
-  sha256sum --strict --check ${CACHED_MACHINE_OS_IMAGE}.sha256sum || ( rm -f "${CACHED_MACHINE_OS_IMAGE}" ; exit 1 )
-
 fi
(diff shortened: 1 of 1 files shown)


04_setup_ironic.sh
```diff
@@ -111,18 +111,18 @@ for IMAGE in ${IRONIC_IMAGE} ${VBMC_IMAGE} ${SUSHY_TOOLS_IMAGE} ; do
 done
 
 CACHED_MACHINE_OS_IMAGE="${IRONIC_DATA_DIR}/html/images/${MACHINE_OS_IMAGE_NAME}"
-if [ ! -f "${CACHED_MACHINE_OS_IMAGE}" ]; then
+if [ ! -f "${CACHED_MACHINE_OS_IMAGE}" -o ! -f "${CACHED_MACHINE_OS_IMAGE}.sha256sum" ]; then
   curl -g --insecure -L -o "${CACHED_MACHINE_OS_IMAGE}" "${MACHINE_OS_IMAGE_URL}"
   echo "${MACHINE_OS_IMAGE_SHA256} ${CACHED_MACHINE_OS_IMAGE}" | tee ${CACHED_MACHINE_OS_IMAGE}.sha256sum
```
zaneb, line 116: I think we probably want this to be unconditional too - we should use the latest checksum even if the checksum file does already exist.

04_setup_ironic.sh
```diff
@@ -111,18 +111,18 @@ for IMAGE in ${IRONIC_IMAGE} ${VBMC_IMAGE} ${SUSHY_TOOLS_IMAGE} ; do
 done
 
 CACHED_MACHINE_OS_IMAGE="${IRONIC_DATA_DIR}/html/images/${MACHINE_OS_IMAGE_NAME}"
-if [ ! -f "${CACHED_MACHINE_OS_IMAGE}" ]; then
+if [ ! -f "${CACHED_MACHINE_OS_IMAGE}" -o ! -f "${CACHED_MACHINE_OS_IMAGE}.sha256sum" ]; then
   curl -g --insecure -L -o "${CACHED_MACHINE_OS_IMAGE}" "${MACHINE_OS_IMAGE_URL}"
```
zaneb, line 115: We could add `|| ( rm -f "${CACHED_MACHINE_OS_IMAGE}" ; exit 1 )` on the end here to ensure that the image file gets deleted if the download doesn't complete, rather than going by the non-existence of the checksum file.

hardys: I'm not sure that will necessarily be robust, e.g if the script is killed mid-download we won't hit the error-path `rm`

We could perhaps add `-C` to curl to continue in the event of a partial download, but it's perhaps safest to just remove any existing file and download again, in the event we didn't reach the point of validating/writing the checksum file?

04_setup_ironic.sh
hardys: Yeah true, in theory the filename always changes when there's a new RHCOS build, but when you're locally hacking an image the checksum could change between runs with the same filename

dtantsur:
No time for this, closing.

PR #1067, opened 2020-07-17
update scripts for running local developer versions of bmo and capbm

dhellmann:


hardys:
This all looks good, but I wonder if you can share some more details about your dev workflow (here, or even better in the docs?)

I've typically been testing MAO and CAPBM changes like:

```
export BAREMETAL_MACHINE_CONTROLLERS_LOCAL_IMAGE="https://github.com/openshift/cluster-api-provider-baremetal"
export MACHINE_API_OPERATOR_LOCAL_IMAGE="https://github.com/openshift/machine-api-operator"
```

with my local checkouts picked up by `04_setup_ironic.sh`  from `$HOME/cluster-api-provider-baremetal and `$HOME/machine-api-operator` respectively

This works quite well (and we should probably add it to the docs), but it's inconvenient if you want to do fast dev-iterations on each repo (you can stop the CVO and hack the image configmap, but mostly I'll do unit tests then an e2e test with dev-scripts).

I suspect these changes enable a more streamlined "restart MAO/CAPBM on live cluster" flow, but it would be helpful to share details e.g perhaps we can update https://github.com/openshift-metal3/dev-scripts/blob/master/docs/custom-mao-and-capbm.md or even add a new make target?  `make redeploy mao` or something could be super helpful when doing repetitive testing?

Not blocking on any of that, I just wanted to better understand the intended workflow with these scripts (and I suspect I'm not the only one who's not super-clear on it?)

dhellmann:
Yeah, good point on the documentation. I've updated the instructions for running from source.

You're right about the scripts. They disable the MAO and then create modified versions of the Deployments it creates before running the component from the local git repository. This provides much better turn-around for development and testing for day 2 operations, and I then use CI to test deployments. The approach you describe sounds useful, too, so having that in the documentation would be great.

PR #1068, opened 2020-07-20
Switch to "mixed" mode for VM management

derekhiggins:
Creates some VMs with ipmi and others with redfish, so
dev-scripts will test both paths.

### DIFF
--- a/01_install_requirements.sh
+++ b/01_install_requirements.sh
@@ -17,7 +17,7 @@ if [ -z "${METAL3_DEV_ENV}" ]; then
   # TODO -- come up with a plan for continuously updating this
   # Note we only do this in the case where METAL3_DEV_ENV is
   # unset, to enable developer testing of local checkouts
-  git reset 15260d6313758d29397f2b14ccac084e8bf8bcdd --hard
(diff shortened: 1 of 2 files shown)


dhellmann:
Having a mix of settings seems useful in CI, but like a nightmare if I'm trying to debug something on my development system. Can we separate those 2 use cases and use mixed in CI and pick one mode to use locally as a default?

derekhiggins:
> Having a mix of settings seems useful in CI, but like a nightmare if I'm trying to debug something on my development system. Can we separate those 2 use cases and use mixed in CI and pick one mode to use locally as a default?

I pushed up a new version that only uses "mixed" in CI.

derekhiggins:
Confirmed working, taken from ocp/ostest/baremetalhosts.json in ci
> "address": "redfish+http://[fd2e:6f44:5dd8:c956::1]:8000/redfish/v1/Systems/eaeacc80-494d-4078-9cff-dedb9fe8fd60",
> "address": "ipmi://[fd2e:6f44:5dd8:c956::1]:6234",

PR #1080, opened 2020-08-07
add a script to get the image settings for the active MAO

dhellmann:
Sometimes when working with the machine-api-operator it is not
necessary to provide a custom set of images, and the images file in
the machine-api-operator repository may not match the cluster where
development is happening. This new script extracts the image settings
from the running machine-api-operator so they can be used with a
custom build.

### DIFF
--- a/mao-images.sh
+++ b/mao-images.sh
@@ -1,0 +1,3 @@
+#!/bin/bash
+
+oc get configmap -n openshift-machine-api machine-api-operator-images -o jsonpath="{.data.images\.json}"
--- a/docs/custom-mao-and-capbm.md
+++ b/docs/custom-mao-and-capbm.md
@@ -98,8 +98,12 @@ Edit `custom-images.json` to have a modified image for the BareMetal case:
 
 ### 4) Now run the MAO
 
-Change `custom-images.json` to `pkg/operator/fixtures/images.json` if you
-didn’t build a custom CAPBM.
+Use `mao-images.sh` to extract the images being used by the current
+machine-api-operator if you have not customized the list:
+
+```sh
+./mao-images.sh > custom-images.json
+```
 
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -7,6 +7,7 @@ config_*.sh
 clouds.yaml
 _clouds_yaml
 metal3-dev/deployment-*.yaml
+custom-images.json
```
zaneb, line 10: Seems a bit weird to add this to gitignore when it's not actually part of the script?

mao-images.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+pod=$(oc get pod -n openshift-machine-api | grep '^machine-api-operator-')
```
zaneb, line 3: This will return the whole status line, I think you want something like `oc get pod -n openshift-machine-api -o name | grep '^pod/machine-api-operator-'`

mao-images.sh
```diff
@@ -0,0 +1,10 @@
...
+
+if [[ -z "$pod" ]]; then
+    echo "ERROR: No machine-api-operator pod found in openshift-machine-api namespace" 1>&2
+    exit 1
+fi
+
+oc exec pod/$pod -n openshift-machine-api -c machine-api-operator -- cat /etc/machine-api-operator-config/images/images.json
```
zaneb, line 10: One downside of this is that once you scale the MAO down, it doesn't work any more. An alternative would be to do:

```
oc get configmap -n openshift-machine-api machine-api-operator-images -o jsonpath="{.data.images\.json}"
```

That gets the content of the configmap provided by the CVO directly (rather than reading it from where it is mounted within the pod).

dhellmann: Good point, done.

.gitignore
dhellmann: Removed

PR #1083, opened 2020-08-17
Fix COPY instruction for custom repositories file

elfosardo:
The paths of files and directories in the COPY instructions inside
the Dockerfile are interpreted as relative to the source of the
context of the build.
Therefore, the custom repository file must be created inside the
base-iamge directory or the build will fail.


hardys:
@bfournie FYI I think anyone can add the /lgtm label (IIRC the OWNERS file only controls the /approve)

/lgtm

PR #1085, opened 2020-08-19
Create all workers with a single ansible run

hardys:
Currently we invoke the ansible roles twice, which results
in missing DHCP reservations in the ostestbm network configuration,
because that network already exists when the second run takes place.

That means on scale out any workers defined via NUM_EXTRA_WORKERS will
fail to join the cluster, since they register as "localhost", but the
machine hostname is "localhost.localdomain", which the machine-approver
rejects.

### DIFF
--- a/02_configure_host.sh
+++ b/02_configure_host.sh
@@ -55,7 +55,7 @@ ansible-playbook \
     -e "baremetal_network_name=${BAREMETAL_NETWORK_NAME}" \
     -e "working_dir=$WORKING_DIR" \
     -e "num_masters=$NUM_MASTERS" \
-    -e "num_workers=$NUM_WORKERS" \
+    -e "num_workers=$((NUM_WORKERS + NUM_EXTRA_WORKERS))" \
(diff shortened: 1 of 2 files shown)


hardys:
I've not yet tested scale-out with this, but we see the reservations correctly now (with NUM_WORKERS AND NUM_EXTRA_WORKERS both set to 2):

```
$ cat /var/lib/libvirt/dnsmasq/ostestbm.hostsfile 
00:54:d7:08:f0:e5,192.168.111.20,master-0
00:54:d7:08:f0:e9,192.168.111.21,master-1
00:54:d7:08:f0:ed,192.168.111.22,master-2
00:54:d7:08:f0:f1,192.168.111.23,worker-0
00:54:d7:08:f0:f5,192.168.111.24,worker-1
00:54:d7:08:f0:f9,192.168.111.25,worker-2
00:54:d7:08:f0:fd,192.168.111.26,worker-3

```

honza:
I just tested a scale out, and it worked great.

PR #1086, opened 2020-08-24
use ROOT_DISK_NAME to set rootDeviceHints

dhellmann:
Always set the rootDeviceHints as a way to test and demonstrate the
feature.


derekhiggins:
We should probably only set this if MASTER_HARDWARE_PROFILE isn't set as it would override anybody who is using profiles.

dhellmann:
> We should probably only set this if MASTER_HARDWARE_PROFILE isn't set as it would override anybody who is using profiles.

I updated it to check both profile variables and to emit warnings if those are used instead of the device name variable.

PR #1087, opened 2020-08-24
preserve the generated assets for debugging in case the cluster fails

dhellmann:
I've found it useful to be able to examine the raw YAML for the generated
assets to understand why a cluster doesn't always come up (especially
when the host resources are invalid).

### DIFF
--- a/utils.sh
+++ b/utils.sh
@@ -72,6 +72,12 @@ function create_cluster() {
       mv ${assets_dir}/worker.ign ${assets_dir}/worker.ign.orig
       jq -s '.[0] * .[1]' ${IGNITION_EXTRA} ${assets_dir}/worker.ign.orig | tee ${assets_dir}/worker.ign
     fi
+
+    # Preserve the assets for debugging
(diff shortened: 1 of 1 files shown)


utils.sh
```diff
@@ -72,6 +72,11 @@ function create_cluster() {
       mv ${assets_dir}/worker.ign ${assets_dir}/worker.ign.orig
       jq -s '.[0] * .[1]' ${IGNITION_EXTRA} ${assets_dir}/worker.ign.orig | tee ${assets_dir}/worker.ign
     fi
+
+    # Preserve the assets for debugging
+    mkdir -p "${assets_dir}/saved-assets"
+    cp -av "${assets_dir}/openshift/" "${assets_dir}/saved-assets"
```
hardys, line 78: There is also the `${assets_dir}/manifests` directory which I think also gets removed after `create cluster` ?

dhellmann: Added

zaneb:
/retitle preserve the generated assets for debugging in case the cluster fails
inspection is overloaded ;)
/lgtm

PR #1089, opened 2020-08-25
metal3-dev: Copy ironic creds for local bmo

zaneb:


metal3-dev/local-bmo.sh
```diff
@@ -71,6 +71,26 @@ for var in IRONIC_ENDPOINT IRONIC_INSPECTOR_ENDPOINT DEPLOY_KERNEL_URL DEPLOY_RA
...
+get_creds() {
+    local svc=$1
+    local cred_dir="${auth_dir}/${svc}"
+    local secret="metal3-${svc}-password"
+    if oc get -n openshift-machine-api secret ${secret} -o name >/dev/null; then
+        if [ ! -d ${auth_dir} ]; then
+            sudo mkdir -p "${auth_dir}"
+            sudo chown -R $USER:$USER "${auth_dir}/.."
+        fi
+        mkdir "${cred_dir}"
```
andfasano, line 84: `-p` requires also here

zaneb: No it isn't.

andfasano: Not sure if I'm missing something, but when I ran it twice (from a clean env, without `/opt/metal3`) the first time it's ok but the second fails with a `mkdir: cannot create directory ‘/opt/metal3/auth/ironic’: File exists`

zaneb: Ah, OK. The parent directory always exists, but `-p` also suppresses the error if the cred_dir itself exists. I'll fix it up, thanks.

PR #1090, opened 2020-08-27
Install the baremetal command along with openstackclient

derekhiggins:
metal3-dev-env already does this, just bringing parity.


iurygregory (approved):

andfasano:
I think you're referring to https://github.com/metal3-io/metal3-dev-env/blob/1cba74fd70d3dc229ec432afbefe7a7f779c1419/02_configure_host.sh#L231

iurygregory:
> @derekhiggins I think you're referring to https://github.com/metal3-io/metal3-dev-env/blob/1cba74fd70d3dc229ec432afbefe7a7f779c1419/02_configure_host.sh#L231

@andfasano yeah he is referring to that, at the moment we don't have it on the dev-scripts here and we found some problems using the openstack cli while running on `http_basic` so it's good to have the baremetal cli available

PR #1092, opened 2020-09-01
Add support to generate clouds.yaml with http_basic credentials

iurygregory:
This PR aims to provide a generic way to generate the `clouds.yaml`, in 4.6 we enabled the `http_basic` for ironic an ironic-inspector.

- metal3-templater now can generate specific config for the bootstrap, and for ironic and ironic-inspector according to the auth_type.
- clouds.yaml.template was updated to support the structure for `http_basic`
- utils.sh was update to generate the correct `clouds.yaml` according to the OCP version.

### DIFF
--- a/clouds.yaml.template
+++ b/clouds.yaml.template
@@ -1,9 +1,24 @@
 clouds:
+{{- if eq .AuthType "none"}}
   metal3-bootstrap:
-    auth_type: none
+    auth_type: {{.AuthType}}
     baremetal_endpoint_override: {{.BootstrapIronicURL}}
     baremetal_introspection_endpoint_override: {{.BootstrapInspectorURL}}
   metal3:
-    auth_type: none
+    auth_type: {{.AuthType}}
     baremetal_endpoint_override: {{.ClusterIronicURL}}
     baremetal_introspection_endpoint_override: {{.ClusterInspectorURL}}
+{{- else if eq .AuthType "http_basic"}}
+  metal3:
+    auth_type: {{.AuthType}}
+    auth:
+      username: {{.IronicUser}}
+      password: {{.IronicPassword}}
+    baremetal_endpoint_override: {{.ClusterIronicURL}}
+  metal3-inspector:
+    auth_type: {{.AuthType}}
+    auth:
+      username: {{.InspectorUser}}
+      password: {{.InspectorPassword}}
+    baremetal_introspection_endpoint_override: {{.ClusterInspectorURL}}
+{{- end}}
--- a/metal3-templater.go
+++ b/metal3-templater.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"flag"
 	"fmt"
 	"github.com/apparentlymart/go-cidr/cidr"
 	"github.com/openshift/installer/pkg/ipnet"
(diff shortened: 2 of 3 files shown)


metal3-templater.go
```diff
@@ -64,6 +70,14 @@ func main() {
 	templateData.ClusterIronicURL = fmt.Sprintf("http://%s", net.JoinHostPort(ip, "6385"))
 	templateData.ClusterInspectorURL = fmt.Sprintf("http://%s", net.JoinHostPort(ip, "5050"))
 
+	// http_basic credentials
+	if numArgs == 11 {
+		templateData.IronicUser = fmt.Sprintf("%s", os.Args[7])
```
dhellmann, line 75: I don't think the `Sprintf()` calls are needed here, are they? Can't we just assign the array element to the field of the template struct?

utils.sh
```diff
@@ -73,6 +73,11 @@ function create_cluster() {
       jq -s '.[0] * .[1]' ${IGNITION_EXTRA} ${assets_dir}/worker.ign.orig | tee ${assets_dir}/worker.ign
     fi
     $OPENSHIFT_INSTALLER --dir "${assets_dir}" --log-level=debug create cluster
+
+    OCP_VERSIONS_NOAUTH="4.3 4.4 4.5"
```
dhellmann, line 77: We're checking versions here and on line 246. Could we move all of this into generate_auth_template so that it does the right thing regardless of the version?

metal3-templater.go
```diff
@@ -23,13 +23,19 @@ type templater struct {
...
+	InspectorPassword     string
 }
 
 func main() {
 	var templateData templater
-
-	if len(os.Args) < 7 {
-		fmt.Printf("usage: <prog> TEMPLATE_FILE INTERFACE NETWORK BOOTSTRAP_IP CLUSTER_IP IMAGE_URL\n")
+	numArgs := len(os.Args)
+	if numArgs == 7 && numArgs == 11  {
```
zaneb, line 35: It seems unlikely that this expression will ever be true.

metal3-templater.go
```diff
@@ -64,6 +70,14 @@ func main() {
 	templateData.ClusterIronicURL = fmt.Sprintf("http://%s", net.JoinHostPort(ip, "6385"))
 	templateData.ClusterInspectorURL = fmt.Sprintf("http://%s", net.JoinHostPort(ip, "5050"))
 
+	// http_basic credentials
+	if numArgs == 11 {
```
zaneb, line 74: Counting numbers of arguments doesn't really seem like a sustainable approach to extending tools like this. We need to be able to specify what the arguments mean. This approach does not give us the flexibility to change to a different auth method (upstream is probably going to implement mTLS auth at some point). The [flag](https://golang.org/pkg/flag/) package in the standard library looks like it provides a relatively simple way to implement proper CLI parsing.

metal3-templater.go
```diff
@@ -64,6 +70,14 @@ func main() {
...
+		templateData.IronicUser = fmt.Sprintf("%s", os.Args[7])
+		templateData.IronicPassword = fmt.Sprintf("%s", os.Args[8])
```
zaneb, line 76: I'd suggest combining username and password into a single argument as `<username>:<password>`.

utils.sh
```diff
@@ -238,7 +243,24 @@ function generate_templates {
     fi
 
     # clouds.yaml
-    go run metal3-templater.go clouds.yaml.template "$CLUSTER_PRO_IF" "$PROVISIONING_NETWORK" "$MACHINE_OS_IMAGE_URL" "$BOOTSTRAP_PROVISIONING_IP" "$CLUSTER_PROVISIONING_IP" > clouds.yaml
+    if [[ "$OPENSHIFT_VERSION" != "4.6" ]]; then
```
zaneb, line 246: This is going to break as soon as 4.7 opens.

utils.sh
iurygregory: the templates are different for old versions (I think 4.3 4.4 and 4.5 doesn't support http_basic) so we just generate the normal template on generate_templates, I also added the new template so it will generate the one to use the http_basic and it will be only call if the versions is not  in the OCP_VERSIONS_NOAUTH

utils.sh
```diff
@@ -238,7 +243,24 @@ function generate_templates {
...
+      # which mounts a config dir into the ironic-client container
+      mkdir -p _clouds_yaml
+      cp clouds.yaml _clouds_yaml
+    fi
+}
+
+function generate_auth_template {
+    MACHINE_OS_IMAGE_URL="http:///$(wrap_if_ipv6 $MIRROR_IP)/images/${MACHINE_OS_IMAGE_NAME}?sha256=${MACHINE_OS_BOOTSTRAP_IMAGE_SHA256}"
+    # Get the auth information for ironic and ironic-inspector
+    export KUBECONFIG="${SCRIPTDIR}/ocp/${CLUSTER_NAME}/auth/kubeconfig"
```
zaneb, line 257: This should be set up by whatever is calling this function; imho this is not something we should be exporting from inside random function calls.
If it's not set already at the time this is called then we should just pass it as a CLI argument rather than changing the environment.

utils.sh
```diff
@@ -238,7 +243,24 @@ function generate_templates {
     fi
 
     # clouds.yaml
-    go run metal3-templater.go clouds.yaml.template "$CLUSTER_PRO_IF" "$PROVISIONING_NETWORK" "$MACHINE_OS_IMAGE_URL" "$BOOTSTRAP_PROVISIONING_IP" "$CLUSTER_PROVISIONING_IP" > clouds.yaml
+    if [[ "$OPENSHIFT_VERSION" != "4.6" ]]; then
+      go run metal3-templater.go clouds.yaml.template "$CLUSTER_PRO_IF" "$PROVISIONING_NETWORK" "$MACHINE_OS_IMAGE_URL" "$BOOTSTRAP_PROVISIONING_IP" "$CLUSTER_PROVISIONING_IP" > clouds.yaml      # For compatibility with metal3-dev-env openstackclient.sh
```
zaneb, line 247: Missing line break before the comment.

utils.sh
dhellmann: Sure, the templates are different. But we're handling templates for those different versions in different places.

metal3-templater.go
iurygregory: sorry it should be an OR... I thought I changed..

utils.sh
zaneb: > the templates are different for old versions

That seems like something that could be templatable :)

iurygregory: it's running in different places because for noauth we don't need oc cli available to get the username password, maybe it would be fine to generate everything in the end...

metal3-templater.go
iurygregory: Done

metal3-templater.go
iurygregory: not sure if it would be worth having them in a single argument and than we need to split..

utils.sh
iurygregory: Fixed

utils.sh
iurygregory: I'm going to test without setting the KUBECONFIG and see how it goes

utils.sh
iurygregory: Done

metal3-templater.go
iurygregory: I've update to use flag, going to do some tests and update the PR

clouds_auth.yaml.template
```diff
@@ -0,0 +1,13 @@
+clouds:
```
derekhiggins, line 1: Can we add metal3-bootstrap to this file also?

derekhiggins: Actually scratch that, stupid idea

derekhiggins:
Some comments above to be addressed but just confirming the resulting file works

```
[derekh@r640-u07 dev-scripts]$ export OS_CLOUD=metal3
[derekh@r640-u07 dev-scripts]$ baremetal node list
+++ echo v0.3.2
+++ echo v0.3.9
Management cluster forced to be minikube when container runtime is not docker
+--------------------------------------+-----------------+--------------------------------------+-------------+--------------------+-------------+
| UUID                                 | Name            | Instance UUID                        | Power State | Provisioning State | Maintenance |
+--------------------------------------+-----------------+--------------------------------------+-------------+--------------------+-------------+
| 023cdbb8-78d2-49f0-a624-01dd6dcf8c1d | ostest-worker-1 | e5d66ec9-266e-4f55-b0b5-e9dfdbcc178f | power on    | active             | False       |
| e6e69e42-9558-484e-ac6c-d67eef3cec22 | ostest-master-0 | b748c0aa-6be9-4e03-b033-e34496224c25 | power on    | active             | False       |
| d08eae55-56bb-4967-b33f-8ba0323fe017 | ostest-master-1 | 5ecb93a6-abb0-41c4-a4b6-eb8209c88584 | power on    | active             | False       |
| dfe8f94f-daff-429a-ae67-e92e8b4b5063 | ostest-master-2 | daab16ad-8ec0-42e1-b24c-81306bafdf20 | power on    | active             | False       |
| a53a5594-538d-47d6-bf5a-cd2493f52520 | ostest-worker-0 | fe40fdac-ed53-4d44-9197-bfb89db4f6bb | power on    | active             | False       |
+--------------------------------------+-----------------+--------------------------------------+-------------+--------------------+-------------+
[derekh@r640-u07 dev-scripts]$ export OS_CLOUD=metal3-inspector
[derekh@r640-u07 dev-scripts]$ baremetal introspection list
+++ echo v0.3.2
+++ echo v0.3.9
Management cluster forced to be minikube when container runtime is not docker
+--------------------------------------+----------------------------+----------------------------+-------+
| UUID                                 | Started at                 | Finished at                | Error |
+--------------------------------------+----------------------------+----------------------------+-------+
| a53a5594-538d-47d6-bf5a-cd2493f52520 | 2020-09-02T10:32:15.429763 | 2020-09-02T10:34:27.717066 | None  |
| 023cdbb8-78d2-49f0-a624-01dd6dcf8c1d | 2020-09-02T10:32:13.450488 | 2020-09-02T10:34:26.571526 | None  |
+--------------------------------------+----------------------------+----------------------------+-------+

```

clouds_auth.yaml.template
iurygregory: @derekhiggins I was wondering about it, any ideas if the bootstrap is deployed using `http_basic` or `noauth` ?

iurygregory:
@dhellmann @zaneb  I'm finishing my tests to generate the configs for the `clouds.yaml`, I've added two flags to the `metal3-templater.go` that represents the auth and I'm trying to using a single file for template, atm it looks like: 
```
[stack@host10 dev-scripts]$ cat clouds.yaml.template 
clouds:
{{- if eq .AuthType "none"}}
  metal3-bootstrap:
    auth_type: {{.AuthType}}
    baremetal_endpoint_override: {{.BootstrapIronicURL}}
    baremetal_introspection_endpoint_override: {{.BootstrapInspectorURL}}
  metal3:
    auth_type: {{.AuthType}}
    baremetal_endpoint_override: {{.ClusterIronicURL}}
    baremetal_introspection_endpoint_override: {{.ClusterInspectorURL}}
{{- else if eq .AuthType "htpp_basic"}}
  metal3:
    auth_type: {{.AuthType}}
    auth:
      username: {{.IronicUser}}
      password: {{.IronicPassword}}
    baremetal_endpoint_override: {{.ClusterIronicURL}}
  metal3-inspector:
    auth_type: {{.AuthType}}
    auth:
      username: {{.InspectorUser}}
      password: {{.InspectorPassword}}
    baremetal_introspection_endpoint_override: {{.ClusterInspectorURL}}
{{- end}}
```

The `metal3-templater.go` would run:
`go run metal3-templater.go clouds.yaml.template "noauth" "$CLUSTER_PRO_IF" "$PROVISIONING_NETWORK" "$MACHINE_OS_IMAGE_URL" "$BOOTSTRAP_PROVISIONING_IP" "$CLUSTER_PROVISIONING_IP" > clouds.yaml`
or
`go run metal3-templater.go clouds_auth.yaml.template "http_basic" "$CLUSTER_PRO_IF" "$PROVISIONING_NETWORK" "$MACHINE_OS_IMAGE_URL" "$BOOTSTRAP_PROVISIONING_IP" "$CLUSTER_PROVISIONING_IP" "$IRONIC_USER" "$IRONIC_PASSWORD" "$INSPECTOR_USER" "$INSPECTOR_PASSWORD" > clouds.yaml`

This would be ok?

clouds_auth.yaml.template
derekhiggins: Its noath but the reason I said its a stupid idea is because its gone by the time this runs....

metal3-templater.go
zaneb: You always have to pass username and password together, so it makes sense to have a single argument like e.g. `--ironic-basic-auth=<user>:<pass>`.
The semantics are well-defined because it's traditional to pass basic auth creds in a url like `http://<user>:<pass>@<host>/<path>`, and the only reason we're not doing that here is that we're passing only the cluster IP and constructing the URLs ourselves because the IP is used in multiple places.

zaneb:
There's a typo (`htpp_basic`).

Also, it doesn't make sense to not provide a metal3-inspector cloud in the noauth case, thus guaranteeing that you have to use different commands for inspector depending on whether auth is enabled. (It probably does still make sense to include both endpoints in the 'metal3' cloud in the noauth case.)

iurygregory:
> etal3-inspector cloud in the noauth case, thus guaranteeing that you have to use different commands for inspector depending on whether auth is enabled. (It probably does still make sense to include bot

For noauth we don't need

metal3-templater.go
iurygregory: this would be for the CLI, but we need to have different variables or we will need to add things in the `clouds.yaml.template` to split `<user>:<pass>` and put in the correct field no?

zaneb: Yeah, I meant combine them in the CLI and split them in code here.

iurygregory: I've updated the code and I will let run to see how it goes using <user>:<password>, tomorrow I will push a new commit in this PR

utils.sh
dhellmann: Yes, I think keeping all of it together in one place will make it easier to follow and maintain later.

utils.sh
dhellmann: `common.sh` already sets `KUBECONFIG`.

iurygregory:
Can someone add the `ok to test`? =)

clouds.yaml.template
```diff
@@ -1,9 +1,24 @@
...
+    auth_type: {{.AuthType}}
     baremetal_endpoint_override: {{.BootstrapIronicURL}}
     baremetal_introspection_endpoint_override: {{.BootstrapInspectorURL}}
   metal3:
-    auth_type: none
+    auth_type: {{.AuthType}}
     baremetal_endpoint_override: {{.ClusterIronicURL}}
     baremetal_introspection_endpoint_override: {{.ClusterInspectorURL}}
+{{- else if eq .AuthType "http_basic"}}
+  metal3:
```
andfasano, line 12: Why `metal3-bootstrap` has been removed?

iurygregory: since I moved the `generate_template` to be at the end of `create_cluster`  the bootstrap will be gone by that time, the reason for this is that the `clouds.yaml` for `http_basic` requires the metal3 pod to be up so we can get the username/password for ironic and inspector

andfasano: Ok, so how it will affect the old procedure for debugging bootstrap VM, as documented at https://github.com/openshift-metal3/dev-scripts/#interacting-with-ironic-directly?

iurygregory: I'm going to update the metal3-templater.go  adding an option to only generate the bootstrap info (so it can be called at the begin of `create_cluster` ) I think this will solve our problems. Will try to test locally and I will update the PR

iurygregory:
@andfasano now you should have the `metal3-bootstrap` cloud available =)

clouds.yaml.template
```diff
@@ -1,9 +1,24 @@
...
+    auth_type: {{.AuthType}}
+    auth:
```
dhellmann, line 14: It seems like the only part that's different between the 2 cases is the fact that for http_basic this auth section is added, so the template could be simplified to avoid repeating the rest of the settings.

metal3-templater.go
```diff
@@ -19,56 +20,122 @@ type templater struct {
...
 
-	templateFile := os.Args[1]
+	var auth = os.Args[1]
+	var templateData templater
+	cmd := make([]string, 6)
 
-	templateData.ProvisioningInterface = os.Args[2]
+	switch auth {
+	case "bootstrap":
+		if len(os.Args) < 4 {
```
dhellmann, line 52: Is there some reason not to define flags for all of the inputs to avoid checking the length of the arguments?

iurygregory: Tbh I didn't think about having flag for each argument, I will do some tests

clouds.yaml.template
iurygregory: not only the `auth` section, we "split" the `metal3` cloud into `metal3` and `metal3-inspector`, in the noauth scenario we have the baremetal_introspection_endpoint_override set for `metal3` .
I will try to come up with a more simplified version

iurygregory:
```
clouds:
  metal3-bootstrap:
    auth_type: none
    baremetal_endpoint_override: {{.BootstrapIronicURL}}
    baremetal_introspection_endpoint_override: {{.BootstrapInspectorURL}}
  metal3:
    auth_type: {{.AuthType}}
{{- if eq .AuthType "http_basic"}}
    auth:
      username: {{.IronicUser}}
      password: {{.IronicPassword}}
    baremetal_endpoint_override: {{.ClusterIronicURL}}
  metal3-inspector:
    auth_type: {{.AuthType}}
    auth:
      username: {{.InspectorUser}}
      password: {{.InspectorPassword}}
    baremetal_introspection_endpoint_override: {{.ClusterInspectorURL}}
{{- end}}
{{- if eq .AuthType "none"}}
    baremetal_endpoint_override: {{.ClusterIronicURL}}
    baremetal_introspection_endpoint_override: {{.ClusterInspectorURL}}
{{- end}}
```

Something like this @dhellmann ?

dhellmann: How about 

```
clouds:
  metal3-bootstrap:
    auth_type: none
    baremetal_endpoint_override: {{.BootstrapIronicURL}}
    baremetal_introspection_endpoint_override: {{.BootstrapInspectorURL}}
  metal3:
    baremetal_endpoint_override: {{.ClusterIronicURL}}
    auth_type: {{.AuthType}}
{{- if eq .AuthType "http_basic"}}
    auth:
      username: {{.IronicUser}}
      password: {{.IronicPassword}}
{{- end}}
  metal3-inspector:
    baremetal_introspection_endpoint_override: {{.ClusterInspectorURL}}
    auth_type: {{.AuthType}}
{{- if eq .AuthType "http_basic"}}
    auth:
      username: {{.IronicUser}}
      password: {{.IronicPassword}}
{{- end}}
```

Do we need credentials for the `metal3-bootstrap` entry?

dhellmann: Ah, I couldn't tell from the template that your goal was to have 2 different implementations for the different auth types. It might be clearer to make each if statement contain a complete set of the values so the reader doesn't have to parse the template to understand it, like this:

```
clouds:
  metal3-bootstrap:
    auth_type: none
    baremetal_endpoint_override: {{.BootstrapIronicURL}}
    baremetal_introspection_endpoint_override: {{.BootstrapInspectorURL}}
{{- if eq .AuthType "none"}}
  metal3:
    auth_type: {{.AuthType}}
    baremetal_endpoint_override: {{.ClusterIronicURL}}
    baremetal_introspection_endpoint_override: {{.ClusterInspectorURL}}
{{- end}}
{{- if eq .AuthType "http_basic"}}
  metal3:
    auth_type: {{.AuthType}}
    auth:
      username: {{.IronicUser}}
      password: {{.IronicPassword}}
    baremetal_endpoint_override: {{.ClusterIronicURL}}
  metal3-inspector:
    auth_type: {{.AuthType}}
    auth:
      username: {{.InspectorUser}}
      password: {{.InspectorPassword}}
    baremetal_introspection_endpoint_override: {{.ClusterInspectorURL}}
{{- end}}
```

And if there is a comment syntax for the template, we could use that to highlight what's going on.

iurygregory: ack!

metal3-templater.go
iurygregory: @dhellmann I've updated adding arguments to each parameter we have.

iurygregory:
CI seems happy with the changes =)

dhellmann:
/approve

This looks OK to me. @zaneb had a comment about always including both cloud definitions, so that a user would always use the same command to access the inspector API. I think that makes sense, from a usability perspective, but having something giving us the passwords automatically in any form is going to help in the short term so I'm willing to let that refinement come in another PR.

clouds.yaml.template
zaneb: > for `metal3-bootstrap` afaik we don't need.

Not yet, but it'll be the next thing.

zaneb:
> @zaneb had a comment about always including both cloud definitions, so that a user would always use the same command to access the inspector API. I think that makes sense, from a usability perspective, but having something giving us the passwords automatically in any form is going to help in the short term so I'm willing to let that refinement come in another PR.

#1111

PR #1093, opened 2020-09-02
fetch a ci token as needed for the pull secret

dhellmann:
The user-specific pull secret granted by logging in to
api.ci.openshift.org expires. Rather than forcing users to manually
get a new one, use the login command to get it ourselves and combine
it with the more static parts of the pull secret.

This change deprecates the PULL_SECRET variable in favor of
PERSONAL_PULL_SECRET and CI_TOKEN variables.

### DIFF
--- a/04_setup_ironic.sh
+++ b/04_setup_ironic.sh
@@ -19,6 +19,7 @@ rm -f assets/templates/99_local-registry.yaml $OPENSHIFT_INSTALL_PATH/data/data/
 
 # Various commands here need the Pull Secret in a file
 export REGISTRY_AUTH_FILE=$(mktemp --tmpdir "pullsecret--XXXXXXXXXX")
+build_pull_secret
 _tmpfiles=$REGISTRY_AUTH_FILE
 { echo "${PULL_SECRET}" ; } 2> /dev/null > $REGISTRY_AUTH_FILE
 
--- a/04_setup_ironic.sh
+++ b/04_setup_ironic.sh
@@ -17,16 +17,7 @@ source utils.sh
 #export BAREMETAL_OPERATOR_LOCAL_IMAGE=192.168.111.1:5000/localimages/bmo:latest
 rm -f assets/templates/99_local-registry.yaml $OPENSHIFT_INSTALL_PATH/data/data/bootstrap/baremetal/files/etc/containers/registries.conf
 
-# Various commands here need the Pull Secret in a file
-export REGISTRY_AUTH_FILE=$(mktemp --tmpdir "pullsecret--XXXXXXXXXX")
-build_pull_secret
-_tmpfiles=$REGISTRY_AUTH_FILE
-{ echo "${PULL_SECRET}" ; } 2> /dev/null > $REGISTRY_AUTH_FILE
-
-# Combine pull-secret with registry's password
-COMBINED_AUTH_FILE=$(mktemp --tmpdir "combined-pullsecret--XXXXXXXXXX")
-_tmpfiles="$_tmpfiles $COMBINED_AUTH_FILE"
-jq -s '.[0] * .[1]' ${REGISTRY_AUTH_FILE} ${REGISTRY_CREDS} > ${COMBINED_AUTH_FILE}
+write_pull_secret
 
 DOCKERFILE=$(mktemp --tmpdir "release-update--XXXXXXXXXX")
 _tmpfiles="$_tmpfiles $DOCKERFILE"
@@ -48,7 +39,7 @@ fi
 for IMAGE_VAR in $(env | grep "_LOCAL_IMAGE=" | grep -o "^[^=]*") ; do
     IMAGE=${!IMAGE_VAR}
 
-    sudo -E podman pull --authfile $COMBINED_AUTH_FILE $OPENSHIFT_RELEASE_IMAGE
+    sudo -E podman pull --authfile $PULL_SECRET_FILE $OPENSHIFT_RELEASE_IMAGE
 
     # Is it a git repo?
     if [[ "$IMAGE" =~ "://" ]] ; then
@@ -67,9 +58,9 @@ for IMAGE_VAR in $(env | grep "_LOCAL_IMAGE=" | grep -o "^[^=]*") ; do
         if [[ -n ${BASE_IMAGE_DIR:-} ]]; then
             sed -i "s/^FROM [^ ]*/FROM ${BASE_IMAGE_DIR}/g" ${IMAGE_DOCKERFILE}
         fi
-        sudo podman build --authfile $COMBINED_AUTH_FILE -t ${!IMAGE_VAR} -f $IMAGE_DOCKERFILE .
+        sudo podman build --authfile $PULL_SECRET_FILE -t ${!IMAGE_VAR} -f $IMAGE_DOCKERFILE .
         cd -
-        sudo podman push --tls-verify=false --authfile $COMBINED_AUTH_FILE ${!IMAGE_VAR} ${!IMAGE_VAR}
+        sudo podman push --tls-verify=false --authfile $PULL_SECRET_FILE ${!IMAGE_VAR} ${!IMAGE_VAR}
     fi
 
     IMAGE_NAME=$(echo ${IMAGE_VAR/_LOCAL_IMAGE} | tr '[:upper:]_' '[:lower:]-')
@@ -88,7 +79,7 @@ if [ ! -z "${MIRROR_IMAGES}" ]; then
 
     oc adm release mirror \
        --insecure=true \
-        -a ${COMBINED_AUTH_FILE}  \
+        -a ${PULL_SECRET_FILE}  \
         --from ${OPENSHIFT_RELEASE_IMAGE} \
         --to-release-image ${LOCAL_REGISTRY_DNS_NAME}:${LOCAL_REGISTRY_PORT}/localimages/local-release-image:${OPENSHIFT_RELEASE_TAG} \
         --to ${LOCAL_REGISTRY_DNS_NAME}:${LOCAL_REGISTRY_PORT}/localimages/local-release-image 2>&1 | tee ${MIRROR_LOG_FILE}
@@ -97,7 +88,7 @@ if [ ! -z "${MIRROR_IMAGES}" ]; then
     #To ensure that you use the correct images for the version of OpenShift Container Platform that you selected,
     #you must extract the installation program from the mirrored content:
     if [ -z "$KNI_INSTALL_FROM_GIT" ]; then
-      oc adm release extract --registry-config "${COMBINED_AUTH_FILE}" \
+      oc adm release extract --registry-config "${PULL_SECRET_FILE}" \
         --command=openshift-baremetal-install --to "${EXTRACT_DIR}" \
         "${LOCAL_REGISTRY_DNS_NAME}:${LOCAL_REGISTRY_PORT}/localimages/local-release-image:${OPENSHIFT_RELEASE_TAG}"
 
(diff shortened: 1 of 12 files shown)


config_example.sh
```diff
@@ -2,7 +2,14 @@
 
 # You can get this secret from https://cloud.redhat.com/openshift/install/pull-secret
 set +x
-export PULL_SECRET=''
+export PERSONAL_SECRET=''
```
zaneb, line 5: Since we're changing this anyway, it would be preferable if this were just a path to a file. Keeping it inline makes the config file hard to edit, and complicates stuff like using Ansible to set up a host. I realise people can just do `PERSONAL_SECRET="$(cat secret.json)"`, but there's never any real reason for it to be inline. We could even provide a default file name and ignore it in .gitignore. Would certainly avoid all the +x/-x stuff as well.

utils.sh
```diff
@@ -380,18 +380,43 @@ EOF
...
+    # Get a current token for registry.svc.ci.openshift.org
+    oc login https://api.ci.openshift.org --kubeconfig=$tmpkubeconfig --token=${CI_TOKEN}
+    token=$(oc registry login --kubeconfig=$tmpkubeconfig --to=- \
+                | jq -r '.auths["registry.svc.ci.openshift.org"].auth')
+
+    # Build the full pull secret
+    export PULL_SECRET=$(echo "${PERSONAL_PULL_SECRET}" \
+        | jq --arg token "$token" \
+             '.auths + {"registry.svc.ci.openshift.org": {"auth": $token}}' \
+           | jq '{"auths": .}')
```
zaneb, line 417: Arguably simpler:

```
export PULL_SECRET=$({ echo "$PERSONAL_PULL_SECRET"; oc registry login --kubeconfig=$tmpkubeconfig --to=-; } | jq -n 'reduce inputs as $in ({}; . * $in)')
```

or if the personal pull secret is in a file:

```
export PULL_SECRET=$(oc registry login --kubeconfig=$tmpkubeconfig --to=- | jq -n 'reduce inputs as $in ({}; . * $in)' ${PERSONAL_PULL_SECRET_FILE} -)
```

utils.sh
```diff
@@ -380,18 +380,43 @@ EOF
...
-      echo "Please get a valid pull secret for cloud.openshift.com."
+  if [[ "${PERSONAL_PULL_SECRET}" != *"cloud.openshift.com"* ]]; then
+      error "Get a valid pull secret (json string) from https://cloud.redhat.com/openshift/install/pull-secret"
       exit 1
   fi
 }
 
+function build_pull_secret() {
+    # Check to see if we already have a full pull secret.
+    # avoid "-z $PULL_SECRET" to ensure the secret is not logged
```
zaneb, line 397: Won't we also have a problem with CI_TOKEN and PERSONAL_PULL_SECRET being logged in `verify_pull_secret` above?

04_setup_ironic.sh
```diff
@@ -19,6 +19,7 @@ rm -f assets/templates/99_local-registry.yaml $OPENSHIFT_INSTALL_PATH/data/data/
 
 # Various commands here need the Pull Secret in a file
 export REGISTRY_AUTH_FILE=$(mktemp --tmpdir "pullsecret--XXXXXXXXXX")
+build_pull_secret
```
zaneb, line 22: Why not pass the filename here and have it write directly to a file instead of going via an environment variable?
Or even create the tempfile inside build_pull_secret and have that function return the filename.

utils.sh
dhellmann: I have no idea what either of those commands is doing. They're shorter, though, so I guess they're simpler? ;-)

utils.sh
dhellmann: We're logging PULL_SECRET every time we use it via echo anyway, so I don't know how accurate this comment really is.

config_example.sh
andfasano: +1 This will help also in the CI builds an minimize the chance of a leak

04_setup_ironic.sh
dhellmann: Good idea. I added another patch to the PR to do that.

utils.sh
zaneb: It's just concatenating the two files and merging all of the objects. So, basically what we want, without having to look inside the data and hard-code what we expect it to look like.

utils.sh
zaneb: Another reason to have it in a file maybe.

dhellmann: Yes, I think moving it to a file resolved that.

utils.sh
dhellmann: I took another stab at simplifying the merge. Now that all of the values are in files, we can use the same `*` operations in both cases.

config_example.sh
dhellmann: The CI_TOKEN value is logged when it is used in the command to login to get the pull secret. I'm not sure we can avoid that, but since we're only going to use the tokens for personal users maybe it doesn't matter?

andfasano: I don't expect to use CI_TOKEN it in the CI builds, since we already have a combined pull secret file prepared by the CI lower infrastructure (ci-operator), see https://github.com/openshift/release/blob/6bbee5476c9a8487fea4d12d227e8e52f480e998/ci-operator/step-registry/baremetalds/devscripts/setup/baremetalds-devscripts-setup-commands.sh#L78

dhellmann: OK, good. I think the latest version of this is probably logging fewer secret values, then.

utils.sh
```diff
@@ -401,26 +401,20 @@ function write_pull_secret() {
...
-    # create the final output.
-    jq -s '.[0] * .[1]' ${tmppullsecret} ${REGISTRY_CREDS} > ${PULL_SECRET_FILE}
+    oc registry login --kubeconfig=$tmpkubeconfig --to=$tmppullsecret1
+
+    # Combine the personal pull secret with the one for the CI registry,
+    # then with the local registry credentials.
+    tmppullsecret2=$(mktemp --tmpdir "pullsecret2--XXXXXXXXXX")
+    _tmpfiles="$_tmpfiles $tmppullsecret2"
+    jq -s '.[0] * .[1]' ${PERSONAL_PULL_SECRET} ${tmppullsecret1} > ${tmppullsecret2}
+    jq -s '.[0] * .[1]' ${tmppullsecret2} ${REGISTRY_CREDS} > ${PULL_SECRET_FILE}
```
zaneb, line 419: That jq syntax is certainly easier to read than what I had. You could eliminate the second tempfile by combining all 3 in one go:

```
jq -s '.[0] * .[1] * .[2]' ${PERSONAL_PULL_SECRET} ${tmppullsecret1} ${REGISTRY_CREDS} >${PULL_SECRET_FILE}
```

(and of course you could eliminate the first tempfile as well by piping from oc registry login and using `-` as the filename.)

config_example.sh
zaneb: config_example is out of date wrt the code now.

ocp_install_env.sh
```diff
@@ -149,10 +145,12 @@ function generate_ocp_install_config() {
...
 
-    # when using local mirror set pull secret to this mirror
-    # also this should ensure we don't accidentally pull from upstream
+    # when using local mirror set pull secret to just this mirror to
+    # ensure we don't accidentally pull from upstream
     if [ ! -z "${MIRROR_IMAGES}" ]; then
-        export PULL_SECRET=$(cat ${REGISTRY_CREDS})
+        export install_config_pull_secret="${REGISTRY_CREDS}"
+    else
+        export install_config_pull_secret="${PULL_SECRET_FILE}"
```
zaneb, line 153: nit: not sure this needs to be exported

utils.sh
```diff
@@ -380,18 +380,45 @@ EOF
 function verify_pull_secret() {
   # Do some PULL_SECRET sanity checking
   if [[ "${OPENSHIFT_RELEASE_IMAGE}" == *"registry.svc.ci.openshift.org"* ]]; then
-      if [[ "${PULL_SECRET}" != *"registry.svc.ci.openshift.org"* ]]; then
-          echo "Please get a valid pull secret for registry.svc.ci.openshift.org."
+      if [[ -z "${CI_TOKEN}"  ]]; then
```
zaneb, line 383: Could be `[ ${#CI_TOKEN} = 0 ]`

utils.sh
```diff
@@ -380,18 +380,45 @@ EOF
...
+          error "Please login to https://api.ci.openshift.org and copy the token from the login command from the menu in the top right corner to set CI_TOKEN."
           exit 1
       fi
   fi
 
-  if [[ "${PULL_SECRET}" != *"cloud.openshift.com"* ]]; then
-      echo "Please get a valid pull secret for cloud.openshift.com."
+  if ! grep -q cloud.openshift.com "${PERSONAL_PULL_SECRET}"; then
+      error "Get a valid pull secret (json string) from https://cloud.redhat.com/openshift/install/pull-secret"
```
zaneb, line 390: Perhaps preface this with something like:

    error "No cloud.openshift.com pull secret in ${PERSONAL_PULL_SECRET}"

utils.sh
```diff
@@ -380,18 +380,45 @@ EOF
...
       exit 1
   fi
 }
 
+function write_pull_secret() {
+    if [ ${#PULL_SECRET} != 0 ]; then
+        echo "Using PULL_SECRET variable. Consider switching to PULL_SECRET and CI_TOKEN."
```
zaneb, line 397: ...switching to PERSONAL_PULL_SECRET ...

zaneb (commented):
This looks good. I feel like it needs to be documented somewhere.

utils.sh
dhellmann: Done

utils.sh
dhellmann: Done

utils.sh
dhellmann: Refactoring is the worst.

utils.sh
dhellmann: I tried that and it didn't work, but I think I was passing the wrong filenames or something because it's working for me now.

dhellmann (commented):
Rebased and comments addressed.

ocp_install_env.sh
```diff
@@ -47,10 +45,8 @@ function extract_rhcos_json() {
     pullsecret_file=$(mktemp --tmpdir "pullsecret--XXXXXXXXXX")
     _tmpfiles="$_tmpfiles $pullsecret_file"
```
hardys, line 46: I think we can remove this tmpfile as it's no longer used?

utils.sh
```diff
@@ -380,18 +380,43 @@ EOF
...
   fi
 }
 
+function write_pull_secret() {
+    if [ ${#PULL_SECRET} != 0 ]; then
+        echo "Using PULL_SECRET variable. Consider switching to PERSONAL_PULL_SECRET and CI_TOKEN."
+        set +x
+        echo "${PULL_SECRET}" > ${PULL_SECRET_FILE}
+        set -x
+        return
```
hardys, line 402: I tested without making changes to my config, e.g with the existing PULL_SECRET and the mirroring fails like:

```
error: unable to upload blob sha256:a03401a44180b6581a149376d6fd2d5bd85d938445fd5b5ad270e14ddde4937c to virthost.ostest.test.metalkube.org:5000/localimages/local-release-image: Post "https://virthost.ostest.test.metalkube.org:5000/v2/localimages/local-release-image/blobs/uploads/": no basic auth credentials
error: unable to upload blob sha256:97c7aeeb52fe8b1a310c9b1ec54763d1affbba48be9298b6476f5bc13896fbf5 to virthost.ostest.test.metalkube.org:5000/localimages/local-release-image: Post "https://virthost.ostest.test.metalkube.org:5000/v2/localimages/local-release-image/blobs/uploads/": no basic auth credentials
info: Mirroring completed in 56.2s (0B/s)
error: one or more errors occurred while uploading images
```

Which I guess is expected, since we return here before doing the oc registry login etc.

I wonder if we should consider just removing the old interface, and forcing everyone to switch to the new interface?  Then at least everyone can use the same path, and we can ensure it stays working in CI?

dhellmann: I thought about that, but wasn't sure how CI was being configured so I wasn't sure what to do to update it. Are the scripts to get the pull secret for CI in this repo somewhere?

andfasano: No, it's part of the of the multistage test here: https://github.com/openshift/release/blob/d741522e927fb687a1357ebfdac7593e9fd353ef/ci-operator/step-registry/baremetalds/devscripts/setup/baremetalds-devscripts-setup-commands.sh#L46

dhellmann: OK, so in order for this PR to pass CI at all we need it to continue to support PULL_SECRET. We could update the job definition separately, then come back and remove the PULL_SECRET support here.

ocp_install_env.sh
dhellmann: Done

utils.sh
dhellmann: I've updated this section to always merge the $PULL_SECRET with the local registry credentials instead of assuming that PULL_SECRET is complete on its own.

PR #1095, opened 2020-09-04
Remove wait for ironic service

andfasano:
BMO now checks explicitly for the availability of Ironic services within the reconcile loop before applying any logic, so the `local-bmo.sh` script can be slightly simplified


zaneb:
It occurs to me this might continue to be a problem when deploying older releases with the latest dev-scripts, but that probably doesn't happen that often with the local-bmo script.

PR #1096, opened 2020-09-07
Local cbo script

andfasano:
This PR adds a new script to test a local Cluster Baremetal Operator checkout. The script copies an existing metal3 deployment into a new one and adds the required annotation to grant the ownership to cbo.

### DIFF
--- a/metal3-dev/local-cbo.sh
+++ b/metal3-dev/local-cbo.sh
@@ -1,0 +1,22 @@
+#!/bin/bash -xe
+
+SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )/.." && pwd )"
+
+LOGDIR=${SCRIPTDIR}/logs
+source $SCRIPTDIR/logging.sh
+
+source $SCRIPTDIR/common.sh
+source $SCRIPTDIR/utils.sh
+
+cbo_path=$GOPATH/src/github.com/openshift/cluster-baremetal-operator
+if [ ! -d $cbo_path ]; then
+    echo "Did not find $cbo_path" 1>&2
+    exit 1
+fi
+
+# Run the operator
+cd $cbo_path
+
+export RUN_NAMESPACE=openshift-machine-api
+make run
+


metal3-dev/local-cbo.sh
```diff
@@ -0,0 +1,61 @@
...
+if [ ! -d $cbo_path ]; then
+    echo "Did not find $cbo_path" 1>&2
+    exit 1
+fi
+
+# Stop the machine-api-operator so it does not try to fix the
+# deployment we are going to change.
+$SCRIPTDIR/stop-mao.sh
+
+# Scale the existing deployment down.
```
dhellmann, line 32: I don't think we need to manipulate the metal3 deployment, do we? The CBO will do that for us when it comes up, right?

metal3-dev/local-cbo.sh
```diff
@@ -0,0 +1,61 @@
...
+    oc wait --for=delete -n openshift-machine-api $metal3pods || true
+fi
+
+OUTDIR=${OCP_DIR}/metal3-cbo-dev
+mkdir -p $OUTDIR
+
+# Save a copy of the full deployment as input
+oc get deployment -n openshift-machine-api -o yaml metal3 > $OUTDIR/bmo-deployment-full.yaml
+
+# Add the annotation for the cbo ownership
```
dhellmann, line 45: The CBO should do this when it comes up. Since the script stops the MAO before starting the CBO there shouldn't be any race condition.

andfasano:
Since the MAO currently also checks the presence of the CBO cluster operator (in addition to the annotations) I had to stop it before patching the existing metal3 deployment, otherwise it will continue to manage the metal3 deployment

metal3-dev/local-cbo.sh
sadasu: I think what we discussed regarding this was that eventually we will need this script to add the annotation until the CBO code catches up to adding its own cboOwnedAnnotation. During testing CBO incrementally, we would want this script to create a dummy metal3 deployment with the cboOwnedAnnotation added to it with replicas = 0.

sadasu: Also, we do not want to remove the maoOwnedAnnotation because when both the cboOwnedAnnotation and the maoOwnedAnnotation are present, the cboOwnedAnnotation should take precedence.

dhellmann: I think we're mixing 2 things together in the work this PR is doing.

For CI, we need normal dev-scripts to have a variable to control creating a completely new Deployment manifest called metal3 with the expected annotation to prevent the MAO from doing anything with metal3 at all. That will let us test bringing up a cluster with the CBO and without MAO's logic for managing metal3. We will want a separate PR for that work.

This PR is creating a separate script for developers to run locally once they have a functioning cluster, to bring up a locally built CBO for testing. That script shouldn't need to do much for now, since the CBO should immediately add the required annotation to the metal3 Deployment and the MAO should stop managing it. Eventually when the CVO is installing the CBO the script will need to stop the copy of the CBO running in the cluster to allow the local replace it.

andfasano: Ok will remove the patch command then

metal3-dev/local-cbo.sh
```diff
@@ -0,0 +1,50 @@
...
+if [ ! -d $cbo_path ]; then
+    echo "Did not find $cbo_path" 1>&2
+    exit 1
+fi
+
+# Stop the machine-api-operator so it does not try to fix the
+# deployment we are going to change. This is required until MAO
+# will continue to check the presence of the CBO operator - in
+# addition to the annotations
+$SCRIPTDIR/stop-mao.sh
```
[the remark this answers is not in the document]
andfasano: If MAO is not stopped during a local CBO development then it's not clear how to bypass this check: https://github.com/openshift/machine-api-operator/blob/f38acc8d312f527cb2cf1505ef8e346a99a9638b/pkg/operator/sync.go#L208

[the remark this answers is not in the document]
dhellmann: It's the responsibility of the CBO to configure its ClusterOperator resource and to annotate the metal3 deployment to tell the MAO to ignore it. Those two stories should be early in the list of implementation priorities, and until they are implemented it doesn't matter that the MAO keeps doing its work because the CBO doesn't do anything, yet.

metal3-dev/local-cbo.sh
```diff
@@ -0,0 +1,34 @@
...
+source $SCRIPTDIR/common.sh
+source $SCRIPTDIR/utils.sh
+
+if ! which kubebuilder 2>&1 >/dev/null ; then
+    echo "Did not find kubebuilder." 1>&2
+    echo "To install it see https://book.kubebuilder.io/quick-start.html." 1>&2
+    exit 1
+fi
+
+if ! which yq 2>&1 >/dev/null ; then
```
dhellmann, line 17: yq is no longer needed, right?

metal3-dev/local-cbo.sh
```diff
@@ -0,0 +1,34 @@
...
+
+SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )/.." && pwd )"
+
+LOGDIR=${SCRIPTDIR}/logs
+source $SCRIPTDIR/logging.sh
+
+source $SCRIPTDIR/common.sh
+source $SCRIPTDIR/utils.sh
+
+if ! which kubebuilder 2>&1 >/dev/null ; then
```
dhellmann, line 11: Is this checking a dependency that the CBO Makefile should handle?

metal3-dev/local-cbo.sh
andfasano: Right, I'll remove it

metal3-dev/local-cbo.sh
andfasano: Not directly, I though at the beginning it could have been useful to ensure its presence during the early stage of development. As it stands right now, I think it could also be removed.

dhellmann (commented):
/approve
/lgtm

PR #1097, opened 2020-09-08
Remove download of headers file when using local ipa-downloader image

bfournie:
This is no longer needed.

### DIFF
--- a/04_setup_ironic.sh
+++ b/04_setup_ironic.sh
@@ -199,6 +199,5 @@ while ! curl --fail -g http://$(wrap_if_ipv6 ${PROVISIONING_HOST_IP})/images/${M
 if [ -n "${IRONIC_IPA_DOWNLOADER_LOCAL_IMAGE:-}" ];
 then
   while ! curl --fail --head -g http://$(wrap_if_ipv6 ${PROVISIONING_HOST_IP})/images/ironic-python-agent.initramfs ; do sleep 1; done
-  while ! curl --fail --head -g http://$(wrap_if_ipv6 ${PROVISIONING_HOST_IP})/images/ironic-python-agent.tar.headers ; do sleep 1; done
   while ! curl --fail --head -g http://$(wrap_if_ipv6 ${PROVISIONING_HOST_IP})/images/ironic-python-agent.kernel ; do sleep 1; done
 fi


04_setup_ironic.sh
```diff
@@ -199,6 +199,5 @@ while ! curl --fail -g http://$(wrap_if_ipv6 ${PROVISIONING_HOST_IP})/images/${M
 if [ -n "${IRONIC_IPA_DOWNLOADER_LOCAL_IMAGE:-}" ];
 then
   while ! curl --fail --head -g http://$(wrap_if_ipv6 ${PROVISIONING_HOST_IP})/images/ironic-python-agent.initramfs ; do sleep 1; done
-  while ! curl --fail --head -g http://$(wrap_if_ipv6 ${PROVISIONING_HOST_IP})/images/ironic-python-agent.tar.headers ; do sleep 1; done
```
hardys, line 202: When testing with the upstream ipa-downloader, this is still used on the bootstrap VM I think?  

https://github.com/metal3-io/ironic-ipa-downloader/blob/master/get-resource.sh#L25

https://github.com/openshift/installer/blob/master/data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template#L97

However it's true that we no longer pass CACHEURL to the container that's run on the cluster ref:

https://github.com/openshift/machine-api-operator/blob/master/pkg/operator/baremetal_pod.go#L356

I guess we need to decide if we should remove all this mirroring code (and the CACHEURL from the installer), or leave it as-is to speed up testing (only on the bootstrap VM) with upstream images?

@derekhiggins IIRC you were in favor of maintaining this previously, any thoughts?

derekhiggins: Ya, we're still using it in the bootstrap node (snippits from recent upstream run)
```
+ '[' -n http://192.168.111.1/images -a '!' -e ironic-python-agent.tar.headers ']'
+ curl -g --verbose --fail -O http://192.168.111.1/images/ironic-python-agent.tar.headers
+ curl -g --verbose --dump-header ironic-python-agent.tar.headers -O https://images.rdoproject.org/centos8/master/rdo_trunk/current-tripleo//ironic-python-agent.tar --header 'If-None-Match: "1f789000-5ae78b5eca4
+ curl -g --verbose -O http://192.168.111.1/images/ironic-python-agent-1f789000-5ae78b5eca443/ironic-python-agent.tar

```

 and in metal3-dev-env, I'm guessing the caching logic is probably more important here as people are more likely to run metal3-dev-env in a home environment
https://github.com/metal3-io/metal3-dev-env/blob/master/03_launch_mgmt_cluster.sh#L179

We could check with some users of metal3-dev-env if they still rely on the caching we should probably keep it, and then if we keep it we may as well keep using it in the boot strap node.

derekhiggins: having said all that, looking at the code, this is the line that creates the headers file on the bootstrap node
https://github.com/openshift/ironic-ipa-downloader/blob/f8523061/get-resource.sh#L61
So if the check for initramfs and kernel works then its unlikely the check for headers with fail. If removing this check allows some custom images to work then I'm ok with removing it.

derekhiggins (commented):
/lgtm

PR #1098, opened 2020-09-08
Explain how to get a CI cluster pull secret

danwinship:
The README currently suggests that you only need a "public" pull secret, but in order to install nightly or CI builds you need a pull secret from the CI cluster as well.


dhellmann:
See also #1093

danwinship:
> See also #1093

That seems nicer
/close

PR #1102, opened 2020-09-14
Empty metal3 deployment for CBO testing

andfasano:
This PR adds an empty metal3 deployment used to test CBO ownership

### DIFF
--- a/assets/metal3-cbo-deployment.yaml
+++ b/assets/metal3-cbo-deployment.yaml
@@ -1,0 +1,50 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  annotations:
+    deployment.kubernetes.io/revision: "1"
+    baremetal.openshift.io/owned: ""
+  generation: 1
+  labels:
+    api: clusterapi
+    k8s-app: controller
+  name: metal3
+  namespace: openshift-machine-api
+spec:
+  replicas: 0
+  selector:
+    matchLabels:
+      api: clusterapi
+      k8s-app: controller
+  template:
+    metadata:
+      labels:
+        api: clusterapi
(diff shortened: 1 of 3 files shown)


assets/metal3-cbo-deployment.yaml
```diff
@@ -0,0 +1,50 @@
...
+    deployment.kubernetes.io/revision: "1"
+    baremetal.openshift.io/owned: ""
+  generation: 1
+  labels:
+    api: clusterapi
+    k8s-app: controller
+  name: metal3
+  namespace: openshift-machine-api
+spec:
+  replicas: 1
```
dhellmann, line 14: This could be 0.

dhellmann (commented):
/retest

assets/metal3-cbo-deployment.yaml
andfasano: Done

utils.sh
```diff
@@ -61,6 +61,15 @@ function create_cluster() {
         cp assets/ipv6-dual-stack-no-upgrade.yaml ${assets_dir}/openshift/.
     fi
 
+    if [[  ! -z "${ENABLE_CBO_TEST}" ]]; then
```
dhellmann, line 64: Does anything look at the images used by the deployment before a pod is created? I don't think it's a problem to create this image, I just wonder if we actually need it.

andfasano: Just wanted to have a fixed simple image to reference in case the pod of the deployment is going to go up (even though now I changed the replicas to 0) - the image spec is anyhow required

dhellmann: OK, let's keep this as it is, then.

PR #1105, opened 2020-09-15
Fix known_hosts cleanup

bcrochet:
The cleanup of known_hosts file in ocp_cleanup.sh requires
EXTERNAL_SUBNET_V{4,6} to be set. If one does not have that set in
config.sh, then the known_hosts cleanup does not happen, since
network.sh isn't included in the cleanup. This adds network.sh to
be sourced in the cleanup script.

### DIFF
--- a/ocp_cleanup.sh
+++ b/ocp_cleanup.sh
@@ -3,6 +3,7 @@ set -x
 
 source logging.sh
 source common.sh
+source network.sh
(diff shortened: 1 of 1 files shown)


hardys:
Thanks, I noticed this was broken too but hadn't yet looked into why :+1: 

/approve

/retest

PR #1111, opened 2020-09-17
Always include metal3-inspector 'cloud' in clouds.yaml

zaneb:
When the auth method is basic-auth, you have to use
`--os-cloud=metal3-inspector` instead of `--os-cloud=metal3` when talking to
ironic-inspector rather than ironic-api. For consistency, ensure this
also works when the auth method is noauth.

### DIFF
--- a/clouds.yaml.template
+++ b/clouds.yaml.template
@@ -8,6 +8,9 @@ clouds:
     auth_type: {{.AuthType}}
     baremetal_endpoint_override: {{.ClusterIronicURL}}
     baremetal_introspection_endpoint_override: {{.ClusterInspectorURL}}
+  metal3-inspector:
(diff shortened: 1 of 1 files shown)


hardys:
We should probably mention this in https://github.com/openshift-metal3/dev-scripts/blob/master/README.md#interacting-with-ironic-directly but it seems like that needs some other updates so we can do it as a follow-up

derekhiggins:
> We should probably mention this in https://github.com/openshift-metal3/dev-scripts/blob/master/README.md#interacting-with-ironic-directly but it seems like that needs some other updates so we can do it as a follow-up

Some updates here, I'll update this to mention inspector
https://github.com/openshift-metal3/dev-scripts/pull/1104

PR #1120, opened 2020-09-29
add gather_audit_logs target

sanchezl:
gather target no longer collects audit logs. Lack of audit logs in CI for metal-upi is blocking progress on debugging an important problems on that platform

### DIFF
--- a/must_gather.sh
+++ b/must_gather.sh
@@ -22,4 +22,4 @@ else
   MUST_GATHER_IMAGE=""
 fi
 
-oc --insecure-skip-tls-verify adm must-gather $MUST_GATHER_IMAGE --dest-dir "$MUST_GATHER_PATH" > "$MUST_GATHER_PATH/must-gather.log"
+oc --insecure-skip-tls-verify adm must-gather $MUST_GATHER_IMAGE --dest-dir "$MUST_GATHER_PATH" "$@" > "$MUST_GATHER_PATH/must-gather.log"
--- a/Makefile
+++ b/Makefile
@@ -29,6 +29,9 @@ ocp_run:
 gather:
 	./must_gather.sh
 
+gather_audit_logs:
+	./must_gather.sh -- /usr/bin/gather_audit_logs
+
 clean: ocp_cleanup ironic_cleanup host_cleanup assisted_deployment_cleanup
 
 assisted_deployment_cleanup:


must_gather.sh
```diff
@@ -22,4 +22,4 @@ else
   MUST_GATHER_IMAGE=""
 fi
 
-oc --insecure-skip-tls-verify adm must-gather $MUST_GATHER_IMAGE --dest-dir "$MUST_GATHER_PATH" > "$MUST_GATHER_PATH/must-gather.log"
+oc --insecure-skip-tls-verify adm must-gather $MUST_GATHER_IMAGE --dest-dir "$MUST_GATHER_PATH" "$@"> "$MUST_GATHER_PATH/must-gather.log"
```
zaneb, line 25: I feel like this would be easier to read with a space before the >

zaneb (commented):
> Lack of audit logs in CI for metal-upi is blocking progress on debugging an important problems on that platform

Did you mean metal-ipi? If not I think you're going to be disappointed ;)

/approve

must_gather.sh
sanchezl: Done.

sanchezl:
> Did you mean metal-ipi? If not I think you're going to be disappointed ;)

Yes, I was actually looking into meta-upi, but looks like metal-ipi needs this too :-) .

andfasano (commented):
What about enriching directly the `must_gather.sh` script with an additional `must-gather` command for the audit logs? My feeling is that, if it's another important piece of logs to be fetched for troubleshooting (in addition to the previous ones), then it could make sense to have everything together.
Moreover, having a single script could simplify the entire chain, since it will not require to change Makefile with an additional target, and then the CI [must-gather multistage step](https://github.com/openshift/release/blob/a64045360b52cb08f7cffad83cd2870114f32d4c/ci-operator/step-registry/baremetalds/devscripts/gather/baremetalds-devscripts-gather-commands.sh#L44)

hardys:
> What about enriching directly the must_gather.sh script with an additional must-gather command for the audit logs?

Good point, should we just always collect the audit logs, e.g so they are captured in CI?

andfasano:
> > What about enriching directly the must_gather.sh script with an additional must-gather command for the audit logs?
> 
> Good point, should we just always collect the audit logs, e.g so they are captured in CI?

I think that was the original idea

PR #1122, opened 2020-09-30
validation improvements

dhellmann:
- Reorganize validation to make it easier to run separate tests when
  building a cluster vs. cleaning up a host.
- Add an early check to ensure the CI_TOKEN is valid.


hardys:
I'm not entirely sure if we still want the separate `sanitychecks.sh` now - it may make sense for those checks to also be done in this new script?

That can be done as a follow up if we decide it makes sense though, lgtm!

/lgtm

PR #1125, opened 2020-10-02
Add network.sh to must_gather.sh

hardys:
Otherwise MIRROR_IMAGES isn't defined when testing, and we see
a failure to pull the image in the CI logs when test ipv6

### DIFF
--- a/must_gather.sh
+++ b/must_gather.sh
@@ -1,10 +1,11 @@
 #!/usr/bin/env bash
 
-set -xe
+set -xeu
 
 source logging.sh
 source common.sh
 source utils.sh
+source network.sh
 
 MUST_GATHER_PATH=${MUST_GATHER_PATH:-$LOGDIR/$CLUSTER_NAME/must-gather}
 if [ ! -d "$MUST_GATHER_PATH" ]; then


hardys:
@stbenjam @trozet @danwinship 

This fixes one problem with must-gather for the e2e-ipi-baremetal ipv6 CI, but it's still possible for the gather to get stuck if the serviceNetwork is completely broken, e.g:

```
[must-gather-rgdnc] POD Unable to connect to the server: dial tcp [fd02::1]:443: i/o timeout
[must-gather-rgdnc] POD Unable to connect to the server: dial tcp [fd02::1]:443: i/o timeout
[must-gather-rgdnc] POD Unable to connect to the server: dial tcp [fd02::1]:443: i/o timeout
[must-gather-rgdnc] POD Unable to connect to the server: dial tcp [fd02::1]:443: i/o timeout
[must-gather-rgdnc] POD Unable to connect to the server: dial tcp [fd02::1]:443: i/o timeout
[must-gather-rgdnc] POD Unable to connect to the server: dial tcp [fd02::1]:443: i/o timeout
[must-gather-rgdnc] POD Unable to connect to the server: dial tcp [fd02::1]:443: i/o timeout
```

Any suggestions on how we might handled that - should we fall back to scp and just copy the pod logs off each node perhaps?

danwinship:
In general, must-gather depends on a minimal level of cluster functionality...

I'd say, if it turns out that "must-gather failed because the service network isn't up" turns out to be a recurring problem, then we should try to work around it, but it's OK to not worry about for now

PR #1129, opened 2020-10-12
Remove deploy validation in requirements step

krsacme:
oc command will be installed on the requirements step.
But deploy validation tries to login with the provided
ci token to validate the token. Allow the requirements to
install oc cli, then proceed with the validation in next step.

### DIFF
--- a/01_install_requirements.sh
+++ b/01_install_requirements.sh
@@ -8,7 +8,7 @@ source sanitychecks.sh
 source utils.sh
 source validation.sh
 
-early_deploy_validation
+early_deploy_validation true
 
 if [ -z "${METAL3_DEV_ENV}" ]; then
   export REPO_PATH=${WORKING_DIR}
--- a/validation.sh
+++ b/validation.sh
@@ -51,6 +51,8 @@ function early_either_validation() {
 # build a cluster.
 function early_deploy_validation() {
 
+    CHECK_OC_TOOL_PRESENCE=${1:-"false"}
+
     early_either_validation
 
     if [ ! -s ${PERSONAL_PULL_SECRET} ]; then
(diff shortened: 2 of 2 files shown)


krsacme:
@dhellmann PTAL

hardys:
Hi @krsacme thanks for the PR!  Makes sense to me, we didn't hit this issue testing due to existing copies of `oc` existing on the local system.

/lgtm

01_install_requirements.sh
```diff
@@ -8,7 +8,7 @@ source sanitychecks.sh
 source utils.sh
 source validation.sh
 
-early_deploy_validation
+early_either_validation
```
dhellmann, line 11: We still want to exit early if the secret and token aren't set. Maybe we can change the validation step to only run if oc is installed? Or pass an argument to the function to disable that 1 check?

krsacme: Added an argument to the check. PTAL.

hardys: This script doesn't depend on those variables, so I was thinking it'd be sufficient to leave that check until the 02 script?

I guess alternatively if we want to check those variables are set always, we could move the checks (but not the oc login) to early_either_validation?

dhellmann: > This script doesn't depend on those variables, so I was thinking it'd be sufficient to leave that check until the 02 script?
> 
> I guess alternatively if we want to check those variables are set always, we could move the checks (but not the oc login) to early_either_validation?

The "either" function is meant to work for cleanup, too, and the variables aren't needed then.

Even though the requirements script doesn't need the variables if someone types `make` on a clean host I wanted them to not have to wait while it updates all of the packages before they see the error.

dhellmann:
/approve

This version looks good. Thanks!

Issue #1131, opened 2020-10-15
Deployment throws error with config OPENSHIFT_VERSION=4.5.0


krsacme:
**Describe the bug**
Providing `export OPENSHIFT_VERSION=4.5.0` as config version results in error post cluster deployment

**To Reproduce**
Deploy the cluster with below config
```
export OPENSHIFT_RELEASE_TYPE=ga
export OPENSHIFT_VERSION=4.5.0
export PERSONAL_PULL_SECRET=/home/kni/pull-secret.json
export IP_STACK=v4
export NUM_WORKERS=0
export PRO_IF=eno2
export INT_IF=eno3
export CLUSTER_NAME=vnf
export BASE_DOMAIN=test.metal3.io
export PROVISIONING_NETWORK=172.51.0.0/16
export MASTER_MEMORY=16384
export MASTER_DISK=40
export MASTER_VCPU=8
export EXTERNAL_SUBNET_V4=192.168.212.0/24
export WORKING_DIR=/home/dev-scripts
```

**Expected/observed behavior**
After the cluster is deployed, it post deployment actions throw below error

```
+(utils.sh:102): create_cluster(): generate_auth_template                                                                       
+(utils.sh:271): generate_auth_template(): OCP_VERSIONS_NOAUTH='4.3 4.4 4.5'
+(utils.sh:272): generate_auth_template(): [[ 4.3 4.4 4.5 == *\4\.\5\.\0* ]]
++(utils.sh:275): generate_auth_template(): oc -n openshift-machine-api get secret/metal3-ironic-password -o template --template
'{{.data.username}}'                                                                                                            
++(utils.sh:275): generate_auth_template(): base64 -d                                                                           
Error from server (NotFound): secrets "metal3-ironic-password" not found                                                        
+(utils.sh:275): generate_auth_template(): IRONIC_USER=
+(utils.sh:1): generate_auth_template(): removetmp                                                                              
+(utils.sh:483): removetmp(): '[' -n ' /tmp/test-token--ObGdpcqX6W' ']'                                                         
+(utils.sh:483): removetmp(): rm -rf /tmp/test-token--ObGdpcqX6W                                                                
make: *** [Makefile:27: ocp_run] Error 1
```

PR #1132, opened 2020-10-15
Auth template generate fails for 4.5.0 version

krsacme:
Using OPENSHIFT_VERSION=4.5.0 results in failure of
the auth template generation because it take wrong path
as the condition only checks for 4.5 only. Trim the version
if it is of format x.y.z before checking.

Fixes: #1131

### DIFF
--- a/utils.sh
+++ b/utils.sh
@@ -269,7 +269,11 @@ function sync_repo_and_patch {
 function generate_auth_template {
     # clouds.yaml
     OCP_VERSIONS_NOAUTH="4.3 4.4 4.5"
-    if [[ "$OCP_VERSIONS_NOAUTH" == *"$OPENSHIFT_VERSION"* ]]; then
+
+    # Fix the version if OPENSHIFT_VERSION follows x.y.z format to change it to x.y
+    OPENSHIFT_VERSION_TRIM=$(echo "$OPENSHIFT_VERSION" | grep -oE '^[[:digit:]]+\.[[:digit:]]+' || echo "$OPENSHIFT_VERSION")
+
+    if [[ "$OCP_VERSIONS_NOAUTH" == *"$OPENSHIFT_VERSION_TRIM"* ]]; then
         go run metal3-templater.go "noauth" -template-file=clouds.yaml.template -provisioning-interface="$CLUSTER_PRO_IF" -provisioning-network="$PROVISIONING_NETWORK" -image-url="$MACHINE_OS_IMAGE_URL" -bootstrap-ip="$BOOTSTRAP_PROVISIONING_IP" -cluster-ip="$CLUSTER_PROVISIONING_IP" > clouds.yaml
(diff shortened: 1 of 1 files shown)


krsacme:
Auth failure on ci, checking if its related to the patch.
```
* 4x kubelet: Failed to pull image "registry.svc.ci.openshift.org/ci/secret-wrapper:latest": rpc error: code = Unknown desc = unable to retrieve auth token: invalid username/password: unauthorized: authentication required
```

hardys:
> Auth failure on ci, checking if its related to the patch.

I think there's a CI outage atm so it's probably not related to the patch.

I made a small comment, but if you prefer to leave it as-is that's fine, we'll need to recheck CI when things are working though.

utils.sh
```diff
@@ -269,7 +269,12 @@ function sync_repo_and_patch {
 function generate_auth_template {
     # clouds.yaml
     OCP_VERSIONS_NOAUTH="4.3 4.4 4.5"
-    if [[ "$OCP_VERSIONS_NOAUTH" == *"$OPENSHIFT_VERSION"* ]]; then
+
+    # Fix the version if OPENSHIFT_VERSION follows x.y.z format to change it to x.y
+    PATTERN="[0-9]+\.[0-9]+\.[0-9]+"
+    OPENSHIFT_VERSION_TRIM=$([[ "$OPENSHIFT_VERSION" =~ $PATTERN ]] && echo ${OPENSHIFT_VERSION%.*} || echo $OPENSHIFT_VERSION)
```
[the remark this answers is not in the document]
krsacme: Its simpler. I will apply it.

utils.sh
```diff
@@ -269,7 +269,11 @@ function sync_repo_and_patch {
 function generate_auth_template {
     # clouds.yaml
     OCP_VERSIONS_NOAUTH="4.3 4.4 4.5"
-    if [[ "$OCP_VERSIONS_NOAUTH" == *"$OPENSHIFT_VERSION"* ]]; then
+
+    # Fix the version if OPENSHIFT_VERSION follows x.y.z format to change it to x.y
+    OPENSHIFT_VERSION_TRIM=$(echo "$OPENSHIFT_VERSION" | grep -o '^[[:digit:]].[[:digit:]]')
```
hardys, line 274: Sorry, my suggestion earlier was incomplete - I forgot to escape the `.` so this will match e.g 4.3 and 4x3, also it won't work when either number reaches double figures, so should probably be:

```
grep -oE '^[[:digit:]]+\.[[:digit:]]+'
```

krsacme:
When `OPENSHIFT_VERSION=""`, grep fails, will fix it.

utils.sh
krsacme: Modified. Aslo added error handing when version is empty string (grep will throw error).

PR #1138, opened 2020-10-29
Add workaround for OKD ironic images

hardys:
Currently OKD has a dummy/stub image in place for all the ironic*
images, so the current approach with *LOCAL_IMAGE doesn't work,
we end up sedding the stub image everywhere with the first image.

Instead, until the OKD ironic images are ready, we can work around
the problem by explicitly replacing each image, then a config like
this should work I think:

```
export UPSTREAM_IRONIC=true
export OPENSHIFT_RELEASE_IMAGE=registry.svc.ci.openshift.org/origin/release:4.6.0-0.okd-2020-10-28-192548
export IRONIC_MACHINE_OS_DOWNLOADER_LOCAL_IMAGE=quay.io/hroyrh/test:latest
export IRONIC_STATIC_IP_MANAGER_LOCAL_IMAGE=quay.io/hroyrh/test-static-ip-manager:latest
```

I also added `RELEASE_IMAGE_DEBUG` so you can add an export to get
a dump of the relevant files during the release image build.

### DIFF
--- a/04_setup_ironic.sh
+++ b/04_setup_ironic.sh
@@ -15,6 +15,30 @@ early_deploy_validation
 # Account for differences in 1.* and 2.* version reporting.
 PODMAN_VERSION=$(sudo podman version -f json | jq -r '.Version,.Client.Version|strings')
 
+OKD_IW_CALLED=""
+function okd_image_workaround() {
+    echo "OKD image workaround $1 $2 $3"
+    if [ -z "${OKD_IW_CALLED}" ]; then
+        echo "RUN dnf -y install jq" >> $3
+        OKD_IW_CALLED="y"
+    fi
+    # Update image-references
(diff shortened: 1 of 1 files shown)


vrutkovs:
Looks fine, iiuc once https://github.com/openshift/release/pull/12910 is merged some hacks could be removed, right?

When all images would have RDO-based versions this change would be reverted?

hardys:
> Looks fine, iiuc once [openshift/release#12910](https://github.com/openshift/release/pull/12910) is merged some hacks could be removed, right?
> 
> When all images would have RDO-based versions this change would be reverted?

Yes and yes :)  

This is just a temporary workaround to allow testing with a custom release payload as @hroyrh ran into problems trying to test with OKD via dev-scripts (with a similar config to that mentioned in the commit message)

vrutkovs:
Perfect, thanks!

dtantsur:
Hi! Is this still up-to-date? I think there have been some progress with OKD?

vrutkovs:
OKD still has dummy Ironic images, we are still interested in this PR

PR #1141, opened 2020-11-04
Clarify using local images in config_example.sh

dtantsur:
Add two notes for things that are easy to forget/misinterpret.

### DIFF
--- a/config_example.sh
+++ b/config_example.sh
@@ -20,6 +20,8 @@ set -x
 #export IRONIC_INSPECTOR_LOCAL_IMAGE=https://github.com/metal3-io/ironic-inspector
 #export IRONIC_LOCAL_IMAGE=quay.io/username/ironic
 #export MACHINE_CONFIG_OPERATOR_LOCAL_IMAGE=https://github.com/openshift/machine-config-operator
+# NOTE: If a checkout already exists in $HOME, it won't be re-created.
+# NOTE: You must set CUSTOM_REPO_FILE to build OpenShift ironic images locally.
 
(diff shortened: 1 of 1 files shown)


config_example.sh
```diff
@@ -20,6 +20,8 @@ set -x
 #export IRONIC_INSPECTOR_LOCAL_IMAGE=https://github.com/metal3-io/ironic-inspector
 #export IRONIC_LOCAL_IMAGE=quay.io/username/ironic
 #export MACHINE_CONFIG_OPERATOR_LOCAL_IMAGE=https://github.com/openshift/machine-config-operator
+# NOTE: If a checkout already exists in $HOME, it won't be re-created.
+# NOTE: You must set CUSTOM_REPO_FILE to build OpenShift ironic images locally.
```
hardys, line 24: I'd say "to build some OpenShift images locally" as it's not only the ironic images that need this

dtantsur: "Some" is too vague, how will somebody know what exactly? Which images require a local repo as well?

derekhiggins: I havn't seen CUSTOM_REPO_FILE before but I think its probably anything that installs packages that isn't an upstream package? Looks like their is about 120 images in a release payload I don't know how many of those this would include.

hardys: Yeah it's anything which installs packages, the developer will know because package install fails due to inaccessible repos ;)

hardys: Maybe we can figure out a way to detect that in future but for now this comment makes it a little clearer, thanks!

PR #1084, opened 2020-08-19
Fetch Pull Request to build local image

elfosardo:
Add the possibility to specify a Pull Request number associated to
an image and build it locally with that change.

PR #1142, opened 2020-11-06
Add a way to install extra packages in the images

elfosardo:
Long due change to provide a way to automatically install extra
packages in the local custom built images.
This can be useful to test different versions of packages or just
test new dependencies or tools.

This depends on https://github.com/metal3-io/ironic-inspector-image/pull/72 and https://github.com/metal3-io/ironic-image/pull/227

### DIFF
--- a/04_setup_ironic.sh
+++ b/04_setup_ironic.sh
@@ -38,6 +38,7 @@ fi
 
 for IMAGE_VAR in $(env | grep "_LOCAL_IMAGE=" | grep -o "^[^=]*") ; do
     IMAGE=${!IMAGE_VAR}
+    BUILD_COMMAND_ARGS=""
 
     sudo -E podman pull --authfile $PULL_SECRET_FILE $OPENSHIFT_RELEASE_IMAGE
 
@@ -67,6 +68,17 @@ for IMAGE_VAR in $(env | grep "_LOCAL_IMAGE=" | grep -o "^[^=]*") ; do
                 git fetch origin pull/${IMAGE_PR}/head:pr${IMAGE_PR}
                 git checkout pr${IMAGE_PR}
         fi
+        
+        # If we want to install extra packages, we set the path to a
+        # file containing the packages (one per line) we want to install
+        EXTRA_PKGS_FILE_PATH=${IMAGE_VAR/_LOCAL_IMAGE}_EXTRA_PACKAGES
+        EXTRA_PKGS_FILE=${!EXTRA_PKGS_FILE_PATH:-}
+        EXTRA_PKGS_FILE=$(cd $OLDPWD; realpath $EXTRA_PKGS_FILE)
+        if [[ -n $EXTRA_PKGS_FILE ]]; then
+            cp $EXTRA_PKGS_FILE "$REPOPATH"
+            EXTRA_PKGS_FILE_NAME=$(basename $EXTRA_PKGS_FILE)
+            BUILD_COMMAND_ARGS+=" --build-arg EXTRA_PKGS_LIST=$EXTRA_PKGS_FILE_NAME"
+        fi
 
         # If we built a custom base image, we should use it as a new base in
         # the Dockerfile to prevent discrepancies between locally built images.
@@ -74,7 +86,8 @@ for IMAGE_VAR in $(env | grep "_LOCAL_IMAGE=" | grep -o "^[^=]*") ; do
         if [[ -n ${BASE_IMAGE_DIR:-} ]]; then
             sed -i "s/^FROM [^ ]*/FROM ${BASE_IMAGE_DIR}/g" ${IMAGE_DOCKERFILE}
         fi
-        sudo podman build --authfile $PULL_SECRET_FILE -t ${!IMAGE_VAR} -f $IMAGE_DOCKERFILE .
+        
+        sudo podman build --authfile $PULL_SECRET_FILE $BUILD_COMMAND_ARGS -t ${!IMAGE_VAR} -f $IMAGE_DOCKERFILE .
         cd -
         sudo podman push --tls-verify=false --authfile $PULL_SECRET_FILE ${!IMAGE_VAR} ${!IMAGE_VAR}
     fi
--- a/config_example.sh
+++ b/config_example.sh
@@ -23,6 +23,12 @@ set -x
 # NOTE: If a checkout already exists in $HOME, it won't be re-created.
 # NOTE: You must set CUSTOM_REPO_FILE to build some OpenShift images, e.g. ironic ones.
 
+# Use <IMAGE_NAME>_EXTRA_PACKAGES to set the path (relative to dev-scripts or
+# absolute) to a file with extra packages to install in the image, one per line.
+# At the moment, this option is supported with ironic-image and ironic-inspector-image
+# For example:
+# export IRONIC_EXTRA_PACKAGES=ironic-extra-pkgs.txt
+
 # Set this variable to point the custom base image to a different location
 # It can be an absolute path or a local path under the dev-scripts dir
 # export BASE_IMAGE_DIR=base-image


04_setup_ironic.sh
```diff
@@ -49,18 +49,32 @@ for IMAGE_VAR in $(env | grep "_LOCAL_IMAGE=" | grep -o "^[^=]*") ; do
...
+        MAIN_PKGS_FILE_NAME=${IMAGE_VAR/_LOCAL_IMAGE}_MAIN_PKGS
+        MAIN_PKGS_FILE=${!MAIN_PKGS_FILE_NAME:-"main-packages-list.txt"}
+        EXTRA_PKGS_FILE_NAME=${IMAGE_VAR/_LOCAL_IMAGE}_EXTRA_PACKAGES
+        EXTRA_PKGS_FILE=${!EXTRA_PKGS_FILE_NAME:-}
+        if [[ -n $EXTRA_PKGS_FILE ]]; then
+            # add newline at the end of file if not present
+            sed -i -e '$a\' "${REPOPATH}/${MAIN_PKGS_FILE}"
+            cat "$EXTRA_PKGS_FILE" >> "${REPOPATH}/${MAIN_PKGS_FILE}"
+        fi
+
```
andfasano, line 63: Hi Riccardo,
not sure if I missed something, but if it's for local testing (and thus there's already a repo checked out), why can't you modify directly the pkg base file? In any case here you have to define an extra file and store it somewhere. Can you please clarify thanks?

elfosardo: if you build an image locally, the image repository needs to be synced the first time, and that's done automatically by the script
you won't have any file to change the first time you run dev-scripts with locally built images, so without this you will need to manually clone the repo, modify the packages file, and then run the script
another option I considered is to modify the images  themselves to accept an "extra package" file directly as argument, that would simplify things, but it requires more work (need to change 3 images first, then dev-scripts).
I will probably do it anyway, depending on which direction we take to build the final images, but for now this is the fastest solution IMHO

andfasano: Am I missing something or #1140 could cover also such case, without the need of passing an extra file (but instead by directly modifying the Dockerfile)?

elfosardo: #1084 is needed just to import and test a PR for one or more images
it doesn't involve changing the Dockerfile

andfasano: As far I as understood this PR it's useful to test manually the changes to the extra packages _without_ checking out and modifying explicitly the target repo, and that's ok.
 
Assuming that the tested changes are ok, a new PR on the target repo will be still required, containing the same new packages defined in the extra file. Under this point of view #1084 appears to be more CI-friendly, allowing to "revert" the logical workflow and to test directly the desired changes in dev-script (ie, a new PR `x` could be opened for `ironic-image` with some changes to https://github.com/openshift/ironic-image/blob/master/main-packages-list.txt, and then another PR in dev-script could be used to test PR `x`, thanks to the fact that extra config variables can now be injected directly).

elfosardo: I think there's a bit of confusion here:
- this PR allows to install extra packages in the images during build time, no matter what they come from, and it's useful for testing packages before doing cross-tags, for example
- the #1084 allows to test PRs from the ironic container images no matter their content, but it can't test new packages if they're not cross-tagged first
they have 2 complete different scopes, this one is mainly for downstream testing, #1084 is more for upstream testing

andfasano: Right, but since technically speaking both of them are applied during the build time and could affect the same files as well, I'd suggest a couple of points:

1) Enrich a little bit the documentation and clarify the scope of usage of the config vars (in addition to the technical description, which is useful) 
2) For this PR, use an env var (instead of a file - or at least not just the file) to specify the extra packages to be tested: such approach could be more easily tested also via CI thanks to the usage of the `[DEVSCRIPTS_CONFIG](https://github.com/openshift/release/blob/0fb4497f32ee508cb8d692b2eae900611bbd889f/ci-operator/step-registry/baremetalds/devscripts/setup/baremetalds-devscripts-setup-ref.yaml#L10)` setting

Minor point: #1084 if defined at the same time could overwrite the context of _EXTRA_PACKAGES, probably it could be useful to add a check once this PR will be landed

elfosardo: > Right, but since technically speaking both of them are applied during the build time and could affect the same files as well, I'd suggest a couple of points:
> 
> 1. Enrich a little bit the documentation and clarify the scope of usage of the config vars (in addition to the technical description, which is useful)

Will do

> 2. For this PR, use an env var (instead of a file - or at least not just the file) to specify the extra packages to be tested: such approach could be more easily tested also via CI thanks to the usage of the `[DEVSCRIPTS_CONFIG](https://github.com/openshift/release/blob/0fb4497f32ee508cb8d692b2eae900611bbd889f/ci-operator/step-registry/baremetalds/devscripts/setup/baremetalds-devscripts-setup-ref.yaml#L10)` setting

If I get it right, I disagree on this.
If we use an env var and we have to specify multiple packages, it will be kind of hard to set and read.
A file with one rpm per line is very easy to prepare, and it's very clear to read
Please consider that the logic in the images repos will slightly change, so this also will become a little different.
I'm actually tempted to go the other way around, change the images first and then here, it probably makes more sense.

> 
> Minor point: #1084 if defined at the same time could overwrite the context of _EXTRA_PACKAGES, probably it could be useful to add a check once this PR will be landed

Yes, I know, unfortunately.
This also makes me wonder if I should just handle the changes to the images first and then come back here, after #1084 is merged.

elfosardo: @andfasano thanks for the discussion and the constructive feedback, it helped making me reconsider my approach
despite working correctly most of the times, in the end I feel like I need one more change on the images before moving forward

andfasano: Thanks, sounds good! :+1:

elfosardo:
this depends now on https://github.com/metal3-io/ironic-image/pull/227 https://github.com/metal3-io/ironic-inspector-image/pull/72
will update logic accordingly

elfosardo:
@stbenjam @andfasano hi! Can you please give another look at this? Thanks!

andfasano (approved):
Changes are fine to me. Leaving to @stbenjam final review

derekhiggins (changes requested):

04_setup_ironic.sh
```diff
@@ -73,14 +74,23 @@ for IMAGE_VAR in $(env | grep "_LOCAL_IMAGE=" | grep -o "^[^=]*") ; do
                 git fetch origin pull/${IMAGE_PR}/head:pr${IMAGE_PR}
                 git checkout pr${IMAGE_PR}
         fi
+        
+        # If we set extra packages, we add them to the main pkgs base file
+        EXTRA_PKGS_FILE_NAME=${IMAGE_VAR/_LOCAL_IMAGE}_EXTRA_PACKAGES
+        EXTRA_PKGS_FILE=${!EXTRA_PKGS_FILE_NAME:-}
+        if [[ -n $EXTRA_PKGS_FILE ]]; then
+            cp $EXTRA_PKGS_FILE "$REPOPATH"
```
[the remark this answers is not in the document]
elfosardo: it was definitely not doing it when I started working on this 3 months ago :)
Thanks for checking, I added a basename command afterwards that should fix that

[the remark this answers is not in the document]
derekhiggins: Thanks, this fixes the second problem, but relative path names still aren't working

```
[derekh@u07 dev-scripts]$ IRONIC_EXTRA_PACKAGES=3eer ./04_setup_ironic.sh 
+(./04_setup_ironic.sh:78): main(): cp 3eer /home/derekh/ironic-image
cp: cannot stat '3eer': No such file or directory

```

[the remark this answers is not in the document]
elfosardo: Relative path names cannot work as it is since we're changing dir into the image repo, unless the file is already present in the image repo, but then it would be useless to copy it there.
The idea is to have the file in a place that is not the image repo and then copy it inside the image repo.
I can add 'absolute' before 'path' to the comment and to the README to be more explicit.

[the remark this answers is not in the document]
derekhiggins: > Relative path names cannot work as it is since we're changing dir into the image repo, unless the file is already present in the image repo, but then it would be useless to copy it there.
> The idea is to have the file in a place that is not the image repo and then copy it inside the image repo.
> I can add 'absolute' before 'path' to the comment and to the README to be more explicit, or define in advance a path where the file should be located, removing some degree of freedom.

We could use OLDPWD and realpath to convert to an absolute path name, e.g.
`EXTRA_PKGS_FILE=$(cd $OLDPWD ; realpath $EXTRA_PKGS_FILE)`

but if that wont work for some reason I'm happy with mentioning it in config_example.sh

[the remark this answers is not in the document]
elfosardo: doesn't this imply that the file would necessary be in $OLDPWD ?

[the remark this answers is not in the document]
derekhiggins: It implies that the EXTRA_PKGS_FILE is either absolute or is relative to  $OLDPWD, The script was run from $OLDPWD so either should work.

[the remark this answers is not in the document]
elfosardo: so if EXTRA_PKGS_FILE is not in the dir where we run the script from (I suppose dev-scripts repo dir), it won't work
if we go for this solution we need to modify the comments and the README, which I don't mind doing
having the EXTRA_PKGS_FILE under dev-scripts was my first intention anyway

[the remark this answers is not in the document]
derekhiggins: > so if EXTRA_PKGS_FILE is not in the dir where we run the script from (I suppose dev-scripts repo dir), it won't work
> if we go for this solution we need to modify the comments and the README, which I don't mind doing
> having the EXTRA_PKGS_FILE under dev-scripts was my first intention anyway

I don't see the case where it wouldn't work, see examples below, once the path is correct relative to where the script was run its ok

```
[derekh@u07 dev-scripts]$ cat test.sh
EXTRA_PKGS_FILE=$1
cd /tmp
EXTRA_PKGS_FILE=$(cd $OLDPWD ; realpath $EXTRA_PKGS_FILE)
echo $EXTRA_PKGS_FILE

[derekh@u07 dev-scripts]$ bash test.sh afile
/home/derekh/dev-scripts/afile
[derekh@u07 dev-scripts]$ bash test.sh ../afile
/home/derekh/afile
[derekh@u07 dev-scripts]$ bash test.sh /tmp/afile
/tmp/afile
[derekh@u07 dev-scripts]$ bash test.sh ~/ironic-image/extra-package
/home/derekh/ironic-image/extra-package
```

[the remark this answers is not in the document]
elfosardo: alright, I'll apply the change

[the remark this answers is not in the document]
derekhiggins: thanks

config_example.sh
```diff
@@ -23,6 +23,11 @@ set -x
 # NOTE: If a checkout already exists in $HOME, it won't be re-created.
 # NOTE: You must set CUSTOM_REPO_FILE to build some OpenShift images, e.g. ironic ones.
 
+# Use <IMAGE_NAME>_EXTRA_PACKAGES to set the path (relative to dev-scripts or
+# absolute) to a file with extra packages to install in the image, one per line.
+# For example:
+# export IRONIC_IPA_DOWNLOADER_EXTRA_PACKAGES=ipa-downloader-extra-pkgs.txt
```
hardys, line 29: I'm not sure this is a good example - AFAICS the IPA downloader Dockerfile doesn't support EXTRA_PACKAGES?

https://github.com/openshift/ironic-ipa-downloader/blob/master/Dockerfile.ocp#L6

vs

https://github.com/openshift/ironic-image/blob/master/Dockerfile.ocp#L22

So unless this is something now supported in the base image, we should probably list the containers which will work with this interface in the example config?

elfosardo: you're right, this is a leftover from another era, thanks, I'll change it

elfosardo:
@derekhiggins @stbenjam @hardys any chance you can have another look at this please? Thanks!

elfosardo:
@stbenjam @hardys looking for approval :)

dtantsur (commented):
/approve

PR #1143, opened 2020-11-06
update local-bmo.sh with new CRD location

dhellmann:
As part of the kubebuilder migration, the location of the CRD in the
baremetal-operator repository moved. Update the script for running the
operator locally to look at the new location.

/cc @zaneb


dhellmann:
We should take #1144 instead.

/close

PR #1144, opened 2020-11-09
Fix local-bmo script

andfasano:
Sync'up with the latest changes on BMO, fixed the script


metal3-dev/local-bmo.sh
```diff
@@ -28,6 +28,7 @@ fi
 
 # Stop the machine-api-operator so it does not try to fix the
 # deployment we are going to change.
+cd $SCRIPTDIR
```
dhellmann, line 31: Is this needed?

andfasano: Without didn't work for me

PR #1145, opened 2020-11-09
Remove ARG default for custom repo file in base-image

elfosardo:
The bash way doesn't always work and could cause issues during build time.
Make TEST_REPO ARG mandatory for base-image.

### DIFF
--- a/base-image/Dockerfile
+++ b/base-image/Dockerfile
@@ -1,8 +1,8 @@
-ARG TestRepo
+ARG TEST_REPO
 
 FROM ubi8
 
 RUN rm -f /etc/yum.repos.d/*
-COPY ${TestRepo:-"custom.repo"} /etc/yum.repos.d/
+COPY $TEST_REPO /etc/yum.repos.d/
 
 RUN dnf upgrade -y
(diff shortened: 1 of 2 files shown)


dtantsur (approved):

base-image/Dockerfile
```diff
@@ -1,8 +1,8 @@
-ARG TestRepo
+ARG TestRepo=custom.repo
```
hardys, line 1: This does seem to fix the issues with recent recent podman versions - I wonder should we just remove the custom.repo default, since it's empty anyway and that may result in a clearer failure if problems happen again?

I think the reason the previous issue wasn't clear is we silently copied custom.repo and ignored the actual TestRepo, so if TestRepo was undefined the COPY would fail and it'd be clearer what the problem was?

hardys:
/approve

This wfm locally and does solve the problems with recent podman versions - one comment added @elfosardo - should we just remove the default to make future failures clearer?

elfosardo:
> This wfm locally and does solve the problems with recent podman versions - one comment added @elfosardo - should we just remove the default to make future failures clearer?

I see what you mean, so we just make TestRepo mandatory
Yeah, I think it makes sense, I will change it here

hardys:
@elfosardo thanks this lgtm - I did notice another issue while testing though - this approach doesn't work for all images for example the baremetal-operator won't build with the base image:

https://github.com/openshift/baremetal-operator/blob/master/Dockerfile.ocp#L1

The issue is the base image doesn't have the tools in the builder image e.g make etc - I wonder if we need to make this code build a different base image based on the FROM in each Dockerfile?

Or we could not replace builder images I guess since in general I don't think they'll install anything?

PR #1146, opened 2020-11-09
Update version detection and make dnsVIP conditional

hardys:
Updates detection of the openshift version because although users can set OPENSHIFT_VERSION, and we can derive it from most user specified release versions, the release images generated in CI don't work because they look like `registry.build02.ci.openshift.org/ci-op-xyz`

Instead we can save the release info locally, and look at the release version directly, which should result in version logic that works both for local testing and in CI.

### DIFF
--- a/03_build_installer.sh
+++ b/03_build_installer.sh
@@ -18,6 +18,8 @@ extract_oc "${OPENSHIFT_RELEASE_IMAGE}"
 
 mkdir -p $OCP_DIR
 
+save_release_info ${OPENSHIFT_RELEASE_IMAGE} ${OCP_DIR}
+
 if [ -z "$KNI_INSTALL_FROM_GIT" ]; then
   # Extract openshift-install from the release image
   extract_installer "${OPENSHIFT_RELEASE_IMAGE}" $OCP_DIR
--- a/ocp_install_env.sh
+++ b/ocp_install_env.sh
@@ -36,6 +36,24 @@ function extract_installer() {
     extract_command openshift-baremetal-install "$1" "$2"
 }
 
+function save_release_info() {
+    local release_image
+    local outdir
+
+    release_image="$1"
+    outdir="$2"
+
+    oc adm release info --registry-config "$PULL_SECRET_FILE" "$release_image" -o json > ${outdir}/release_info.json
+}
+
+function openshift_release_version() {
+    jq -r ".config.config.Labels.\"io.openshift.release\"" ${OCP_DIR}/release_info.json
+}
+
+function image_for() {
+    jq -r ".references.spec.tags[] | select(.name == \"$1\") | .from.name" ${OCP_DIR}/release_info.json
+}
+
 function extract_rhcos_json() {
     local release_image
     local outdir
@@ -43,7 +61,7 @@ function extract_rhcos_json() {
     release_image="$1"
     outdir="$2"
 
-    baremetal_image=$(oc adm release info --image-for=baremetal-installer --registry-config "$PULL_SECRET_FILE" "$release_image")
+    baremetal_image=$(image_for baremetal-installer)
     baremetal_container=$(podman create --authfile "$PULL_SECRET_FILE" "$baremetal_image")
 
     # This is OK to fail as rhcos.json isn't available in every release,
--- a/ocp_install_env.sh
+++ b/ocp_install_env.sh
@@ -114,6 +114,15 @@ EOF
   fi
 }
 
+function dnsvip() {
+  # dnsVIP was removed from 4.5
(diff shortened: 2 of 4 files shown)


03_build_installer.sh
```diff
@@ -18,6 +18,8 @@ extract_oc "${OPENSHIFT_RELEASE_IMAGE}"
 
 mkdir -p $OCP_DIR
 
+save_release_info ${OPENSHIFT_RELEASE_IMAGE} ${OCP_DIR}
+
```
hardys, line 22: Perhaps I should remove OCP_DIR here - I was aligning with the style used here but noticed after that OCP_DIR is actually already consumed directly in several functions defined in `ocp_install_env.sh`

ocp_install_env.sh
```diff
@@ -36,14 +36,39 @@ function extract_installer() {
...
+
+    release_image="$1"
+    outdir="$2"
+
+    oc adm release info --registry-config "$PULL_SECRET_FILE" "$release_image" -o json > ${outdir}/release_info.json
+}
+
+# Gives e.g 4.7.0-0.nightly-2020-10-27-051128
+function openshift_release_version() {
+    jq -r ".config.config.Labels.\"io.openshift.release\"" ${OCP_DIR}/release_info.json
```
andfasano, line 51: What about using ".metadata.version" ?

hardys: AFAICS they return the same value so sure we could do that, any idea why there are both keys?

```
$ jq .metadata.version,.config.config.Labels.\"io.openshift.release\" release_info.json 
"4.7.0-0.nightly-2020-11-10-061115"
"4.7.0-0.nightly-2020-11-10-061115"
```

FWIW I just copied this existing path from the `04` script where it parsed the output of `oc adm release info`

andfasano: AFAIK the `.metadata.version` should describe the current version as per the [Cincinnati upgrade protocol](https://github.com/openshift/cincinnati/blob/master/docs/design/cincinnati.md#update-graph) - but probably as you noted they contain the very same version

ocp_install_env.sh
```diff
@@ -186,7 +223,7 @@ $(baremetal_network_configuration)
     clusterOSImage: http://$(wrap_if_ipv6 $MIRROR_IP)/images/${MACHINE_OS_IMAGE_NAME}?sha256=${MACHINE_OS_IMAGE_SHA256}
     apiVIP: ${API_VIP}
     ingressVIP: ${INGRESS_VIP}
-    dnsVIP: ${DNS_VIP}
+$(dnsvip)
```
zaneb, line 226: Might be safer to do the indentation here rather than in the function.

ocp_install_env.sh
```diff
@@ -89,6 +114,18 @@ EOF
   fi
 }
 
+function dnsvip() {
+  local version
+
+  # dnsVIP was removed from 4.5
+  version=$(openshift_version | sed "s/\.//")
+  if [[ $version < 45 ]]; then
```
zaneb, line 122: technically `(( ))` compares numbers and `[[ ]]` compares strings, so the former appears to be what you were going for by removing the `.`?

zaneb (commented):
/approve

ocp_install_env.sh
hardys: Ah yes you're right, or I guess we could test it with sort like in https://github.com/openshift-metal3/dev-scripts/blob/master/04_setup_ironic.sh#L176

ocp_install_env.sh
hardys: Yeah I did think the same but all of the other functions adding install-config snippets work like this so I decided to align for consistency.

ocp_install_env.sh
hardys: Updated to use sort instead

ocp_install_env.sh
```diff
@@ -36,14 +36,39 @@ function extract_installer() {
...
+
+# Gives e.g 4.7.0-0.nightly-2020-10-27-051128
+function openshift_release_version() {
+    jq -r ".metadata.version" ${OCP_DIR}/release_info.json
+}
+
+# Gives us e.g 4.7 because although OPENSHIFT_VERSION is set by users,
+# but is not set in CI
+function openshift_version() {
+    jq -r ".metadata.version" ${OCP_DIR}/release_info.json | head -c3
```
andfasano, line 57: Probably not very urgent, but what about instead using something like:

`jq -r ".metadata.version" ${OCP_DIR}/release_info.json | grep -oP "\d\.\d+"` ?

It will avoid report a wrong number when version will be > 4.9

hardys: Good point, will do!

PR #1152, opened 2020-11-17
Additional trust bundle

Avielyo10:
Adding the option to add certificate bundle that will be added to the nodes' trusted certificate store.

### DIFF
--- a/config_example.sh
+++ b/config_example.sh
@@ -188,3 +188,8 @@ set -x
 
 # Image reference for installing hive. See hive_install.sh.
 #export HIVE_DEPLOY_IMAGE="registry.svc.ci.openshift.org/openshift/hive-v4.0:hive"
+
+# PEM-encoded X.509 certificate bundle that will be added to the nodes' trusted 
+# certificate store. This trust bundle may also be used when a proxy has 
+# been configured.
+# export ADDITIONAL_TRUST_BUNDLE=$(cat ca)
--- a/ocp_install_env.sh
+++ b/ocp_install_env.sh
@@ -98,10 +98,10 @@ EOF
 }
 
 function additional_trust_bundle() {
-    if [ ! -z "$ADDITIONAL_TRUST_BUNDLE" ]; then
+    if [ ! -z "$ADDITIONAL_TRUST_BUNDLE" ] && [ -z $MIRROR_IMAGES ]; then
 cat <<EOF
 additionalTrustBundle: |
-  $(echo ${ADDITIONAL_TRUST_BUNDLE} | sed -e ':a;N;$!ba;s/\n/\n  /g')
+$(echo ${ADDITIONAL_TRUST_BUNDLE} | awk '{ print " ", $0 }')
 EOF
     fi
 }
--- a/ocp_install_env.sh
+++ b/ocp_install_env.sh
@@ -103,6 +103,10 @@ cat <<EOF
 additionalTrustBundle: |
 $(echo ${ADDITIONAL_TRUST_BUNDLE} | awk '{ print " ", $0 }')
 EOF
+    elif [ ! -z "$ADDITIONAL_TRUST_BUNDLE" ] && [ ! -z "$MIRROR_IMAGES" ]; then
+cat <<EOF
+$(echo ${ADDITIONAL_TRUST_BUNDLE} | awk '{ print " ", $0 }')
+EOF
     fi
 }
 
--- a/ocp_install_env.sh
+++ b/ocp_install_env.sh
@@ -97,6 +97,15 @@ EOF
     fi
 }
 
+function additional_trust_bundle() {
+    if [ ! -z "$ADDITIONAL_TRUST_BUNDLE" ]; then
+cat <<EOF
+additionalTrustBundle: |
+  $(echo ${ADDITIONAL_TRUST_BUNDLE} | sed -e ':a;N;$!ba;s/\n/\n  /g')
+EOF
+    fi
+}
+
 function cluster_network() {
   if [[ "${IP_STACK}" == "v4" ]]; then
 cat <<EOF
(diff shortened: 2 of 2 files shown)


ocp_install_env.sh
```diff
@@ -97,6 +97,15 @@ EOF
     fi
 }
 
+function additional_trust_bundle() {
+    if [ ! -z "$ADDITIONAL_TRUST_BUNDLE" ]; then
+cat <<EOF
+additionalTrustBundle: |
+  $(echo ${ADDITIONAL_TRUST_BUNDLE} | sed -e ':a;N;$!ba;s/\n/\n  /g')
```
hardys, line 104: This will create a duplicate section when `image_mirror_config` is called (when MIRROR_IMAGES or ipv6 is specified), we'll need some way to merge the content.

hardys:
Thanks for the PR @Avielyo10 - I made a comment since we do already configure this when using a local registry, can you add this additional var in a way that's compatible with that please?

Also interested to understand the reason this is needed, can you expand on what you're testing at all?

ocp_install_env.sh
Avielyo10: Why duplicated?does `image_mirror_config` adds this section?

Avielyo10:
@hardys We are testing OCP deployment & upgrades with our operators, so instead of deploying the cluster and then Adding the certs and waiting for the MC to finish we can do it in one shot

ocp_install_env.sh
hardys: Yes exactly, it already adds this section https://github.com/openshift-metal3/dev-scripts/blob/master/utils.sh#L331

Avielyo10:
@hardys Please review :)

ocp_install_env.sh
```diff
@@ -98,10 +98,10 @@ EOF
 }
 
 function additional_trust_bundle() {
-    if [ ! -z "$ADDITIONAL_TRUST_BUNDLE" ]; then
+    if [ ! -z "$ADDITIONAL_TRUST_BUNDLE" ] && [ -z $MIRROR_IMAGES ]; then
```
hardys, line 101: This still won't work when MIRROR_IMAGES is set (this is a very common option) - image_mirror_config writes the mirror config and it ends like:

```
additionalTrustBundle: |
${INDENTED_CERT}
```

So here we need to not write the `additionalTrustBundle: |` part when MIRROR_IMAGES is set, something like:

```
    if [ ! -z "$ADDITIONAL_TRUST_BUNDLE" ] && [ -z "$MIRROR_IMAGES" ]; then
cat <<EOF
additionalTrustBundle: |
    $(echo ${ADDITIONAL_TRUST_BUNDLE} | awk '{ print " ", $0 }')
EOF
    elif [ ! -z "$ADDITIONAL_TRUST_BUNDLE" ] && [ ! -z "$MIRROR_IMAGES" ]; then
cat <<EOF
    $(echo ${ADDITIONAL_TRUST_BUNDLE} | awk '{ print " ", $0 }')
EOF
    fi
```

Avielyo10: @hardys With this code you actually say that  `$MIRROR_IMAGES` is not affecting at all.

By looking at [image_mirror_config function](https://github.com/openshift-metal3/dev-scripts/blob/master/utils.sh#L316-L343) we are skipping the whole function if `$MIRROR_IMAGES` is empty.
On my function we are adding `additionalTrustBundle` iff `$MIRROR_IMAGES` is empty. Therefore if `$MIRROR_IMAGES` is not empty, my function will be skipped & `image_mirror_config` won't.

Avielyo10: @hardys Understood, done.

hardys: Thanks!

Avielyo10:
@hardys @dhellmann @zaneb Can this get /lgtm?

ocp_install_env.sh
```diff
@@ -97,6 +97,19 @@ EOF
...
+cat <<EOF
+additionalTrustBundle: |
+$(echo ${ADDITIONAL_TRUST_BUNDLE} | awk '{ print " ", $0 }')
+EOF
+    elif [ ! -z "$ADDITIONAL_TRUST_BUNDLE" ] && [ ! -z "$MIRROR_IMAGES" ]; then
+cat <<EOF
+$(echo ${ADDITIONAL_TRUST_BUNDLE} | awk '{ print " ", $0 }')
+EOF
+    fi
```
zaneb, line 110: A less complicated version of this function:
```
if [ -n "${ADDITIONAL_TRUST_BUNDLE}" ]; then
    if [ -z "${MIRROR_IMAGES}" ]; then
        echo "additionalTrustBundle: |"
    fi
    echo "${ADDITIONAL_TRUST_BUNDLE}" | awk '{ print " ", $0 }'
fi
```

config_example.sh
```diff
@@ -188,3 +188,8 @@ set -x
 
 # Image reference for installing hive. See hive_install.sh.
 #export HIVE_DEPLOY_IMAGE="registry.svc.ci.openshift.org/openshift/hive-v4.0:hive"
+
+# PEM-encoded X.509 certificate bundle that will be added to the nodes' trusted 
+# certificate store. This trust bundle may also be used when a proxy has 
+# been configured.
+# export ADDITIONAL_TRUST_BUNDLE=$(cat ca)
```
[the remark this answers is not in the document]
Avielyo10: We can say the same on ssh keys

ocp_install_env.sh
Avielyo10: Agree

config_example.sh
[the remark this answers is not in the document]
Avielyo10: @zaneb I believe you are right, I'll work to fix this

PR #1153, opened 2020-11-19
Cleanup /var/lib/libvirt/openshift-images

hardys:
For some reason the virsh --remove-all-storage isn't actually removing
the storage, this may be because we moved to a non-default pool?

This can quickly use up space as each leaked directory is about 2.2G
so lets clean it up until we can find a more robust way to handle
this in the installer.

### DIFF
--- a/ocp_cleanup.sh
+++ b/ocp_cleanup.sh
@@ -49,6 +49,12 @@ for vm in $(sudo virsh list --all --name | grep "^${CLUSTER_NAME}.*bootstrap"); 
   sudo virsh destroy $vm
   sudo virsh undefine $vm --remove-all-storage
 done
+
+# For some reason --remove-all-storage doesn't actually remove the storage...
+if [ -d /var/lib/libvirt/openshift-images ]; then
+  sudo rm -fr /var/lib/libvirt/openshift-images/${CLUSTER_NAME}-*
+fi
+
 # The .ign volume isn't deleted via --remove-all-storage
 VOLS="$(sudo virsh vol-list --pool default | awk '{print $1}' | grep "^${CLUSTER_NAME}.*bootstrap")"
 for v in $VOLS; do


ocp_cleanup.sh
```diff
@@ -49,6 +49,12 @@ for vm in $(sudo virsh list --all --name | grep "^${CLUSTER_NAME}.*bootstrap");
   sudo virsh destroy $vm
   sudo virsh undefine $vm --remove-all-storage
 done
+
+# For some reason --remove-all-storage doesn't actually remove the storage...
+if [ -d /var/lib/libvirt/openshift-images ]; then
+  sudo rm -f /var/lib/libvirt/openshift-images/${CLUSTER_NAME}-*
+fi
+
```
hardys, line 57: When https://github.com/openshift-metal3/dev-scripts/pull/1146 lands we can perhaps do something smarter since the pool isn't default from 4.6 onwards so the logic below could handle this?

celebdor: In my dsal machine I see
```
/var/lib/libvirt/openshift-images/ostest-nj5l9-bootstrap/ostest-nj5l9-bootstrap
```
Doesn't that mean that we should have rm work recursively?

hardys: Yes, good point - thanks!

cybertron:
Interestingly, I no longer see this happening in my environments. On the node I already cleaned up, I have (with a cluster actively running):

$ ls -lah /var/lib/libvirt/openshift-images/
total 8.0K
drwxr-xr-x.  2 root root 4.0K Nov 18 17:50 .
drwxr-xr-x. 10 root root 4.0K Nov 18 15:03 ..
$

On another node that I didn't clean up I just have a bunch of empty directories there. I wonder if there was a bug that affected deployments during a certain window of time that left a bunch of cruft, but has since been fixed.

hardys:
> Interestingly, I no longer see this happening in my environments. On the node I already cleaned up, I have (with a cluster actively running):

I think it's a corner case where if the installer fails during the terraform apply phase, it then doesn't clean up properly in the `destroy cluster` step - if your deployment succeeded then I'd expect the existing cleanup to work, but if you somehow cause the deployment of masters to fail I suspect it won't?

cybertron:
/lgtm

Fair enough. It shouldn't hurt anything to clean these up in any case.

elfosardo (approved):

Issue #1154, opened 2020-11-19
Error when chown runs and user name differs from group name


mrhillsman:
There are various locations where chown runs with $USER:$USER but the group is not the same as the username. An example is where this causes the initial copying of dev-scripts to /opt directory to set improper permissions and the make command fails.

PR #1155, opened 2020-11-19
Add GROUP variable for chown where different from USER

mrhillsman:
`whoami` does not always match the group name for
the user. Use `id -gn` to set the group name properly.
Also create GROUP environment variable and replace
USER where appropriate for chown commands.

Fix openshift-metal3/dev-scripts#1154

Signed-off-by: Melvin Hillsman <mrhillsman@redhat.com>

### DIFF
--- a/02_configure_host.sh
+++ b/02_configure_host.sh
@@ -80,7 +80,7 @@ ansible-playbook \
 if [ ${NUM_EXTRA_WORKERS} -ne 0 ]; then
   ORIG_NODES_FILE="${NODES_FILE}.orig"
   cp -f ${NODES_FILE} ${ORIG_NODES_FILE}
-  sudo chown -R $USER:$USER ${NODES_FILE}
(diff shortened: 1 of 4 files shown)


dhellmann:
Hi, @mrhillsman!

/ok-to-test

hardys:
Thanks for the fix @mrhillsman! LGTM except please can you fix the commit message, it should be something like:

```
    Fix incorrect group name

     `whoami` does not always match the group name for
    the user. Use `id -gn` to set the group name properly.
    Also create GROUP environment variable and replace
    USER where appropriate for chown commands.
```

mrhillsman:
> Thanks for the fix @mrhillsman! LGTM except please can you fix the commit message, it should be something like:
> 
> ```
>     Fix incorrect group name
> 
>      `whoami` does not always match the group name for
>     the user. Use `id -gn` to set the group name properly.
>     Also create GROUP environment variable and replace
>     USER where appropriate for chown commands.
> ```

Done

PR #1157, opened 2020-11-20
Fix the URL for GA releases

pliurh:


common.sh
```diff
@@ -115,7 +115,7 @@ function get_latest_ci_image() {
 # if we provide OPENSHIFT_RELEASE_IMAGE, do not curl. This is needed for offline installs
 if [ -z "${OPENSHIFT_RELEASE_IMAGE:-}" ]; then
   if [[ "$OPENSHIFT_RELEASE_TYPE" == "ga" ]]; then
-    LATEST_CI_IMAGE=$(curl -L https://mirror.openshift.com/pub/openshift-v4/clients/ocp/${OPENSHIFT_VERSION}/release.txt  | grep -o 'quay.io/openshift-release-dev/ocp-release[^"]\+')
+    LATEST_CI_IMAGE=$(curl -L https://mirror.openshift.com/pub/openshift-v4/clients/ocp/stable-${OPENSHIFT_VERSION}/release.txt  | grep -o 'quay.io/openshift-release-dev/ocp-release[^"]\+')
```
hardys, line 118: This will break folks using a specific z-stream release e.g see https://mirror.openshift.com/pub/openshift-v4/clients/ocp/ - currently it's valid to say `OPENSHIFT_VERSION=4.6.5`

Perhaps we should make it possible to set `OPENSHIFT_VERSION=stable-4.6` as an alternative?

PR #1160, opened 2020-11-30
Improve clouds.yaml generation

hardys:
Currently this is only generated on a successful deployment, but the time you need it most is during/after a failed deploy for debugging.

So try to generate a clouds.yaml (including for the bootstrap VM) even on a failed deploy, and add a script that allows us to generate the file during a deployment if needed.

### DIFF
--- a/generate_clouds_yaml.sh
+++ b/generate_clouds_yaml.sh
@@ -1,0 +1,9 @@
+#!/usr/bin/env bash
+set -xe
+
+source common.sh
+source utils.sh
+source network.sh
+source ocp_install_env.sh
+
+generate_auth_template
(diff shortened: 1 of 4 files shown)


generate_clouds_yaml.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+set -xe
+
+source common.sh
+source utils.sh
+source network.sh
+source ocp_install_env.sh
+
+generate_auth_template
```
andfasano, line 9: I think this script could be really useful, maybe it's worth adding some notes on the documentation about it?

andfasano (commented):
In general looks a pretty useful feature

generate_clouds_yaml.sh
hardys: Good point, I added a docs update

PR #1167, opened 2020-12-03
Fix CUSTOM_REPO_FILE arg

elfosardo:
When passed to the build command, the CUSTOM_REPO_FILE arg should
not be enclosed in quotes, otherwise the parser in base-image will
give an error.


elfosardo:
obsoleted by #1165

PR #1168, opened 2020-12-03
add host cleanup commands to revert NetworkManager dnsmasq config

dustinblack:
A deployment would lay down the `/etc/NetworkManager/conf.d/dnsmasq.conf` file to enable the NetworkManager dnsmasq service, but a `make clean` would not clean up this configuration and revert the system back to a state without NetworkManager dnsmasq. This simply adds the additional cleanup to the `host_cleanup.sh` script.

Also corrected a function name typo.

### DIFF
--- a/host_cleanup.sh
+++ b/host_cleanup.sh
@@ -32,6 +32,14 @@ ansible-playbook \
     -b -vvv ${VM_SETUP_PATH}/teardown-playbook.yml
 
 sudo rm -rf /etc/NetworkManager/dnsmasq.d/openshift-${CLUSTER_NAME}.conf /etc/yum.repos.d/delorean*
+sudo rm -rf /etc/NetworkManager/conf.d/dnsmasq.conf
+sudo rm -rf /etc/NetworkManager/dnsmasq.d/upstream.conf
+if systemctl is-active --quiet NetworkManager; then
+  sudo systemctl reload NetworkManager
+else
+  sudo systemctl restart NetworkManager
+fi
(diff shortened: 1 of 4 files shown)


host_cleanup.sh
```diff
@@ -32,6 +32,14 @@ ansible-playbook \
     -b -vvv ${VM_SETUP_PATH}/teardown-playbook.yml
 
 sudo rm -rf /etc/NetworkManager/dnsmasq.d/openshift-${CLUSTER_NAME}.conf /etc/yum.repos.d/delorean*
+sudo rm -rf /etc/NetworkManager/conf.d/dnsmasq.conf
```
andfasano, line 35: A question (@tsorya ?). The code below using the Ansible module `ini_file` seems to indicate the intention to append a new entry in case of existing file; in such case maybe for consistency it could be useful to revert just that change (ie `ansible localhost -b -m ini_file -a "path=/etc/NetworkManager/conf.d/dnsmasq.conf section=main option=dns value=dnsmasq state=absent", rather than removing the entire config file at all?

hardys: Yeah I thought the same but looking at the file there's no other content so removing seemed OK - also note if you're switching between dev-scripts and metal3-dev-env that it's removed here https://github.com/metal3-io/metal3-dev-env/blob/master/host_cleanup.sh#L56

andfasano: Ok.

/lgtm

PR #1172, opened 2020-12-10
Make online status of extra workers configurable

stuggi:
The generated ocp host manifest for the BMH extra workers specifies
the hosts to be powered on. This introduce EXTRA_WORKERS_ONLINE_STATUS
to config_$USER.sh to make the online status of the NUM_EXTRA_WORKERS
configurable. Default stays "online: true".


hardys:
Thanks @stuggi - instead of changing the hard-coded value, could you perhaps add a variable?

In my testing I find it preferable to have the hosts stay powered on, for example to look at the console output after [booting a coreos live iso](https://github.com/metal3-io/metal3-docs/pull/150) - if we add a variable folks can choose the behavior which suits their requirements?

hardys:
/ok-to-test
/approve

Thanks!

PR #1176, opened 2021-01-05
Add support for infinite DHCP leases

cybertron:
The proper way to do this would be to set the lease length in the
libvirt network config, but unfortunately the version of libvirt we
have in CentOS 8 right now doesn't support that. This adds a script
that manually modifies the libvirt dnsmasq configuration to make the
leases infinite. It is started in the background, and killed when
make clean is run. It can be enabled by setting INFINITE_LEASES=1.

### DIFF
--- a/02_configure_host.sh
+++ b/02_configure_host.sh
@@ -243,3 +243,13 @@ if [ "$DISABLE_MULTICAST" == "true" ]; then
         sudo ebtables -A OUTPUT --pkttype-type multicast -p ip6 --ip6-dst ${dst} -j DROP
     done
 fi
+
+# Hack to force use of infinite leases in versions of libvirt that don't
+# support setting the lease time in the network config.
+# This should be removed when we have a new enough version of libvirt to
+# do it properly. I believe any version 6.3.0 or above should have the feature.
+if [ ${INFINITE_LEASES:-0} -eq 1 ]
+then
+    ./infinite-leases.sh &
+    echo "$!" > "${WORKING_DIR}/infinite-lease-pid"
+fi
--- a/infinite-leases.sh
+++ b/infinite-leases.sh
@@ -1,0 +1,19 @@
+#!/bin/bash
+
+NETWORK_NAME=${CLUSTER_NAME}bm
+FILENAME=/var/lib/libvirt/dnsmasq/${NETWORK_NAME}.hostsfile
+
+# We need to keep running even after setting the infinite leases because
+# they are overwritten multiple times in the deployment process.
+while :
+do
+    if grep -q -v infinite $FILENAME
+    then
+        sudo perl -pi -e 's/(.*?)(,infinite|)$/\1,infinite/' $FILENAME
+
+        pid=$(ps aux | grep dnsmasq | grep "$NETWORK_NAME" | grep -v root | awk '{print $2}')
+        sudo kill -s SIGHUP $pid
+        echo "Added infinite leases to $FILENAME"
+    fi
+    sleep 10
+done
--- a/host_cleanup.sh
+++ b/host_cleanup.sh
@@ -60,3 +60,9 @@ sudo ebtables --flush
 
 # Kill any lingering proxy
 sudo pkill -f oc.*proxy
+
+# Kill the infinite lease script
+if [ -s "${WORKING_DIR}/infinite-lease-pid" ]; then
+    kill $(cat "${WORKING_DIR}/infinite-lease-pid")
+    rm "${WORKING_DIR}/infinite-lease-pid"
+fi


02_configure_host.sh
```diff
@@ -243,3 +243,13 @@ if [ "$DISABLE_MULTICAST" == "true" ]; then
...
 fi
+
+# Hack to force use of infinite leases in versions of libvirt that don't
+# support setting the lease time in the network config.
+# This should be removed when we have a new enough version of libvirt to
+# do it properly.
+if [ ${INFINITE_LEASES:-0} -eq 1 ]
+then
+    ./infinite-leases.sh &
+    echo "$!" > infinite-lease-pid
```
hardys, line 254: It'd probably be better to put the pid file into `$WORKING_DIR` I think?

https://github.com/openshift-metal3/dev-scripts/blob/master/common.sh#L72

cybertron: Yep, good point.

cybertron:
/retest

The error message includes approximately all the errors, so let's see if we can roll the dice and get a more useful one this time... :-P

infinite-leases.sh
```diff
@@ -0,0 +1,19 @@
...
+NETWORK_NAME=${CLUSTER_NAME}bm
+FILENAME=/var/lib/libvirt/dnsmasq/${NETWORK_NAME}.hostsfile
+
+# We need to keep running even after setting the infinite leases because
+# they are overwritten multiple times in the deployment process.
+while :
+do
+    if grep -q -v infinite $FILENAME
+    then
+        sudo perl -pi -e 's/(.*?)(,infinite|)$/\1,infinite/' $FILENAME
```
derekhiggins, line 12: adding this to the hostsfile means that we cant restart the libvirt network (I guess the lease-init script can't parse it), maybe we don't care but I thought it was worth pointing out.

```
[derekh@u07 dev-scripts]$ sudo virsh net-start ostestbm
error: Failed to start network ostestbm
error: internal error: Child process (VIR_BRIDGE_NAME=ostestbm /usr/sbin/dnsmasq --conf-file=/var/lib/libvirt/dnsmasq/ostestbm.conf --leasefile-ro --dhcp-script=/usr/libexec/libvirt_leaseshelper) unexpected exit status 11: 
dnsmasq: lease-init script returned exit code 1
[derekh@u07 dev-scripts]$ kill: not enough arguments
Added infinite leases to /var/lib/libvirt/dnsmasq/ostestbm.hostsfile

```
Also the scripts that generates the clouds.yaml no longer works, as it can't get the host IP addresses with virsh
```
[derekh@u07 dev-scripts]$ virsh net-dhcp-leases ostestbm
error: failed to get network 'ostestbm'
error: Network not found: no network with matching name 'ostestbm'
```

02_configure_host.sh
```diff
@@ -243,3 +243,13 @@ if [ "$DISABLE_MULTICAST" == "true" ]; then
         sudo ebtables -A OUTPUT --pkttype-type multicast -p ip6 --ip6-dst ${dst} -j DROP
     done
 fi
+
+# Hack to force use of infinite leases in versions of libvirt that don't
+# support setting the lease time in the network config.
+# This should be removed when we have a new enough version of libvirt to
```
derekhiggins, line 249: Can we add what that version number would be?

derekhiggins (changes requested):
Some notes inline, also wondering if you can mention briefly why its needed.

infinite-leases.sh
```diff
@@ -0,0 +1,19 @@
...
+
+        pid=$(ps aux | grep dnsmasq | grep "$NETWORK_NAME" | grep -v root | awk '{print $2}')
+        sudo kill -s SIGHUP $pid
+        echo "Added infinite leases to $FILENAME"
```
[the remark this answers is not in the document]
cybertron: This isn't really designed to be run after deployment. The problem the script was designed to solve is the fact that the hosts file gets rewritten multiple times during the deployment, so if you manually added the leases they might get removed again. After deployment that's no longer a problem.

cybertron (commented):
This is to support a feature where we convert infinite DHCP leases to static configuration on the node. It's sort of a workaround for the fact that OpenShift doesn't have good support for static node networking configuration. The feature has already merged and this script is what I used while I was working on it.

02_configure_host.sh
```diff
@@ -243,3 +243,13 @@ if [ "$DISABLE_MULTICAST" == "true" ]; then
...
+# do it properly.
+if [ ${INFINITE_LEASES:-0} -eq 1 ]
+then
+    ./infinite-leases.sh &
```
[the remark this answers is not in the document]
derekhiggins: Sorry, I think I was mistaken when I mentioned 02_configure_host.sh , what I'm seeing is that after the deployment infinite-leases.sh is still running. If for some reason the loop to edit the leases file is executed then the terminal where I had run the make command from gets the output
"Added infinite leases to /var/lib/libvirt/dnsmasq/ostestbm.hostsfile"

infinite-leases.sh
derekhiggins: I guess if its not on by default then its not a big problem, and it'll have to be sorted somehow before getting switched on.

infinite-leases.sh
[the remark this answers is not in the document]
derekhiggins: So would it be ok to just kill the script at the end of the deployment?

cybertron:
It looks like we do have new enough libvirt on centos 8 now if we use the advanced-virtualization repo. That's probably better than this method. New PR pushed at https://github.com/openshift-metal3/dev-scripts/pull/1191

PR #1178, opened 2021-01-12
local-bmo.sh make BMO path configurable

hardys:
Adds the option to override the default path via the
BAREMETAL_OPERATOR_PATH variable.

This is useful when you want to modify the checkout consumed
via BAREMETAL_OPERATOR_LOCAL_IMAGE, which ends up in $HOME
not the regular GOPATH location.

### DIFF
--- a/metal3-dev/local-bmo.sh
+++ b/metal3-dev/local-bmo.sh
@@ -15,9 +15,9 @@ if ! which yq 2>&1 >/dev/null ; then
     exit 1
 fi
 
-bmo_path=$GOPATH/src/github.com/metal3-io/baremetal-operator
-if [ ! -d $bmo_path ]; then
-    echo "Did not find $bmo_path" 1>&2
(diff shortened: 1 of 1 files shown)


metal3-dev/local-bmo.sh
```diff
@@ -15,9 +15,9 @@ if ! which yq 2>&1 >/dev/null ; then
     exit 1
 fi
 
-bmo_path=$GOPATH/src/github.com/metal3-io/baremetal-operator
-if [ ! -d $bmo_path ]; then
-    echo "Did not find $bmo_path" 1>&2
+BMO_PATH=${BAREMETAL_OPERATOR_PATH:-$GOPATH/src/github.com/metal3-io/baremetal-operator}
```
andfasano, line 18: For documentation purposes it could be useful to add it also in the `config_example.sh` file

hardys: I did consider that, but it's not part of the regular `make` flow, so I figured it was more of an "advanced" case where folks will just look at the script?

PR #1179, opened 2021-01-18
Edit local-cbo to create required paths

ardaguclu:
* Add stopping of already running CBO.
* cluster-baremetal-operator fails when there is no path `/etc/cluster-baremetal-operator/images` in local. Added path creation when running `local-cbo.sh`
* `local-cbo.sh` also applies `0000_31_cluster-baremetal-operator_06_deployment.yaml`, which might conflict with the local operator and it is filtered.

### DIFF
--- a/metal3-dev/local-cbo.sh
+++ b/metal3-dev/local-cbo.sh
@@ -9,19 +9,8 @@ source $SCRIPTDIR/common.sh
 source $SCRIPTDIR/utils.sh
 
 # Stop already running CBO.
-pushd $SCRIPTDIR
+cd $SCRIPTDIR
 $SCRIPTDIR/stop-cbo.sh
-
-CBO_LOCAL_IMAGE_PATH="/etc/cluster-baremetal-operator/images"
-
-# Copy cbo-images.json into the CBO_LOCAL_IMAGE_PATH to run CBO locally. 
-# CBO needs image.json file in this path.
-if [[ ! -f "${CBO_LOCAL_IMAGE_PATH}/images.json" ]]; then
-    sudo mkdir -p $CBO_LOCAL_IMAGE_PATH
-    sudo cp ${OCP_DIR}/cbo-images.json ${CBO_LOCAL_IMAGE_PATH}/images.json
-fi
-
-popd
 
 cbo_path=$GOPATH/src/github.com/openshift/cluster-baremetal-operator
 if [ ! -d $cbo_path ]; then
@@ -33,11 +22,5 @@ fi
 cd $cbo_path
 
 export RUN_NAMESPACE=openshift-machine-api
-oc apply -f manifests/0000_31_cluster-baremetal-operator_01_images.configmap.yaml
-oc apply -f manifests/0000_31_cluster-baremetal-operator_02_metal3provisioning.crd.yaml
-oc apply -f manifests/0000_31_cluster-baremetal-operator_03_baremetalhost.crd.yaml
-oc apply -f manifests/0000_31_cluster-baremetal-operator_04_serviceaccount.yaml
-oc apply -f manifests/0000_31_cluster-baremetal-operator_05_rbac.yaml
-oc apply -f manifests/0000_31_cluster-baremetal-operator_07_clusteroperator.cr.yaml
-make run
+make run -e ${OCP_DIR}/cbo-images.json
 
--- a/metal3-dev/local-cbo.sh
+++ b/metal3-dev/local-cbo.sh
@@ -8,6 +8,21 @@ source $SCRIPTDIR/logging.sh
 source $SCRIPTDIR/common.sh
 source $SCRIPTDIR/utils.sh
 
+# Stop already running CBO.
+pushd $SCRIPTDIR
+$SCRIPTDIR/stop-cbo.sh
+
+CBO_LOCAL_IMAGE_PATH="/etc/cluster-baremetal-operator/images"
+
+# Copy cbo-images.json into the CBO_LOCAL_IMAGE_PATH to run CBO locally. 
+# CBO needs image.json file in this path.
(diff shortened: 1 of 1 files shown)


metal3-dev/local-cbo.sh
```diff
@@ -8,6 +8,21 @@ source $SCRIPTDIR/logging.sh
 source $SCRIPTDIR/common.sh
 source $SCRIPTDIR/utils.sh
 
+# Stop already running CBO.
+pushd $SCRIPTDIR
+$SCRIPTDIR/stop-cbo.sh
```
andfasano, line 13: nit: I would keep in line with https://github.com/openshift-metal3/dev-scripts/blob/21ec5582c73c4887f7bc85f5de079a2620a24b3b/metal3-dev/local-bmo.sh#L24-L27

metal3-dev/local-cbo.sh
```diff
@@ -8,6 +8,21 @@ source $SCRIPTDIR/logging.sh
...
+
+CBO_LOCAL_IMAGE_PATH="/etc/cluster-baremetal-operator/images"
+
+# Copy cbo-images.json into the CBO_LOCAL_IMAGE_PATH to run CBO locally. 
+# CBO needs image.json file in this path.
+if [[ ! -f "${CBO_LOCAL_IMAGE_PATH}/images.json" ]]; then
+    sudo mkdir -p $CBO_LOCAL_IMAGE_PATH
+    sudo cp ${OCP_DIR}/cbo-images.json ${CBO_LOCAL_IMAGE_PATH}/images.json
+fi
```
andfasano, line 22: What about instead passing directly to the make the location of the files? Ie

`make run -e ${OCP_DIR}/cbo-images.json`

This should save some code here

metal3-dev/local-cbo.sh
```diff
@@ -8,6 +8,21 @@ source $SCRIPTDIR/logging.sh
...
 
+# Stop already running CBO.
+pushd $SCRIPTDIR
+$SCRIPTDIR/stop-cbo.sh
+
+CBO_LOCAL_IMAGE_PATH="/etc/cluster-baremetal-operator/images"
+
+# Copy cbo-images.json into the CBO_LOCAL_IMAGE_PATH to run CBO locally. 
+# CBO needs image.json file in this path.
+if [[ ! -f "${CBO_LOCAL_IMAGE_PATH}/images.json" ]]; then
```
dhellmann, line 19: The operator takes a command line argument to pass this file, so we could avoid cluttering the real system files of the local host by putting it in /opt/dev-scripts or using the version in the OCP_DIR if we could pass that value to `make` on line 42. That may require changes in the cluster-baremetal-operator Makefile to update the `run` target.

metal3-dev/local-cbo.sh
```diff
@@ -18,6 +33,11 @@ fi
 cd $cbo_path
 
 export RUN_NAMESPACE=openshift-machine-api
-oc apply -f manifests/
+oc apply -f manifests/0000_31_cluster-baremetal-operator_01_images.configmap.yaml
+oc apply -f manifests/0000_31_cluster-baremetal-operator_02_metal3provisioning.crd.yaml
+oc apply -f manifests/0000_31_cluster-baremetal-operator_03_baremetalhost.crd.yaml
```
dhellmann, line 38: This CRD needs to match the version used by the BMO in the cluster. If we're not changing the BMO image in this script, we should keep the CRD that has already been installed when the cluster was built. If we're running a local BMO, the script to do that will update the CRD itself.

metal3-dev/local-cbo.sh
ardaguclu: Agree, that would save some code.

metal3-dev/local-cbo.sh
```diff
@@ -18,6 +33,11 @@ fi
...
+oc apply -f manifests/0000_31_cluster-baremetal-operator_04_serviceaccount.yaml
+oc apply -f manifests/0000_31_cluster-baremetal-operator_05_rbac.yaml
+oc apply -f manifests/0000_31_cluster-baremetal-operator_07_clusteroperator.cr.yaml
```
dhellmann, line 41: This will discard the state already managed by the CBO in its CO, and I don't think we want to do that.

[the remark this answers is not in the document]
ardaguclu: Out of sync problem exists in that case. However, there should be a way to filter `0000_31_cluster-baremetal-operator_06_deployment.yaml` to overcome the conflict problem. Probably I can list the yamls under `manifest` directory and apply.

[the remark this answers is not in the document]
ardaguclu: @andfasano Thanks for review.

ardaguclu: @dhellmann thanks for review. I understand your concerns about these commands. As far as I understand, your concerns are also valid for the previous version `oc apply -f manifests/`.

dhellmann: Yes, good point. We probably shouldn't have been doing that either. :-)

andfasano: Initially such approach was required as a way also to test the generated manifests (during a local development or CI). I think it could be reviewed now (maybe to be more granular), considering both the latest developments done on CBO and the fact our CI is slightly more robust on this side - thanks to the upgrade jobs (but maybe in a different PR)

ardaguclu: @andfasano @dhellmann now I remove `oc apply` to prevent possible incompatibilities in this PR and will work on granular changes in different PR. What do you think?

asalkeld:
looks ok to me

asalkeld (approved):
/lgtm

PR #1182, opened 2021-02-02
Split out HOST_IP_STACK from IP_STACK for dual-stack conversion testing

danwinship:
I'm starting to test single-stack-to-dual-stack conversion, which requires bringing up nodes that _could_ be dual-stack but aren't.

There are various things blocking the actual conversion from working, but I believe this part of it is correct.


network.sh
```diff
@@ -67,11 +68,26 @@ elif [[ -n "${CLUSTER_SUBNET}" ]]; then
 fi
 
 
-if [[ "$IP_STACK" = "v4" ]]
+if [[ "$HOST_IP_STACK" = "v4" ]]
```
elfosardo, line 71: this works, but maybe could be a good case in which we can use case :)

hardys: This is consistent with the current IP_STACK code so while arguably a case might be cleaner, such refactoring is probably best handled in a followup?

elfosardo: a follow up sounds good

PR #1183, opened 2021-02-08
Add ENABLE_LOCAL_REGISTRY config var to setup the local registry

andfasano:
The introduction of ENABLE_LOCAL_REGISTRY will allow then to setup the local registry both explicitly - by setting such variable - or in all the other previous situations where it was required).
This approach will be used to expand the OpenShift CI tests coverage, since it's required to run the e2e suites in disconnected mode (see [1]). To do that, the test workflow will mirror the required images to the local dev-scripts private registry.

[1] https://github.com/openshift/origin/pull/24887

### DIFF
--- a/02_configure_host.sh
+++ b/02_configure_host.sh
@@ -50,8 +50,11 @@ fi
 
 export ANSIBLE_FORCE_COLOR=true
 
+
 # Setup the registry for mirroring images
-if [[ ! -z "${MIRROR_IMAGES}" || $(env | grep "_LOCAL_IMAGE=")  || ! -z "${ENABLE_CBO_TEST}" ]]; then
+export ENABLE_LOCAL_REGISTRY=${ENABLE_LOCAL_REGISTRY:-${MIRROR_IMAGES:-${ENABLE_CBO_TEST:-$(env | grep "_LOCAL_IMAGE=")}}}
+
+if [[ ! -z "${ENABLE_LOCAL_REGISTRY}" ]]; then
     setup_local_registry
 fi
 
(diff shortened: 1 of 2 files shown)


02_configure_host.sh
```diff
@@ -50,8 +50,11 @@ fi
 
 export ANSIBLE_FORCE_COLOR=true
 
+
 # Setup the registry for mirroring images
-if [[ ! -z "${MIRROR_IMAGES}" || $(env | grep "_LOCAL_IMAGE=")  || ! -z "${ENABLE_CBO_TEST}" ]]; then
+export ENABLE_LOCAL_REGISTRY=${ENABLE_LOCAL_REGISTRY:-${MIRROR_IMAGES:-${ENABLE_CBO_TEST:-$(env | grep "_LOCAL_IMAGE=")}}}
```
hardys, line 55: Why do we need to add this new variable, when `MIRROR_IMAGES` already provides an interface to enable the local registry?

andfasano: The fact is that `MIRROR_IMAGES` in dev-scripts is enabled by default for IPv6 and dualstack configuration, but not for IPv4.
Since we need to enable the disconnected run mode for the CI tests in all the configurations, this approach would allow to have the registry available also in such case - without the need of mirroring the release images.

andfasano: Note: we could also enable `MIRROR_IMAGES` always for the CI tests, but this means paying additional 15 minutes also when not strictly required.

hardys: But you can set MIRROR_IMAGES in the IPv4 case, it's just not enabled by default, so why do we need a new variable that does basically the same thing?  Sorry, perhaps I'm missing something obvious here :)

andfasano: Yes, setting `MIRROR_IMAGES` for IPv4 case technically speaking will let [this change](https://github.com/openshift/release/blob/e569b6f4bf6575c6fbda0593236162f483a5f156/ci-operator/step-registry/baremetalds/e2e/test/baremetalds-e2e-test-commands.sh#L37-L38) work properly, but it will also fill the private registry with the _release_ images - something that is not really required for the IPv4 case.
The real feature from dev-scripts that we're really interested into it's just the local private registry, to be consumed by the `openshift-tests` runner for some of the e2e tests (which requires ie images like `"k8s.gcr.io/pause:3.2` that cannot be fetched by a Packet server)

hardys: Ah yes I see now, thanks for clarifying, so we can enable the registry, but skip the release mirroring that happens in `04_setup_ironic.sh`

PR #1186, opened 2021-02-09
Fix exporting provisioning ip for ipv4/ipv6 cases in local-bmo.sh

andfasano:
Env variables weren't exported correctly in the `local-bmo.sh` script, thus provoking an inspection failure due a parsing error.

### DIFF
--- a/metal3-dev/local-bmo.sh
+++ b/metal3-dev/local-bmo.sh
@@ -67,8 +67,9 @@ oc apply -f $OUTDIR/bmo-deployment-dev.yaml -n openshift-machine-api
 # Set some variables the operator expects to have in order to work
 export OPERATOR_NAME=baremetal-operator
 
+CLUSTER_IP=$(wrap_if_ipv6 $CLUSTER_PROVISIONING_IP)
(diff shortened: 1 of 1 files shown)


metal3-dev/local-bmo.sh
```diff
@@ -67,8 +67,13 @@ oc apply -f $OUTDIR/bmo-deployment-dev.yaml -n openshift-machine-api
 # Set some variables the operator expects to have in order to work
 export OPERATOR_NAME=baremetal-operator
 
+CLUSTER_IP=${CLUSTER_PROVISIONING_IP}
+if [[ "${IP_STACK}" == "v6" ]]; then 
+  CLUSTER_IP="["${CLUSTER_PROVISIONING_IP}"]"
```
derekhiggins, line 72: utils.sh has a function you can use directly, see wrap_if_ipv6

andfasano: Nice one, thanks!

elfosardo (approved):

iurygregory (approved):

PR #1188, opened 2021-02-16
Add ENABLE_LOCAL_REGISTRY check in image mirror config

ardaguclu:
For IPv4 environments, when `ENABLE_LOCAL_REGISTRY` is enabled, we are getting `x509: certificate signed by unknown authority` in CI workflows. This is due to the fact that `additionalTrustBundle` should be added as well as `imageContentSources` into the install-config, even when `MIRROR_IMAGES` is disabled.

This PR adds enabling this even when MIRROR_IMAGES is disabled environments.

### DIFF
--- a/ocp_install_env.sh
+++ b/ocp_install_env.sh
@@ -133,7 +133,7 @@ EOF
 
 function additional_trust_bundle() {
   if [ ! -z "$ADDITIONAL_TRUST_BUNDLE" ]; then
-    if [ -z $MIRROR_IMAGES ]; then
+    if [ -z "${MIRROR_IMAGES}" && -z "${ENABLE_LOCAL_REGISTRY}" ]; then
       echo "additionalTrustBundle: |"
     fi
     awk '{ print " ", $0 }' "${ADDITIONAL_TRUST_BUNDLE}"
--- a/utils.sh
+++ b/utils.sh
@@ -290,12 +290,12 @@ function generate_metal3_config {
 }
 
 function image_mirror_config {
-    if [ ! -z "${MIRROR_IMAGES}" ]; then
-        . /tmp/mirrored_release_image
-        TAGGED=$(echo $MIRRORED_RELEASE_IMAGE | sed -e 's/release://')
-        RELEASE=$(echo $MIRRORED_RELEASE_IMAGE | grep -o 'registry.ci.openshift.org[^":\@]\+')
+    if [ ! -z "${MIRROR_IMAGES}" || ! -z "${ENABLE_LOCAL_REGISTRY}" ]; then
         INDENTED_CERT=$( cat $REGISTRY_DIR/certs/$REGISTRY_CRT | awk '{ print " ", $0 }' )
-        if [ ! -s ${MIRROR_LOG_FILE} ]; then
+        if [ ! -z "${MIRROR_IMAGES}" && ! -s ${MIRROR_LOG_FILE} ]; then
+            . /tmp/mirrored_release_image
+            TAGGED=$(echo $MIRRORED_RELEASE_IMAGE | sed -e 's/release://')
+            RELEASE=$(echo $MIRRORED_RELEASE_IMAGE | grep -o 'registry.ci.openshift.org[^":\@]\+')
             cat << EOF
 imageContentSources:
 - mirrors:
--- a/02_configure_host.sh
+++ b/02_configure_host.sh
@@ -50,11 +50,7 @@ fi
 
 export ANSIBLE_FORCE_COLOR=true
 
-
-# Setup the registry for mirroring images
-export ENABLE_LOCAL_REGISTRY=${ENABLE_LOCAL_REGISTRY:-${MIRROR_IMAGES:-${ENABLE_CBO_TEST:-$(env | grep "_LOCAL_IMAGE=")}}}
-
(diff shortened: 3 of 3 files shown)


utils.sh
```diff
@@ -290,7 +290,7 @@ function generate_metal3_config {
 }
 
 function image_mirror_config {
-    if [ ! -z "${MIRROR_IMAGES}" ]; then
+    if [ ! -z "${ENABLE_LOCAL_REGISTRY:-${MIRROR_IMAGES:-}}" ]; then
```
andfasano, line 293: Is this whole block really required also in case of plain registry only? It seems it contains references to some stuff used only in case of mirroring release image (ie `local-release-mage`)

ocp_install_env.sh
```diff
@@ -133,7 +133,7 @@ EOF
 
 function additional_trust_bundle() {
   if [ ! -z "$ADDITIONAL_TRUST_BUNDLE" ]; then
-    if [ -z $MIRROR_IMAGES ]; then
+    if [ -z "${ENABLE_LOCAL_REGISTRY:-${MIRROR_IMAGES:-}}" ]; then
```
andfasano, line 136: `ENABLE_LOCAL_REGISTRY` is already defined also in terms of `MIRROR_IMAGES`. In case this can be reviewed https://github.com/openshift-metal3/dev-scripts/blob/5b8ec2c49578f4a47338b81471277664501665de/02_configure_host.sh#L55 and CBO check extracted

utils.sh
ardaguclu: Most probably you are right. For example, we might be using the inner blocks of `if [ ! -s ${MIRROR_LOG_FILE} ]; then` only. But do you think that we should separate the logic for MIRROR_IMAGES and ENABLE_LOCAL_REGISTRY?

ocp_install_env.sh
ardaguclu: What do you mean by extracting CBO?, extracting from 
`export ENABLE_LOCAL_REGISTRY=${ENABLE_LOCAL_REGISTRY:-${MIRROR_IMAGES:-${ENABLE_CBO_TEST:-$(env | grep "_LOCAL_IMAGE=")}}} ` or extracting at additional_trust_bundle function?

ardaguclu: Because if we extract CBO from here `export ENABLE_LOCAL_REGISTRY=${ENABLE_LOCAL_REGISTRY:-${MIRROR_IMAGES:-${ENABLE_CBO_TEST:-$(env | grep "_LOCAL_IMAGE=")}}} `, we need to add CBO_TEST into https://github.com/openshift-metal3/dev-scripts/blob/5b8ec2c49578f4a47338b81471277664501665de/02_configure_host.sh#L57

utils.sh
andfasano: Yes, they are two different features for two different scenarios. `MIRROR_IMAGES` is required for disconnected _release_ images (which is a mandatory condition for installing ipv6 scenario, but it's not for ipv4).
`ENABLE_LOCAL_REGISTRY` instead is meant to support disconnected _test_ images (which is something always required for all the test configurations we're supporting right now).

ocp_install_env.sh
andfasano: Sorry, meant to split the current definition of `ENABLE_LOCAL_REGISTRY` and keep it just as `export ENABLE_LOCAL_REGISTRY=${ENABLE_LOCAL_REGISTRY:-${MIRROR_IMAGES}}`, in the assumption that MIRROR_IMAGES indicates a super-feature of ENABLE_LOCAL_REGISTRY

dev-scripts-additional-config
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+export ENABLE_LOCAL_REGISTRY=true
```
andfasano, line 3: To test in isolation just ENABLE_LOCAL_REGISTRY I'd suggest setting also:
```
export IP_STACK=v4
export NETWORK_TYPE=OpenShiftSDN
```

Otherwise IPv6 will be enabled by default, and thus also MIRROR_IMAGES

utils.sh
```diff
@@ -290,12 +290,14 @@ function generate_metal3_config {
 }
 
 function image_mirror_config {
-    if [ ! -z "${MIRROR_IMAGES}" ]; then
-        . /tmp/mirrored_release_image
-        TAGGED=$(echo $MIRRORED_RELEASE_IMAGE | sed -e 's/release://')
-        RELEASE=$(echo $MIRRORED_RELEASE_IMAGE | grep -o 'registry.ci.openshift.org[^":\@]\+')
+    export ENABLE_LOCAL_REGISTRY=${ENABLE_LOCAL_REGISTRY:-${MIRROR_IMAGES:-}}
```
andfasano, line 293: Looks like this one could be removed?

ardaguclu: The best place is to export ENABLE_LOCAL_REGISTRY common.sh. However, network.sh runs after common.sh and overwrites MIRROR_IMAGES;
https://github.com/openshift-metal3/dev-scripts/blob/67a2d84d7cc22665202359288a85583a2fc9b99f/network.sh#L106

We can export ENABLE_LOCAL_REGISTRY in network.sh or in here.

PR #1191, opened 2021-02-17
Support lease expiry option

cybertron:
This was added to metal3-dev-env in [0]. Expose the setting for
dev-scripts too.

With the current version of libvirt available, we are not able to run
virsh net-dhcp-leases when infinite leases are enabled. I've added
logic to skip that step when LEASE_EXPIRY=0. It just means we
can't retrieve clouds.yaml from the bootstrap.
    
0: https://github.com/metal3-io/metal3-dev-env/pull/599

### DIFF
--- a/01_install_requirements.sh
+++ b/01_install_requirements.sh
@@ -19,7 +19,7 @@ if [ -z "${METAL3_DEV_ENV}" ]; then
   # TODO -- come up with a plan for continuously updating this
   # Note we only do this in the case where METAL3_DEV_ENV is
   # unset, to enable developer testing of local checkouts
-  git reset 8908da5241d52e25a7e1b2e60d6d604cf797f890 --hard
+  git reset 184d0afaab77424872f0441c9e11e23a620e5b57 --hard
   popd
 fi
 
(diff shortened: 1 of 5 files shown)


hardys:
Sorry this has been around for so long @cybertron - I'm just testing a metal3-dev-env patch, and noticed that we need this because the lease_expiry option is mandatory now in metal3-dev-env.

Could you update https://github.com/openshift-metal3/dev-scripts/blob/master/01_install_requirements.sh#L22 to pin to a recent/latest version which contains the lease-expiry change please?

I also needed the following galaxy role:

```
--- a/01_install_requirements.sh
+++ b/01_install_requirements.sh
@@ -43,7 +43,7 @@ sudo alternatives --set python /usr/bin/python3
 
 pushd ${METAL3_DEV_ENV_PATH}
 ansible-galaxy install -r vm-setup/requirements.yml
-ansible-galaxy collection install ansible.netcommon ansible.posix
+ansible-galaxy collection install ansible.netcommon ansible.posix community.general
```

hardys:
Also it'd probably be good to add the new variable to `config_example.sh` - and do you think we should perhaps prefix the name e.g `DHCP_LEASE_EXPIRY` ?

cybertron:
Okay, once I remembered that you need a newer libvirt for this to have any effect I was able to verify that infinite leases still work as expected after the changes we discussed above. I think this should be good to go once I push the new version.

PR #1196, opened 2021-03-03
using slash delimiters in sysctl for handling vlan NICs

osherdp:
Happened to me on a scalelab machine.
Figured out we can handle it in this way.

/hold


hardys:
Tested some sysctl commands locally and seems to work OK, I've not got vlans setup to test with the nic.vlan naming though.

/approve

osherdp:
/unhold
Yes, it worked on my setup that has vlans

elfosardo (approved):

PR #1198, opened 2021-03-09
Set IMAGES_JSON when running a local CBO

derekhiggins:
stop-cbo.sh writes it to $OCP_DIR


metal3-dev/local-cbo.sh
```diff
@@ -14,10 +14,12 @@ if [ ! -d $cbo_path ]; then
     exit 1
 fi
 
+export IMAGES_JSON=$(realpath ${OCP_DIR}/cbo-images.json)
```
hardys, line 17: This won't work when running the script from the metal3-dev dir, should we perhaps use `$SCRIPTDIR` instead?

derekhiggins: > This won't work when running the script from the metal3-dev dir, should we perhaps use `$SCRIPTDIR` instead?

Done, thanks.

PR #1201, opened 2021-03-18
Fix error building images without extra pkgs

cybertron:
The extra pkg change broke image builds that don't use it because
it was unconditionally calling realpath on the variable. If it wasn't
set then this resulted in an invalid command. To fix this we just
need to move the realpath call inside the -n check so we only do it
if the variable is populated.


hardys:
Any idea why CI on https://github.com/openshift-metal3/dev-scripts/pull/1142 didn't catch this?

elfosardo:
> Any idea why CI on #1142 didn't catch this?
> 
> Edit: I guess it's because the `set -e` on the script doesn't catch the subshell error, so you only noticed this by looking at the logs vs the parent script failing?

that's probably it, I still need to move the build images code to a separate script as done for the base image, that should help catching future errors

cybertron:
A dev-scripts patch doesn't build any images so it wouldn't have hit this code path. I believe set -e should catch errors from a subshell.

PR #1204, opened 2021-03-23
Apply NUM_EXTRA_WORKERS if explicitly set in CI

ardaguclu:
Some CI tests try to test machine scale in and out like;
`Managed cluster should grow and decrease when scaling different machineSets simultaneously`.
If the CI suite set NUM_EXTRA_WORKERS field to test this, we automatically apply it.

### DIFF
--- a/06_create_cluster.sh
+++ b/06_create_cluster.sh
@@ -36,4 +36,20 @@ if [[ ! -z "${ENABLE_LOCAL_REGISTRY}" ]]; then
       --type merge --patch "{\"spec\":{\"samplesRegistry\":\"${LOCAL_REGISTRY_DNS_NAME}\"}}"
 fi
 
+if [[ -n "${APPLY_EXTRA_WORKERS}" ]]; then
+    if [[ ${NUM_EXTRA_WORKERS} -ne 0 && -s "${OCP_DIR}/extra_host_manifests.yaml" ]]; then
+        oc apply -f "${OCP_DIR}/extra_host_manifests.yaml"
+
+        for h in $(jq -r '.[].name' ${EXTRA_BAREMETALHOSTS_FILE}); do
+            while ! oc get baremetalhost -n openshift-machine-api $h 2>/dev/null; do
+                echo "Waiting for $h"
+                sleep 5
+            done
+            echo "$h is successfully applied"
+        done
+    else
+        echo "NUM_EXTRA_WORKERS should be set and extra_host_manifests.yaml should exist"
+    fi
+fi
+
 echo "Cluster up, you can interact with it via oc --config ${KUBECONFIG} <command>"
--- a/config_example.sh
+++ b/config_example.sh
@@ -85,6 +85,9 @@ set -x
 # Indicate number of extra VMs to create but not deploy
 #export NUM_EXTRA_WORKERS=0
 
+# Apply extra workers automatically if NUM_EXTRA_WORKERS is set
+#export APPLY_EXTRA_WORKERS=true
+
 # Indicate the online status of the NUM_EXTRA_WORKERS set in extra_host_manifests.yaml
 #export EXTRA_WORKERS_ONLINE_STATUS=true
 


06_create_cluster.sh
```diff
@@ -34,6 +34,12 @@ if [[ ! -z "${ENABLE_LOCAL_REGISTRY}" ]]; then
     
     oc patch configs.samples cluster \
       --type merge --patch "{\"spec\":{\"samplesRegistry\":\"${LOCAL_REGISTRY_DNS_NAME}\"}}"
+    
+    # Some CI tests try to test scale in and out of machines. For that purpose,
+    # if NUM_EXTRA_WORKERS is set, we enable extra worker for that test.
+    if [ -f "${OCP_DIR}/extra_host_manifests.yaml" ]; then
```
andfasano, line 40: While it could make sense to apply automatically the extra manifests during the CI, I wouldn't like to have this feature for the local development, so I'd suggest to enable it only if `OPENSHIFT_CI` is set (see https://github.com/openshift/release/blob/6ef9fb413a75b8f3cc967acc9a532dc7c92b51a4/ci-operator/step-registry/baremetalds/devscripts/setup/baremetalds-devscripts-setup-commands.sh#L77)

06_create_cluster.sh
```diff
@@ -34,6 +34,12 @@ if [[ ! -z "${ENABLE_LOCAL_REGISTRY}" ]]; then
...
+	    oc apply -f "${OCP_DIR}/extra_host_manifests.yaml"
+    fi
```
andfasano, line 42: A wait condition to verify that all the hosts were setup could be useful, and it allow to detect issues earlier. In addition, there wouldn't be the risk to run the tests while the hosts are coming up

06_create_cluster.sh
ardaguclu: This check is in the ENABLE_LOCAL_REGISTRY condition and not affecting local development. Do you mean using new environment variable `OPENSHIFT_CI` (comes from CI) in dev-scripts instead using ENABLE_LOCAL_REGISTRY?

06_create_cluster.sh
ardaguclu: Good point. Because test failed when bmh is not in ready state.

06_create_cluster.sh
andfasano: Yes: potentially `ENABLE_LOCAL_REGISTRY` could be enabled also during local development, so I don't think it's related. The extra worker feature it's instead something independent - I think it could even be worth to have a dedicated variable setting to automatically apply the manifest, but since right now it seems just a CI requirement I suggested `OPENSHIFT_CI`

hardys: Sometimes I would like this behavior locally (for example testing e2e deployments where the live-iso flow adds extra workers ref https://github.com/openshift-metal3/dev-scripts/pull/1193)

Could we add some new variable which controls this behavior, e.g `APPLY_EXTRA_WORKERS` or similar, then set that in the CI config?

andfasano: Seems like a good idea

ardaguclu: Agree. Thanks to that we will offer a new dev-script feature instead using specific to CI.

06_create_cluster.sh
```diff
@@ -36,4 +36,15 @@ if [[ ! -z "${ENABLE_LOCAL_REGISTRY}" ]]; then
       --type merge --patch "{\"spec\":{\"samplesRegistry\":\"${LOCAL_REGISTRY_DNS_NAME}\"}}"
 fi
 
+if [[ ! -z "${APPLY_EXTRA_WORKERS}" && -f "${OCP_DIR}/extra_host_manifests.yaml" ]]; then
```
andfasano, line 39: My expectation would be that if `APPLY_EXTRA_WORKERS` has been set then the lack of `extra_hosts_manifest` should be considered an error (at least, something to be logged).
Maybe it could be useful to have a preliminary sanity check that if `APPLY_EXTRA_WORKERS` is set then `NUM_EXTRA_WORKERS` must be greater than 0?

06_create_cluster.sh
```diff
@@ -36,4 +36,19 @@ if [[ ! -z "${ENABLE_LOCAL_REGISTRY}" ]]; then
       --type merge --patch "{\"spec\":{\"samplesRegistry\":\"${LOCAL_REGISTRY_DNS_NAME}\"}}"
 fi
 
+if [[ ! -z "${APPLY_EXTRA_WORKERS}" ]]; then
+    if [[ ${NUM_EXTRA_WORKERS} -ne 0 && -f "${OCP_DIR}/extra_host_manifests.yaml" ]]; then
+        oc apply -f "${OCP_DIR}/extra_host_manifests.yaml"
+
+        set +x
+        while oc get baremetalhost -n openshift-machine-api -o json | jq '.items[].status.poweredOn' | grep -q "false"; do
+            sleep 30
```
hardys, line 45: If we're going to check the power state here, it should be [${EXTRA_WORKERS_ONLINE_STATUS}](https://github.com/openshift-metal3/dev-scripts/blob/master/ocp_install_env.sh#L284)

But I don't think that's necessarily going to be reliable in all cases, since it's possible that the workers in some other state.

Instead, I'd suggest checking for the extra hosts explicitly, e.g something like:

```
for h in $(jq -r '.[].name' ${EXTRA_BAREMETALHOSTS_FILE}); do
  while ! oc get baremetalhost -n openshift-machine-api $h 2>/dev/null; do
    echo "Waiting for $h"
    sleep 1
  done
done
```

Also note the reduced sleep time - the apply should happen pretty fast so I don't think it's necessary to wait 30s?

ardaguclu: Although apply is pretty fast, host stays in inspecting state for a while and if the scale in/out test runs in this phase(pretty low probability but it might create flakiness), test fails.

hardys: Ack, good point - we could add a check for a specific provisioning state then (just checking the power status won't be enough), and we may want to make that configurable since e.g in https://github.com/openshift-metal3/dev-scripts/pull/1193 I'm adding options which will result in the BMH state being `provisioned` vs just ready when no image is specified?

hardys: re the sleep time perhaps we can compromise and use 5s or something - my motivation is I don't want to wait 30s unnecessarily in the live-iso flow (where inspection may be disabled).

ardaguclu: is it meaningful to create a new environment variable like `APPLY_EXTRA_WORKERS_WAIT_STATE` default is `ready`?, or maybe we can continue with your proposal above and if we see any flakiness, we add state check later.

andfasano: > Although apply is pretty fast, host stays in inspecting state for a while and if the scale in/out test runs in this phase(pretty low probability but it might create flakiness), test fails.

I don't think it's correct to push at this level a test requirement. If the test requires a waiting condition (and that's fully reasonable), it must be in the test code, and not in the setup (since it could change in future, or be completely replaced by a different setup mechanism). So I guess that 5s delay it's acceptable as a starting point

ardaguclu: I partially disagree. Test tries to scale in and out without extra paying attention of aws, gcp or dev-scripts. I agree, this is not dev-scripts mission to be implemented in the way of specific case of this test. If @hardys will not need in live-iso flow as a general need, I think we can move forward with 5s delay.

06_create_cluster.sh
```diff
@@ -36,4 +36,20 @@ if [[ ! -z "${ENABLE_LOCAL_REGISTRY}" ]]; then
       --type merge --patch "{\"spec\":{\"samplesRegistry\":\"${LOCAL_REGISTRY_DNS_NAME}\"}}"
 fi
 
+if [[ ! -z "${APPLY_EXTRA_WORKERS}" ]]; then
```
elfosardo, line 39: let's use -n instead ?

06_create_cluster.sh
```diff
@@ -36,4 +36,20 @@ if [[ ! -z "${ENABLE_LOCAL_REGISTRY}" ]]; then
...
+    if [[ ${NUM_EXTRA_WORKERS} -ne 0 && -f "${OCP_DIR}/extra_host_manifests.yaml" ]]; then
```
elfosardo, line 40: maybe use -s to check that the extra_host_manifests.yaml file is not empty?

06_create_cluster.sh
```diff
@@ -36,4 +36,20 @@ if [[ ! -z "${ENABLE_LOCAL_REGISTRY}" ]]; then
...
+
+        for h in $(jq -r '.[].name' ${EXTRA_BAREMETALHOSTS_FILE}); do
+            while ! oc get baremetalhost -n openshift-machine-api $h 2>/dev/null; do
+                echo "Waiting for $h"
+                sleep 5
+            done
+            echo "$h is successfully applied"
+        done
+    else
+        echo "NUM_EXTRA_WORKERS should be set and extra_host_manifests.yaml should exist"
```
elfosardo, line 51: should this be enough to fail ?

06_create_cluster.sh
```diff
@@ -36,4 +36,20 @@ if [[ ! -z "${ENABLE_LOCAL_REGISTRY}" ]]; then
...
 fi
 
+if [[ ! -z "${APPLY_EXTRA_WORKERS}" ]]; then
+    if [[ ${NUM_EXTRA_WORKERS} -ne 0 && -f "${OCP_DIR}/extra_host_manifests.yaml" ]]; then
+        oc apply -f "${OCP_DIR}/extra_host_manifests.yaml"
+
+        for h in $(jq -r '.[].name' ${EXTRA_BAREMETALHOSTS_FILE}); do
+            while ! oc get baremetalhost -n openshift-machine-api $h 2>/dev/null; do
+                echo "Waiting for $h"
+                sleep 5
```
elfosardo, line 46: sleep is fine at the beginning but  checking the status of the host would be better

elfosardo (changes requested):

06_create_cluster.sh
ardaguclu: We discussed checking the status, but what is the status we wait seems to be a intricate.

06_create_cluster.sh
hardys: > I don't think it's correct to push at this level a test requirement. If the test requires a waiting condition (and that's fully reasonable), it must be in the test code, and not in the setup

Would those tests support checking the BMH state as a precursor to scaling up the MachineSet, or are they completely platform agnostic?  I was assuming the latter, hence thinking we would check the BMH state here.

Another option is just don't check the BMH state at all, then the test will still eventually fail when the new Machine fails to move to `Provisioned` - it'll just take longer.

06_create_cluster.sh
ardaguclu: I think, thanks to just logging instead exiting, user still has a chance to apply manually.

06_create_cluster.sh
ardaguclu: It seems that this usage is more common in dev-scripts

06_create_cluster.sh
elfosardo: sounds good!

06_create_cluster.sh
elfosardo: alright, I guess it's fine for now

06_create_cluster.sh
elfosardo: that's true, and IMHO that's unfortunate
Here (and in other places in dev-scripts) we want to check if the variable is set and not empty, the canonical way to do that in Bourne-like shell is using -n, or you could use -v in modern versions (>=4.2)
the rest is just a complication

elfosardo (approved):

PR #1206, opened 2021-03-24
Disable TLS verification in clouds.yaml

dtantsur:
In theory it should be possible to load the public key from secrets, but
I don't have an environment to figure it out yet.


flaper87 (commented):
/lgtm

hardys:
It'd be nice to get this working with TLS but for now this seems reasonable as it's only for developer usage

/approve

PR #1208, opened 2021-03-24
Add some extra image/volume cleanup

hardys:
When testing the virtualmedia flow, I noticed that we're
leaking boot images, and also that on realclean we don't
clean any dangling podman volumes.

### DIFF
--- a/ocp_cleanup.sh
+++ b/ocp_cleanup.sh
@@ -50,13 +50,13 @@ for vm in $(sudo virsh list --all --name | grep "^${CLUSTER_NAME}.*bootstrap"); 
   sudo virsh undefine $vm --remove-all-storage
 done
 
-# For some reason --remove-all-storage doesn't actually remove the storage...
+# For some reason --remove-all-storage doesn't actually remove the storage
+# so we do some extra cleanup of volumes
 if [ -d /var/lib/libvirt/openshift-images ]; then
   sudo rm -fr /var/lib/libvirt/openshift-images/${CLUSTER_NAME}-*
 fi
 
-# The .ign volume isn't deleted via --remove-all-storage
-VOLS="$(sudo virsh vol-list --pool default | awk '{print $1}' | grep "^${CLUSTER_NAME}.*bootstrap")"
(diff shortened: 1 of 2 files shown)


elfosardo (approved):

ocp_cleanup.sh
```diff
@@ -50,13 +50,13 @@ for vm in $(sudo virsh list --all --name | grep "^${CLUSTER_NAME}.*bootstrap");
...
-# For some reason --remove-all-storage doesn't actually remove the storage...
+# For some reason --remove-all-storage doesn't actually remove the storage
+# so we do some extra cleanup of volumes
 if [ -d /var/lib/libvirt/openshift-images ]; then
   sudo rm -fr /var/lib/libvirt/openshift-images/${CLUSTER_NAME}-*
 fi
 
-# The .ign volume isn't deleted via --remove-all-storage
-VOLS="$(sudo virsh vol-list --pool default | awk '{print $1}' | grep "^${CLUSTER_NAME}.*bootstrap")"
+VOLS="$(sudo virsh vol-list --pool default | awk '{print $1}' | grep -e "^${CLUSTER_NAME}.*bootstrap" -e "^configdrive-" -e "boot-*-iso-*")"
```
derekhiggins, line 59: "boot--iso-" - is this intended to match files of the format below?
boot-ea7f3cad-5130-4e42-896b-3a391837ad9d-iso-34f547df-726d-4c00-b8fa-c1a26d2959b5.img
I seem to have a bunch of them left behind, if this is the files it should be removing then prehaps ""^boot-.-iso-""

hardys: Thanks @derekhiggins yes those are the files I wanted to clean - I'm not entirely sure if this is working around a sushy-tools bug, or if the issue is we just don't detach the iso when tearing down vla `make clean`, but as you say the files build up pretty quickly when doing virtualmedia testing.

I'll adjust the pattern as you suggest :+1:

hardys: Resolved in the latest version, thanks!

asalkeld (commented):
/lgtm

PR #1212, opened 2021-03-30
Always create the provisioning network

kirankt:
With the day-2 capability to manage the provisioning network, it is desirable to always create the provisioning bridge in dev-scripts.


imain (approved):
Looks good to me!  Thanks Kiran!!

PR #1214, opened 2021-04-06
Add support for Day 2 management of the provisioning network

kirankt:
This PR enables the day 2 management of the provisioning network. The script manipulates the provisioning-configuration CR and sets the values in the spec based on whether the provisioning network is managed, unmanaged or disabled.

Usage:
./provisioning-network.sh [Managed|Unmanaged|Disabled]

### DIFF
--- a/provisioning-network.sh
+++ b/provisioning-network.sh
@@ -1,0 +1,44 @@
+#!/bin/bash
+
+# This script manages the day-2 changes that can be made to the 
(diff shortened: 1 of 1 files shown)


sadasu:
Couple of nits:
1. File name has a typo. I think you intended it to be provisioning-network.sh
2. Can we add a comment at the top of the file saying that this script can be used only for day-2 functionality because it expects a valid Provisioning CR to be already present on the cluster.

PR #1219, opened 2021-04-09
Make CBO TLS files available on the host

honza:
These files are now required when running the local CBO.

### DIFF
--- a/metal3-dev/local-cbo.sh
+++ b/metal3-dev/local-cbo.sh
@@ -14,6 +14,17 @@ if [ ! -d $cbo_path ]; then
     exit 1
 fi
 
+# Make the TLS files available in the expected location on the present host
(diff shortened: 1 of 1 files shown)


metal3-dev/local-cbo.sh
```diff
@@ -14,6 +14,17 @@ if [ ! -d $cbo_path ]; then
     exit 1
 fi
 
+# Make the TLS files available in the expected location on the present host
+mkdir -p /etc/cluster-baremetal-operator/tls
```
hardys, line 18: It'd be nice if this could be overridden so the certs are in e.g OCP_DIR - but it looks like it's hard-coded in the CBO atm https://github.com/openshift/cluster-baremetal-operator/blob/master/main.go#L97

hardys:
lgtm, small comment re the global path, maybe we can improve that in future, not blocking on it

PR #1222, opened 2021-04-15
Fix VM_EXTRADISKS

hardys:
Since https://github.com/metal3-io/metal3-dev-env/pull/129
the extradisks variable actually doesn't work, so instead we
need to set the value in vm_setup_vars.yml directly

### DIFF
--- a/02_configure_host.sh
+++ b/02_configure_host.sh
@@ -63,7 +63,6 @@ ansible-playbook \
     -e "working_dir=$WORKING_DIR" \
     -e "num_masters=$NUM_MASTERS" \
     -e "num_workers=$((NUM_WORKERS + NUM_EXTRA_WORKERS))" \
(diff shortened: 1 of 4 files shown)


flaper87 (approved):
@lranjbar FYI 

As a follow-up to this PR, I think it would be nice to have an env var to also set the number of desired disks per node. Use case for this would be an AI deployment, which requires 2 PVs

:spaghetti:

hardys:
> As a follow-up to this PR, I think it would be nice to have an env var to also set the number of desired disks per node. Use case for this would be an AI deployment, which requires 2 PVs

That sounds reasonable to me - such a variable could be easily added in a similar way to this PR, the relevant variable is here https://github.com/metal3-io/metal3-dev-env/blob/master/vm-setup/roles/libvirt/defaults/main.yml#L24

elfosardo (approved):

PR #1233, opened 2021-04-28
assisted-installer: do not set starting CSV

djzager:
With there being multiple releases of the assisted-service-operator,
starting at v0.0.1 is only going to cause headaches. This removes it
entirely and leaves it up to OLM which version to install first...which
if we set up our manifests correctly should always be the latest.


flaper87 (approved):
:spaghetti: 

/lgtm
/retest

PR #1244, opened 2021-05-20
remove workaround of unset XDG_RUNTIME_DIR

osherdp:
RHEL 8.2 already having podman >= 1.4.4, so this workaround shouldn't be needed anymore.
/cc @stbenjam 
/hold


osherdp:
/unhold
@stbenjam tests seem to pass :)

PR #1245, opened 2021-05-23
when creating the image sha256sum remove the path to the image file

asalkeld:
This is needed because if the checksum is provided to the BMH like:

```
image:
    checksum: http://192.168.111.1/images/rhcos-48.84.202104271417-0-openstack.x86_64.qcow2.gz.sha256sum
    checksumType: sha256
    format: qcow2
    url: http://192.168.111.1/images/rhcos-48.84.202104271417-0-openstack.x86_64.qcow2.gz
```

We see an error like:

```
    Error downloading image: Download of image bdab49e61715b72c5e937c54a232677cbf821ebcd71aeb3ba58028683cccad07
    failed: Checksum file does not contain name rhcos-48.84.202104271417-0-openstack.x86_64.qcow2.gz
```

It seems that Ironic is checking for the filename/basename and fails to find the checksum if a fully qualified path is used.

This isn't an issue in regular IPI deployments because we use the cluster cached copy of RHCOS which calculates the checksum differently ref  https://github.com/openshift/ironic-rhcos-downloader/blob/master/get-resource.sh#L67 

/cc @hardys

### DIFF
--- a/04_setup_ironic.sh
+++ b/04_setup_ironic.sh
@@ -68,7 +68,7 @@ for IMAGE_VAR in $(env | grep "_LOCAL_IMAGE=" | grep -o "^[^=]*") ; do
                 git fetch origin pull/${IMAGE_PR}/head:pr${IMAGE_PR}
                 git checkout pr${IMAGE_PR}
         fi
-        
+
         # If we want to install extra packages, we set the path to a
         # file containing the packages (one per line) we want to install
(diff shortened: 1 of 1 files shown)


hardys:
LGTM, I added some details to the PR summary based on our slack discussion figuring this out.

Ideally we'd have a bit more detail in the commit message body, but I'm not sure it's worth the CI cycles to update it so I went with adding details to the PR instead so folks can see why this was needed.

@andfasano can you check and lgtm if happy?  Thanks!

PR #1247, opened 2021-05-25
NO-ISSUE Store artifacts in saved-assets/assisted-installer-manifests

mkowalski:
This PR makes the `assisted_deployment.sh` script save all the manifests
into `saved-assets/` so that operators have an ability to inspect them
after (un)successful installation.

The previous behavior was to generate the manifest on fly and pass it
directly to `oc apply`. From now on the manifest will be saved on a
local filesystem and only afterwards applied from the file that has been
created.

This PR creates as well a patched copy of `extra_host_manifests.yaml`
given that the namespace of the created resource must match the
namespace of all the other resources we are creating.

### DIFF
--- a/assisted_deployment.sh
+++ b/assisted_deployment.sh
@@ -19,13 +19,11 @@ ASSISTED_OPENSHIFT_VERSIONS="${ASSISTED_OPENSHIFT_VERSIONS:-}"
 ASSISTED_NAMESPACE="${ASSISTED_NAMESPACE:-assisted-installer}"
 ASSISTED_OPERATOR_INDEX="${ASSISTED_OPERATOR_INDEX:-quay.io/ocpmetal/assisted-service-index:latest}"
 
+ASSETS_DIR="${OCP_DIR}/saved-assets/assisted-installer-manifests"
 
-function deploy_local_storage() {
-  oc adm new-project openshift-local-storage || true
 
-  oc annotate project openshift-local-storage openshift.io/node-selector=''
-
-  cat <<EOF | oc apply -f -
+function generate_local_storage() {
+  cat >"${ASSETS_DIR}/01-local-storage-operator.yaml" <<EOF
 apiVersion: operators.coreos.com/v1alpha2
 kind: OperatorGroup
 metadata:
@@ -47,10 +45,7 @@ spec:
   sourceNamespace: openshift-marketplace
 EOF
 
-  wait_for_crd "localvolumes.local.storage.openshift.io"
-
-  echo "Creating local volume and storage class..."
-  cat <<EOCR | oc apply -f -
+  cat >"${ASSETS_DIR}/02-local-volume.yaml" <<EOCR
 apiVersion: local.storage.openshift.io/v1
 kind: LocalVolume
 metadata:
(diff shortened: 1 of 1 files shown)


assisted_deployment.sh
```diff
@@ -68,11 +63,24 @@ spec:
 EOCR
 }
 
+function deploy_local_storage() {
+  oc adm new-project openshift-local-storage || true
```
flaper87, line 67: I think the `local-storage` operator will take care of this (since we are passing the target namespace)

assisted_deployment.sh
```diff
@@ -208,11 +213,30 @@ spec:
...
+  wait_for_crd "agentserviceconfigs.agent-install.openshift.io"
+
+  oc apply -f "${ASSETS_DIR}/05-assisted-service-config.yaml"
+}
+
+
+function patch_extra_host_manifests() {
+  cp "${OCP_DIR}/extra_host_manifests.yaml" "${ASSETS_DIR}/06-extra-host-manifests.yaml"
 
+  sed -i s/"namespace: openshift-machine-api"/"namespace: ${ASSISTED_NAMESPACE}"/ "${ASSETS_DIR}/06-extra-host-manifests.yaml"
```
flaper87, line 234: Could we also add the inspect annotation and disable the cleaning mode?

flaper87 (changes requested):
Yes, this is exactly what I was hoping for! I like that we can write these manifests down. 

It may be worth printing some debug information at the end of the installation process telling users where the manifests are and that they have to apply the BMH one manually.

Also, I wonder if the `extra_hosts_manifest` file always exists, even when no extra hosts are requested.

assisted_deployment.sh
mkowalski: This code has just been moved around, but I'll test how it works without this piece

assisted_deployment.sh
mkowalski: Yes, but I'd rather do it in a separate PR. It won't be the prettiest code ever (looking at you, Bash), so I'd rather keep it out of this one

assisted_deployment.sh
mkowalski: Seems like that's not the case

```
Creating local storage operator group and subscription...
+(./assisted_deployment.sh:83): deploy_local_storage(): oc apply -f ocp/ostest/saved-assets/assisted-installer-manifests/01-local-storage-operator.yaml                                                                              
Error from server (NotFound): error when creating "ocp/ostest/saved-assets/assisted-installer-manifests/01-local-storage-operator.yaml": namespaces "openshift-local-storage" not found                                              
```

flaper87: > Seems like that's not the case
> 
> ```
> Creating local storage operator group and subscription...
> +(./assisted_deployment.sh:83): deploy_local_storage(): oc apply -f ocp/ostest/saved-assets/assisted-installer-manifests/01-local-storage-operator.yaml                                                                              
> Error from server (NotFound): error when creating "ocp/ostest/saved-assets/assisted-installer-manifests/01-local-storage-operator.yaml": namespaces "openshift-local-storage" not found                                              
> ```

Gotcha! thanks!

Let's add the debug/prints I mentioned and get this in then :smiley:

flaper87 (approved):

flaper87 (approved):
:pizza:

assisted_deployment.sh
hardys: I have some commits in https://github.com/openshift-metal3/dev-scripts/pull/1193/commits which may help, I can rebase that or break out some of the commits if you'd like to use that approach vs modifying in assisted_deployment.sh

hardys:
> It may be worth printing some debug information at the end of the installation process telling users where the manifests are and that they have to apply the BMH one manually.

Note there is an APPLY_EXTRA_WORKERS variable if you want to avoid manually applying the manifest?

https://github.com/openshift-metal3/dev-scripts/blob/master/config_example.sh#L89

> Also, I wonder if the `extra_hosts_manifest` file always exists, even when no extra hosts are requested.

It doesn't https://github.com/openshift-metal3/dev-scripts/blob/master/02_configure_host.sh#L78

mkowalski:
>Note there is an APPLY_EXTRA_WORKERS variable if you want to avoid manually applying the manifest?

That's right, but right now we need to tune the manifest a bit so using `APPLY_EXTRA_WORKERS` is not yet desirable

>>Also, I wonder if the extra_hosts_manifest file always exists, even when no extra hosts are requested.

>It doesn't https://github.com/openshift-metal3/dev-scripts/blob/master/02_configure_host.sh#L78

Ah yeah, I had the change to detect if the file exists but lost it in some rebase. Just added it back

PR #1250, opened 2021-05-27
Create Hive operand in assisted deployment

rwsu:
Currently, the assisted deployment only creates the Hive operator.
The operand also needs to be created so that it can reconcile
CRs like cluster deployments.

Signed-off-by: Richard Su <rwsu@redhat.com>


flaper87 (approved):
Thanks! :pizza:

flaper87 (commented):
/approve
/lgtm

PR #1251, opened 2021-05-31
Enable configuring specific specs for extraworkers

osherdp:
Added another flavor ``extraworker`` so that we can configure memory, vCPUs count, and disk size for extra workers.


elfosardo (approved):

cybertron (approved):
/lgtm

Seems fine to me. Default behavior should be the same as before, but this provides a little extra flexibility for configuring the extra workers.

osherdp:
@elfosardo @cybertron can any of you approve?
Thanks in advance guys! 😁

PR #1257, opened 2021-06-08
[WIP] Bug 1968754 - CI: e2e-metal-ipi-upgrade failing on KubeletHasDiskPressure, which triggers machine-config RequiredPoolsFailed 

rhjanders:
Increasing VM disk size to 50GB in attempt to troubleshoot CI issues covered
in BZ1968754.


hardys:
It'd be helpful to answer the following questions before modifying the default I think:

* Does this `KubeletHasDiskPressure` only happen in the e2e-metal-ipi-upgrade job?
* Is this a new issue (specific to upgrading to the latest version), if so can we identify when it started failing?
* Can we identify what's using the disk space? (Not sure this will be possible without a local reproduce)
* What root disk size is used for other platforms, has there been a recent increase in requirements elsewhere?

I'm not opposed to bumping this if we have to, but it'll increase the requirements for all developer boxes, and every CI job, so it'd be good to ensure we're not papering over some underlying issue before proceeding

rhjanders:
@hardys all good points! I'm kicking off an install on a lab machine to get more info/context as well.

hardys:
This was resolved via https://github.com/openshift-metal3/dev-scripts/pull/1261 and https://github.com/openshift/ironic-rhcos-downloader/pull/44

PR #1261, opened 2021-06-10
Increase WORKER_DISK to 30

ardaguclu:
This PR increases default `WORKER_DISK` to 30 GB. CI jobs are running with this default value and after running ~1000 parallel suite tests in parallel, storage becomes insufficient and occasionally some tests fail because of getting worker DiskPressure error.

Reference: https://search.ci.openshift.org/?maxAge=24h&type=build-log&search=KubeletHasDiskPressure


hardys:
This seems reasonable, and matches the current master disk size, we can then continue investigation into the master disk-pressure issues and decide if the master size should increase too ref https://github.com/openshift-metal3/dev-scripts/pull/1257

dtantsur:
/approve

20G sounds very small indeed

PR #1266, opened 2021-06-18
Change IPv6DualStack feature version check

ardaguclu:
This PR changes version extraction method for IPv6DualStack feature apply. In CI, `OPENSHIFT_VERSION` is always unset and for 4.6 and 4.7 versions, this condition is always false. Thereby, dualstack jobs are failing in these versions.

### DIFF
--- a/utils.sh
+++ b/utils.sh
@@ -67,7 +67,7 @@ function create_cluster() {
 
     find assets/generated -name '*.yaml' -exec cp -f {} ${assets_dir}/openshift \;
 
-    if [[ "${IP_STACK}" == "v4v6" && "$OPENSHIFT_VERSION" =~ 4.[67] ]]; then
+    if [[ "${IP_STACK}" == "v4v6" && "$(openshift_version $OCP_DIR)" =~ 4.[67] ]]; then
         # The IPv6DualStack feature is not on by default in 4.6 and 4.7 and needs
         # to be manually enabled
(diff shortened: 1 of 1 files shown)


utils.sh
```diff
@@ -67,7 +67,7 @@ function create_cluster() {
 
     find assets/generated -name '*.yaml' -exec cp -f {} ${assets_dir}/openshift \;
 
-    if [[ "${IP_STACK}" == "v4v6" && "$OPENSHIFT_VERSION" =~ 4.[67] ]]; then
+    if [[ "${IP_STACK}" == "v4v6" && "$(openshift_version $OCP_DIR)" =~ 4.[67] ]]; then
```
hardys, line 70: Ah good catch, I wonder if we can either remove OPENSHIFT_VERSION, or make it local to the places it's really needed (only the `oc` download IIRC, everywhere else we should use the release-derived version since #1146)

ardaguclu: Thanks @hardys 

I'll check the other places as well, if it does not need dramatic changes, I can add it to this PR.

ardaguclu: I removed `OPENSHIFT_VERSION` usage only https://github.com/openshift-metal3/dev-scripts/blob/82293c97caddfd13829b97d13c110e389acf5fa4/common.sh#L93 remained.

PR #1267, opened 2021-06-22
retry downloading of oc client from mirror.openshift.com

osherdp:
We're having those kind of errors in all kinds of jobs:
https://search.ci.openshift.org/?search=OpenSSL+SSL_connect%3A+SSL_ERROR_SYSCALL+in+connection+to+mirror.openshift.com%3A443&maxAge=48h&context=1&type=all&name=&excludeName=&maxMatches=5&maxBytes=20971520&groupBy=job

I think it will help us deal with those kind of flakes on other places as well, but let's start with the oc installation.

### DIFF
--- a/utils.sh
+++ b/utils.sh
@@ -5,12 +5,12 @@ set -o pipefail
 function retry_with_timeout() {
   retries=$1
   timeout_duration=$2
-  command=${@:3}
+  command=${*:3}
 
-  for i in $(seq $retries); do
+  for _ in $(seq "$retries"); do
     exit_code=0
-    timeout $timeout_duration bash -c "$command" || exit_code=$?
-    if (( $exit_code == 0 )); then
+    timeout "$timeout_duration" bash -c "$command" || exit_code=$?
+    if (( exit_code == 0 )); then
       return 0
     fi
   done
--- a/utils.sh
+++ b/utils.sh
@@ -2,6 +2,22 @@
 
 set -o pipefail
 
+function retry_with_timeout() {
+  retries=$1
+  timeout_duration=$2
+  command=${@:3}
+
+  for i in $(seq $retries); do
+    exit_code=0
+    timeout $timeout_duration bash -c "$command" || exit_code=$?
+    if (( $exit_code == 0 )); then
+      return 0
+    fi
+  done
+
+  return $(( exit_code ))
+}
+
 function generate_assets() {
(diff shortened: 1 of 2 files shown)


elfosardo (approved):

utils.sh
```diff
@@ -2,6 +2,22 @@
...
 
+function retry_with_timeout() {
+  retries=$1
+  timeout_duration=$2
+  command=${@:3}
+
+  for i in $(seq $retries); do
+    exit_code=0
+    timeout $timeout_duration bash -c "$command" || exit_code=$?
+    if (( $exit_code == 0 )); then
```
celebdor, line 13: Arithmetic ops do not need the dollar.
```suggestion
    if (( exit_code == 0 )); then
```

utils.sh
```diff
@@ -2,6 +2,22 @@
...
 set -o pipefail
 
+function retry_with_timeout() {
+  retries=$1
+  timeout_duration=$2
+  command=${@:3}
+
+  for i in $(seq $retries); do
+    exit_code=0
+    timeout $timeout_duration bash -c "$command" || exit_code=$?
```
celebdor, line 12: you should quote function parameters:
```suggestion
    timeout "$timeout_duration" bash -c "$command" || exit_code=$?
```

utils.sh
```diff
@@ -2,6 +2,22 @@
 
 set -o pipefail
 
+function retry_with_timeout() {
+  retries=$1
+  timeout_duration=$2
+  command=${@:3}
+
+  for i in $(seq $retries); do
```
celebdor, line 10: i is never used and retries should be quoted, being a function parameter.

```suggestion
  for _ in $(seq "$retries"); do
```

utils.sh
```diff
@@ -2,6 +2,22 @@
 
 set -o pipefail
 
+function retry_with_timeout() {
+  retries=$1
+  timeout_duration=$2
+  command=${@:3}
```
celebdor, line 8: $@ is an array, but here you are assigning to a string variable, because later on you use it as a string parameter to bash. Better use $*

```suggestion
  command=${*:3}
```

celebdor (changes requested):

osherdp:
/hold
to test it locally

osherdp:
/unhold
yep, works pretty well

celebdor (approved):

PR #1268, opened 2021-06-22
Fix make target for Assisted Installer

mkowalski:
`make configure` is not an idempotent action and in the current form
makes it impossible to use `make assisted` without breaking a
deployment, i.e.

```
make
make assisted
```

leaves the environment in a broken state. This PR changes the `assisted`
target so it only executes the installation script for the assisted
service.

Please note it implicitly assumes that the user previously executed
`make all`, however this is not a change in the behaviour as currently
this is also required.


flaper87 (approved):
/approve
/lgtm

PR #1269, opened 2021-06-23
Improve script to install assisted-service

mkowalski:
This PR adds additional ways to use `assisted_deployment.sh` so that it
can not only perform a complete installation of the Assisted Service but
can also be used to install only its prerequisites, i.e. LSO and Hive.

This is a useful workflow for development where operator-sdk is used to
install the operator itself, but we don't want to handle the base
requirements manually.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -62,3 +62,6 @@ bell:
 
 assisted_deployment:
 	./assisted_deployment.sh install_assisted_service
+
+assisted_deployment_requirements:
+	./assisted_deployment.sh install_prerequisites_assisted_service
--- a/assisted_deployment.sh
+++ b/assisted_deployment.sh
@@ -23,6 +23,8 @@ ASSETS_DIR="${OCP_DIR}/saved-assets/assisted-installer-manifests"
 
 
 function generate_local_storage() {
+  mkdir -p "${ASSETS_DIR}"
+
   cat >"${ASSETS_DIR}/01-local-storage-operator.yaml" <<EOF
 apiVersion: operators.coreos.com/v1alpha2
 kind: OperatorGroup
@@ -55,14 +57,14 @@ spec:
   logLevel: Normal
   managementState: Managed
   storageClassDevices:
-$(storage_devices_config)
+$(fill_local_storage)
       storageClassName: assisted-service
       volumeMode: Filesystem
 EOCR
 }
 
 
-function storage_devices_config() {
+function fill_local_storage() {
   if [ ! -z "${VM_EXTRADISKS_LIST}" ]; then
 cat <<EOF
     - devicePaths:
@@ -94,6 +96,8 @@ function deploy_local_storage() {
 
 
 function generate_hive() {
+  mkdir -p "${ASSETS_DIR}"
+
   cat >"${ASSETS_DIR}/03-hive.yaml" <<EOCR
 apiVersion: operators.coreos.com/v1alpha1
 kind: Subscription
@@ -135,7 +139,7 @@ EOF
 }
 
 
-function subscription_config() {
+function fill_assisted_operator() {
     if [ -n "${ASSISTED_SERVICE_IMAGE}" ]; then
 cat <<EOF
     - name: SERVICE_IMAGE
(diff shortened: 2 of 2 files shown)


assisted_deployment.sh
```diff
@@ -273,26 +281,38 @@ function patch_extra_host_manifests() {
...
- deploy_assisted_operator
+  install_prerequisites_assisted_service
+  deploy_assisted_operator
 
- oc wait -n "$ASSISTED_NAMESPACE" --for=condition=Ready pod -l app=assisted-service --timeout=90s
+  oc wait -n "$ASSISTED_NAMESPACE" --for=condition=Ready pod -l app=assisted-service --timeout=90s
+
+  echo "Installation finished..."
+  echo "For debugging purposes all the manifests have been saved in ${OCP_DIR}/saved-assets/assisted-installer-manifests"
+  echo "Please remember to manually apply BareMetalHost manifest available in the directory above."
```
pawanpinjarkar, line 291: It will be handy to just also print the command `oc apply -f ${OCP_DIR}/saved-assets/assisted-installer-manifests/bmh-manifest-filename`

assisted_deployment.sh
```diff
@@ -273,26 +281,38 @@ function patch_extra_host_manifests() {
...
+  mkdir -p "${ASSETS_DIR}"
+  patch_extra_host_manifests
+  deploy_local_storage
+  deploy_hive
+  generate_assisted_operator
 
- echo "Installation finished..."
- echo "For debugging purposes all the manifests have been saved in ${OCP_DIR}/saved-assets/assisted-installer-manifests"
- echo "Please remember to manually apply BareMetalHost manifest available in the directory above."
+  echo "LSO and Hive have been deployed. Useful manifests are available in ${OCP_DIR}/saved-assets/assisted-installer-manifests"
```
pawanpinjarkar, line 304: Better to use Local Storage Operator here

assisted_deployment.sh
pawanpinjarkar: Also, can we not execute the oc apply from the script?

mkowalski: Done

assisted_deployment.sh
mkowalski: Done

assisted_deployment.sh
```diff
@@ -273,26 +281,39 @@ function patch_extra_host_manifests() {
...
+
+# For a development workflow where we want to deploy the assisted service using operator-sdk it is
+# useful to have a process installing required dependencies, i.e. LSO and Hive as well as generating
+# AgentServiceConfig manifest that is required later on.
+function install_prerequisites_assisted_service() {
+  mkdir -p "${ASSETS_DIR}"
+  patch_extra_host_manifests
+  deploy_local_storage
+  deploy_hive
+  generate_assisted_operator
```
rwsu, line 303: Does the operator-sdk deployment use all of the artifacts from generate_assisted_operator? If it only uses the AgentServiceConfig, perhaps that should be separated from generate_assisted_operator? Or the operator specific parts of generate_assisted_operator should be moved to deploy_assisted_operator?

mkowalski: Ok, I split `generate_assisted_service_config` out of the `generate_assisted_operator` so that in operator-sdk prerequisites we generate only AgentServiceConfig. Other resources like CatalogSource, OperatorGroup and Subscription are not needed to be created as a manifest because operator-sdk takes care of them, so indeed AgentServiceConfig is the only resource we need in the manual flow.

Makefile
```diff
@@ -60,5 +60,8 @@ podman_cleanup:
 bell:
 	@echo "Done!" $$'\a'
 
-assisted_deployment:
+assisted_deployment_complete:
```
rwsu, line 63: There is a possibility that changing this target name will break some existing workflow. I would suggest leaving it unchanged.

rwsu (commented):
Not sure if target renaming is a concern. Otherwise ltgm.

Makefile
flaper87: Agreed!  Let's perhaps keep the current name (and deprecated it)

flaper87 (changes requested):

Makefile
mkowalski: Yeah, I just kept the current one. We never know what may explode...

rwsu (approved):

flaper87 (approved):
/lgtm

PR #1271, opened 2021-06-29
Fix image_mirror_config test condition bug

ardaguclu:
When there is no `MIRROR_IMAGES` or `ENABLE_LOCAL_REGISTRY` environment variables, image_mirror_config function gets an `-z: command not found` error and https://github.com/openshift-metal3/dev-scripts/blob/d3249ae3e97d7a9cb473b17b3e0fab43f27ab6ff/utils.sh#L340 is never executed.

This PR fixes by adding extra brackets to test condition.

### DIFF
--- a/utils.sh
+++ b/utils.sh
@@ -334,9 +334,9 @@ function generate_metal3_config {
 }
 
 function image_mirror_config {
-    if [ ! -z "${MIRROR_IMAGES}" || ! -z "${ENABLE_LOCAL_REGISTRY}" ]; then
+    if [[ ! -z "${MIRROR_IMAGES}" || ! -z "${ENABLE_LOCAL_REGISTRY}" ]]; then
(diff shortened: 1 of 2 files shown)


utils.sh
```diff
@@ -334,9 +334,9 @@ function generate_metal3_config {
 }
 
 function image_mirror_config {
-    if [ ! -z "${MIRROR_IMAGES}" || ! -z "${ENABLE_LOCAL_REGISTRY}" ]; then
+    if [[ ! -z "${MIRROR_IMAGES}" || ! -z "${ENABLE_LOCAL_REGISTRY}" ]]; then
```
andfasano, line 337: Looks like the same problem could also be here: https://github.com/openshift-metal3/dev-scripts/blob/41b9cfb172af92788ab9c3395b2d60ae90f886f0/ocp_install_env.sh#L136?

ardaguclu: I think so

PR #1273, opened 2021-07-05
Update base-image logic

elfosardo:
Introduce a variable to explicitly trigger the build of a custom base
image for the ironic images.
Do not make that depends from the CUSTOM_REPO_FILE variable anymore.
The CUSTOM_REPO_FILE is not mandatory anymore to build a custom base
image.
Also a custom base image is not required anymore to build local images as
the openshift base images are pullable from the openshift registry.

### DIFF
--- a/04_setup_ironic.sh
+++ b/04_setup_ironic.sh
@@ -29,11 +29,12 @@ DOCKERFILE=$(mktemp --tmpdir "release-update--XXXXXXXXXX")
 _tmpfiles="$_tmpfiles $DOCKERFILE"
 echo "FROM $OPENSHIFT_RELEASE_IMAGE" > $DOCKERFILE
 
-# To build custom images is highly recommended to build a base image first.
-# Build a base image if we set a custom repo file in the config file and
-# the file exists
-if [[ -n ${CUSTOM_REPO_FILE:-} ]]; then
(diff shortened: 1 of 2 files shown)


04_setup_ironic.sh
```diff
@@ -29,11 +29,12 @@ DOCKERFILE=$(mktemp --tmpdir "release-update--XXXXXXXXXX")
...
-# To build custom images is highly recommended to build a base image first.
-# Build a base image if we set a custom repo file in the config file and
-# the file exists
-if [[ -n ${CUSTOM_REPO_FILE:-} ]]; then
-    ./build-base-image.sh ${BASE_IMAGE_DIR:-base-image} $CUSTOM_REPO_FILE
+# Build a custom base image for the ironic images.
+# This may be necessary in case we want to rebuild the base image from
+# scratch or can't for some reason get the base openshift image from the
+# openshift registry.
+if [ "$CUSTOM_BASE_IMAGE" == "true" ]; then
```
dtantsur, line 36: Maybe use non-empty CUSTOM_REPO_FILE as a flag?

elfosardo: CUSTOM_REPO_FILE is not required for a custom base image, we may want to rebuild the base image also for other reasons

kirankt (approved):
/lgtm

PR #1275, opened 2021-07-09
Fix run_e2e.sh script

s1061123:


s1061123:
will override e2e-metal-ipi Monday because we need the fix and the PR does not brake CI (metallb e2e is optional and never invoked in e2e-metal-ipi*).

PR #1279, opened 2021-07-19
RHCOS Live Image caching support

kirankt:
This PR introduces a new variable USE_RHCOS_LIVE_IMAGES which enables caching of 4 RHCOS live images: kernel, rootfs, initramfs and iso.

### DIFF
--- a/04_setup_ironic.sh
+++ b/04_setup_ironic.sh
@@ -162,6 +162,29 @@ if [ ! -f "${CACHED_MACHINE_OS_IMAGE}" ]; then
   pushd $(dirname ${CACHED_MACHINE_OS_IMAGE})
   sha256sum --strict --check ${CACHED_MACHINE_OS_IMAGE}.sha256sum || ( rm -f "${CACHED_MACHINE_OS_IMAGE}" ; exit 1 )
   popd
+  # If using the live images, fetch the kernel, rootfs and initramfs images also
+  if [ "${RHCOS_LIVE_IMAGES}"  == "true" ]; then
+     CACHED_MACHINE_OS_LIVE_KERNEL="${IRONIC_DATA_DIR}/html/images/${MACHINE_OS_LIVE_KERNEL}"
+     curl -g --insecure -L -o "${CACHED_MACHINE_OS_LIVE_KERNEL}" "${LIVE_KERNEL_IMAGE_URL}"
+     echo "${MACHINE_OS_LIVE_KERNEL_SHA256} $(basename ${CACHED_MACHINE_OS_LIVE_KERNEL})" | tee ${CACHED_MACHINE_OS_LIVE_KERNEL}.sha256sum
+     pushd $(dirname ${CACHED_MACHINE_OS_LIVE_KERNEL})
+     sha256sum --strict --check ${CACHED_MACHINE_OS_LIVE_KERNEL}.sha256sum || ( rm -f "${CACHED_MACHINE_OS_LIVE_KERNEL}" ; exit 1 )
+     popd
+     #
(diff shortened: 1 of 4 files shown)


04_setup_ironic.sh
```diff
@@ -159,6 +159,29 @@ if [ ! -f "${CACHED_MACHINE_OS_IMAGE}" ]; then
   pushd $(dirname ${CACHED_MACHINE_OS_IMAGE})
   sha256sum --strict --check ${CACHED_MACHINE_OS_IMAGE}.sha256sum || ( rm -f "${CACHED_MACHINE_OS_IMAGE}" ; exit 1 )
   popd
+  # If using the live images, fetch the kernel, rootfs and initramfs images also
+  if [ "${RHCOS_LIVE_IMAGES}"  == "true" ]; then
+     CACHED_MACHINE_OS_LIVE_KERNEL="${IRONIC_DATA_DIR}/html/images/${MACHINE_OS_LIVE_KERNEL}"
+     curl -g --insecure -L -o "${CACHED_MACHINE_OS_LIVE_KERNEL}" "${LIVE_KERNEL_IMAGE_URL}"
```
elfosardo, line 165: the change is ok, just wondering if we could transform this in a function since we have the exact same code repeated 3 times

elfosardo (approved):

04_setup_ironic.sh
```diff
@@ -159,6 +159,29 @@ if [ ! -f "${CACHED_MACHINE_OS_IMAGE}" ]; then
   pushd $(dirname ${CACHED_MACHINE_OS_IMAGE})
   sha256sum --strict --check ${CACHED_MACHINE_OS_IMAGE}.sha256sum || ( rm -f "${CACHED_MACHINE_OS_IMAGE}" ; exit 1 )
   popd
+  # If using the live images, fetch the kernel, rootfs and initramfs images also
+  if [ "${RHCOS_LIVE_IMAGES}"  == "true" ]; then
```
zaneb, line 163: This needs to be something like `"${RHCOS_LIVE_IMAGES:-false}"` to prevent an error if it is not defined in the config.
Or alternatively we could put `RHCOS_LIVE_IMAGES=${RHCOS_LIVE_IMAGES:-false}` in common.sh.

ocp_install_env.sh
```diff
@@ -181,6 +181,21 @@ EOF
   fi
 }
 
+function os_images() {
+	# If using the live images, fetch the kernel, rootfs and initramfs images also
+	if [ "${RHCOS_LIVE_IMAGES}"  == "true" ]; then
```
zaneb, line 186: ditto

rhcos.sh
```diff
@@ -52,3 +52,30 @@ else
   export MACHINE_OS_INSTALLER_BOOTSTRAP_IMAGE_UNCOMPRESSED_SHA256=$(jq -r '.images.qemu["uncompressed-sha256"]' $OCP_DIR/rhcos.json)
   export MACHINE_OS_BOOTSTRAP_IMAGE_UNCOMPRESSED_SHA256=${MACHINE_OS_BOOTSTRAP_IMAGE_UNCOMPRESSED_SHA256:-${MACHINE_OS_INSTALLER_BOOTSTRAP_IMAGE_UNCOMPRESSED_SHA256}}
 fi
+
+if [ "${RHCOS_LIVE_IMAGES}"  == "true" ]; then
```
zaneb, line 56: ditto

kirankt:
@sadasu PTAL. Looks like all the tests passed. Thanks.

hardys:
/lgtm

I agree it'd be nice to de-duplicate some of this, but we can perhaps handle that as a follow-up

PR #1280, opened 2021-07-24
Do not remove upstream DNS configuraton in cleanup

mkowalski:
This PR changes the logic of host cleanup so that the upstream DNS
configuration is not removed. This comes from the fact that by default
dev-scripts configure NetworkManager to use dnsmasq with DHCP-provided
DNS servers and removes the existing /etc/resolv.conf configuration.

This causes problems in environments configured without DHCP with DNS
servers configured manually without use of the NetworkManager. With this
PR user can use `ADDN_DNS` variable to provide their own DNS servers
that will be passed to the dnsmasq and will not be removed when
dev-scripts are cleaned up.

Even though this does not cause the changes to be completely reverted
(user will still end up with NetworkManager + dnsmasq), it does not
leave the machine with no DNS servers at all.


flaper87 (commented):
/approve

I will leave the final approval to @hardys since you guys debugged this together!

hardys:
The only problem with doing this unconditionally is some folks expect the dev-scripts changes to be reverted on `make clean` ref https://github.com/openshift-metal3/dev-scripts/pull/1168

I wonder if we can somehow detect the resolv.conf isn't NM generated and save it somewhere to restore during host_cleanup.sh? (perhaps looking for presence of `Generated by NetworkManager` ?)

mkowalski:
>The only problem with doing this unconditionally is some folks expect the dev-scripts changes to be reverted

Yes, this is the most reasonable assumption, but it is not the case already now - i.e. running `make` and then `make clean` leaves the host in a broken state if there is no DHCP server in the network.

Honestly, I'm not sure if there is a silver bullet here... 

>I wonder if we can somehow detect the resolv.conf isn't NM generated and save it somewhere to restore during host_cleanup.sh

Yes, if `/etc/resolv.conf` is NM-generated then it's marked as such. But if we go down this way, we will face the next question - "whether we have configured NM in such a way or was it the user before". Overall I'm trying to think bigger here - how do people use dev-scripts ? In it's current form I already have a dedicated machine that is not used for any other setups, so I kinda accept the fact that dev-scripts may make changes to the system that are difficult (if not impossible) to revert. Are there even people running dev-scripts on their day-to-day workstations ?

hardys:
> > The only problem with doing this unconditionally is some folks expect the dev-scripts changes to be reverted
> 
> Yes, this is the most reasonable assumption, but it is not the case already now - i.e. running `make` and then `make clean` leaves the host in a broken state if there is no DHCP server in the network.

Right, we're in agreement this is something which needs to be fixed, it's just a question of how :)

> Honestly, I'm not sure if there is a silver bullet here...
> 
> > I wonder if we can somehow detect the resolv.conf isn't NM generated and save it somewhere to restore during host_cleanup.sh
> 
> Yes, if `/etc/resolv.conf` is NM-generated then it's marked as such. But if we go down this way, we will face the next question - "whether we have configured NM in such a way or was it the user before". Overall I'm trying to think bigger here - how do people use dev-scripts ? In it's current form I already have a dedicated machine that is not used for any other setups, so I kinda accept the fact that dev-scripts may make changes to the system that are difficult (if not impossible) to revert. Are there even people running dev-scripts on their day-to-day workstations ?

I think the majority of folks do have a dedicated machine, but clearly @dustinblack had a reason for pushing his cleanup PR, so I'm trying to find a way to solve your problem without reintroducing his ;)

I don't think any of these changes are impossible to revert, we just need to backup and restore the resolv.conf, right?

That could even be done unconditionally, since in the NM-managed case it'll just get overwritten, but in the static-config case we'll restore the previous config?

hardys:
Another point to note is that even when folks have a dedicated machine, sometimes they will want to switch between dev-scripts testing and other setups (upstream metal3-dev-env, maybe assisted-test-infra etc)

flaper87:
@mkowalski I think there is value in revisiting this PR, if you have the bandwidth.

mkowalski:
A simple change now should be enough to handle multiple scenarios with as simple logic as possible. I tried to avoid using specific markers in the `/etc/resolv.conf` because I'm already aware of 2 possibilities (and there are probably many more)

* `# Generated by NetworkManager`
* `# This is /run/systemd/resolve/stub-resolv.conf managed by man:systemd-resolved(8).`

mkowalski:
/retest

None of them looks like failing because of this change

elfosardo:
@mkowalski this looks stuck since a while, you want to keep it open?

PR #1281, opened 2021-07-29
Enable alternative CI_SERVER login

hardys:
When testing cluster-bot builds, the server (and CI_TOKEN) is
different, so it's necessary to modify the server login URL from
that used for regular CI builds.

### DIFF
--- a/utils.sh
+++ b/utils.sh
@@ -518,7 +518,7 @@ function write_pull_secret() {
     # Get a current pull secret for registry.ci.openshift.org using the token
     tmpkubeconfig=$(mktemp --tmpdir "kubeconfig--XXXXXXXXXX")
     _tmpfiles="$_tmpfiles $tmpkubeconfig"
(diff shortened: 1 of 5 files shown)


utils.sh
```diff
@@ -518,7 +518,7 @@ function write_pull_secret() {
     # Get a current pull secret for registry.ci.openshift.org using the token
     tmpkubeconfig=$(mktemp --tmpdir "kubeconfig--XXXXXXXXXX")
     _tmpfiles="$_tmpfiles $tmpkubeconfig"
-    oc login https://api.ci.l2s4.p1.openshiftapps.com:6443 --kubeconfig=$tmpkubeconfig --token=${CI_TOKEN}
```
dtantsur, line 521: I wonder if you want to include the port in the variable.

hardys: It didn't prove necessary to enable testing cluster-bot builds, and AFAIK this is the common port for API access, so I'd say we can add that later if it proves necessary?

asalkeld:
/retest
looks great to me :-)

PR #1284, opened 2021-08-06
WIP: Disable chrony for connected jobs to test

ardaguclu:
e2e-metal-ipi jobs are failing in waterfall pattern. In logs, there are apiserver restarts, handshake failures, etcd connection errors, just a hypothesis triggering e2e-metal-ipi jobs after disabling chrony.

### DIFF
--- a/utils.sh
+++ b/utils.sh
@@ -102,7 +102,7 @@ function create_cluster() {
     mkdir -p ${assets_dir}/openshift
     generate_assets
 
-    if [ -z "${NTP_SERVERS}" ];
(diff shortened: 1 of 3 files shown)


utils.sh
```diff
@@ -102,7 +102,7 @@ function create_cluster() {
     mkdir -p ${assets_dir}/openshift
     generate_assets
 
-    if [ -z "${NTP_SERVERS}" ];
+    if [[ ! -z "${MIRROR_IMAGES}" && -z "${NTP_SERVERS}" ]];
```
elfosardo, line 105: maybe "-n" instead of "! -z" ?

PR #1286, opened 2021-08-17
Add support for additional networks

cybertron:
For testing kubernetes-nmstate on OpenShift baremetal, we need two
extra nics per node. This patch adds network definitions for them
and a new env var ADDITIONAL_NETWORKS, which if set to "y" will
trigger creation of the new networks.

### DIFF
--- a/vm_setup_vars.yml
+++ b/vm_setup_vars.yml
@@ -76,4 +76,48 @@ external_network:
         - domain: "apps.{{ cluster_domain }}"
           addr: "127.0.0.1"
 
-networks: "{{ provisioning_network + external_network }}"
+nmstate_network_1_cidr_v4: "{{ lookup('env', 'NMSTATE_NETWORK_1_SUBNET_V4') | default('192.168.221.0/24',true) }}"
+nmstate_network_1_cidr_v6: "{{ lookup('env', 'NMSTATE_NETWORK_1_SUBNET_V6') | default('fd2e:6f44:5dd8:ca56::/120',true) }}"
+nmstate_network_1_cidr: "{{ nmstate_network_1_cidr_v4 | default(nmstate_network_1_cidr_v6, true) }}"
+
+nmstate_network_2_cidr_v4: "{{ lookup('env', 'NMSTATE_NETWORK_2_SUBNET_V4') | default('192.168.222.0/24',true) }}"
+nmstate_network_2_cidr_v6: "{{ lookup('env', 'NMSTATE_NETWORK_2_SUBNET_V6') | default('fd2e:6f44:5dd8:cc56::/120',true) }}"
+nmstate_network_2_cidr: "{{ nmstate_network_2_cidr_v4 | default(nmstate_network_2_cidr_v6, true) }}"
+
+additional_networks:
+  - name: "nmstate1"
+    bridge: "nmstate1"
+    forward_mode: nat
+    lease_expiry: 60
+    address_v4: "{{ nmstate_network_1_cidr_v4|nthhost(1)|default('', true) }}"
+    netmask_v4: "{{ nmstate_network_1_cidr_v4|ipaddr('netmask') }}"
+    address_v6: "{{ nmstate_network_1_cidr_v6|nthhost(1)|default('', true) }}"
+    prefix_v6: "{{ nmstate_network_1_cidr_v6|ipaddr('prefix') }}"
+    dhcp_range_v4:
+      - "{{ nmstate_network_1_cidr_v4|nthhost(20) }}"
+      - "{{ nmstate_network_1_cidr_v4|nthhost(60) }}"
+    dhcp_range_v6:
+      - "{{ nmstate_network_1_cidr_v6|nthhost(20) }}"
+      - "{{ nmstate_network_1_cidr_v6|nthhost(60) }}"
+    nat_port_range:
+      - 1024
+      - 65535
+  - name: "nmstate2"
+    bridge: "nmstate2"
+    forward_mode: nat
+    lease_expiry: 60
+    address_v4: "{{ nmstate_network_2_cidr_v4|nthhost(1)|default('', true) }}"
+    netmask_v4: "{{ nmstate_network_2_cidr_v4|ipaddr('netmask') }}"
+    address_v6: "{{ nmstate_network_2_cidr_v6|nthhost(1)|default('', true) }}"
+    prefix_v6: "{{ nmstate_network_2_cidr_v6|ipaddr('prefix') }}"
+    dhcp_range_v4:
+      - "{{ nmstate_network_2_cidr_v4|nthhost(20) }}"
+      - "{{ nmstate_network_2_cidr_v4|nthhost(60) }}"
+    dhcp_range_v6:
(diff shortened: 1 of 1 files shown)


hardys:
I can imagine quite a few other use-cases for multi-network testing, where it'd be really useful to have the ability to specify arbitrary networks/names - similar to what I did for flavors in https://github.com/metal3-io/metal3-dev-env/pull/129

Not blocking on it, but what do you think about perhaps collaborating on a more flexible solution, vs this hard-coded approach?

I'm thinking something like we'd use the same ipmath approach to calculate sane defaults (e.g increment a subnet by 1 for each network like you've done here in the defaults) https://github.com/metal3-io/metal3-dev-env/blob/master/vm-setup/roles/libvirt/templates/network.xml.j2#L44

That would admittedly further complicate the j2 template, but I'm thinking it may be worth it to enable a more flexible end result?

hardys:
Another option (if we want to avoid changing metal3-dev-env) would be to calculate `additional_networks` e.g in a playbook task that runs before including the existing `setup-playbook.yml` - then we could have a dev-scripts variable which just accepts a list of additional network names, with default for everything (which can still be overridden per-network basis but we wouldn't require the hard-coded per-network vars)

cybertron:
I'm open to making this more general, although we need to be a little careful with defaulting too much. I've had issues with using this patch in ipv6 deployments because the existence of ipv4 addresses on some interfaces confuses (or confused, I haven't checked recently) some of the address logic. I guess if there's some way to specify that a given field should be blank it should be fine.

The question of where to do it depends on whether we think this will be useful to other metal3 users. I could see an argument from metal3 that an interface is already provided to create an arbitrary number of networks. This would potentially simplify usage of that interface, but otherwise it would be functionally identical.

Either way it would be nice to have a more general way to do this in dev-scripts. Even if we wire it into metal3-dev-env we'll need some way to specify the additional networks here too so maybe that's a good starting point? My Ansible design is not super strong so I don't have strong opinions about how we do that.

creydr:
@cybertron can you rebase this an add the `lease_expiry` attribute to `additional_networks`?
-> [slack discussion](https://coreos.slack.com/archives/CFP6ST0A3/p1635868138101500)

hardys:
Sorry for the delay, but https://github.com/metal3-io/metal3-dev-env/pull/852 outlines a potentially more flexible approach - the same as defined here can be achieved with something like:

```
export EXTRA_NETWORK_NAMES="nmstate1 nmstate2"
export NMSTATE1_NETWORK_SUBNET_V4='192.168.221.0/24'
export NMSTATE1_NETWORK_SUBNET_V6='fd2e:6f44:5dd8:ca56::/120'
export NMSTATE2_NETWORK_SUBNET_V4='192.168.222.0/24'
export NMSTATE2_NETWORK_SUBNET_V6='fd2e:6f44:5dd8:cc56::/120'
```
However the network names (and number of networks) becomes configurable (and we gain the feature in the upstream environment, which may be useful on occasion).

cybertron:
I tried the metal3-dev-env patch today and didn't get extra networks defined. I'll take another look tomorrow and see if I missed anything, but there may be more needed for that. Is there any dev-scripts integration required?

hardys:
> I tried the metal3-dev-env patch today and didn't get extra networks defined. I'll take another look tomorrow and see if I missed anything, but there may be more needed for that. Is there any dev-scripts integration required?

Hmm, it's working for me via metal3-dev-env, not yet tried via dev-scripts - did you [export METAL3_DEV_ENV](https://github.com/openshift-metal3/dev-scripts/blob/master/01_install_requirements.sh#L13) to use a local checkout with my changes applied?

```
[shardy@master-1 metal3-dev-env]$ grep NETWORK config_shardy.sh 
export EXTRA_NETWORK_NAMES="nmstate1 nmstate2"
export NMSTATE1_NETWORK_SUBNET_V4='192.168.221.0/24'
export NMSTATE1_NETWORK_SUBNET_V6='fd2e:6f44:5dd8:ca56::/120'
export NMSTATE2_NETWORK_SUBNET_V4='192.168.222.0/24'
export NMSTATE2_NETWORK_SUBNET_V6='fd2e:6f44:5dd8:cc56::/120'
[shardy@master-1 metal3-dev-env]$ sudo virsh net-list 
 Name           State    Autostart   Persistent
-------------------------------------------------
 baremetal      active   yes         yes
 default        active   yes         yes
 mk-minikube    active   yes         yes
 nmstate1       active   yes         yes
 nmstate2       active   yes         yes
 provisioning   active   yes         yes

[shardy@master-1 metal3-dev-env]$ cat /var/lib/libvirt/dnsmasq/
baremetal.addnhosts    default.hostsfile      nmstate1.hostsfile     virbr0.macs
baremetal.conf         mk-minikube.addnhosts  nmstate1.status        virbr0.status
baremetal.hostsfile    mk-minikube.conf       nmstate2.addnhosts     virbr1.macs
baremetal.status       mk-minikube.hostsfile  nmstate2.conf          virbr1.status
default.addnhosts      nmstate1.addnhosts     nmstate2.hostsfile     
default.conf           nmstate1.conf          nmstate2.status        
[shardy@master-1 metal3-dev-env]$ cat /var/lib/libvirt/dnsmasq/nmstate1.hostsfile 
00:7f:f7:5e:06:f3,192.168.221.20,node-0
00:7f:f7:5e:06:fb,192.168.221.21,node-1
id:00:03:00:01:00:7f:f7:5e:06:f3,node-0,[fd2e:6f44:5dd8:ca56::14]
id:00:03:00:01:00:7f:f7:5e:06:fb,node-1,[fd2e:6f44:5dd8:ca56::15]
```

hardys:
@cybertron note the environment variable names I used are slightly different to this PR e.g `<network name, uppercase>_NETWORK_SUBNET_V4` and it must match the name in `EXTRA_NETWORK_NAMES` (but name can be lowercase)

hardys:
Ok, can confirm it's also not working for me via dev-scripts, will look into why and report back

cybertron:
I went ahead and rebased this because I needed it to test the kubernetes-nmstate rebase anyway. I haven't had a chance to dig into Steve's patch any further to figure out why it isn't applying in dev-scripts, although I can see at least some of the related tasks being run in the logs. :-/

hardys:
> I went ahead and rebased this because I needed it to test the kubernetes-nmstate rebase anyway. I haven't had a chance to dig into Steve's patch any further to figure out why it isn't applying in dev-scripts, although I can see at least some of the related tasks being run in the logs. :-/

Sorry @cybertron I forgot to report back - I think it's a variable precedence issue, because in the metal3-dev-env PR I append to `networks` but then the `--extra-vars` CLI value passed from dev-scripts overrides the value (surprisingly this happens silently, even if you use set_fact it seems to be silently ignored).

I think the fix is to rename some of the things in the metal3-dev-env PR to avoid use of the existing variable name, but I've not yet had time to make that change.

hardys:
Ok I think I worked around the variable precedence issues, see  https://github.com/metal3-io/metal3-dev-env/pull/852 updated, and can be tested with https://github.com/openshift-metal3/dev-scripts/pull/1312 as an alternative approach

cybertron:
Addressed by #1312.

PR #1287, opened 2021-09-01
Bump go version to 1.16

hardys:
Since https://github.com/openshift/installer/pull/5120
we need >=1.16 to build the installer from source


hardys:
/hold 

Looks like we need to update metal3-templater to use modules

hardys:
Obsoleted by https://github.com/openshift-metal3/dev-scripts/pull/1292

PR #1288, opened 2021-09-06
Use https in the generated clouds.yaml for the cluster

dtantsur:
We have switched to TLS long ago, the generated clouds.yaml still has HTTP.

### DIFF
--- a/metal3-templater.go
+++ b/metal3-templater.go
@@ -3,13 +3,14 @@ package main
 import (
 	"flag"
 	"fmt"
-	"github.com/apparentlymart/go-cidr/cidr"
-	"github.com/openshift/installer/pkg/ipnet"
 	"net"
 	"net/url"
 	"os"
 	"strings"
 	"text/template"
+
+	"github.com/apparentlymart/go-cidr/cidr"
+	"github.com/openshift/installer/pkg/ipnet"
 )
 
 type templater struct {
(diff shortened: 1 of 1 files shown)


metal3-templater.go
```diff
@@ -143,8 +144,8 @@ func main() {
 		// ProvisioningIP
 		size, _ := ipnet.IPNet.Mask.Size()
 		templateData.ProvisioningIP = fmt.Sprintf("%s/%d", clusterIP, size)
-		templateData.ClusterIronicURL = fmt.Sprintf("http://%s", net.JoinHostPort(clusterIP, "6385"))
-		templateData.ClusterInspectorURL = fmt.Sprintf("http://%s", net.JoinHostPort(clusterIP, "5050"))
+		templateData.ClusterIronicURL = fmt.Sprintf("https://%s", net.JoinHostPort(clusterIP, "6385"))
+		templateData.ClusterInspectorURL = fmt.Sprintf("https://%s", net.JoinHostPort(clusterIP, "5050"))
```
hardys, line 148: Looking at CBO this will only work for 4.8 and onwards, I'm not sure that many folks are testing older versions locally now, but we might want to add a conditional if so (looking at https://access.redhat.com/support/policy/updates/openshift we may have to sometimes test 4.7/4.6EUS)

hardys:
I'm fine to merge this as is, but if folks need this to work also on older versions (not sure that's necessarily the case) we might want to add a conditional as a followup

derekhiggins:
> I'm fine to merge this as is, but if folks need this to work also on older versions (not sure that's necessarily the case) we might want to add a conditional as a followup

Agreed, supporting the current use case makes sense, we can add a conditional for 4.7 if it becomes a problem.

/lgtm

flaper87 (approved):
/lgtm

dtantsur:
/retest

Given that there workaround is quite simple, I'm not sure we need a conditional here.

PR #1290, opened 2021-09-10
Switch to nmcli commands, allow MTU to be specified

mrbojangles3:
@hardys , this is a draft PR, but I was hoping you would have time to take a look at it, make some comments, and I could fix them as my work hours are shifted from yours. This code results in a working cluster for me, using OVNKubernetes and no baremetal workers. Also while the MTU attaches to the interface, not the bridge. Assuming the bridge will select the lowest mtu of all the ports / interfaces attached to it, I must be missing a port somewhere. 
If you have time, I would love feedback on this PR.


02_configure_host.sh
```diff
@@ -123,63 +123,64 @@ if [ "$MANAGE_PRO_BRIDGE" == "y" ]; then
...
-    if [ ! -e /etc/sysconfig/network-scripts/ifcfg-${PROVISIONING_NETWORK_NAME} ] ; then
+    sudo nmcli con add type bridge con-name ${PROVISIONING_NETWORK_NAME} ifname ${PROVISIONING_NETWORK_NAME}
+    sudo nmcli con modify ${PROVISIONING_NETWORK_NAME} connection.zone ${ZONE}
+    sudo nmcli con modify ${PROVISIONING_NETWORK_NAME} connection.autoconnect TRUE
+    # This setting will bring the bridge ports up when the bridge itself is brought up.
+    sudo nmcli con modify ${PROVISIONING_NETWORK_NAME} connection.autoconnect-slaves 1
         if [[ "$(ipversion $PROVISIONING_HOST_IP)" == "6" ]]; then
-            echo -e "DEVICE=${PROVISIONING_NETWORK_NAME}\nTYPE=Bridge\nONBOOT=yes\nNM_CONTROLLED=no\nIPV6_AUTOCONF=no\nIPV6INIT=yes\nIPV6ADDR=${PROVISIONING_HOST_IP}/64${ZONE}" | sudo dd of=/etc/sysconfig/network-scripts/ifcfg-${PROVISIONING_NETWORK_NAME}
+	    sudo nmcli con modify ${PROVISIONING_NETWORK_NAME} ipv6.address $PROVISIONING_HOST_IP/64
+	    sudo nmcli con modify ${PROVISIONING_NETWORK_NAME} ipv6.addr-gen-mode stable-privacy
```
hardys, line 133: Given that the bridge is statically configured, so we need the addr-gen-mode?

Not blocking, just wondered why it was added - currently we only support DHCPv6 for ipv6 testing via dev-scripts, and we configure the [libvirt DHCP reservations](https://github.com/metal3-io/metal3-dev-env/blob/master/vm-setup/roles/libvirt/templates/network.xml.j2#L93) so the IPs are predictable (relatedly we configure the target cluster hosts to have a [predictable DUID/IAID](https://github.com/openshift/machine-config-operator/blob/master/templates/common/on-prem/files/NetworkManager-onprem.conf.yaml#L9..L10))

hardys:
Thanks for starting this @mrbojangles3 overall looks good - one question added, and I'm wondering how we influence the nic MTU for the deployed RHCOS hosts - I was expecting to either have to modify the libvirt-managed dnsmasq config to pass the MTU option, or pass a MachineConfig to explicitly configure the nic MTU?

When that part is figured out, we'll need to also set the SDN MTU in the install-config here: https://github.com/openshift-metal3/dev-scripts/blob/master/ocp_install_env.sh#L143

02_configure_host.sh
mrbojangles3: I am glad you asked, as I was unsure when I put this in. My goal was to get something working on my local machine and I was hitting a lot of "gotchas" about the order of the commands and what not. so I will test with this line removed to make sure it still works.

mrbojangles3:
I saw the section in the ocp_install_env.sh. I haven't had much experience with "here scripts / here docs" inside of a bash file I want to add the MTU regardless of the v4,v6,v4v6 option. I can add the MTU call in each of the if branches, but ideally it would be in one place. How can that be accomplished with the here doc? or is adding in 3 places sufficient ?

mrbojangles3:
In addition to the above test failing. I am also not getting any dhcp offer for the bare metal scale out. Sine I was able to get this in the past, I think this is related to my changes.

mrbojangles3:
@hardys I am not sure why the most recent test failed. It looks like the scripts tried to reach out via ipv6 address and were unable to make contact. What I think is the relevant error is reproduced below. It seems the local mirror isn't routable? My ipv6 change is supposed to only apply to prov network, I only intend to set the interface as the ipv6 route for the given ipv6 range, as I don't want to modify the routes for the whole box. It seems narrowly setting the route is not correct? 
```+(utils.sh:535): switch_to_internal_dns(): echo server=147.75.207.207
+(utils.sh:535): switch_to_internal_dns(): sudo tee /etc/NetworkManager/dnsmasq.d/upstream.conf
server=147.75.207.207
+(utils.sh:538): switch_to_internal_dns(): sudo systemctl reload NetworkManager
+(./02_configure_host.sh:215): main(): sudo sed -i /virthost.ostest.test.metalkube.org/d /etc/hosts
+(./02_configure_host.sh:216): main(): echo 'fd2e:6f44:5dd8:c956::1 virthost.ostest.test.metalkube.org'
+(./02_configure_host.sh:216): main(): sudo tee -a /etc/hosts
fd2e:6f44:5dd8:c956::1 virthost.ostest.test.metalkube.org
+(./02_configure_host.sh:220): main(): rm -f /root/private-mirror-ostest.json
+(./02_configure_host.sh:221): main(): [[ ! -z true ]]
+(./02_configure_host.sh:223): main(): sudo podman login --authfile /root/private-mirror-ostest.json -u ocp-user -p ocp-pass virthost.ostest.test.metalkube.org:5000
Error: authenticating creds for "virthost.ostest.test.metalkube.org:5000": error pinging docker registry virthost.ostest.test.metalkube.org:5000: Get "https://virthost.ostest.test.metalkube.org:5000/v2/": dial tcp [fd2e:6f44:5dd8:c956::1]:5000: connect: no route to host```

hardys:
> @hardys I am not sure why the most recent test failed. It looks like the scripts tried to reach out via ipv6 address and were unable to make contact. What I think is the relevant error is reproduced below. It seems the local mirror isn't routable? My ipv6 change is supposed to only apply to prov network, I only intend to set the interface as the ipv6 route for the given ipv6 range, as I don't want to modify the routes for the whole box. It seems narrowly setting the route is not correct?

Hmm I may have to test locally if you need some help to debug, but note all of the changes which touch e.g `BAREMETAL_NETWORK_NAME` are relevant here, because fd2e:6f44:5dd8:c956::1 is the bridge IP of the baremetal network (where we normally reach the local container registry)

I'd start by looking for that IP e.g in `ip a`, then prove the registry is accessible from the virthost itself, then debug with e.g tracerout/nmap from a VM connected to the baremetal network (a live image running on one of the dev-scripts created virsh VMs may be helpful here).

Let me know how you get on - if needed I'll pull the PR and kick the tyres locally so we can make progress - thanks for your efforts so far!

hardys:
Another suggestion is to bisect this PR, e.g it may make sense to decouple the nmcli refactoring from the MTU update?.

PR #1292, opened 2021-09-17
Bump go_version to 1.17.1

honza:
baremetal-operator depends on 1.16, and there is no reason to not use
the latest

### DIFF
--- a/01_install_requirements.sh
+++ b/01_install_requirements.sh
@@ -47,7 +47,7 @@ ansible-galaxy collection install ansible.netcommon ansible.posix community.gene
 ANSIBLE_FORCE_COLOR=true ansible-playbook \
   -e "working_dir=$WORKING_DIR" \
   -e "virthost=$HOSTNAME" \
-  -e "go_version=1.14.4" \
+  -e "go_version=1.17.1" \
   -i vm-setup/inventory.ini \
   -b -vvv vm-setup/install-package-playbook.yml
 popd
--- a/go.mod
+++ b/go.mod
@@ -1,0 +1,10 @@
+module github.com/openshift-metal3/dev-scripts/metal3-templater
+
+go 1.17
+
+require (
+	github.com/apparentlymart/go-cidr v1.1.0
+	github.com/openshift/installer v0.16.1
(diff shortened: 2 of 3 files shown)


hardys:
As mentioned in https://github.com/openshift-metal3/dev-scripts/pull/1287 we need to fix `metal3-templater.go` so it plays nicely with the latest go version, see CI failures.

In my local env I worked around it with `GO111MODULE=off go run metal3-templater.go ...` but suspect we should fix things to actually use modules instead - not found time to do that yet, so if you're happy to do so in this PR @honza I can close mine?

I don't have a strong opinion on which version we choose, but FWIW I was going for 1.16 to align with the installer requirements ref https://github.com/openshift/installer/pull/5120

hardys:
> /retest-required

The CI failures are real - see [logs](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift-metal3_dev-scripts/1292/pull-ci-openshift-metal3-dev-scripts-master-e2e-metal-ipi/1439869917632401408/artifacts/e2e-metal-ipi/baremetalds-devscripts-setup/artifacts/root/dev-scripts/logs/06_create_cluster-2021-09-20-084751.log)

andfasano:
> As mentioned in #1287 we need to fix `metal3-templater.go` so it plays nicely with the latest go version, see CI failures.
> 
> In my local env I worked around it with `GO111MODULE=off go run metal3-templater.go ...` but suspect we should fix things to actually use modules instead - not found time to do that yet, so if you're happy to do so in this PR @honza I can close mine?
> 
> I don't have a strong opinion on which version we choose, but FWIW I was going for 1.16 to align with the installer requirements ref [openshift/installer#5120](https://github.com/openshift/installer/pull/5120)

I've added a version with the modules here: https://github.com/openshift-metal3/dev-scripts/pull/1295

honza:
Sorry for the late reply, somehow I missed the notifications. I just pushed the go module bits.

elfosardo (approved):

PR #1293, opened 2021-09-18
Small fixes to run remote_nodes.sh in Openshift CI

lranjbar:
The goal is to run this script in Openshift CI as it's own step outside of `baremetalds-devscripts-setup`.

### Summary of changes:
- Adding `APPLY_REMOTE_NODES` similar to `APPLY_EXTRA_WORKERS` I want to do this later on. Specifically I want to allow Assisted Installer to apply the remote nodes in my tests.
- Splitting out to `REMOTE_IP_STACK` I believe unlinking this from the dev-scripts `IP_STACK` will work better since it will be in a different step than `baremetalds-devscripts-setup`.
- Moving `TEARDOWN_PLAYBOOK` is a small fix otherwise the name of the playbook is `-teardown_playbook.yml`
- OCP doesn't like changes to the namespace `openshift-machine-api` and with good reason. Here is the error I encountered before putting in both if statements:
```
+(remote_nodes.sh:102): cleanup_remote_cluster(): oc delete ns openshift-machine-api
Error from server (Forbidden): namespaces "openshift-machine-api" is forbidden: this namespace may not be deleted
```
- Changes to `REMOTE_CLUSTER_NAME` to be something much shorter. The `baremetal_network_name` has a requiement to be pretty short. Without this change I got the following errors:
```
TASK [Start libvirt networks] *************************************************************************************************
task path: /opt/dev-scripts/metal3-dev-env/vm-setup/roles/libvirt/tasks/network_setup_tasks.yml:38
.
.
.
    "msg": "error creating bridge interface ostest-openshift-machine-api: Numerical result out of range"
}
```
and
```
+(remote_nodes.sh:93): setup_remote_cluster(): sudo iptables -I FORWARD 1 -o ostest-openshift-machine-api -i ostestbm -j ACCEPT
iptables v1.8.4 (nf_tables): interface name `ostest-openshift-machine-api' must be shorter than IFNAMSIZ (15)
Try `iptables -h' or 'iptables --help' for more information.
```

On a different note it won't be needed for Openshift CI but I wasn't able to get the cleanup function to really work. Ansible wasn't importing the tasks from metal3-dev-env correctly. I've just been manually tearing down stuff locally. 🤷🏼‍♀️


remote_nodes.sh
```diff
@@ -91,13 +93,24 @@ function setup_remote_cluster() {
...
+		oc apply -f ${OCP_DIR}/remote_host_manifests.yaml
+	fi
+
 }
 
 function cleanup_remote_cluster() {
-	# Remove ns
-	oc delete ns ${NS}
+	# Remove namespace - OCP will not allow you to remove openshift-machine-api
+	if [[ $NS != "openshift-machine-api" ]]
```
flaper87, line 110: For the sake of overengineering and future prediction, what do you think about having a list of namespaces (or space-separated string of namespaces) that should be "protected"?

remote_nodes.sh
```diff
@@ -7,12 +7,13 @@ source ocp_install_env.sh
...
 NS=${2:-openshift-machine-api}
-TEARDOWN_PLAYBOOK=/tmp/${REMOTE_CLUSTER_NAME}-teardown-playbook.yml
 IPTABLES=iptables
 
-REMOTE_CLUSTER_NAME=${REMOTE_CLUSTER_NAME:-${CLUSTER_NAME}${NS}}
+APPLY_REMOTE_NODES=${APPLY_REMOTE_NODES:-"true"}
+REMOTE_CLUSTER_NAME=${REMOTE_CLUSTER_NAME:-"${CLUSTER_NAME}spoke"}
 REMOTE_CLUSTER_NUM_MASTERS=${REMOTE_CLUSTER_NUM_MASTERS:-1}
 REMOTE_CLUSTER_NUM_WORKERS=${REMOTE_CLUSTER_NUM_WORKERS:-0}
+REMOTE_IP_STACK=${REMOTE_IP_STACK:-"v4v6"}
```
flaper87, line 16: What do you think about using `$IP_STACK` as the default? This should also help us preserve compatibility with the current workflow, if `REMOTE_IP_STACK` is not set.

flaper87 (changes requested):
Couple of comments, otherwise lgtm

remote_nodes.sh
lranjbar: Doing some reason on this.. we might be able to just if the namespace starts with "openshift" but having a running list is good too.

remote_nodes.sh
lranjbar: I don't really see an issue with this would make running it locally easier too. :)

remote_nodes.sh
flaper87: Either works for me! Considering all the `openshift-` namespaces as "protected" sounds sane.

PR #1297, opened 2021-10-04
Add ability to inject static ips into install-config

imain:
Set up a STATIC_IPS configuration flag that configures dev-scripts
to use static IPs (v4 and/or v6) for the baremetal network.

Co-Authored-By: Caleb Boylan <cboylan@redhat.com>

### DIFF
--- a/01_install_requirements.sh
+++ b/01_install_requirements.sh
@@ -19,7 +19,7 @@ if [ -z "${METAL3_DEV_ENV}" ]; then
   # TODO -- come up with a plan for continuously updating this
   # Note we only do this in the case where METAL3_DEV_ENV is
   # unset, to enable developer testing of local checkouts
-  git reset ee1b2fec44e22700c8d250d5e0e371e3fd0aba17 --hard
+  #git reset 67afcb3315b44e2c68121493a819819966e588a2 --hard
   popd
 fi
 
--- a/common.sh
+++ b/common.sh
@@ -56,11 +56,22 @@ export CLUSTER_NAME=${CLUSTER_NAME:-ostest}
 
 export PROVISIONING_NETWORK_PROFILE=${PROVISIONING_NETWORK_PROFILE:-"Managed"}
 
+# Set to 1 if you want dev scripts to configure static IPs for your deployment.
+export STATIC_IPS=${STATIC_IPS:-""}
+# Set to false to disable DHCP on all networks.
+export DHCP_ENABLED=${DHCP_ENABLED:-"true"}
+
 # Network interface names can only be 15 characters long, so
 # abbreviate provisioning and baremetal and add them as suffixes to
 # the cluster name.
 export PROVISIONING_NETWORK_NAME=${PROVISIONING_NETWORK_NAME:-${CLUSTER_NAME}pr}
 export BAREMETAL_NETWORK_NAME=${BAREMETAL_NETWORK_NAME:-${CLUSTER_NAME}bm}
+
+if [[ "$PROVISIONING_NETWORK_PROFILE" == "Disabled" ]]; then
+    export PROVISIONING_NETWORK_INTERFACE="enp2s1"
+else
+    export PROVISIONING_NETWORK_INTERFACE="enp2s0"
+fi
 
 export BASE_DOMAIN=${BASE_DOMAIN:-test.metalkube.org}
 export CLUSTER_DOMAIN="${CLUSTER_NAME}.${BASE_DOMAIN}"
@@ -157,7 +168,7 @@ else
   export BMC_DRIVER=${BMC_DRIVER:-mixed}
 fi
 
-if [[ "$PROVISIONING_NETWORK_PROFILE" == "Disabled" ]]; then
+if [[ "$PROVISIONING_NETWORK_PROFILE" == "Disabled" || -n "$STATIC_IPS" ]]; then
   export BMC_DRIVER="redfish-virtualmedia"
 fi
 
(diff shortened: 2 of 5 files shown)


imain:
Requires https://github.com/metal3-io/metal3-dev-env/pull/820

Same comment applies here.  We could plumb this with a new vm_setup_vars.yaml that sets up a static IP configuration for the network.  That may be a better approach.  Let me know.

utils.sh
```diff
@@ -212,6 +212,10 @@ function node_map_to_install_config_hosts() {
 
     for ((idx=$start_idx;idx<$(($1 + $start_idx));idx++)); do
       name=$(node_val ${idx} "name")
+      ipv4_addr=$(node_val ${idx} "static_ipv4")
+      ipv4_addr_prefix=$(node_val ${idx} "static_ipv4_prefix")
+      ipv6_addr=$(node_val ${idx} "static_ipv6")
+      ipv6_addr_prefix=$(node_val ${idx} "static_ipv6_prefix")
```
hardys, line 218: If we don't use metal3-dev-env we could use the [nth_ip function](https://github.com/openshift-metal3/dev-scripts/blob/master/network.sh#L3) instead here?

utils.sh
```diff
@@ -230,6 +234,10 @@ function node_map_to_install_config_hosts() {
       if [[ "$boot_mode" == "null" ]]; then
              boot_mode="UEFI"
       fi
+      static_interface=enp2s0
+      if [[ "$PROVISIONING_NETWORK_PROFILE" == "Disabled" ]]; then
+        static_interface=enp1s0
+      fi
```
hardys, line 240: Maybe be cleaner if we created a variable in `common.sh` and assigned it based on PROVISIONING_NETWORK_PROFILE?

(That way this could still work if testing with baremetal servers)

imain: done.

01_install_requirements.sh
```diff
@@ -19,7 +19,7 @@ if [ -z "${METAL3_DEV_ENV}" ]; then
   # TODO -- come up with a plan for continuously updating this
   # Note we only do this in the case where METAL3_DEV_ENV is
   # unset, to enable developer testing of local checkouts
-  git reset ee1b2fec44e22700c8d250d5e0e371e3fd0aba17 --hard
+  #git reset 67afcb3315b44e2c68121493a819819966e588a2 --hard
```
hardys, line 22: This needs to be removed  - note you can use `METAL3_DEV_ENV` to reference a local source checkout for testing with latest metal3-dev-env  :)

common.sh
```diff
@@ -157,7 +168,7 @@ else
   export BMC_DRIVER=${BMC_DRIVER:-mixed}
 fi
 
-if [[ "$PROVISIONING_NETWORK_PROFILE" == "Disabled" ]]; then
+if [[ "$PROVISIONING_NETWORK_PROFILE" == "Disabled" || -n "$STATIC_IPS" ]]; then
```
hardys, line 171: I think we should support both virtualmedia and PXE paths now?

vm_setup_vars.yml
```diff
@@ -51,6 +51,8 @@ provisioning_network:
     bridge: "{{ provisioning_network_name  }}"
     forward_mode: bridge
 
+virsh_dhcp_enabled: "{{ lookup('env', 'DHCP_ENABLED')|default('', true) }}"
```
hardys, line 54: I don't think we need this, since the environment variable is already respected in the defaults ref https://github.com/metal3-io/metal3-dev-env/pull/879 ?

utils.sh
```diff
@@ -241,7 +241,41 @@ function node_map_to_install_config_hosts() {
           disableCertificateVerification: ${disable_certificate_verification}
         bootMACAddress: ${mac}
         bootMode: ${boot_mode}
+        networkConfig: |
+          routes:
+            config:
+            - destination: 0.0.0.0/0
+              next-hop-address: ${PROVISIONING_HOST_EXTERNAL_IP}
+              next-hop-interface: ${PROVISIONING_NETWORK_INTERFACE}
```
derekhiggins, line 249: We normally would have the default route on the External network, is the change to the provisioning network intentional?

hardys:
@imain any updates on this?  It'd be useful to get this done so we can attempt reproduces for bugs (for example https://bugzilla.redhat.com/show_bug.cgi?id=2036677) - I think other than the comments already added, we'll also need to inject static configuration for the bootstrap VM?

It'd be great to either create a new periodic CI job too based on this work, or switch the config of one of the existing ones to provide coverage.

hardys:
Note there are also a couple of comments on https://github.com/metal3-io/metal3-dev-env/pull/879 which need resolving, then we can merge that and update the dev-scripts pin

vm_setup_vars.yml
```diff
@@ -59,6 +61,7 @@ external_network:
     netmask_v4: "{{ baremetal_network_cidr_v4|ipaddr('netmask') }}"
     address_v6: "{{ baremetal_network_cidr_v6|nthhost(1)|default('', true) }}"
     prefix_v6: "{{ baremetal_network_cidr_v6|ipaddr('prefix') }}"
+    dhcp_enabled: "{{ virsh_dhcp_enabled|bool }}"
```
hardys, line 64: Another option is we conditionally define `dhcp_range*` - then we won't need the new metal3-dev-env flag since the `<dhcp` stanza is already skipped when the range is omitted

hardys:
I pushed https://github.com/openshift-metal3/dev-scripts/pull/1336 which handles the bootstrap VM ip configuration, which we'll need for this

vm_setup_vars.yml
hardys: I pushed https://github.com/metal3-io/metal3-dev-env/pull/902 which means we can just set the `dhcp_range` variables to an empty list, which avoids the need for a new variable

derekhiggins:
Closing as this has been handled by https://github.com/openshift-metal3/dev-scripts/pull/1418

PR #1298, opened 2021-10-05
Redact some info installer logs after creation

ardaguclu:


utils.sh
```diff
@@ -162,7 +162,11 @@ function create_cluster() {
     fi
 
     trap auth_template_and_removetmp EXIT
-    $OPENSHIFT_INSTALLER --dir "${assets_dir}" --log-level=debug create cluster
+    $OPENSHIFT_INSTALLER --dir "${assets_dir}" --log-level=debug create cluster 2>&1 | grep --line-buffered -v 'password\|X-Auth-Token\|UserData:' &
+    wait "$!"
+    ret="$?"
```
hardys, line 167: Would it be better to sanitize this in the installer directly?

Also can you clarify why you need to background the installer process here vs just grep the output?

ardaguclu: Other platforms use this way to redact these information https://github.com/openshift/release/blob/af5031f1a342d0a3ee858a2075910c4fbcf030ea/ci-operator/step-registry/ipi/install/install/ipi-install-install-commands.sh#L111. Only reason is to be compatible with them. 

I don't know, but maybe there is a reason not hiding in installer instead redacting in other places.

ardaguclu: We might not need this PR, if https://github.com/openshift/release/pull/22495 fills all gaps.
cc: @andfasano

andfasano: This PR is required to sanitize the stdout (until it will be directly done by installer). In the meantime seems an acceptable solution for me

derekhiggins: > Also can you clarify why you need to background the installer process here vs just grep the output?

Also wondering this?

andfasano: Ok, probably it's not required in our case, and we could get rid of both the background and wait command

derekhiggins: The background thing comes from here https://github.com/openshift/release/pull/2680/files
It seems to be so the command can be killed reliably

andfasano: Thanks @derekhiggins for the link. I guess it could be fine since our installation part is executed [via ssh](https://github.com/openshift/release/blob/a8191dd3778ad2fec4bd1029318ad51326a89de0/ci-operator/step-registry/baremetalds/devscripts/setup/baremetalds-devscripts-setup-commands.sh#L103) and not via a direct command in the prow pod

PR #1301, opened 2021-10-08
Update metal3-dev-env pin, and ansible/python version

hardys:
Currently switching between dev-scripts and metal3-dev-env is inconvenient
because metal3-dev-env requires a much newer ansible version than we have
packaged (at least in the RHEL case)

Newer ansible also requires python 3.8+ (which is available along with 3.9,
but the default is 3.6)

So, update to install ansible pinned to the same version as metal3-dev-env
via pip, also upgrade python and set alternatives to use the newer version.

### DIFF
--- a/01_install_requirements.sh
+++ b/01_install_requirements.sh
@@ -19,9 +19,13 @@ if [ -z "${METAL3_DEV_ENV}" ]; then
   # TODO -- come up with a plan for continuously updating this
   # Note we only do this in the case where METAL3_DEV_ENV is
   # unset, to enable developer testing of local checkouts
-  git reset 184d0afaab77424872f0441c9e11e23a620e5b57 --hard
+  git reset 67afcb3315b44e2c68121493a819819966e588a2 --hard
   popd
 fi
+
+# This must be aligned with the metal3-dev-env pinned version above, see
+# https://github.com/metal3-io/metal3-dev-env/blob/master/lib/common.sh
+export ANSIBLE_VERSION=${ANSIBLE_VERSION:-"4.6.0"}
 
 # Update to latest packages first
 sudo dnf -y upgrade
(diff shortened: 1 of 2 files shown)


dtantsur:
/lgtm

I need this for https://github.com/openshift/cluster-baremetal-operator/pull/206, so thank you!

On the other hand, we should probably convince upstream to install ansible in a virtual env (similarly to how Bifrost does it).

01_install_requirements.sh
```diff
@@ -33,17 +37,26 @@ export DISTRO="${ID}${VERSION_ID%.*}"
...
 # Install ansible, other packages are installed via
 # vm-setup/install-package-playbook.yml
-sudo dnf -y install python3 ansible
-sudo alternatives --set python /usr/bin/python3
+# Note recent ansible needs python >= 3.8 so we install 3.9 here
+sudo dnf -y install python39
+sudo alternatives --set python /usr/bin/python3.9
+sudo alternatives --set python3 /usr/bin/python3.9
+sudo update-alternatives --install /usr/bin/pip3 pip3 /usr/bin/pip3.9 1
+sudo pip3 install ansible=="${ANSIBLE_VERSION}" netaddr
```
dtantsur, line 52: duplicated netaddr here

hardys:
> On the other hand, we should probably convince upstream to install ansible in a virtual env (similarly to how Bifrost does it).

Yeah I was thinking it'd be better to avoid the root pip install, but for now I went with it - if we could consume the exact same script from metal3-dev-env that would be ideal to avoid future PRs like this to resync things

01_install_requirements.sh
hardys: Oops, thanks, now fixed!

zaneb:
Could we maybe change both upstream and here to use `pip install --user` do you think?
/approve

PR #1306, opened 2021-11-01
Allow creating VMs with emulated TPM device

mkowalski:


mkowalski:
Together with https://github.com/metal3-io/metal3-dev-env/pull/848, I'm getting the following inside a VM

```
[root@master-0 ~]# dmesg | grep -i tpm
[    0.000000] efi:  SMBIOS=0x7e9d2000  TPMFinalLog=0x7ebf6000  ACPI=0x7eb7d000  ACPI 2.0=0x7eb7d014  MEMATTR=0x7d4a8018  MOKvar=0x7e975000  TPMEventLog=0x7d4a5018 
[    0.000000] ACPI: TPM2 0x000000007EB75000 00004C (v04 BOCHS  BXPCTPM2 00000001 BXPC 00000001)
[    2.233085] tpm_tis MSFT0101:00: 2.0 TPM (device-id 0x1, rev-id 1)
```

mkowalski:
/test e2e-metal-ipi-ovn-ipv6

Does not seem like related issue at all

```
Message: "secrets is forbidden: User \"e2e-test-templates-2cqw2-edituser\" cannot create resource \"secrets\" in API group \"\" in the namespace \"e2e-test-templates-2cqw2\"",
```

hardys:
lgtm, we'll need to update the [metal3-dev-env pin](https://github.com/openshift-metal3/dev-scripts/blob/master/01_install_requirements.sh#L22) when https://github.com/metal3-io/metal3-dev-env/pull/848 lands

hardys:
See comments in https://github.com/metal3-io/metal3-dev-env/pull/848 - if we add support for the environment variable directly in the metal3-dev-env role I don't think we need this PR at all (we'd still need to bump the pin though of course)

mkowalski:
> See comments in [metal3-io/metal3-dev-env#848](https://github.com/metal3-io/metal3-dev-env/pull/848) - if we add support for the environment variable directly in the metal3-dev-env role I don't think we need this PR at all (we'd still need to bump the pin though of course)

Right, I'll just keep an example of the config.sh so that the usage is documented

mkowalski:
/test e2e-metal-ipi-ovn-ipv6

The failure looks unrelated

hardys:
@mkowalski apologies but the metal3-dev-env pin got bumped via https://github.com/openshift-metal3/dev-scripts/pull/1312 so this now conflicts, could you remove the `01_install_requirements.sh` changes (since the bump from my PR will include your TPM changes)?

We can then override CI if needed since it'll be a non-functional change to the config example, thanks!

hardys:
/approve
/lgtm

Thanks!

PR #1308, opened 2021-11-17
Delete BMO ValidatingWebhookConfiguration on local runs

ardaguclu:
CBO creates BMO resources as well as ValidatingWebhookConfiguration
and service. However, on local runs, BMO ValidatingWebhook is disabled
by default. To allow requests go to BMO API, this PR deletes
ValidatingWebhookConfiguration.

### DIFF
--- a/metal3-dev/local-bmo.sh
+++ b/metal3-dev/local-bmo.sh
@@ -36,6 +36,10 @@ if oc get pod -o name -n openshift-machine-api | egrep -v 'metal3-(development|i
     oc wait --for=delete -n openshift-machine-api $metal3pods || true
 fi
 
+# Delete validatingwebhookconfiguration if exist. Because on local runs,
+# baremetal operator validatingwebhook API is disabled by default
+oc delete validatingwebhookconfiguration/baremetal-operator-validating-webhook-configuration --ignore-not-found=true
+
 # Save a copy of the full deployment as input
 oc get deployment -n openshift-machine-api -o yaml metal3 > $OUTDIR/bmo-deployment-full.yaml
 


metal3-dev/local-bmo.sh
```diff
@@ -36,6 +36,10 @@ if oc get pod -o name -n openshift-machine-api | egrep -v 'metal3-(development|i
     oc wait --for=delete -n openshift-machine-api $metal3pods || true
 fi
 
+# Delete validatingwebhookconfiguration if exist. Because on local runs,
+# baremetal operator validatingwebhook API is disabled by default
+oc delete validatingwebhookconfiguration/baremetal-operator-validating-webhook-configuration --ignore-not-found=true
```
andfasano, line 41: Could be possible, to restore the previous cluster state once done, to save the resource in the `$OUTDIR` (ie, to be reused in `mao-bmo.sh`)? Not sure if it could require some cleanup

ardaguclu: As far as I can see, this scale https://github.com/openshift-metal3/dev-scripts/blob/7df2c42e2bde60dde87a88de1c4b8fa1be14fba5/metal3-dev/mao-bmo.sh#L18 is managed by CBO. Thus, CBO will automatically create `ValidatingWebhookConfiguration` if not exist.

andfasano: Thanks, if it is automatically managed by CBO and doesn't require any other action for restoring the previous state then it's fine for me.

/lgtm

ardaguclu:
@sadasu could you please have a look at this?, thanks.

PR #1310, opened 2021-11-22
Enable EPEL repo on RHEL8

markmc:
A metal3-dev-env change ([Enable authentication in sushy-tools](https://github.com/metal3-io/metal3-dev-env/pull/798)) added a requirement for python3-passlib and python3-bcrypt.

Apparently these aren't available in RHEL and some folks on slack have been workinga around this by using EPEL. They are also available in OSP 16 repos.

### DIFF
--- a/01_install_requirements.sh
+++ b/01_install_requirements.sh
@@ -37,6 +37,9 @@ export DISTRO="${ID}${VERSION_ID%.*}"
 if [[ $DISTRO == "centos8" ]]; then
     sudo dnf -y install epel-release dnf --enablerepo=extras
 elif [[ $DISTRO == "rhel8" ]]; then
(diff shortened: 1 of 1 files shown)


andfasano:
Thanks @markmc for the patch

/lgtm

hardys:
/test e2e-metal-ipi
/approve

> Apparently these aren't available in RHEL and some folks on slack have been workinga around this by using EPEL. They are also available in OSP 16 repos.

In the past we avoided enabling EPEL for RHEL due to conflicts with the downstream ansible (and OSP IIRC) repos, but since we moved to installing ansible from pip, this should be fine I think, provided we're happy the EPEL versions are new enough

PR #1313, opened 2021-11-24
Refactor to enable local storage

hardys:
Refactor code from assisted_deployment.sh to enable_local_storage.sh
so that the local-storage operator can be enabled for deployed clusters
independent of the other assisted dependencies.
    
Note this currently only works for ga builds (not nightly/ci)

Also fix the default for `VM_EXTRADISKS_LIST` - this is wrong so just specifying `VM_EXTRADISKS=true` doesn't work with this script (the operator ends up looking for `vdb` but the disk actually shows up as `vda`


flaper87 (approved):
:spaghetti:

enable_local_storage.sh
```diff
@@ -0,0 +1,90 @@
+#!/usr/bin/env bash
+set -ex
```
elfosardo, line 2: would probably set also -u and -o pipefail

hardys:
> Is this change still only for assisted deployments?, because `deploy_local_storage` is called only in
> 
> https://github.com/openshift-metal3/dev-scripts/blob/41bafcec7d86e47d620b366310cf7f94b5a98e3e/assisted_deployment.sh#L303

The point of this PR is to make it possible to enable local storage on a dev-scripts deployed cluster independent of the assisted stuff - so someone can run `make` then run `enable_local_storage.sh` if it's needed (for example if you want to try hypershift on a deployed cluster, or anything else which requires persistent storage)

ardaguclu:
That makes sense, thanks @hardys 
/lgtm

flaper87 (approved):
/lgtm

andfasano:
Given that the patch seems ok, wouldn't be recommended to have an assisted related job in the [CI dev-scripts repo](https://github.com/openshift/release/blob/master/ci-operator/config/openshift-metal3/dev-scripts/openshift-metal3-dev-scripts-master.yaml) just to ensure that the change is not breaking anything?

enable_local_storage.sh
```diff
@@ -0,0 +1,114 @@
...
+
+if [[ "$OPENSHIFT_RELEASE_TYPE" == "ga" ]]; then
+  generate_subscription
+  echo "Creating local storage operator group and subscription..."
+  oc apply -f "${ASSETS_DIR}/01-local-storage-operator.yaml"
+else
+  oc project openshift-local-storage
+  LSO_PATH=${LOCAL_STORAGE_OPERATOR_PATH:-$GOPATH/src/github.com/openshift/local-storage-operator}
+  if [ ! -d $LSO_PATH ]; then
+      echo "Did not find $LSO_PATH" 1>&2
```
dtantsur, line 86: Follow-up idea: just clone or `go get` it?

enable_local_storage.sh
```diff
@@ -0,0 +1,114 @@
...
+  generate_local_volume
+  echo "Creating local volume and storage class..."
+  oc apply -f "${ASSETS_DIR}/02-local-volume.yaml"
+}
+
+
+if [ "${VM_EXTRADISKS}" != "false" ]; then
+  deploy_local_storage
+else
+  echo "Cannot deploy local storage unless VM_EXTRADISKS is enabled"
```
dtantsur, line 112: Looks like this logic is no longer needed now that it's a separate script?

hardys: No, you must have extra disks or the LSO can't create PV's from them, so we error out here if VM_EXTRADISKS isn't defined in the config

hardys:
> Given that the patch seems ok, wouldn't be recommended to have an assisted related job in the [CI dev-scripts repo](https://github.com/openshift/release/blob/master/ci-operator/config/openshift-metal3/dev-scripts/openshift-metal3-dev-scripts-master.yaml) just to ensure that the change is not breaking anything?

I checked with @flaper87 and he said the script I'm changing here isn't used in AI CI (which presumably explains why we don't have any gating coverage) - IIUC https://github.com/openshift/assisted-test-infra is used instead, but I admit I'm not at all familiar with exactly how the AI CI jobs are implemented

elfosardo:
/close
obsolete

PR #1315, opened 2021-11-25
Store extraworkers info in a secret

andfasano:
This patch stores the yaml generated for the extra workers in a secret, so that [e2e baremetal tests](https://github.com/openshift/origin/tree/master/test/extended/baremetal) could use it to dynamically add/delete extra workers when required

### DIFF
--- a/06_create_cluster.sh
+++ b/06_create_cluster.sh
@@ -58,6 +58,11 @@ if [[ -n "${APPLY_EXTRA_WORKERS}" ]]; then
     fi
 fi
 
+# Create a secret containing extraworkers info for the e2e tests
+if [[ ${NUM_EXTRA_WORKERS} -ne 0 && -d "${OCP_DIR}/extras" ]]; then
+    oc create secret generic extraworkers-secret --from-file="${OCP_DIR}/extras/" -n openshift-machine-api
+fi
+
 if [[ ! -z "${ENABLE_METALLB}" ]]; then
 
 	if [[ -z ${METALLB_IMAGE_BASE} ]]; then
--- a/ocp_install_env.sh
+++ b/ocp_install_env.sh
@@ -283,14 +283,17 @@ function generate_ocp_host_manifest() {
     mkdir -p "${outdir}"
     rm -f "${outdir}/extra_hosts.yaml"
 
+    mkdir -p "${outdir}/extras"
+    rm -f "${outdir}/extras/*"
+
+    worker_index=0
     jq --raw-output '.[] | .name + " " + .ports[0].address + " " + .driver_info.username + " " + .driver_info.password + " " + .driver_info.address' $host_input \
        | while read name mac username password address ; do
 
         encoded_username=$(echo -n "$username" | base64)
         encoded_password=$(echo -n "$password" | base64)
 
-    cat >> "${outdir}/${host_output}" << EOF
----
+        secret="---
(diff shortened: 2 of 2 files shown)


ocp_install_env.sh
```diff
@@ -312,8 +315,15 @@ spec:
...
+        echo "${secret}${bmh}" >> "${outdir}/${host_output}"
+
+        echo "${secret}" >> "${outdir}/extras/extraworker-${worker_index}-secret"
+        echo "${bmh}" >> "${outdir}/extras/extraworker-${worker_index}-bmh"
+        ((worker_index+=1))
     done
-}
+
+    # This secret will be used by the e2e test to create/destroy additional workers when required
+    oc create secret generic extraworkers-secret --from-file="${outdir}/extras/" --dry-run=client --output=yaml > "${outdir}/extraworkers-secret.yaml"
```
derekhiggins, line 328: This can't run before the cluster is up
```
+(ocp_install_env.sh:328): generate_ocp_host_manifest(): oc create secret generic extraworkers-secret --from-file=ocp/ostest/extras/ --dry-run=client --output=yaml                                               
W1125 12:35:05.692447 1833488 loader.go:221] Config not found: /home/derekh/dev-scripts/ocp/ostest/auth/kubeconfig                 
```

andfasano: Ok thanks, was under the impression that `--dry-run=client` was enough, will fix it

ocp_install_env.sh
```diff
@@ -312,8 +315,13 @@ spec:
   bootMACAddress: $mac
   bmc:
     address: $address
-    credentialsName: ${name}-bmc-secret
-EOF
+    credentialsName: ${name}-bmc-secret"
 
+        echo "${secret}${bmh}" >> "${outdir}/${host_output}"
+
+        # Extra files will be used later to generate a secret used by e2e tests
```
ardaguclu, line 322: Maybe we can do this only if `OPENSHIFT_CI` is set?. Because it looks like apart from CI, this will not be used. For local or any other runs, extra-hosts yaml file is sufficient.

06_create_cluster.sh
```diff
@@ -58,6 +58,11 @@ if [[ -n "${APPLY_EXTRA_WORKERS}" ]]; then
     fi
 fi
 
+# Create a secret containing extraworkers info for the e2e tests
```
ardaguclu, line 61: Same question about `OPENSHIFT_CI` below also is in here.

ardaguclu (commented):
That looks good to me.

06_create_cluster.sh
derekhiggins: If it isn't very resource intensive then I'd add it regardless,
if we do decide to only add it in CI then we should make sure that the error message from openshift-tests makes it clear whats missing (for anybody trying to reproduce tests manually)

ocp_install_env.sh
andfasano: In reality I guess it could be useful also for normal development: sometimes I allocate more extra workers, and deploy them one-by-one or in batch, so I ended up cutting&paste the global file

06_create_cluster.sh
andfasano: Yep, looks something that we could keep it anyway

andfasano:
Fixed a typo

PR #1316, opened 2021-11-30
Get CoreOS data from stream

zaneb:
The file `data/data/rhcos.json` [no longer exists](https://github.com/openshift/installer/pull/5252) (replaced by
`data/data/coreos/rhcos.json`). If the installer binary is able to output
the new (stream) file, use that. Only fall back to using the old file
for older releases.

### DIFF
--- a/ocp_install_env.sh
+++ b/ocp_install_env.sh
@@ -87,7 +87,7 @@ function build_installer() {
   cd $OPENSHIFT_INSTALL_PATH
   TAGS="libvirt baremetal" hack/build.sh
   popd
-  cp "$OPENSHIFT_INSTALL_PATH/data/data/rhcos.json" "$OCP_DIR"
+  "$OPENSHIFT_INSTALL_PATH/bin/openshift-install" coreos print-stream-json >"$OCP_DIR/rhcos.json" || cp "$OPENSHIFT_INSTALL_PATH/data/data/rhcos.json" "$OCP_DIR"
 }
 
 function baremetal_network_configuration() {


ocp_install_env.sh
```diff
@@ -87,7 +87,7 @@ function build_installer() {
   cd $OPENSHIFT_INSTALL_PATH
   TAGS="libvirt baremetal" hack/build.sh
   popd
-  cp "$OPENSHIFT_INSTALL_PATH/data/data/rhcos.json" "$OCP_DIR"
+  "$OPENSHIFT_INSTALL_PATH/bin/openshift-install" coreos print-stream-json >"$OCP_DIR/rhcos.json" || cp "$OPENSHIFT_INSTALL_PATH/data/data/rhcos.json" "$OCP_DIR"
```
hardys, line 90: We already create this file via print-stream-json [here](https://github.com/openshift-metal3/dev-scripts/blob/master/rhcos.sh#L2) so we could perhaps only copy if the "old" file location is found here?

I actually found this issue a few days ago but forgot to push the patch (apologies!) https://github.com/hardys/dev-scripts/commit/c4b30ffb5c70c54760ab4141bdd06a3646d2fed7 so if you agree that approach is preferable I can just create a PR, or we can go with this approach.

zaneb: I didn't bother to check in to how those two scripts interact. I assume _you_ know what you're doing, so your fix is preferable since it's simpler :) Happy to close this one.

hardys: I don't have a strong preference, but with my approach we avoid running the openshift-install command twice, which I guess may be slightly faster and less noisy in the logs - it also means we can remove the copy completely when 4.7 reaches EOL.

I pushed https://github.com/openshift-metal3/dev-scripts/pull/1317 - if you're happy with that we can close this one.

PR #1317, opened 2021-11-30
Conditionally copy rhcos.json

hardys:
This is only needed in rhcos.sh for old versions which lack the
openshift-install coreos-print-stream-json option, and the
file moved in https://github.com/openshift/installer/pull/5252
so we should only copy if the "old" location is detected

### DIFF
--- a/ocp_install_env.sh
+++ b/ocp_install_env.sh
@@ -87,7 +87,14 @@ function build_installer() {
   cd $OPENSHIFT_INSTALL_PATH
   TAGS="libvirt baremetal" hack/build.sh
   popd
-  cp "$OPENSHIFT_INSTALL_PATH/data/data/rhcos.json" "$OCP_DIR"
+  # This is only needed in rhcos.sh for old versions which lack the
+  # openshift-install coreos-print-stream-json option
+  # That landed in 4.8, and in 4.10 this file moved, so just
+  # skip copying it if it's not in the "old" location ref
+  # https://github.com/openshift/installer/pull/5252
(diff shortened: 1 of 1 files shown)


bfournie (approved):

ocp_install_env.sh
```diff
@@ -87,7 +87,14 @@ function build_installer() {
   cd $OPENSHIFT_INSTALL_PATH
   TAGS="libvirt baremetal" hack/build.sh
   popd
-  cp "$OPENSHIFT_INSTALL_PATH/data/data/rhcos.json" "$OCP_DIR"
+  # This is only needed in rhcos.sh for old versions which lack the
+  # openshift-install coreos-print-stream-json option
+  # That landed in 4.8, and in 4.10 this file moved, so just
+  # skip copying it if it's not in the "old" location ref
+  # https://github.com/openshift/installer/pull/5252
+  if [ -f "$OPENSHIFT_INSTALL_PATH/data/data/rhcos.json" ]; then
```
elfosardo, line 95: nit: "$OPENSHIFT_INSTALL_PATH/data/data/rhcos.json" should probably be yet another variable

elfosardo (approved):

ocp_install_env.sh
hardys: It wasn't a variable before, and the path is decided by the installer repo - I don't understand under what circumstances someone would want to override this path, can you elaborate?

hardys:
/test e2e-metal-ipi

Looks like an issue with the image registry

PR #1321, opened 2021-12-02
MetalLB E2E tests: skip L2 metrics test

sabinaaledort:
Need to skip L2 metrics test because the pod that's running the tests is not in the same subnet of the cluster nodes, so the arp request that's done in the test won't work.


msherif1234:
failures not related to changes in this PR I checked pod logs and it showing api/auth pods existing

PR #1327, opened 2021-12-08
MetalLB E2E tests: Add use-operator flag

sabinaaledort:
Adding the flag to support the last change in the operator, as we are deploying MetalLB here through the operator.

In the new operator configuration the operator takes control of the ConfigMap, meaning if an external thing (like the e2e tests) changes the ConfigMap the Operator will reconcile it to what it thinks it should be - making the e2e tests useless.


sabinaaledort:
/hold
should be merged after https://github.com/openshift/metallb/pull/19

PR #1332, opened 2021-12-23
Add support for mirroring OLM operators

cybertron:
In disconnected deployments (like ipv6) we need to mirror any OLM
operators that we want to install because the OLM images aren't
otherwise available. This patch adds a MIRROR_OLM configuration
variable that will create a pruned index containing only the
relevant operators in the local registry.

The mirroring process generates a couple of manifests needed to use
the mirrored index, and these files are added to the deployment
manifests directory so it will be usable immediately.

### DIFF
--- a/04_setup_ironic.sh
+++ b/04_setup_ironic.sh
@@ -136,11 +136,18 @@ if [ ! -z "${MIRROR_IMAGES}" ]; then
     IRONIC_LOCAL_IMAGE=${IRONIC_LOCAL_IMAGE:-"${LOCAL_REGISTRY_PREFIX}@${IRONIC_RELEASE_IMAGE}"}
 
     if [ -n "${MIRROR_OLM:-}" ]; then
-        echo "Mirroring OLM operator(s): ${MIRROR_OLM}"
+        echo "Installing OPM client"
         VERSION="$(openshift_version ${OCP_DIR})"
         OLM_DIR=$(mktemp --tmpdir -d "mirror-olm--XXXXXXXXXX")
         _tmpfiles="$_tmpfiles $OLM_DIR"
 
+        pushd $OLM_DIR
+        curl -O https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp/${VERSION}.0/opm-linux.tar.gz
+        tar xf opm-linux.tar.gz
+        sudo mv -f opm /usr/local/bin
+        popd
+
+        echo "Mirroring OLM operator(s): ${MIRROR_OLM}"
         REGISTRY_AUTH_FILE=${PULL_SECRET_FILE} opm index prune -f registry.redhat.io/redhat/redhat-operator-index:v${VERSION} -p ${MIRROR_OLM} -t ${LOCAL_REGISTRY_DNS_NAME}:${LOCAL_REGISTRY_PORT}/localimages/redhat-operator-index:v${VERSION}
 
         podman push --authfile ${PULL_SECRET_FILE} ${LOCAL_REGISTRY_DNS_NAME}:${LOCAL_REGISTRY_PORT}/localimages/redhat-operator-index:v${VERSION}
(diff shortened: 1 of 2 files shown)


cybertron:
Note that this requires the opm client installed. Since it's an optional feature, I don't know if we need to include it as part of the install or if we can just link to https://docs.openshift.com/container-platform/4.9/cli_reference/opm/cli-opm-install.html#cli-opm-install in the docs.

hardys:
> Note that this requires the opm client installed. Since it's an optional feature, I don't know if we need to include it as part of the install or if we can just link to https://docs.openshift.com/container-platform/4.9/cli_reference/opm/cli-opm-install.html#cli-opm-install in the docs.

Nice, this lgtm - but yeah perhaps it would be good to install the client when `MIRROR_OLM` is specified to avoid the manual step?

cybertron:
Sounds good. I had planned to do that, but ran out of time before the break and wanted to get the initial version pushed in case my server crashed while I was out. :-)

cybertron:
Okay, added opm client install and documented this in config_example since there are a couple of gotchas with using it.

04_setup_ironic.sh
```diff
@@ -134,6 +134,30 @@ if [ ! -z "${MIRROR_IMAGES}" ]; then
     IRONIC_RELEASE_IMAGE=$(image_for ironic | cut -d '@' -f2)
     LOCAL_REGISTRY_PREFIX="${LOCAL_REGISTRY_DNS_NAME}:${LOCAL_REGISTRY_PORT}/localimages/local-release-image"
     IRONIC_LOCAL_IMAGE=${IRONIC_LOCAL_IMAGE:-"${LOCAL_REGISTRY_PREFIX}@${IRONIC_RELEASE_IMAGE}"}
+
+    if [ ! -z "${MIRROR_OLM}" ]; then
```
hardys, line 138: The CI job is failing because this is undefined by default:

```
./04_setup_ironic.sh: line 138: MIRROR_OLM: unbound variable
```

So we need to either default it in `common.sh` or here e.g `${MIRROR_OLM:-}`

cybertron: Ah, oops. I didn't realize this script was set -u. I've fixed that and the double-negation ! -z to just -n.

ardaguclu:
As far as I understand, this should not run if MIRROR_ORM is enabled(disables all catalog resources);

https://github.com/openshift-metal3/dev-scripts/blob/4a3816bdd7e9b156c03c36c6d972c0b1cb76d340/06_create_cluster.sh#L41

But what I don't understand that user will choose which ones to be enabled. What will happen to others ones?, they will be disabled or they'll stay ImagePullErr state.

cybertron:
> As far as I understand, this should not run if MIRROR_ORM is enabled(disables all catalog resources);
> 
> https://github.com/openshift-metal3/dev-scripts/blob/4a3816bdd7e9b156c03c36c6d972c0b1cb76d340/06_create_cluster.sh#L41
> 
> But what I don't understand that user will choose which ones to be enabled. What will happen to others ones?, they will be disabled or they'll stay ImagePullErr state.

It's still fine to disable the default sources even if we're mirroring OLM content. The mirror index is created as a separate, non-default source so it's not affected by this, and the default sources still won't work.

ardaguclu:
@cybertron it is better to override serial-ovn-ipv6. Unfortunately, it is broken currently.

cybertron:
> @cybertron it is better to override serial-ovn-ipv6. Unfortunately, it is broken currently.

Works for me. Who has permission to do that in this repo?

ardaguclu:
I think you can https://github.com/openshift-metal3/dev-scripts/blob/d842f1bd3ad5d144b582fe36f17d0b5600ba3703/OWNERS#L6 :)

cybertron:

Ha, I guess I've never done this before so I didn't realize you just need to be an owner. :-)

cybertron:
/skip ci/prow/e2e-metal-ipi-serial-ovn-ipv6

It's not gonna pass this time either, tide.

PR #1336, opened 2022-01-12
Add ENABLE_BOOTSTRAP_STATIC_IP option

hardys:
This configures a static IP for the bootstrap external NIC

Requires https://github.com/openshift/installer/pull/5787

### DIFF
--- a/common.sh
+++ b/common.sh
@@ -375,3 +375,7 @@ fi
 
 # Defaults the variable to enable testing a custom machine-api-operator image
 export TEST_CUSTOM_MAO=${TEST_CUSTOM_MAO:-false}
+
+# Set to configure bootstrap VM baremetal network with static IP
+# (Currently this just expects a non-empty value, the IP is fixed to .9)
+export ENABLE_BOOTSTRAP_STATIC_IP=${ENABLE_BOOTSTRAP_STATIC_IP:-}
--- a/ocp_install_env.sh
+++ b/ocp_install_env.sh
@@ -98,6 +98,19 @@ cat <<EOF
     provisioningNetworkInterface: $CLUSTER_PRO_IF
 EOF
   fi
+
+  if [ -n "${ENABLE_BOOTSTRAP_STATIC_IP}" ]; then
+    if [[ "${IP_STACK}" = "v6" ]]; then
+      BOOTSTRAP_IP=$(nth_ip $EXTERNAL_SUBNET_V6 $((idx + 9)))
+    else
+      # Note we assume v4 for dual-stack v4v6 since it's the primary network
+      BOOTSTRAP_IP=$(nth_ip $EXTERNAL_SUBNET_V4 $((idx + 9)))
+    fi
+cat <<EOF
+    bootstrapExternalStaticIP: "${BOOTSTRAP_IP}"
+    bootstrapExternalStaticGateway: "${PROVISIONING_HOST_EXTERNAL_IP}"
+EOF
+  fi
 }
 
 function dnsvip() {
--- a/config_example.sh
+++ b/config_example.sh
@@ -342,6 +342,11 @@ set -x
 #export EXTERNAL_SUBNET_V4="11.0.0.0/24"
 #export EXTERNAL_SUBNET_V6="fd2e:6f44:5dd8:c956::/120"
 
+# ENABLE_BOOTSTRAP_STATIC_IP -
+# Configure a static IP for the bootstrap VM external NIC
+# (Currently this just expects a non-empty value, the IP is fixed to .9)
+# export ENABLE_BOOTSTRAP_STATIC_IP=true
+
 ################################################################################
 ## VM Settings
 ##


hardys:
Ok this is working for me now with ipv4, still needs ipv6 support before removing the WIP

utils.sh
```diff
@@ -161,6 +189,14 @@ function create_cluster() {
       jq -s '.[0] * .[1]' ${IGNITION_EXTRA} ${assets_dir}/worker.ign.orig | tee ${assets_dir}/worker.ign
     fi
 
+    if [ -n "${BOOTSTRAP_STATIC_IP}" ]; then
+      bootstrap_static_ip_ignition ${assets_dir}
+      mv ${assets_dir}/bootstrap.ign ${assets_dir}/bootstrap.ign.orig
+      jq '.storage.files += $input' ${assets_dir}/bootstrap.ign.orig --slurpfile input ${assets_dir}/bootstrap_network_config.ign | tee ${assets_dir}/bootstrap.ign
```
hardys, line 195: FIXME - if we're going to run this in CI we probably want to remove the `tee` and redirect to avoid logging the ignition file

ocp_install_env.sh
```diff
@@ -117,6 +117,12 @@ cat <<EOF
     provisioningBridge: ${PROVISIONING_NETWORK_NAME}
     provisioningNetworkCIDR: $PROVISIONING_NETWORK
     provisioningNetworkInterface: $CLUSTER_PRO_IF
+EOF
+  fi
+  # A predictable MAC is needed to configure a static IP
```
derekhiggins, line 122: I don't think this is needed, the interface is being specified by name, can re remove all the MAC generation?

hardys: Ah good point, we probably can - unless we want to gain coverage for the `externalMACAddress` field?

Provided this works without I'll remove it, and perhaps propose another variable that allows the predictable MAC (and we can decide later if/where to enable that in CI)

andfasano (approved):

common.sh
```diff
@@ -331,3 +331,7 @@ export TEST_CUSTOM_MAO=${TEST_CUSTOM_MAO:-false}
 
 # Set to true to enable RHCOS Live Image based installs
 export RHCOS_LIVE_IMAGES=${RHCOS_LIVE_IMAGES:-false}
+
+# Set to configure bootstrap VM baremetal network with static IP
+# (Currently this just expects a non-empty value, the IP is fixed to .9)
+export BOOTSTRAP_STATIC_IP=${BOOTSTRAP_STATIC_IP:-}
```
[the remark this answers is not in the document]
hardys: Done, thanks for the review!

derekhiggins:
@hardys once this is merged I've pushed up a PR to update imain's static-ip work
https://github.com/openshift-metal3/dev-scripts/pull/1342

utils.sh
```diff
@@ -61,6 +61,40 @@ function configure_chronyd() {
...
+[${BOOTSTRAP_PROTOCOL}]
+method=manual
+addresses=${BOOTSTRAP_IP}/${BOOTSTRAP_PREFIX}
+gateway=${PROVISIONING_HOST_EXTERNAL_IP}
+dns=${PROVISIONING_HOST_EXTERNAL_IP}"
+  # Note the mode is specified as decimal, not octal
+  # https://coreos.github.io/ignition/configuration-v3_2/
+  cat > ${ASSESTS_DIR}/bootstrap_network_config.ign << EOF
+      {
+        "path": "/etc/NetworkManager/system-connections/ens3.nmconnection",
```
reihl, line 89: should be ${interface-name}

utils.sh
```diff
@@ -61,6 +61,40 @@ function configure_chronyd() {
...
+  BOOTSTRAP_PROTOCOL="ipv6"
+else
+  # Note we assume v4 for dual-stack v4v6 since it's the primary network
+  BOOTSTRAP_IP=$(nth_ip $EXTERNAL_SUBNET_V4 $((idx + 9)))
+  BOOTSTRAP_PREFIX=$(ipcalc --prefix $EXTERNAL_SUBNET_V4 | cut -d= -f2)
+  BOOTSTRAP_PROTOCOL="ipv4"
+fi
+  BOOTSTRAP_CONFIG="[connection]
+type=ethernet
+interface-name=ens3
```
reihl, line 78: change to variable and add env variable for boot ex interface

hardys: This is the VM NIC (not the provisioning bridge on the provisioning host), so it never changes and can be hard-coded?

hardys:
/cc @honza 

Updated to align with https://github.com/openshift/installer/pull/5787

pamoedom:
Hi @hardys, could be possible to obtain some status of this PR? do we have any blockade?

FYI, we need it in order to cover METAL-175 because we don't have a fully static IPI BM env where to test it, we are using dev-scripts approach to achieve that, thanks.

Best Regards.

PR #1337, opened 2022-01-15
Stop passing clusterOSImage

zaneb:
In 4.10 we don't need to pass a clusterOSImage in the install config, so don't.

### DIFF
--- a/04_setup_ironic.sh
+++ b/04_setup_ironic.sh
@@ -162,29 +162,6 @@ if [ ! -f "${CACHED_MACHINE_OS_IMAGE}" ]; then
   pushd $(dirname ${CACHED_MACHINE_OS_IMAGE})
   sha256sum --strict --check ${CACHED_MACHINE_OS_IMAGE}.sha256sum || ( rm -f "${CACHED_MACHINE_OS_IMAGE}" ; exit 1 )
   popd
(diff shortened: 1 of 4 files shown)


flaper87 (commented):
This is exciting! :scream: 

/approve

zaneb:
Dualstack job is broken for [known reasons](https://bugzilla.redhat.com/show_bug.cgi?id=2040671) but should get [further now](https://bugzilla.redhat.com/show_bug.cgi?id=2034527). Retesting to verify that.
/test e2e-metal-ipi-ovn-dualstack

ardaguclu:
One thing is not clear for me. Dev-scripts is also used for <4.10 versions in CI jobs. Will that change have any effect on these versions?

hardys:
> One thing is not clear for me. Dev-scripts is also used for <4.10 versions in CI jobs. Will that change have any effect on these versions?

I think that's handled - see the `is_lower_version $OPENSHIFT_VERSION 4.10` conditional

zaneb:
There's no version where we pass URLs for the live CoreOS ISO/PXE.

ardaguclu:
Serial ipv6 job is failing due to a known issue and until it is fixed it might be overidden. 

/override e2e-metal-ipi-serial-ovn-ipv6

PR #1338, opened 2022-01-16
MetalLB: Use metallb-operator image from CI registry

sabinaaledort:
Instead of building the operator image we can use the latest image in the CI registry.

Doing this so when running the CI in metallb-operator repo the deployment will be done with the committed changes instead of latest main.

### DIFF
--- a/metallb/deploy_operator.sh
+++ b/metallb/deploy_operator.sh
@@ -5,12 +5,7 @@ source ${metallb_dir}/metallb_common.sh
 
 METALLB_OPERATOR_REPO=${METALLB_OPERATOR_REPO:-"https://github.com/metallb/metallb-operator.git"}
 METALLB_OPERATOR_COMMIT=${METALLB_OPERATOR_COMMIT:-"4d52b86"}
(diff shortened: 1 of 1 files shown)


metallb/deploy_operator.sh
```diff
@@ -5,12 +5,7 @@ source ${metallb_dir}/metallb_common.sh
 
 METALLB_OPERATOR_REPO=${METALLB_OPERATOR_REPO:-"https://github.com/metallb/metallb-operator.git"}
 METALLB_OPERATOR_COMMIT=${METALLB_OPERATOR_COMMIT:-"4d52b86"}
-LOCAL_REGISTRY_DNS_NAME=${LOCAL_REGISTRY_DNS_NAME:-"virthost.ostest.test.metalkube.org"}
```
fedepaol, line 8: these are removed because not necessary anymore?

sabinaaledort: yes

sabinaaledort: i going to remove them from the CI conf as well

PR #1339, opened 2022-01-25
MetalLB: Remove "\" from SKIP var in E2E tests script 

sabinaaledort:
Since metallb/metallb#1145 we don't need to explicitly set "\" in the SKIP.


sabinaaledort:
This is not relevant anymore since the MetalLB deployment scripts for OCP CI are now part of the openshift/metallb repo:
https://github.com/openshift/metallb/tree/main/openshift-ci

https://github.com/openshift-metal3/dev-scripts/pull/1390

PR #1341, opened 2022-01-25
[WIP] Allow installer to use a proxy 

andfasano:
This patch adds a new config var `INSTALLER_PROXY` so that the installer could use a local proxy (in our CI jobs, we already deploy a proxy in the [baremetalds-devscript-proxy](https://github.com/openshift/release/blob/master/ci-operator/step-registry/baremetalds/devscripts/proxy/baremetalds-devscripts-proxy-commands.sh) step). In addition, all outgoing traffic will be disabled

cc @hardys


hardys:
> (in our CI jobs, we already deploy a proxy in the baremetalds-devscript-proxy step

How do we expect developers to deploy the proxy for local testing?  I was expecting to still need something similar to #1192 so we can easily enable a proxy on the virthost, even if that's not required in CI?

andfasano:
> > (in our CI jobs, we already deploy a proxy in the baremetalds-devscript-proxy step
> 
> How do we expect developers to deploy the proxy for local testing? I was expecting to still need something similar to #1192 so we can easily enable a proxy on the virthost, even if that's not required in CI?

I think we could add another block to kick off the squid container, using something similar to CI [test](https://github.com/openshift/release/blob/master/ci-operator/step-registry/baremetalds/devscripts/proxy/baremetalds-devscripts-proxy-commands.sh). The immediate goal anyhow now was to support the CI.

andfasano:
> > > (in our CI jobs, we already deploy a proxy in the baremetalds-devscript-proxy step
> > 
> > 
> > How do we expect developers to deploy the proxy for local testing? I was expecting to still need something similar to #1192 so we can easily enable a proxy on the virthost, even if that's not required in CI?
> 
> I think we could add another block to kick off the squid container, using something similar to CI [test](https://github.com/openshift/release/blob/master/ci-operator/step-registry/baremetalds/devscripts/proxy/baremetalds-devscripts-proxy-commands.sh). The immediate goal anyhow now was to support the CI.

update: waiting for the job results, but very likely the already existing proxy defined by the CI cannot be used as it is, so another one should be required

andfasano:
Revamping https://github.com/openshift-metal3/dev-scripts/pull/1192

06_create_cluster.sh
```diff
@@ -12,6 +12,12 @@ source validation.sh
 
 early_deploy_validation
 
+if [[ ! -z "$INSTALLER_PROXY" && "$IP_STACK" = "v4" ]]; then
```
ardaguclu, line 15: My question is just curiosity and to learn;

Why we only support IPv4?

andfasano: Just for the sake of simplicity right now, but the long-term goal is to cover ipv6 as well

ocp_install_env.sh
```diff
@@ -276,6 +276,15 @@ sshKey: |
 fips: ${FIPS_MODE:-false}
 EOF
 
+  if [[ ! -z "$INSTALLER_PROXY" && "$IP_STACK" = "v4" ]]; then
```
derekhiggins, line 279: Should we output an error if INSTALLER_PROXY and ipv6 ? 
otherwise its silently ignored

andfasano:
Added ipv6 support

/hold

andfasano:
/hold cancel

@hardys @derekhiggins The patch looks ready, could you please review it? Thanks

elfosardo (approved):

06_create_cluster.sh
```diff
@@ -12,6 +12,18 @@ source validation.sh
 
 early_deploy_validation
 
+if [[ ! -z "$INSTALLER_PROXY" ]]; then
+  export HTTP_PROXY=http://${PROVISIONING_HOST_EXTERNAL_IP}:${INSTALLER_PROXY_PORT}
+  export HTTPS_PROXY=http://${PROVISIONING_HOST_EXTERNAL_IP}:${INSTALLER_PROXY_PORT}
```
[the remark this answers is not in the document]
andfasano: Not sure if it's really required (also in the original issue plain http was used)

.gitignore
```diff
@@ -15,3 +15,4 @@ assets/templates/99_master-chronyd-redhat.yaml
 assets/templates/99_worker-chronyd-redhat.yaml
 
 pull_secret.json
+squid.conf
```
hardys, line 18: It might be better to put this in `WORKING_DIR` somewhere?

02_configure_host.sh
```diff
@@ -54,6 +54,18 @@ if [[ ! -z "${MIRROR_IMAGES}" || $(env | grep "_LOCAL_IMAGE=")  || ! -z "${ENABL
     setup_local_registry
 fi
 
+# Configure a local proxy to be used for the installation
+if [[ ! -z "${INSTALLER_PROXY}" ]]; then
+  generate_proxy_conf > squid.conf
```
hardys, line 59: This redirect won't work unless the script is run from `$PWD` - as mentioned above perhaps writing to `WORKING_DIR` would be better, but if we want it in the dev-scripts dir this should probably use `SCRIPTDIR`

02_configure_host.sh
```diff
@@ -54,6 +54,18 @@ if [[ ! -z "${MIRROR_IMAGES}" || $(env | grep "_LOCAL_IMAGE=")  || ! -z "${ENABL
...
+
+  sudo podman run -d --rm \
+     --net host \
+     --volume $PWD/squid.conf:/etc/squid/squid.conf \
+     --name ds-squid \
+     --dns 127.0.0.1 \
+     quay.io/sameersbn/squid:latest
```
hardys, line 66: This image was last updated 3 years ago, is there a reason this was chosen or is there a more appropriate (maintained) upstream/downstream image we can use?

ocp_install_env.sh
```diff
@@ -276,6 +276,21 @@ sshKey: |
...
+    EXT_SUBNET=${EXTERNAL_SUBNET_V6}
+    if [[ "$IP_STACK" = "v4" ]]; then
+      EXT_SUBNET=${EXTERNAL_SUBNET_V4}
+    fi
+
+    cat >> "${outdir}/install-config.yaml" << EOF
+proxy:
+  httpProxy: http://${PROVISIONING_HOST_EXTERNAL_IP}:${INSTALLER_PROXY_PORT}
+  httpsProxy: http://${PROVISIONING_HOST_EXTERNAL_IP}:${INSTALLER_PROXY_PORT}
+  noProxy: ${PROVISIONING_NETWORK},${EXT_SUBNET}
```
hardys, line 290: If we calculate `NO_PROXY` in this file instead of `06` we could reuse it here?

utils.sh
```diff
@@ -584,6 +584,28 @@ function wait_for_crd() {
...
+  then
+    echo "acl all src $EXTERNAL_SUBNET_V4"
+  fi
+
+  if [[ ! -z "${EXTERNAL_SUBNET_V6}" ]]
+  then
+    echo "acl all src $EXTERNAL_SUBNET_V6"
+  fi
+
+  cat <<EOF
```
hardys, line 600: If you move the `EXT_SUBNET` calculation into this (or some other common) file, the conditionals could be replaced with e.g:

```
acl all src $PROVISIONING_NETWORK
acl all src $EXT_SUBNET
http_access allow all
http_port ${INSTALLER_PROXY_PORT}
debug_options ALL,2
dns_v4_first on
coredump_dir /var/spool/squid
EOF
```

Although having said that, do we need to omit the PROVISIONING_NETWORK when `PROVISIONING_NETWORK_PROFILE` is `Disabled` ?

hardys:
Overall lgtm, just a few minor comments/suggestions :)

ocp_install_env.sh
andfasano: Maybe I could also centralize also all of them, will try it out

andfasano: PS: Just noticed a problem (a spurios check for ipv4 that should have been removed), fixing it

02_configure_host.sh
hardys: AFAICS it's an old/unmaintained image e.g see https://github.com/sameersbn/docker-squid and compare the squid version with the latest stable one - we don't have to block this PR on it, but IMO we should look at building an updated version, or find an actively maintained alternative?

andfasano: Agree, as long as we keep in sync with the CI to avoid the burden of maintaining multiple version. Will look after for something newer.

network.sh
```diff
@@ -151,3 +166,19 @@ else
...
+  if [[ "$IP_STACK" = "v4" ]]; then
+    EXT_SUBNET=${EXTERNAL_SUBNET_V4}
+  fi
+
+  HTTP_PROXY=http://$(wrap_if_ipv6 ${PROVISIONING_HOST_EXTERNAL_IP}):${INSTALLER_PROXY_PORT}
+  HTTPS_PROXY=http://$(wrap_if_ipv6 ${PROVISIONING_HOST_EXTERNAL_IP}):${INSTALLER_PROXY_PORT}
+  NO_PROXY=${PROVISIONING_NETWORK},9999,${EXT_SUBNET}
+
+  if [[ "$PROVISIONING_NETWORK_PROFILE" != "Disabled" ]]; then
+    NO_PROXY=${EXT_SUBNET},9999
```
hardys, line 182: I think this conditional is inverted, e.g we should omit `PROVISIONING_NETWORK` when `PROVISIONING_NETWORK_PROFILE` *is* `Disabled`?

hardys:
lgtm, except one conditional which I think may be wrong due to copy/paste?

network.sh
andfasano: Oh, yes, good catch @hardys thanks!

network.sh
```diff
@@ -151,3 +166,19 @@ else
   export BOOTSTRAP_PROVISIONING_IP=${BOOTSTRAP_PROVISIONING_IP:-$(nth_ip $PROVISIONING_NETWORK 2)}
   export CLUSTER_PROVISIONING_IP=${CLUSTER_PROVISIONING_IP:-$(nth_ip $PROVISIONING_NETWORK 3)}
 fi
+
+# Proxy related configuration
+if  [[ ! -z "$INSTALLER_PROXY" ]]; then
+  export EXT_SUBNET=${EXTERNAL_SUBNET_V6}
+  if [[ "$IP_STACK" = "v4" ]]; then
```
elfosardo, line 173: nit: double = for consistency

02_configure_host.sh
```diff
@@ -54,6 +54,18 @@ if [[ ! -z "${MIRROR_IMAGES}" || $(env | grep "_LOCAL_IMAGE=")  || ! -z "${ENABL
     setup_local_registry
 fi
 
+# Configure a local proxy to be used for the installation
+if [[ ! -z "${INSTALLER_PROXY}" ]]; then
```
elfosardo, line 58: nit: -n instead of ! -z
guess it's too late as we use ! -z everywhere!

elfosardo (approved):

02_configure_host.sh
hardys: Yeah we should perhaps consider a future PR to align on one convention (and other issues like quoting and `[` vs `[[` - not a blocker for this though IMO since it's a cosmetic (and as you say pre-existing) issue

network.sh
andfasano: I used the same approach for the preovious ip_stack comparisons in the same file, maybe we could also address such kind of more global changes in some other prs as also @hardys suggested

andfasano:
`e2e-metal-ipi-serial-ovn-ipv6` failure does not seem related to the current patch, retrying once

/test e2e-metal-ipi-serial-ovn-ipv6

andfasano:
> > /test e2e-metal-ipi-serial-ovn-ipv6
> 
> Looking at sippy this seems to still be perma-failing, seems like we need the https://bugzilla.redhat.com/show_bug.cgi?id=2049117 pending timeout fixes? If you're happy that's the cause I'd suggest we override this job to make progress.

Yes, I've also prepared a PR (https://github.com/openshift/origin/pull/26833) to be used for fixing it (by @bfournie https://github.com/openshift/origin/pull/26810), we can override it

PR #1342, opened 2022-01-26
WIP: Static ip

derekhiggins:
Cluster eventually comes up and completes

Continuing the work started here 
https://github.com/openshift-metal3/dev-scripts/pull/1297

TODO:
- bootstrap consistently times out when using static IPs, cluster eventually comes up after the installer has died
- ~~need to make dhcp config configurable~~
- test ipv6

### DIFF
--- a/02_configure_host.sh
+++ b/02_configure_host.sh
@@ -79,6 +79,20 @@ if [[ $(uname -m) == "aarch64" ]]; then
 fi
 
 
+> $VM_SETUP_PATH/vm_setup_extrahosts.yml
+# If running with static IPs nodes need to get an IP with reverse DNS lookup
+if [ -n "${STATIC_IPS}" ]; then
+    for x in $(seq 0 $(( $NUM_MASTERS + $NUM_WORKERS + $NUM_EXTRA_WORKERS - 1 )) ) ; do
+        _HOSTNAME=master-$x
+        if [ $x -ge $NUM_MASTERS ] ; then
+            _HOSTNAME=worker-$(( $x - $NUM_MASTERS ))
+        fi
+        echo "- {ip: \"{{ baremetal_network_cidr | nthhost($(( $STATIC_IP_CIDR_OFFSET + $x )) ) }}\", hostnames: [\"$_HOSTNAME\"]}" >> $VM_SETUP_PATH/vm_setup_extrahosts.yml
+    done
+else
+    echo '[]' > $VM_SETUP_PATH/vm_setup_extrahosts.yml
+fi
+
 ansible-playbook \
     -e @vm_setup_vars.yml \
     -e "ironic_prefix=${CLUSTER_NAME}_" \
--- a/common.sh
+++ b/common.sh
@@ -338,3 +338,7 @@ fi
 
 # Defaults the variable to enable testing a custom machine-api-operator image
 export TEST_CUSTOM_MAO=${TEST_CUSTOM_MAO:-false}
+
+# Set to configure bootstrap VM baremetal network with static IP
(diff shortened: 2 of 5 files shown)


hardys:
> need to make dhcp config configurable

FYI I started looking at that via https://github.com/metal3-io/metal3-dev-env/pull/902 (now merged) - with that we can do something like this:

```
diff --git a/vm_setup_vars.yml b/vm_setup_vars.yml
index a345ce5..f3b60ed 100644
--- a/vm_setup_vars.yml
+++ b/vm_setup_vars.yml
@@ -61,13 +61,8 @@ external_network:
     netmask_v4: "{{ baremetal_network_cidr_v4|ipaddr('netmask') }}"
     address_v6: "{{ baremetal_network_cidr_v6|nthhost(1)|default('', true) }}"
     prefix_v6: "{{ baremetal_network_cidr_v6|ipaddr('prefix') }}"
-    dhcp_enabled: "{{ virsh_dhcp_enabled|bool }}"
-    dhcp_range_v4:
-      - "{{ baremetal_network_cidr_v4|nthhost(20) }}"
-      - "{{ baremetal_network_cidr_v4|nthhost(60) }}"
-    dhcp_range_v6:
-      - "{{ baremetal_network_cidr_v6|nthhost(20) }}"
-      - "{{ baremetal_network_cidr_v6|nthhost(60) }}"
+    dhcp_range_v4: "{{ [ baremetal_network_cidr_v4|nthhost(20), baremetal_network_cidr_v4|nthhost(60) ] if lookup('env', 'STATIC_IPS') == '' else [] }}"
+    dhcp_range_v6: "{{ [ baremetal_network_cidr_v6|nthhost(20), baremetal_network_cidr_v6|nthhost(60) ] if lookup('env', 'STATIC_IPS') == '' else [] }}"
     lease_expiry: "{{ lookup('env', 'DHCP_LEASE_EXPIRY') }}"
     nat_port_range:
       - 1024

```

cybertron:
Having a shortcut to use static ips is useful, but I'm more interested in an interface to provide arbitrary network configs for the nodes so we can test things like bridge and bond configs. This is fairly important so we can duplicate (as close as possible, anyway) qe and customer network configurations.

Since networkConfigs are all multi-line, I don't think we want to try to pass them as env vars. I think we need an env var that points to a file or files containing the relevant networkConfigs. My first thought is to have the env var point to a directory (like ASSETS_EXTRA_FOLDER) and inside that there would be specific files - master-0.yaml, master-1.yaml, etc. containing the networkConfig for the respective node, as well as a vm_setup_extrahosts.yml for defining necessary DNS records.

None of this eliminates the need for a static ip shortcut option, but I'm wondering if it makes more sense to implement the arbitrary interface and then do the static ip config using that instead of as a one-off like it is now.

cybertron:
I tried this locally and for some reason my workers are stuck on the agent image. I see a few errors in the logs, but it mostly looks like optional stuff and the agent run seems to complete so I'm not sure what's going on there. It's not specific to this patch because when I manually configured static IPs it did the exact same thing.

cybertron:
To wrap up my previous comment, I believe there was an actual bug with networkConfig causing the failure. That's fixed now.

I pushed https://github.com/openshift-metal3/dev-scripts/pull/1418 as an example of what I was thinking for a generic networkConfig interface. It still needs something to add DNS records for static IPs, but with a little vm_setup_vars hacking it worked well. There should be some way to get the dns_extrahosts from the same directory.

derekhiggins:
Closing as this was handled in https://github.com/openshift-metal3/dev-scripts/pull/1418

PR #1343, opened 2022-01-31
Get Centos packages from Vault

derekhiggins:
Centos8 is EOL


ardaguclu:
We know what is the problem about e2e-metal-ipi-serial-ovn-ipv6 and it is not related to this PR. 

/skip e2e-metal-ipi-serial-ovn-ipv6

PR #1346, opened 2022-02-01
Update README.md

pawanpinjarkar:
CentOS8 went EOL in December 2021. Update readme to add additional details to migrate from CentOS Linux 8 to CentOS Stream 8. Using CentOS Linux 8 results in `Error: Failed to download metadata for repo 'appstream': Cannot prepare internal mirrorlist: No URLs in mirrorlist`


celebdor:
This has become stale. Reopen against the current readme if necessary

PR #1347, opened 2022-02-03
Disable ansible clean up for CI instances

ardaguclu:
Dev-scripts setup fails in CI instances when the operating system is RHEL 8.

```
$ sudo dnf erase -y ansible
No match for argument: ansible
$ sudo subscription-manager repos --disable=ansible-2-for-rhel-8-x86_64-rpms
This system has no repositories available through subscriptions.
```

This PR disables clean up for CI jobs, since it already does not exist.

(note: there is no way to test this fix is the only portion left to use RHEL 8 in CI jobs, unlikely but there might be others)


01_install_requirements.sh
```diff
@@ -40,9 +40,13 @@ elif [[ $DISTRO == "rhel8" ]]; then
...
 
-    # The packaged 2.x ansible is too old for compatibility with metal3-dev-env
-    sudo dnf erase -y ansible
-    sudo subscription-manager repos --disable=ansible-2-for-rhel-8-x86_64-rpms
+    # RHEL 8 running on CI instances does not have ansible repo subscribed
+    # and installed. Thus, we can do ansible clean process for only local runs.
+    if [[ -z "$OPENSHIFT_CI" ]]; then
+      # The packaged 2.x ansible is too old for compatibility with metal3-dev-env
+      sudo dnf erase -y ansible
+      sudo subscription-manager repos --disable=ansible-2-for-rhel-8-x86_64-rpms
```
hardys, line 48: Perhaps a more general solution would be to only perform these actions if we find a packaged ansible installed?

IIRC I only added these to allow folks using the previous packaged ansible to upgrade to later dev-scripts (that uses a pip installed version), so we could even consider removing this since #1301 which introduced this landed several months ago, so everyone should have migrated to the pip ansible by now?

ardaguclu: You are right @hardys.

More general solution looks elegant. But this check still might be useful(in case ansible is enabled by default) for the first dev-script installers.

I changed by checking the ansible repo is enabled in subscription-manager, and if yes, deletion process works.

ardaguclu:
Current CI jobs are not testing anything in that patch. On local tests, changes are working.

elfosardo (approved):

derekhiggins:
lgtm but would you mind squishing the 2 patches into 1?

PR #1348, opened 2022-02-03
Add Alma and Rocky Linux as supported host OSes

cybertron:
These two distributions are essentially what CentOS was pre-Stream,
so they should generally be compatible as well. They are documented
as "best effort" support though since I don't think we'll be running
them in CI.

### DIFF
--- a/01_install_requirements.sh
+++ b/01_install_requirements.sh
@@ -51,6 +51,8 @@ case $DISTRO in
     sudo dnf install -y network-scripts
     if [[ $DISTRO == "centos8" ]]; then
       echo "CentOS is not supported anymore. Please switch to CentOS Stream / RHEL / Rocky Linux"
+    fi
+    if [[ $DISTRO == "centos8" || $DISTRO == "almalinux8" || $DISTRO == "rocky8" ]]; then
       sudo dnf -y install epel-release dnf --enablerepo=extras
     elif [[ $DISTRO == "rhel8" ]]; then
       # Enable EPEL for python3-passlib and python3-bcrypt required by metal3-dev-env
@@ -104,6 +106,7 @@ ANSIBLE_FORCE_COLOR=true ansible-playbook \
   -e "virthost=$HOSTNAME" \
   -e "go_version=1.18.3" \
   -e "GOARCH=$GOARCH" \
+  $ALMA_PYTHON_OVERRIDE \
   -i vm-setup/inventory.ini \
   -b -vvv vm-setup/install-package-playbook.yml
 popd
--- a/02_configure_host.sh
+++ b/02_configure_host.sh
@@ -119,6 +119,7 @@ ansible-playbook \
     -e "worker_hostname_format=$WORKER_HOSTNAME_FORMAT" \
     -e "libvirt_arch=$(uname -m)" \
     -e "enable_vnc_console=$VNC_CONSOLE" \
+    $ALMA_PYTHON_OVERRIDE \
     -i ${VM_SETUP_PATH}/inventory.ini \
     -b -vvv ${VM_SETUP_PATH}/setup-playbook.yml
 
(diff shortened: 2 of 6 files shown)


cybertron:
/hold

This got it past the validation stage, but something failed later on. Looks like I need to make further changes.

andfasano:
Note that the default packet setup step creates an instance with Centos8.
If you want something else, you'll need at least a temporary test where you could set the [`PACKET_OS` variable](https://github.com/openshift/release/blob/cb80be4641325733af23c5b7320e92e068f1184d/ci-operator/step-registry/baremetalds/packet/setup/baremetalds-packet-setup-ref.yaml#L16) as required

cybertron:
Okay, this one is working for me locally on Alma. Had to get a little hacky with the ansible python interpreter, but that should be a temporary thing until Ansible fixes it properly.

/hold cancel

01_install_requirements.sh
```diff
@@ -34,8 +34,10 @@ sudo dnf -y upgrade
 # shellcheck disable=SC1091
 source /etc/os-release
 export DISTRO="${ID}${VERSION_ID%.*}"
-if [[ $DISTRO == "centos8" ]]; then
-    echo "CentOS is not supported anymore. Please switch to CentOS Stream / RHEL / Rocky Linux"
+if [[ $DISTRO == "centos8" || $DISTRO == "almalinux8" || $DISTRO == "rocky8" ]]; then
+    if [[ $DISTRO == "centos8" ]]; then
```
elfosardo, line 38: let's move this before L37 ?

elfosardo (changes requested):

01_install_requirements.sh
cybertron: Done

cybertron:
/retest

Some of these jobs are no longer required, and I think all of the required ones have passed.

cybertron:
/retest-required

Looks like issues getting a test environment.

danwinship:
any reason this hasn't merged? CentOS 8 on Equinix no longer seems to work for me:

```
[root@dwinship ~]# dnf -y install git jq make
CentOS Linux 8 - AppStream                                            1.5 kB/s |  38  B     00:00    
Error: Failed to download metadata for repo 'appstream': Cannot prepare internal mirrorlist: No URLs in mirrorlist
```

danwinship:
hm... nor does Rocky even with this PR:

```
The full traceback is:
Traceback (most recent call last):
  File "/tmp/ansible_ansible.posix.selinux_payload_btu2jecb/ansible_ansible.posix.selinux_payload.zip/ansible_collections/ansible/posix/plugins/modules/selinux.py", line 100, in <module>
ModuleNotFoundError: No module named 'selinux'
fatal: [localhost]: FAILED! => {
    "changed": false,
    "invocation": {
        "module_args": {
            "configfile": "/etc/selinux/config",
            "policy": "targeted",
            "state": "permissive",
            "update_kernel_param": false
        }
    },
    "msg": "Failed to import the required Python library (libselinux-python) on dwinship's Python /bin/python3.9. Please read the module documentation and install it in the appropriate location. If the required library is installed, but Ansible is using the wrong Python interpreter, please consult the documentation on ansible_python_interpreter"
}
```

`python3-libselinux` is installed though...

cybertron:
Ah, that sounds like the same problem I ran into with Alma. Maybe the Python interpreter workaround is needed for Rocky too. I thought someone had already fixed that for Rocky but maybe not. I need to rebase this anyway, so I'll add Rocky to the conditional so you can try it.

danwinship:
> I'll add Rocky to the conditional so you can try it.

I retried using Alma and it worked, and I managed to test what I was testing. I do need to go back and test an updated patch at some point but I'm not sure when I'm going to get to that...

danwinship:
/lgtm
works with Rocky now too

cybertron:
/retest-required

Infra failure

cybertron:
Rebased to account for some recent changes in OS setup.

PR #1349, opened 2022-02-08
Fix epel installation for centos and rhel

elfosardo:
Correct way to install/enable epel repos can be found at [1]

[1] https://docs.fedoraproject.org/en-US/epel/#_el8


elfosardo:
/hold
this can't work when using vault as the base repositories are not the same

PR #1350, opened 2022-02-08
Add ARM support

bn222:


hardys:
/ok-to-test

Thanks for starting this! :)

02_configure_host.sh
```diff
@@ -81,6 +81,8 @@ ansible-playbook \
     -e "virtualbmc_base_port=$VBMC_BASE_PORT" \
     -e "master_hostname_format=$MASTER_HOSTNAME_FORMAT" \
     -e "worker_hostname_format=$WORKER_HOSTNAME_FORMAT" \
+    -e "libvirt_arch=$(uname -m)" \
+    -e "enable_vnc_console=false" \
```
hardys, line 85: I think this will break tunnelling via SSH to the VNC console, which is very useful for debugging boot-time issues (I regularly use virt-manager on my laptop connected via `qemu+ssh` to a remote hypervisor, e.g to get info about ramdisk boot issues, or even to debug networking if you enable a user login)

I'd suggest we either remove this, or if you really want to disable vnc, add a new shell variable folks can set in their `config_$USER.sh` ?

ocp_install_env.sh
```diff
@@ -117,6 +125,7 @@ cat <<EOF
     provisioningBridge: ${PROVISIONING_NETWORK_NAME}
     provisioningNetworkCIDR: $PROVISIONING_NETWORK
     provisioningNetworkInterface: $CLUSTER_PRO_IF
+    bootstrapProvisioningInterfaceOverride: "${BOOTSTRAP_PROVISIONING_INTERFACE_OVERRIDE}"
```
hardys, line 128: Looks like this depends on https://github.com/openshift/installer/pull/5554 - I'm wondering if there's a way to detect the required nic in the installer, vs expecting the user to know what the nic inside the (installer generated) bootstrap VM is?

01_install_requirements.sh
```diff
@@ -49,7 +49,7 @@ elif [[ $DISTRO == "rhel8" ]]; then
```
hardys, line 49: Since https://github.com/metal3-io/metal3-dev-env/pull/912 merged, we'll need to update the pin on Line 22 above (we fix dev-scripts to a specific metal3-dev-env commit to avoid the risk of CI breakage in the event some upstream change breaks us)

rhcos.sh
```diff
@@ -1,19 +1,19 @@
 if $OPENSHIFT_INSTALLER coreos print-stream-json >/dev/null 2>&1; then
     $OPENSHIFT_INSTALLER coreos print-stream-json > $OCP_DIR/rhcos.json
-    export MACHINE_OS_INSTALLER_IMAGE_URL=$(jq -r '.architectures.x86_64.artifacts.openstack.formats["qcow2.gz"].disk.location' $OCP_DIR/rhcos.json)
-    export MACHINE_OS_INSTALLER_IMAGE_SHA256=$(jq -r '.architectures.x86_64.artifacts.openstack.formats["qcow2.gz"].disk["sha256"]' $OCP_DIR/rhcos.json)
+    export MACHINE_OS_INSTALLER_IMAGE_URL=$(jq -r .architectures.$(uname -m).artifacts.qemu.formats[\"qcow2.gz\"].disk.location $OCP_DIR/rhcos.json)
+    export MACHINE_OS_INSTALLER_IMAGE_SHA256=$(jq -r .architectures.$(uname -m).artifacts.qemu.formats[\"qcow2.gz\"].disk[\"sha256\"] $OCP_DIR/rhcos.json)
```
hardys, line 4: There's a copy/paste mistake here - s/qemu/openstack

02_configure_host.sh
elfosardo: I'm strongly against disabling vnc_console by default, we should have a way to switch, what @hardys suggest is viable or can be triggered by architecture if this needs to be false for arm64

04_setup_ironic.sh
```diff
@@ -147,7 +147,7 @@ if  sudo podman pod exists ironic-pod ; then
 fi
 
 # Create pod
-sudo podman pod create -n ironic-pod
+sudo podman pod create -n ironic-pod --infra-image k8s.gcr.io/pause:3.2
```
elfosardo, line 150: what's the reason to not use the default pause container?

elfosardo (changes requested):

rhcos.sh
bn222: Thanks for noticing this.

02_configure_host.sh
bn222: I couldn't get this working on arm64 without disabling vnc. I will make it a switch that is set to false only on ARM.

04_setup_ironic.sh
bn222: This was left over from testing

ocp_install_env.sh
bn222: @jeffdyoung is working on the installer to make the detection of the interface automatic. I've removed this option for now from this PR.

elfosardo (changes requested):
sorry, just spot a thing

common.sh
```diff
@@ -265,7 +265,7 @@ export EXTRA_WORKER_DISK=${EXTRA_WORKER_DISK:-${WORKER_DISK}}
 export EXTRA_WORKER_VCPU=${EXTRA_WORKER_VCPU:-${WORKER_VCPU}}
 
 # Ironic vars (Image can be use <NAME>_LOCAL_IMAGE to override)
-export IRONIC_IMAGE="quay.io/metal3-io/ironic:master"
+export IRONIC_IMAGE=${IRONIC_IMAGE:-"quay.io/metal3-io/ironic:master"}
```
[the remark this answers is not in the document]
bn222: there are many occurrences of master branch. Do you want me to change all?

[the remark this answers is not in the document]
elfosardo: it's still "master" should be changed to
export IRONIC_IMAGE=${IRONIC_IMAGE:-"quay.io/metal3-io/ironic:main"}
same for L178

elfosardo (changes requested):

01_install_requirements.sh
```diff
@@ -60,6 +60,12 @@ sudo pip3 install ansible=="${ANSIBLE_VERSION}"
 # and lxml for the pyxpath script
 sudo pip3 install netaddr lxml
 
+
+GOARCH=$(uname -m)
+if [[ $GOARCH == "aarch64" ]]; then
+    GOARCH="arm64"
+fi
+
```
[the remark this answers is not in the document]
bn222: Thanks!

hardys:
/override ci/prow/e2e-metal-ipi-proxy-ipv6
/override ci/prow/e2e-metal-ipi-proxy-ipv4

These jobs are failing, investigation/fixes ongoing but unrelated to this PR https://github.com/openshift/origin/pull/27093 is one fix but I'm not yet sure what the root-cause is on the ipv6 job cc @elfosardo @andfasano

PR #1354, opened 2022-02-17
Revert "Get Centos packages from Vault"

osherdp:
Reverts openshift-metal3/dev-scripts#1343

Since moving to Rocky Linux, this code changes package repositories back to using vault mirror even in Rocky Linux.
Regular repositories are good enough for us, so this change no longer needed.

/cc @derekhiggins 
/hold
(waiting to manually investigate CI logs)

### DIFF
--- a/01_install_requirements.sh
+++ b/01_install_requirements.sh
@@ -35,13 +35,7 @@ sudo dnf -y upgrade
 source /etc/os-release
 export DISTRO="${ID}${VERSION_ID%.*}"
 if [[ $DISTRO == "centos8" ]]; then
-    if [ "$NAME" != "CentOS Stream" -a ! -e /etc/yum.repos.d/centos-vault.repo ] ; then
-        # Centos8 is EOL
-        mv /etc/yum.repos.d /etc/yum.repos.d_$(date +%s)
-        mkdir /etc/yum.repos.d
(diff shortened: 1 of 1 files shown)


01_install_requirements.sh
```diff
@@ -35,13 +35,6 @@ sudo dnf -y upgrade
 source /etc/os-release
 export DISTRO="${ID}${VERSION_ID%.*}"
 if [[ $DISTRO == "centos8" ]]; then
-    if [ "$NAME" != "CentOS Stream" -a ! -e /etc/yum.repos.d/centos-vault.repo ] ; then
-        # Centos8 is EOL
-        mv /etc/yum.repos.d /etc/yum.repos.d_$(date +%s)
```
andfasano, line 40: What about adding a deprecation message in such case, ie "CentOS is not supported anymore"?

osherdp: sounds good. added

osherdp:
Looks good. I do see the deprecation notice here (because it thinks it has a centos environment) but at least it doesn't manipulate the repositories configurations
/unhold

andfasano:
Thanks @osherdp for catching it!

/lgtm

osherdp:
seems to not pass lately at all, so I guess it's broken

PR #1358, opened 2022-03-03
Add registry to the no_proxy when is configured

andfasano:
Especially when testing a disconnected environment (ie ipv6) where the local registry is deployed, it is required to add it to the no_proxy list if a proxy based installation is performed


andfasano:
Required by https://github.com/openshift/release/pull/25912

PR #1359, opened 2022-03-08
Add support for "main" and "master" branches

mkowalski:
This PR fixes function responsible for synchronizing repositories so
that both "main" and "master" branches are supported. Current behaviour
assumes that there always exists a "master" branch what is no longer the
case given repositories follow a patern of using "main" instead.

### DIFF
--- a/utils.sh
+++ b/utils.sh
@@ -267,9 +267,9 @@ function sync_repo_and_patch {
     pushd $DEST
 
     git am --abort || true
-    git checkout master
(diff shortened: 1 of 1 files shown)


javipolo (approved):

andfasano:
Failures on ipv6 jobs appear not to be related to the current patch

/override e2e-metal-ipi-serial-ovn-ipv6
/override e2e-metal-ipi-ovn-ipv6

PR #1360, opened 2022-03-08
Migrate metal3-dev-env branch to main

andfasano:
Default metal3-dev-env branch switched to main https://github.com/metal3-io/metal3-dev-env/pull/897


andfasano:
Addressed by https://github.com/openshift-metal3/dev-scripts/pull/1359

PR #1363, opened 2022-03-10
AGENT-20: Add fleeting Makefile target

lranjbar:
Very basic start to adding this into dev-scripts. The fleeting.iso is built using [openshift-agent-team/fleeting](https://github.com/openshift-agent-team/fleeting).

### DIFF
--- a/02_configure_host.sh
+++ b/02_configure_host.sh
@@ -83,6 +83,7 @@ ansible-playbook \
     -e "num_masters=$NUM_MASTERS" \
     -e "num_workers=$NUM_WORKERS" \
     -e "num_extraworkers=$NUM_EXTRA_WORKERS" \
+    -e "num_fleetings=$NUM_FLEETINGS" \
     -e "libvirt_firmware=uefi" \
     -e "virthost=$HOSTNAME" \
     -e "vm_platform=$NODES_PLATFORM" \
--- a/Makefile
+++ b/Makefile
@@ -68,3 +68,11 @@ assisted_deployment:
 
 assisted_deployment_requirements:
 	./assisted_deployment.sh install_prerequisites_assisted_service
+
+fleeting: fleeting_configure requirements configure fleeting_deploy_iso bell
+
+fleeting_configure:
+	./fleeting_configure.sh
+
+fleeting_deploy_iso:
+	./fleeting_deploy_iso.sh
--- a/fleeting_configure.sh
+++ b/fleeting_configure.sh
@@ -1,0 +1,18 @@
+#!/bin/bash
+
+# FIXME(lranjbar): Replace with Ansible role to define configuration for Multi/Compact/SNO
+source config_example_compact.sh
+
+# USAGE: Update this variable to where your fleeting directory is here or on the command line
+#export FLEETING_REPO_DIR=${$FLEETING_REPO_DIR:-$HOME/github/openshift-agent-team/fleeting/}
+export FLEETING_REPO_DIR=${$FLEETING_REPO_DIR}
+
+# Sets the default locations of the ISO built by fleeting
+export FLEETING_ISO_NAME=${FLEETING_ISO_NAME:-fleeting.iso}
+export FLEETING_ISO_LOCATION=${FLEETING_ISO_LOCATION:-$FLEETING_REPO_DIR/$FLEETING_ISO_NAME}
+
+# Sets the configuration for the fleeting VM
+export NUM_FLEETINGS=1
+export FLEETING_MEMORY=8192
+export FLEETING_DISK=20
+export FLEETING_VCPU=4
--- a/fleeting_deploy_iso.sh
+++ b/fleeting_deploy_iso.sh
@@ -1,0 +1,6 @@
+#!/bin/bash
+
+# FIXME(lranjbar): Look into a cleaner way to do this
+# Attach the fleeting.iso to the VM and start it
+virsh attach-disk ${CLUSTER_NAME}_fleeting_0 ${FLEETING_ISO_LOCATION} hdc --type cdrom --mode readonly
+virsh start ${CLUSTER_NAME}_fleeting_0 --force-boot
--- a/vm_setup_vars.yml
+++ b/vm_setup_vars.yml
@@ -10,6 +10,7 @@ ironic_prefix: "{{ ironic_prefix }}"
 num_masters: 3
 num_workers: 1
 num_extraworkers: 0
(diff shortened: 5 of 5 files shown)


dtantsur:
/ok-to-test

Please ask someone to add you to the org.

02_configure_host.sh
```diff
@@ -83,6 +83,7 @@ ansible-playbook \
     -e "num_masters=$NUM_MASTERS" \
     -e "num_workers=$NUM_WORKERS" \
     -e "num_extraworkers=$NUM_EXTRA_WORKERS" \
+    -e "num_fleetings=$NUM_FLEETINGS" \
```
hardys, line 86: Do we need a new node type for this, or should the install ISO be booted in-place on one of the masters (to replicate what customer cluster installs would do)?

vm_setup_vars.yml
```diff
@@ -30,7 +31,12 @@ flavors:
     disk: "{{ lookup('env', 'EXTRA_WORKER_DISK') }}"
     vcpu: "{{ lookup('env', 'EXTRA_WORKER_VCPU') }}"
     extradisks: "{{ lookup('env', 'VM_EXTRADISKS') | default(false) }}"
-
+  
+  fleeting:
+    memory: "{{ lookup('env', 'FLEETING_MEMORY') }}"
+    disk: "{{ lookup('env', 'FLEETING_DISK') }}"
+    vcpu: "{{ lookup('env', 'FLEETING_VCPU') }}"
```
hardys, line 38: Assuming we can't just use one of the masters, you could default these to e.g `MASTER_MEMORY` etc

fleeting_deploy_iso.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+# FIXME(lranjbar): Look into a cleaner way to do this
+# Attach the fleeting.iso to the VM and start it
+virsh attach-disk ${CLUSTER_NAME}_fleeting_0 ${FLEETING_ISO_LOCATION} hdc --type cdrom --mode readonly
+virsh start ${CLUSTER_NAME}_fleeting_0 --force-boot
```
hardys, line 6: I had to do [something similar](https://github.com/hardys/dev-scripts/blob/hypershift_wip/hypershift/03_hypershift_nodes.sh#L51..L62) for hypershift testing recently and ran into the problem that after install the VM rebooted from cdrom, not disk - does the AI workflow ensure the boot order here?

I'm not super happy with that WIP approach so happy to collaborate on a better solution which works for both cases :)

Makefile
```diff
@@ -68,3 +68,11 @@ assisted_deployment:
 
 assisted_deployment_requirements:
 	./assisted_deployment.sh install_prerequisites_assisted_service
+
+fleeting: fleeting_configure requirements configure fleeting_deploy_iso bell
```
hardys, line 72: I don't have a strong opinion on the name, but would it be more intuitive to call this `agent_deployment` or something?

Also the order looks inconsistent with our existing sequence, should probably be `requirements configure fleeting_configure fleeting_deploy_iso bell` - that way you can always depend on the prerequisites being installed (like ansible for example)

Makefile
```diff
@@ -68,3 +68,11 @@ assisted_deployment:
...
+
+fleeting_configure:
+	./fleeting_configure.sh
+
+fleeting_deploy_iso:
+	./fleeting_deploy_iso.sh
```
hardys, line 78: It might be a good idea to create a subdirectory, e.g similar to how I've been doing hypershift testing:

https://github.com/hardys/dev-scripts/tree/hypershift_wip/hypershift

Then we can also have separate cleanup scripts, README, any non-shared ansible roles etc

hardys:
> Please ask someone to add you to the org.

@lranjbar FYI I sent you an invite to join the org which will avoid the ok-to-test step in future :)

fleeting_configure.sh
```diff
@@ -0,0 +1,18 @@
...
+
+# Sets the default locations of the ISO built by fleeting
+export FLEETING_ISO_NAME=${FLEETING_ISO_NAME:-fleeting.iso}
+export FLEETING_ISO_LOCATION=${FLEETING_ISO_LOCATION:-$FLEETING_REPO_DIR/$FLEETING_ISO_NAME}
+
+# Sets the configuration for the fleeting VM
+export NUM_FLEETINGS=1
+export FLEETING_MEMORY=8192
+export FLEETING_DISK=20
+export FLEETING_VCPU=4
```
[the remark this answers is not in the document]
lranjbar: Yeah, you would need to do this prior to build the ISO. When I was testing this I had already cloned the fleeting repo and had built the ISO.  So I just needed to configure the folder location of that repository and ISO. I'd like to keep the instructions for building that ISO in fleeting like it is today. 🤔

02_configure_host.sh
lranjbar: I don't have a strong opinion on it, it just seemed like the easiest way to integrate at the time.

fleeting_deploy_iso.sh
lranjbar: AI makes changes to the boot order during install. However using a CDROM is kind of a known issue with it due to the how this standard is defined. Now that I think about it using a separate fleeting VM is a workaround by using a VM that is separate from the installation VMs for the cluster. You are right that it isn't following the customers flow exactly so I probably need to think on this a bit more.

I was thinking about going the route of ensuring we have the right libvirt domain XML for the VMs down the road. There is an [Ansible collection for libvirt](https://galaxy.ansible.com/community/libvirt) we can use directly for a lot of these operations but it requires the definitions to be in those libvirt domain XML. So I haven't started refactoring really but I'm thinking if we get the right domain XML we might be able to make this more smooth to use. Alot of what we get from the metal-dev-env is the domain XML so this is where I'm thinking I guess. 🤔 I apologize that this is a rambling thought I'm still working through the details.

Makefile
lranjbar: I originally did call it agent. Since we also have assisted-installer-agent I thought that might be confusing so I went with fleeting instead. I don't have a strong opinion either. We are still trying to figure out what to call ourselves. 🤣

lranjbar:
> > Please ask someone to add you to the org.
> 
> @lranjbar FYI I sent you an invite to join the org which will avoid the ok-to-test step in future :)

I caught it before it expired! 🎉

lranjbar:
Closing since this work is being continued in #1372

PR #1366, opened 2022-03-16
El8 compat 01 installation

celebdor:
el8 01-script compatibility improvements

There's some el8 distros out there that were failing the 01 script due
to:
* Reporting differently their ID in os-release
* Not setting XDG_*
* Missing py39 dependencies

This commit should sort out those things in a more generic way.

It also adds support for having the config out of tree, in the
freedesktop expected place for configurations. Note that since one can
use the config to set the pull secret location, I did not touch its
default.

### DIFF
--- a/01_install_requirements.sh
+++ b/01_install_requirements.sh
@@ -69,6 +69,12 @@ ansible-galaxy install -r vm-setup/requirements.yml
 # temporarily capping it until the issue is fixed
 ansible-galaxy collection install 'ansible.netcommon:<2.6.0'
 ansible-galaxy collection install --upgrade ansible.posix community.general
+
+if ! python -c 'import selinux'; then
+   dnf copr enable -y apuimedo/libselinux-py39
+   sudo dnf install -y python39-libselinux
+fi
+
 ANSIBLE_FORCE_COLOR=true ansible-playbook \
   -e "working_dir=$WORKING_DIR" \
   -e "virthost=$HOSTNAME" \
--- a/common.sh
+++ b/common.sh
@@ -1,20 +1,27 @@
 #!/bin/bash
 
-export PATH="/usr/local/go/bin:$HOME/.local/bin:$PATH"
+if ! [[ "PATH" =~ "$(systemd-path user-binaries)" ]]; then
+    PATH="$(systemd-path user-binaries):$PATH"
+fi
+if ! [[ "PATH" =~ "/usr/local/go/bin" ]]; then
+    PATH="/usr/local/go/bin:$PATH"
+fi
+export PATH
 
 # Set a PS4 value which logs the script name and line #.
 export PS4='+(${BASH_SOURCE}:${LINENO}): ${FUNCNAME[0]:+${FUNCNAME[0]}(): }'
 
-eval "$(go env)"
-
-export PATH="${GOPATH}/bin:$PATH"
-
-# Ensure if a go program crashes we get a coredump
-#
-# To get the dump, use coredumpctl:
-#   coredumpctl -o oc.coredump dump /usr/local/bin/oc
-#
-export GOTRACEBACK=crash
+if command -v go; then
+    eval "$(go env)"
+    export PATH="${GOPATH}/bin:$PATH"
+    # Ensure if a go program crashes we get a coredump
+    #
+    # To get the dump, use coredumpctl:
+    #   coredumpctl -o oc.coredump dump /usr/local/bin/oc
+    #
+    export GOTRACEBACK=crash
+    export OPENSHIFT_INSTALL_PATH="$GOPATH/src/github.com/openshift/installer"
+fi
 
 # Do not use pigz due to race condition in vendored docker code
 # that oc uses.
@@ -38,12 +45,17 @@ is_lower_version () {
   fi
 }
 
+XDG_CONFIG_HOME=${XDG_CONFIG_HOME:-$(systemd-path user-configuration)}
+
 # Get variables from the config file
 if [ -z "${CONFIG:-}" ]; then
     # See if there's a config_$USER.sh in the SCRIPTDIR
     if [ -f ${SCRIPTDIR}/config_${USER}.sh ]; then
         echo "Using CONFIG ${SCRIPTDIR}/config_${USER}.sh" 1>&2
         CONFIG="${SCRIPTDIR}/config_${USER}.sh"
+    elif [[ -f "${XDG_CONFIG_HOME}/dev-scripts/config" ]]; then
+        echo "Using CONFIG ${XDG_CONFIG_HOME}/dev-scripts/config" 1>&2
+        CONFIG="${XDG_CONFIG_HOME}/dev-scripts/config"
     else
         error "Please run with a configuration environment set."
         error "eg CONFIG=config_example.sh ./01_all_in_one.sh"
--- a/validation.sh
+++ b/validation.sh
@@ -4,7 +4,7 @@ set -o pipefail
 
 # Perform some validation steps that we always want done.
 function early_either_validation() {
-    if [ "$USER" != "root" -a "${XDG_RUNTIME_DIR:-}" == "/run/user/0" ] ; then
+    if [ "$USER" != "root" -a "$(systemd-path user-runtime)" == "/run/user/0" ] ; then
         error "Please use a non-root user, WITH a login shell (e.g. su - USER)"
         exit 1
     fi
@@ -16,7 +16,7 @@ function early_either_validation() {
     fi
 
     # Check OS
-    if [[ ! $(awk -F= '/^ID=/ { print $2 }' /etc/os-release | tr -d '"') =~ ^(centos|rhel)$ ]]; then
+    if ! (source /etc/os-release; [[ "$ID" =~ ^(centos|rhel)$ || "$ID_LIKE" =~ (centos|rhel) ]]); then
         error "Unsupported OS"
         exit 1
     fi


01_install_requirements.sh
```diff
@@ -69,6 +69,12 @@ ansible-galaxy install -r vm-setup/requirements.yml
 # temporarily capping it until the issue is fixed
 ansible-galaxy collection install 'ansible.netcommon:<2.6.0'
 ansible-galaxy collection install --upgrade ansible.posix community.general
+
+if ! python -c 'import selinux'; then
+   dnf copr enable -y apuimedo/libselinux-py39
+   sudo dnf install -y python39-libselinux
```
hardys, line 75: Out of interest what is this used for?

celebdor: One of the ansible roles that dev-scripts uses. At some point there's usage of the ansible selinux module. Since we install ansible with pip for python version 3.9, the system installed python library is in a different library site-package dir

andfasano: Which one? If it's something not of a global interest/use, could be possible to move it closer to that role?

common.sh
```diff
@@ -1,20 +1,27 @@
...
+if command -v go; then
+    eval "$(go env)"
+    export PATH="${GOPATH}/bin:$PATH"
+    # Ensure if a go program crashes we get a coredump
+    #
+    # To get the dump, use coredumpctl:
+    #   coredumpctl -o oc.coredump dump /usr/local/bin/oc
+    #
+    export GOTRACEBACK=crash
+    export OPENSHIFT_INSTALL_PATH="$GOPATH/src/github.com/openshift/installer"
```
andfasano, line 23: Isn't already set in [common.sh](https://github.com/openshift-metal3/dev-scripts/blob/081b22db68430997dd7a416114213d87c1c751bb/common.sh#L140)?

01_install_requirements.sh
hardys: Yeah AFAICS I didn't need this when upgrading to 3.9, so it'd be good to understand exactly why (@celebdor can you find the error from dev-scripts/logs perhaps and paste it here?)

celebdor: vm-setup/roles/packages_installation/tasks/centos_required_packages.yml:  ansible.posix.selinux:

common.sh
celebdor: mmm. 22-23 should not be in the PR. I don't know how they sneaked in there

01_install_requirements.sh
andfasano: Filed a PR to fix it: https://github.com/metal3-io/metal3-dev-env/pull/971

hardys: Ok I think I can see the issue - in my environment at least, it seems ansible isn't always using the latest python version, which means I don't hit this error:

```
ansible --version
ansible [core 2.11.5] 
  config file = /home/dev-scripts/metal3-dev-env/ansible.cfg
  configured module search path = ['/home/shardy/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/shardy/.local/lib/python3.9/site-packages/ansible
  ansible collection location = /home/shardy/.ansible/collections:/usr/share/ansible/collections
  executable location = /home/shardy/.local/bin/ansible
  python version = 3.9.6 (default, Aug 11 2021, 06:39:25) [GCC 8.5.0 20210514 (Red Hat 8.5.0-3)]
  jinja version = 3.0.2
  libyaml = True
```

That says 3.9.6, and running a test playbook causes an error:

```
<localhost> EXEC /bin/sh -c 'sudo -H -S -n  -u root /bin/sh -c '"'"'echo BECOME-SUCCESS-aexzjxqhkozopjrvzkbqhiczdrtgpacn ; /usr/bin/python /home/shardy/.ansible/tmp/ansible-tmp-1647610289.3830082-1552414-208748671236489/AnsiballZ_selinux.py'"'"' && sleep 0'
<localhost> EXEC /bin/sh -c 'rm -f -r /home/shardy/.ansible/tmp/ansible-tmp-1647610289.3830082-1552414-208748671236489/ > /dev/null 2>&1 && sleep 0'
The full traceback is:
Traceback (most recent call last):
  File "/tmp/ansible_ansible.posix.selinux_payload_yqrncjpg/ansible_ansible.posix.selinux_payload.zip/ansible_collections/ansible/posix/plugins/modules/selinux.py", line 93, in <module>
ModuleNotFoundError: No module named 'selinux'
fatal: [localhost]: FAILED! => {
    "changed": false,
    "invocation": {
        "module_args": {
            "configfile": "/etc/selinux/config",
            "policy": "targeted",
            "state": "permissive"
        }
    },
    "msg": "Failed to import the required Python library (libselinux-python) on master-1.dev7.kni.lab.eng.bos.redhat.com's Python /usr/bin/python. Please read the module documentation and install it in the appropriate location. If the required library is installed, but Ansible is using the wrong Python interpreter, please consult the documentation on ansible_python_interpreter"
}

PLAY RECAP *************************************************************************************************************************************************************************************
localhost                  : ok=1    changed=0    unreachable=0    failed=1    skipped=0    rescued=0    ignored=0   

$ /usr/bin/python --version
Python 3.9.6
```

However, when running via dev-scripts, I see it works:

```
2022-03-18 09:58:12 ^[[0;34m<localhost> EXEC /bin/sh -c 'sudo -H -S -n  -u root /bin/sh -c '"'"'echo BECOME-SUCCESS-gbyjagyavoyzlnrpbwilchoikenvuhtn ; /usr/libexec/platform-python /home/shardy/.ansible/tmp/ansible-tmp-1647597492.7535546-1549496-123693126113812/AnsiballZ_selinux.py'"'"' && sleep 0'^[[0m
2022-03-18 09:58:13 ^[[0;34m<localhost> EXEC /bin/sh -c 'rm -f -r /home/shardy/.ansible/tmp/ansible-tmp-1647597492.7535546-1549496-123693126113812/ > /dev/null 2>&1 && sleep 0'^[[0m
2022-03-18 09:58:13 ^[[0;32mok: [localhost] => {^[[0m
2022-03-18 09:58:13 ^[[0;32m    "changed": false,^[[0m
2022-03-18 09:58:13 ^[[0;32m    "configfile": "/etc/selinux/config",^[[0m
2022-03-18 09:58:13 ^[[0;32m    "invocation": {^[[0m
2022-03-18 09:58:13 ^[[0;32m        "module_args": {^[[0m
2022-03-18 09:58:13 ^[[0;32m            "configfile": "/etc/selinux/config",^[[0m
2022-03-18 09:58:13 ^[[0;32m            "policy": "targeted",^[[0m
2022-03-18 09:58:13 ^[[0;32m            "state": "permissive"^[[0m
2022-03-18 09:58:13 ^[[0;32m        }^[[0m
2022-03-18 09:58:13 ^[[0;32m    },^[[0m
2022-03-18 09:58:13 ^[[0;32m    "msg": "",^[[0m
2022-03-18 09:58:13 ^[[0;32m    "policy": "targeted",^[[0m
2022-03-18 09:58:13 ^[[0;32m    "reboot_required": false,^[[0m
2022-03-18 09:58:13 ^[[0;32m    "state": "permissive"^[[0m
2022-03-18 09:58:13 ^[[0;32m}^[[0m
```

The difference is ansible is using `/usr/libexec/platform-python` which is still 3.6:

```
$ /usr/libexec/platform-python --version
Python 3.6.8
```

I'm not entirely sure why that is happening since [alternatives](https://github.com/openshift-metal3/dev-scripts/blob/master/01_install_requirements.sh#L55) should be pointing to 3.9 :thinking: 

```
$ ls -l /usr/bin | grep python
-rwxr-xr-x. 1 root root        3467 Nov 27  2018 activate-global-python-argcomplete
lrwxrwxrwx. 1 root root          36 Oct 30  2020 python -> /etc/alternatives/unversioned-python
lrwxrwxrwx. 1 root root          25 Jul  7  2020 python3 -> /etc/alternatives/python3
lrwxrwxrwx. 1 root root          31 Aug 11  2021 python3.6 -> /usr/libexec/platform-python3.6
lrwxrwxrwx. 1 root root          32 Aug 11  2021 python3.6m -> /usr/libexec/platform-python3.6m
-rwxr-xr-x. 1 root root        7760 Aug 11  2021 python3.9
-rwxr-xr-x. 1 root root        2570 Nov 27  2018 python-argcomplete-check-easy-install-script
-rwxr-xr-x. 1 root root         314 Nov 16  2017 python-argcomplete-tcsh
-rwxr-xr-x. 1 root root        2936 Nov 27  2018 register-python-argcomplete
lrwxrwxrwx. 1 root root          24 Oct 30  2020 unversioned-python -> /etc/alternatives/python
$ ls -l /etc/alternatives/ | grep python
lrwxrwxrwx. 1 root root 18 Nov 30 11:13 python -> /usr/bin/python3.9
lrwxrwxrwx. 1 root root 18 Oct  8 11:26 python3 -> /usr/bin/python3.9
lrwxrwxrwx. 1 root root 34 Oct  8 11:26 python3-man -> /usr/share/man/man1/python3.9.1.gz
lrwxrwxrwx. 1 root root 18 Nov 30 11:13 unversioned-python -> /usr/bin/python3.9
lrwxrwxrwx. 1 root root 34 Nov 30 11:13 unversioned-python-man -> /usr/share/man/man1/python3.9.1.gz
```

hardys: Seems like we may need to set [ansible_python_interpreter](https://docs.ansible.com/ansible/latest/porting_guides/porting_guide_2.8.html#python-interpreter-discovery) to something other than `auto` since that may be [causing this behavior](https://bugzilla.redhat.com/show_bug.cgi?id=1670188) (it's weird that ansible --version reports 3.9 though)

andfasano: @hardys confirmed that the same things it's happening on my instance (works with dev-scripts, but not in a standalone way)

hardys: > @hardys confirmed that the same things it's happening on my instance (works with dev-scripts, but not in a standalone way)

I think you'll find if you add `-e ansible_python_interpreter="auto"` standalone it will work (but only because it's using the 3.6 platform-python) - in dev-scripts that is set via the [inventory](https://github.com/metal3-io/metal3-dev-env/blob/main/vm-setup/inventory.ini#L2) which it seems may not be what we want in this case

andfasano: @elfosardo do you maybe have any suggestion for this specific problem? Despite the "confusing" Ansible output, it looks like that the combo rhel8 / python3.6 works fine

elfosardo: The ansible_python_interpreter can be mischieving, if configured to "auto" it sets the python binary based on its internal map https://github.com/ansible/ansible/blob/devel/lib/ansible/config/base.yml#L1502
@hardys suggestion is doable if you know in advance which python interpreter you're going to use (and the exact path), but I would not suggest to use a different python version then the system one because you could miss packaged depencies and have to install them with pip system-wide, which is absolutely not recommended.
Another solution could be using a virtualenv, but we're adding unneeded complexity.
Considering that the combo python3.6/rhel8 works fine, just stick with that for the time being.
If you really need/want to change python version, consider also changing the base operating system, using CentOS Stream 9 with metal3-dev-env does work (please see https://github.com/metal3-io/metal3-dev-env/pull/950), and next step will be converting dev-scripts.

hardys: > If you really need/want to change python version, consider also changing the base operating system

Yes, but one of the advantages to maintaining RHEL support is users don't have to switch OS variants or versions as often ;)

hardys: > it looks like that the combo rhel8 / python3.6 works fine

As I mentioned in https://github.com/openshift-metal3/dev-scripts/pull/1301 3.8+ is the requirement - I don't recall if I found a specific issue at that time, but in the [ansible docs](https://docs.ansible.com/ansible/latest/installation_guide/intro_installation.html#control-node-requirements) it says `Ansible 2.12 will require Python 3.8 or newer to function on the control node`

Also I'm not sure if we're actually using 3.6 consistently (ansible --version reports 3.9, but we see from the logs that at least in some cases 3.6 gets used due to the `auto`)

elfosardo: @hardys the "support" topic probably deserves a different discussion
Anyway, if you really want to use python3.9, I guess you can set `ansible_python_interpreter` to `/usr/bin/python` as it's defined as default alternative in L55, then install the selinux library using pip with `python -m pip install selinux`

andfasano: FYI https://github.com/openshift-metal3/dev-scripts/pull/1371

PR #1368, opened 2022-03-24
Use host network when building images

elfosardo:
In case the host where dev-scripts is running has only localhost as
resolver, the dns resolution during build fails as it's removed from
/etc/resolv.conf since it's not reachable from the container.
If we need to reach external urls during image building, the entire
build fails, as the urls can't be resolved.
As a workaround use the --network host option only during build.
For more info see https://github.com/containers/podman/issues/13599


dtantsur:
/approve

Works for me, thank you!

elfosardo:
/retest
packet setup failed

elfosardo:
/test e2e-metal-ipi-ovn-ipv6
doesn't look like the job is failing because of a baremetal-related issue

PR #1370, opened 2022-03-24
Use kubeconfig option in cluster up message

creydr:
oc option `--config` does not exist. 
````
$ oc --config <path> get no             
Error: unknown flag: --config
See 'oc get --help' for usage.
````

This PR addresses it and uses `--kubeconfig` in the cluster-up message instead.


hardys:
/approve

Thanks for the fix!

creydr:
Seems `e2e-metal-ipi-ovn-ipv6` is perma failing. Could someone with permissions do an `/override` to get this off the plate? :)
@hardys

PR #1372, opened 2022-03-28
AGENT-20: Add agent target to deploy agent installer

andfasano:
This patch adds an initial support for the agent based installer workflow, so that the agent team could start using it. Further improvementes and Ansible refactoring(s) will be addressed in subsequent PRs.

### DIFF
--- a/agent/03_agent_deployment.sh
+++ b/agent/03_agent_deployment.sh
@@ -112,15 +112,7 @@ EOF
 
 function generate_fleeting_iso() {
     export REPO_PATH=${WORKING_DIR}
-    sync_repo_and_patch fleeting https://github.com/openshift-agent-team/fleeting
-
-    # If set, use the configured fleeting PR
-    if [[ -n ${FLEETING_PR:-} ]]; then
-      pushd ${FLEETING_PATH}
-      git fetch origin pull/${FLEETING_PR}/head:pr${FLEETING_PR}
-      git checkout pr${FLEETING_PR}
-      popd
-    fi
+    sync_repo_and_patch fleeting https://github.com/openshift-agent-team/fleeting ${FLEETING_PR}
 
     generate_fleeting_manifests ${FLEETING_PATH}
     
--- a/agent/03_agent_deployment.sh
+++ b/agent/03_agent_deployment.sh
@@ -1,0 +1,164 @@
+#!/usr/bin/env bash
+set -xe
+
+SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )/.." && pwd )"
+
+LOGDIR=${SCRIPTDIR}/logs
+source $SCRIPTDIR/logging.sh
+source $SCRIPTDIR/common.sh
+source $SCRIPTDIR/network.sh
+source $SCRIPTDIR/utils.sh
+source $SCRIPTDIR/validation.sh
+
+### Recommended configuration (compact scenario):
+### export IP_STACK=v4
+### export NETWORK_TYPE=OpenShiftSDN
+### export NUM_WORKERS=0
+### export MASTER_DISK=120
+### export MASTER_MEMORY=16384
+
+early_deploy_validation
+
+export FLEETING_PATH=${FLEETING_PATH:-$WORKING_DIR/fleeting}
+export FLEETING_ISO=${FLEETING_ISO:-$FLEETING_PATH/output/fleeting.iso}
+
+node0=${CLUSTER_NAME}_master_0
+node0_ip=$(nth_ip $EXTERNAL_SUBNET_V4 20)
+
+function generate_fleeting_manifests() {
+
+    FLEETING_MANIFESTS_PATH="${FLEETING_PATH}/manifests"
+
+    mkdir -p ${FLEETING_MANIFESTS_PATH}
+
+    if [[ "$IP_STACK" = "v4" ]]; then
+      CLUSTER_NETWORK=${CLUSTER_SUBNET_V4}
+      SERVICE_NETWORK=${SERVICE_SUBNET_V4}
+      CLUSTER_HOST_PREFIX=${CLUSTER_HOST_PREFIX_V4}
+    elif [[ "$IP_STACK" = "v6" ]]; then
+      CLUSTER_NETWORK=${CLUSTER_SUBNET_V6}
+      SERVICE_NETWORK=${SERVICE_SUBNET_V6}
+      CLUSTER_HOST_PREFIX=${CLUSTER_HOST_PREFIX_V6}
+    fi
+
+    cat > "${FLEETING_MANIFESTS_PATH}/agent-cluster-install.yaml" << EOF
+apiVersion: extensions.hive.openshift.io/v1beta1
+kind: AgentClusterInstall
+metadata:
+  name: test-agent-cluster-install
+  namespace: cluster0
+spec:
+  apiVIP: 192.168.111.10
+  ingressVIP: 192.168.111.11
+  clusterDeploymentRef:
+    name: compact-cluster
+  imageSetRef:
+    name: openshift-v4.10.0
+  networking:
+    clusterNetwork:
+    - cidr: ${CLUSTER_NETWORK}
+      hostPrefix: ${CLUSTER_HOST_PREFIX}
+    serviceNetwork:
+    - ${SERVICE_NETWORK}
+  provisionRequirements:
+    controlPlaneAgents: ${NUM_MASTERS}
+  sshPublicKey: ${SSH_PUB_KEY}
+EOF
+
+    cat > "${FLEETING_MANIFESTS_PATH}/cluster-deployment.yaml" << EOF
+apiVersion: hive.openshift.io/v1
+kind: ClusterDeployment
+metadata:
+  name: compact-cluster
+  namespace: cluster0
+spec:
+  baseDomain: agent.example.com
+  clusterInstallRef:
+    group: extensions.hive.openshift.io
+    kind: AgentClusterInstall
+    name: test-agent-cluster-install
+    version: v1beta1
+  clusterName: compact-cluster
+  controlPlaneConfig:
+    servingCertificates: {}
+  platform:
+    agentBareMetal:
+      agentSelector:
+        matchLabels:
+          bla: aaa
+  pullSecretRef:
+    name: pull-secret
+EOF
+
+    cat > "${FLEETING_MANIFESTS_PATH}/infraenv.yaml" << EOF
+apiVersion: agent-install.openshift.io/v1beta1
+kind: InfraEnv
+metadata:
+  name: myinfraenv
+  namespace: cluster0
+spec:
+  clusterRef:
+    name: compact-cluster  
+    namespace: cluster0
+  pullSecretRef:
+    name: pull-secret
+  sshAuthorizedKey: ${SSH_PUB_KEY}
+  ignitionConfigOverride: '{"ignition": {"version": "3.1.0"}, "storage": {"files": [{"path": "/etc/someconfig", "contents": {"source": "data:text/plain;base64,aGVscGltdHJhcHBlZGluYXN3YWdnZXJzcGVj"}}]}}'
+  nmStateConfigLabelSelector:
+    matchLabels:
+      cluster0-nmstate-label-name: cluster0-nmstate-label-value
+EOF
+
+    set +x
+    pull_secret=$(cat $PULL_SECRET_FILE)
+    cat > "${FLEETING_MANIFESTS_PATH}/pull-secret.yaml" << EOF
+apiVersion: v1
+kind: Secret
+type: kubernetes.io/dockerconfigjson
+metadata:
+  name: pull-secret
+  namespace: cluster0
+stringData:
+  .dockerconfigjson: '${pull_secret}'
+
+EOF
+    set -x
+}
+
+function generate_fleeting_iso() {
+    export REPO_PATH=${WORKING_DIR}
+    sync_repo_and_patch fleeting https://github.com/openshift-agent-team/fleeting
+
+    # If set, use the configured fleeting PR
+    if [[ -n ${FLEETING_PR:-} ]]; then
+      pushd ${FLEETING_PATH}
+      git fetch origin pull/${FLEETING_PR}/head:pr${FLEETING_PR}
+      git checkout pr${FLEETING_PR}
+      popd
+    fi
+
+    generate_fleeting_manifests ${FLEETING_PATH}
+    
+    pushd ${FLEETING_PATH}
+    export NODE_ZERO_IP=$1
+    make iso 
+    popd
+}
+
+function attach_fleeting_iso() {
+    for (( n=0; n<${2}; n++ ))
+    do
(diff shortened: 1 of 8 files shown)


03_agent_deployment.sh
```diff
@@ -0,0 +1,152 @@
+#!/usr/bin/env bash
+set -xe
```
elfosardo, line 2: let's also set "unset variables as an error" and "exit on any error in pipeline"

03_agent_deployment.sh
```diff
@@ -0,0 +1,152 @@
...
+function generate_fleeting_iso() {
+    export REPO_PATH=${WORKING_DIR}
+    sync_repo_and_patch fleeting https://github.com/openshift-agent-team/fleeting
+
+    # If set, use the configured fleeting PR
+    if [[ -n ${FLEETING_PR:-} ]]; then
+      pushd ${FLEETING_PATH}
+      git fetch origin pull/${FLEETING_PR}/head:pr${FLEETING_PR}
+      git checkout pr${FLEETING_PR}
+      popd
```
hardys, line 123: I wonder if we should add this logic to `sync_repo_and_patch` - e.g some generic approach so you can define REPO_NAME_PR=123 or similar?

We have [some old logic](https://github.com/openshift-metal3/dev-scripts/blob/master/utils.sh#L274) to apply specific patches, but that's no longer used, so we could replace it with a PR based approach?

03_agent_deployment.sh
```diff
@@ -0,0 +1,152 @@
...
+    fi
+
+    generate_fleeting_manifests ${FLEETING_PATH}
+    
+    pushd ${FLEETING_PATH}
+    export NODE_ZERO_IP=$1
+    make iso 
+    sudo mv output/fleeting.iso /tmp
```
hardys, line 131: `/tmp` probably isn't the best location, should probably be moved somewhere in `$WORKING_DIR` instead?

03_agent_deployment.sh
```diff
@@ -0,0 +1,152 @@
...
+        sudo virt-xml ${name} --edit target=sdc --disk="boot_order=2" --start
+    done
+}
+
+write_pull_secret
+generate_fleeting_iso ${node0_ip}
+
+attach_fleeting_iso master $NUM_MASTERS
+attach_fleeting_iso worker $NUM_WORKERS
+attach_fleeting_iso extraworker $NUM_EXTRA_WORKERS
```
hardys, line 150: In most previous test flows the "extra" workers weren't used in the initial deploy (so they're available for scale-out tests etc), so perhaps we shouldn't attach the extraworker hosts here?

03_agent_deployment.sh
```diff
@@ -0,0 +1,152 @@
...
+
+    cat > "${FLEETING_MANIFESTS_PATH}/agent-cluster-install.yaml" << EOF
+apiVersion: extensions.hive.openshift.io/v1beta1
+kind: AgentClusterInstall
+metadata:
+  name: test-agent-cluster-install
+  namespace: cluster0
+spec:
+  apiVIP: 192.168.111.10
+  ingressVIP: 192.168.111.11
```
hardys, line 38: I wonder if we can break out the logic in `05_create_install_config.sh` so you can use API_VIP/INGRESS_VIP here instead?

Regardless of how we calculate the values, it would also probably be less confusing if the IPs chosen were consistent with the existing IPI install-config values? :)

03_agent_deployment.sh
```diff
@@ -0,0 +1,152 @@
...
+  imageSetRef:
+    name: openshift-v4.10.0
+  networking:
+    clusterNetwork:
+    - cidr: 10.128.0.0/14
+      hostPrefix: 23
+    serviceNetwork:
+    - 172.30.0.0/16
+  provisionRequirements:
+    controlPlaneAgents: 3
```
hardys, line 50: I guess this should be NUM_MASTERS - relatedly, do we expect the agent flow to allow SNO deployments?

hardys:
Looks like a great start, added a few comments :) 

/cc @lranjbar since this may replace https://github.com/openshift-metal3/dev-scripts/pull/1363 I think?

03_agent_deployment.sh
andfasano: Very likely they shouldn't be started also, so postponing this decision in a later moment

03_agent_deployment.sh
andfasano: I was using them but there are some problems in the setup that I'm still troubleshooting, so I'm leaving the hard-coded values for now

03_agent_deployment.sh
andfasano: Sounds as a good idea. I've experimented a little bit, and replacing the old logic seems a pretty natural approach (I've added the changes for this specific patch as a separate commit)

agent/03_agent_deployment.sh
```diff
@@ -0,0 +1,144 @@
+#!/usr/bin/env bash
+set -euxo pipefail
+
+SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )/.." && pwd )"
+
+LOGDIR=${SCRIPTDIR}/logs
+source $SCRIPTDIR/logging.sh
```
celebdor, line 7: the scriptdir could have spaces, better double quote it

agent/03_agent_deployment.sh
```diff
@@ -0,0 +1,144 @@
...
+metadata:
+  name: test-agent-cluster-install
+  namespace: cluster0
+spec:
+  apiVIP: 192.168.111.10
+  ingressVIP: 192.168.111.11
+  clusterDeploymentRef:
+    name: compact-cluster
+  imageSetRef:
+    name: openshift-v4.10.0
```
celebdor, line 42: Is there a 4.11 nightly we could point to instead?

agent/03_agent_deployment.sh
```diff
@@ -0,0 +1,144 @@
...
+  name: myinfraenv
+  namespace: cluster0
+spec:
+  clusterRef:
+    name: compact-cluster  
+    namespace: cluster0
+  pullSecretRef:
+    name: pull-secret
+  sshAuthorizedKey: ${SSH_PUB_KEY}
+  ignitionConfigOverride: '{"ignition": {"version": "3.1.0"}, "storage": {"files": [{"path": "/etc/someconfig", "contents": {"source": "data:text/plain;base64,aGVscGltdHJhcHBlZGluYXN3YWdnZXJzcGVj"}}]}}'
```
celebdor, line 92: what is this override for? It just puts "helpimtrappedinaswaggerspec" in /etc/someconfig

agent/03_agent_deployment.sh
```diff
@@ -0,0 +1,144 @@
...
+    set -x
+}
+
+function generate_fleeting_iso() {
+    export REPO_PATH=${WORKING_DIR}
+    sync_repo_and_patch fleeting https://github.com/openshift-agent-team/fleeting ${FLEETING_PR}
+
+    generate_fleeting_manifests ${FLEETING_PATH}
+    
+    pushd ${FLEETING_PATH}
```
celebdor, line 120: all these vars in the function could have spaces in them, they should be double quoted.

agent/README.md
```diff
@@ -0,0 +1,24 @@
...
+    # Configure master nodes specs
+    export MASTER_DISK=120
+    export MASTER_MEMORY=16384
+
+    # In compact scenario, no workers are required
+    export NUM_WORKERS=0
+
+Then run the `agent` target:
+
+    $ make agent
```
celebdor, line 24: It would be good to have a follow up PR listing the generated manifests and explaining that you can modify them after they are generated before generating the ISO.

agent/cleanup.sh
```diff
@@ -0,0 +1,21 @@
...
+SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )/.." && pwd )"
+
+LOGDIR=${SCRIPTDIR}/logs
+source $SCRIPTDIR/logging.sh
+source $SCRIPTDIR/common.sh
+source $SCRIPTDIR/utils.sh
+source $SCRIPTDIR/validation.sh
+source $SCRIPTDIR/agent/common.sh
+
+early_cleanup_validation
```
celebdor, line 13: Is this destroying and undefining the libvirt domains?

config_example.sh
```diff
@@ -312,3 +312,15 @@ set -x
...
+## Agent Deployment
+##
+
+# To use a local checkout of the fleeting repo, instead of fetching the remote
+# main branch
+# export FLEETING_PATH=~/go/src/github.com/openshift-agent-team/fleeting
+
+# When set, the changes associated to the specified pull request (for the fleeting repo)
+# are fetched, instead of using the main branch
+# export FLEETING_PR=12
```
celebdor, line 326: I think it might be better to have a less extensive agent/config_example.sh

agent/03_agent_deployment.sh
andfasano: IINW currently it's still hard-coded in fleeting, I was about to open a new PR for that

agent/03_agent_deployment.sh
andfasano: Was coming from the original @rwsu demo files (maybe to verify that ignition was working?). If not strictly required now we could remove it

agent/cleanup.sh
andfasano: No, that's in the `host_cleanup` target

config_example.sh
andfasano: I'd like to followup this part in another PR, especially when we'll be defining better the different scenario. For the moment here we have just two optional settings

agent/03_agent_deployment.sh
zaneb: Looking forward to the Ansible version already.

config_example.sh
celebdor: agreed

agent/03_agent_deployment.sh
celebdor: Let's remove it for now

agent/03_agent_deployment.sh
celebdor: you're right, @zaneb

hardys: I don't have any objection to this being hard-coded for now, but in future IMO it would be better to derive the image version from the same variables we use to determine `OPENSHIFT_RELEASE_IMAGE`

In that case you have the option of getting latest (default), setting a specific `OPENSHIFT_RELEASE_IMAGE`, or asking for a specific version, e.g:

```
export OPENSHIFT_RELEASE_TYPE="ga"
export OPENSHIFT_VERSION="latest-4.10"
```

agent/03_agent_deployment.sh
hardys: In future this could handle the existing IGNITION_EXTRA variable perhaps

agent/README.md
```diff
@@ -0,0 +1,24 @@
...
+
+Use the following configuration in your `config_<user>.sh` to deploy a 
+compact cluster:
+
+    # Configure ipv4 environment
+    export IP_STACK=v4
+    export NETWORK_TYPE=OpenShiftSDN
+
+    # Configure master nodes specs
+    export MASTER_DISK=120
```
hardys, line 16: Is this a typo?  the default is 30G per master, 120G is going to fill up even large virthost disks pretty fast?

utils.sh
```diff
@@ -270,15 +270,15 @@ function sync_repo_and_patch {
...
-        git checkout -b metalkube
-
-        shift; shift;
-        for arg in "$@"; do
-            curl -L $arg | git am
-        done
+      REPO_PR=$3
+      echo "Fetching PR ${REPO_PR}"
+      git fetch origin pull/${REPO_PR}/head:pr${REPO_PR}
+      git checkout pr${REPO_PR}
```
hardys, line 279: This seems fine for now, but in future we could make this more generic but dynamically evaluating `<REPO_NAME>_PR` e.g:

```
export REPO_PR_VAR="$(basename "$2" | tr "a-z" "A-Z")_PR"
REPO_PR=${!REPO_PR}
```

agent/README.md
lranjbar: This is a requirement that is enforced by Assisted Installer otherwise it wont start the install. However, the defaults can be overidden using the environment variable `HW_VALIDATOR_REQUIREMENTS` - docs for this are [here](https://github.com/openshift/assisted-service/blob/4b4f0f45d897430037039a9282225bf6bd02acfe/docs/dev/hardware-requirements.md)

bfournie:
I've tested this in my environment and was able to bring up a cluster with *make agent* and clean it up with *make clean*. I've talked with Andrea about also removing coreos.iso in the clean in a future update.

/lgtm

agent/README.md
zaneb: Can we get libvirt to create them as sparse QCOWs?

agent/03_agent_deployment.sh
zaneb: TBH we could just ignore this field, since we're always going to use the version matching the installer binary.

lranjbar:
I was also able to get an environment up using this. I think we can update the part about the disk space in a follow up PR.

/lgtm

rwsu (commented):
Other than FLEETING_PR not being optional, it worked well.

/lgtm

agent/03_agent_deployment.sh
andfasano: I think it's fine for the moment to ignore it, and review it in a later moment

config_example.sh
andfasano: Thanks @rwsu, fixed it

bfournie:
I retested this with the latest change.
/lgtm

rwsu (commented):
/lgtm

hardys:
/approve

Sounds like this is working so lets merge and iterate on the remaining issues via follow-ups

zaneb:
I expected these test failures to be fixed by #1374. Do we have to rebase this branch before CI will pick up those changes?

andfasano:
> I expected these test failures to be fixed by #1374. Do we have to rebase this branch before CI will pick up those changes?

The branch is merged automatically by Prow with the latest master when building the `src` image (ie https://prow.ci.openshift.org/log?container=test&id=1510894302388031488&job=pull-ci-openshift-metal3-dev-scripts-master-e2e-metal-ipi, `Resolved source https://github.com/openshift-metal3/dev-scripts to master@fd16386a, merging: #1372 169a5ad6 @andfasano`), so unless a conflict is present then it should be ok now

Relaunching another round...

andfasano:
@hardys @elfosardo would be fine to override the ipv4/ipv6 jobs to land the PR, as it's requested by the agent team? None of the current changes seem to be related to the failures, which instead look related to some bootstrap problems (and occasionally packet failure). Thanks!

hardys:
/override ci/prow/e2e-metal-ipi-ovn-dualstack
/override ci/prow/e2e-metal-ipi-virtualmedia

CI is still in bad shape, and this new agent flow isn't covered by any of the existing jobs - we can see from the e2e-metal-ipi-serial-ipv4 that this doesn't break the exising non-agent flow, so approving to enable the agent team to make progress

PR #1373, opened 2022-03-30
Avoid unbound variables

elfosardo:
They're not nice

### DIFF
--- a/02_configure_host.sh
+++ b/02_configure_host.sh
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-set -xe
+set -euxo pipefail
 
 source logging.sh
 source common.sh
@@ -307,7 +307,7 @@ if ! command -v baremetal | grep -v "${IRONICCLIENT_PATH}"; then
 fi
 
 # Block Multicast with ebtables
-if [ "$DISABLE_MULTICAST" == "true" ]; then
+if [ "${DISABLE_MULTICAST:-false}" == "true" ]; then
     for dst in 224.0.0.251 224.0.0.18; do
         sudo ebtables -A INPUT --pkttype-type multicast -p ip4 --ip-dst ${dst} -j DROP
         sudo ebtables -A FORWARD --pkttype-type multicast -p ip4 --ip-dst ${dst} -j DROP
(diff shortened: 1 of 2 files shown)


02_configure_host.sh
```diff
@@ -88,7 +88,7 @@ ansible-playbook \
     -e "vm_platform=$NODES_PLATFORM" \
     -e "sushy_ignore_boot_device=$REDFISH_EMULATOR_IGNORE_BOOT_DEVICE" \
     -e "manage_baremetal=$MANAGE_BR_BRIDGE" \
-    -e "provisioning_url_host=$PROVISIONING_URL_HOST" \
+    -e "provisioning_url_host=${PROVISIONING_URL_HOST:-}" \
```
hardys, line 91: I think this is only used to template the kernel/ramdisk URL in metal3-dev-env which isn't actually used in dev-scripts:

```
$ grep -Ri PROVISIONING_URL_HOST ./*
./roles/libvirt/defaults/main.yml:provisioning_url_host: 172.22.0.1
./roles/libvirt/templates/ironic_nodes.json.j2:        "deploy_kernel": "http://{{ provisioning_url_host }}/images/ironic-python-agent.kernel",
./roles/libvirt/templates/ironic_nodes.json.j2:        "deploy_ramdisk": "http://{{ provisioning_url_host }}/images/ironic-python-agent.initramfs"
```

However for consistency, I think we should just set this to `PROVISIONING_HOST_IP` which is defined, and would be the correct IP if someone wanted to test via Ironic directly when also defining `IRONIC_IPA_DOWNLOADER_LOCAL_IMAGE` to add the IPA images to the virthost httpd?

02_configure_host.sh
```diff
@@ -88,7 +88,7 @@ ansible-playbook \
     -e "vm_platform=$NODES_PLATFORM" \
     -e "sushy_ignore_boot_device=$REDFISH_EMULATOR_IGNORE_BOOT_DEVICE" \
     -e "manage_baremetal=$MANAGE_BR_BRIDGE" \
-    -e "provisioning_url_host=$PROVISIONING_URL_HOST" \
+    -e "provisioning_url_host=${PROVISIONING_URL_HOST:-$PROVISIONING_HOST_IP}" \
```
hardys, line 91: Can we just replace `PROVISIONING_URL_HOST` with `PROVISIONING_HOST_IP` here?

PR #1374, opened 2022-03-30
Fix default ironic image quay branch

elfosardo:
It was changed from master to main recently


osherdp:
> lgtm, although as a follow-up we should probably consider consuming IRONIC_IMAGE from the payload, instead of having this dependency on upstream images (even in the case where UPSTREAM_IRONIC isn't specified)

I think that's how ipv6 jobs still passing

PR #1376, opened 2022-04-04
Silence git checkout of master

cybertron:
On metal3-dev-env where the default branch has been renamed to main,
this command causes an ugly error message in the logs that has
resulted in a lot of wasted debug time. Since we expect that this
command is going to fail some of the time, let's suppress that
output. If there is a legitimate failure on this line, the second
command will also fail so we'll know where to look.

### DIFF
--- a/utils.sh
+++ b/utils.sh
@@ -267,9 +267,9 @@ function sync_repo_and_patch {
     pushd $DEST
 
     git am --abort || true
-    git checkout master || git checkout main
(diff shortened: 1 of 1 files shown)


hardys:
Seems fine but perhaps https://github.com/openshift-metal3/dev-scripts/pull/1377 is a more flexible approach?

cybertron:
This isn't working anyway. I see the error is still output even with --quiet.

PR #1377, opened 2022-04-05
Determine default git branch when syncing repo

elfosardo:
Because noone could be stopped from naming a default branch "goofy"

### DIFF
--- a/utils.sh
+++ b/utils.sh
@@ -266,11 +266,13 @@ function sync_repo_and_patch {
 
     pushd $DEST
 
+    DEFAULT_BRANCH=$(git remote show origin | sed -n '/HEAD branch/s/.*: //p')
+
     git am --abort || true
-    git checkout master || git checkout main
+    git checkout ${DEFAULT_BRANCH}
     git fetch origin
-    git rebase origin/master || git rebase origin/main
-    
+    git rebase origin/${DEFAULT_BRANCH}
+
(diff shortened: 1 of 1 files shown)


hardys:
/approve

lgtm - in future we might want to allow for per-repo override similar to [this suggestion](https://github.com/openshift-metal3/dev-scripts/pull/1372#discussion_r838318920)

cybertron:
/lgtm

The silver lining to ci being a disaster right now is that we have plenty of opportunities to confirm that the bogus error message isn't there. :-)

utils.sh
```diff
@@ -266,11 +266,13 @@ function sync_repo_and_patch {
 
     pushd $DEST
 
+    DEFAULT_BRANCH=$(git remote show origin | sed -n '/HEAD branch/s/.*: //p')
```
andfasano, line 269: What about using `git ls-remote` to check if the remote branch exists or not, ie `git ls-remote -q --exit-code --heads $1 main`?

hardys: But then we'd be back to hard-coding a branch name, vs this approach which finds the HEAD branch from the origin remote (which should handle any branch name)?

andfasano: Agreed. The only benefit of the ls-remote is that it doesn't require to clone the repo for the check, but in this case it's defintely not required.

PR #1380, opened 2022-04-07
Support for agent static network test

bfournie:
Add support for testing the agent static network configuration. Currently only the IP for Node 0 can be configured but the framework is there to use static IPs for all nodes when that support is available.


pawanpinjarkar:
Tested this PR and /lgtm

agent/03_agent_deployment.sh
```diff
@@ -13,6 +13,13 @@ source $SCRIPTDIR/agent/common.sh
 
 early_deploy_validation
 
+function add_host_dns_entries() {
+    ip_list=$1
+
+    sudo virsh net-update ostestbm add dns-host  "<host ip='${ip_list[0]}'> <hostname>master-0-static</hostname> </host>"  --live --config
```
dhellmann, line 19: Doesn't the network name depend on the cluster name? I think there's a variable in common.sh.

bfournie: Yes changed to use variable for network name

agent/03_agent_deployment.sh
```diff
@@ -13,6 +13,18 @@ source $SCRIPTDIR/agent/common.sh
 
 early_deploy_validation
 
+function add_host_dns_entries() {
+    ip_list=$1
+    num_ips=${#ip_list[@]}
+
+    for (( i=0; i<$num_ips; i++ ))
+    do
+        hostname=${FLEETING_HOSTNAME}${i}
```
andfasano, line 22: Especially if in future we're going to add more entries for the other nodes, then I guess it could be better to be coherent with the current dev-scripts naming conventions and use `MASTER_HOSTNAME_FORMAT` and `WORKER_HOSTNAME_FORMAT`

agent/03_agent_deployment.sh
```diff
@@ -13,6 +13,18 @@ source $SCRIPTDIR/agent/common.sh
...
+        sudo virsh net-update ${BAREMETAL_NETWORK_NAME} add dns-host  "<host ip='${ip_list[i]}'> <hostname>${hostname}</hostname> </host>"  --live --config
+    done
```
andfasano, line 24: As per the current imlementation, the same hostname could have an entry in the dhcp as well. What about removing such entry form the dhcp section, in order to have a cleaner/simpler environment?

agent/03_agent_deployment.sh
```diff
@@ -107,17 +119,70 @@ stringData:
...
+
+    # Get the generated mac addresses to set up nmstate
+    mac_list=($(sudo virsh net-dumpxml ${BAREMETAL_NETWORK_NAME} | grep -Po '(?<=host mac=).*' | cut -d' ' -f1))
+    ip_list=$2
+
+    add_host_dns_entries ${ip_list}
+
+    num_ips=${#ip_list[@]}
+
+    echo -n "" > ${FLEETING_MANIFESTS_PATH}/nmstateconfig.yaml
```
andfasano, line 131: Is this effectively required?

agent/03_agent_deployment.sh
```diff
@@ -107,17 +119,70 @@ stringData:
...
+    interfaces:
+      - name: eth0
+        type: ethernet
+        state: up
+        mac-address: ${mac_list[i]}
+        ipv4:
+          enabled: true
+          address:
+            - ip: ${ip_list[i]}
+              prefix-length: 24
```
andfasano, line 155: Here you could use `CLUSTER_HOST_PREFIX`

agent/03_agent_deployment.sh
```diff
@@ -134,10 +199,34 @@ function attach_fleeting_iso() {
 
 write_pull_secret
 
-node0_ip=$(nth_ip $EXTERNAL_SUBNET_V4 20)
+# needed for assisted-service to run nmstatectl
+# This is temporary and will go away when https://github.com/nmstate/nmstate is used
+sudo yum install -y nmstate
+
+if [[ "$FLEETING_STATIC_IP_NODE0_ONLY" = "true" ]]; then
```
andfasano, line 206: Is this config variable required as a transient one?

agent/03_agent_deployment.sh
```diff
@@ -107,17 +119,70 @@ stringData:
   .dockerconfigjson: '${pull_secret}'
 
 EOF
+
+    # Get the generated mac addresses to set up nmstate
+    mac_list=($(sudo virsh net-dumpxml ${BAREMETAL_NETWORK_NAME} | grep -Po '(?<=host mac=).*' | cut -d' ' -f1))
+    ip_list=$2
+
+    add_host_dns_entries ${ip_list}
```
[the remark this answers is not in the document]
bfournie: Sure, moved to after call to generate_fleeting_iso()

agent/03_agent_deployment.sh
```diff
@@ -107,17 +119,70 @@ stringData:
   .dockerconfigjson: '${pull_secret}'
 
 EOF
+
+    # Get the generated mac addresses to set up nmstate
+    mac_list=($(sudo virsh net-dumpxml ${BAREMETAL_NETWORK_NAME} | grep -Po '(?<=host mac=).*' | cut -d' ' -f1))
+    ip_list=$2
+
```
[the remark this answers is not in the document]
bfournie: Good idea to tie the macs to the master/worker names. Made change per suggestion.

agent/03_agent_deployment.sh
```diff
@@ -13,6 +13,18 @@ source $SCRIPTDIR/agent/common.sh
 
 early_deploy_validation
 
+function add_host_dns_entries() {
+    ip_list=$1
+    num_ips=${#ip_list[@]}
+
+    for (( i=0; i<$num_ips; i++ ))
```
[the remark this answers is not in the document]
bfournie: I do like that syntax better but in this case the index is being used to get the hostname.

agent/03_agent_deployment.sh
bfournie: OK, changed to use the MASTER_HOSTNAME_FORMAT and WORKER_HOSTNAME_FORMAT instead of redefining it

agent/03_agent_deployment.sh
bfournie: I've run into some intermittent issues when testing this since it requires restarting the network after that change. I don't want to make this fix now and introduce instability but will take it as a follow-on.

agent/03_agent_deployment.sh
bfournie: Since the cat now appends to the file I wanted to make sure the file doesn't or is empty, but it looks like 'make clean' removes all of the manifest files so yes I can remove this line.

agent/03_agent_deployment.sh
bfournie: Good idea, changed to use that.

agent/03_agent_deployment.sh
bfournie: We need this now as the support for static ips on all states hasn't merged yet but I think it could also be used in the future as a case with static IP on node0 and dhcp addresses on all other hosts is a valid use case.

agent/03_agent_deployment.sh
```diff
@@ -107,17 +134,64 @@ stringData:
...
 
 function generate_fleeting_iso() {
     export REPO_PATH=${WORKING_DIR}
+    ip_list=$1
+
     sync_repo_and_patch fleeting https://github.com/openshift-agent-team/fleeting ${FLEETING_PR}
 
-    generate_fleeting_manifests ${FLEETING_PATH}
-    
+    generate_fleeting_manifests ${FLEETING_PATH} $ip_list $2 $3
```
rwsu, line 192: I'm having trouble testing it. Seeing this error:
`./agent/03_agent_deployment.sh: line 192: $2: unbound variable`

bfournie: Thanks @rwsu, that should be fixed with latest update.

agent/03_agent_deployment.sh
```diff
@@ -134,10 +208,40 @@ function attach_fleeting_iso() {
...
+    ip=${base_ip}+${n}
+    ip_list+=($(nth_ip ${external_subnet} ${ip}))
+done
+
 set_api_and_ingress_vip
 
-generate_fleeting_iso ${node0_ip}
+mac_list=()
+set_mac_addresses master $NUM_MASTERS
+set_mac_addresses worker $NUM_WORKERS
```
andfasano, line 240: I think there could be some space to improve the management of the static ips and mac lists. The two lists are synchronized, as their nth index refers to the same host, but there's no clue about that in the name. Also, their creation is performed in separate places (and with different approaches), and I think that this could make easier to introduce some errors in future.
I'd suggest to:
1) Use a more consistent namings for them, ie something like global vars `FLEETING_NODES_IPS` and `FLEETING_NODES_MACS` (btw simplify also the parameters passing)
2) Have just one function for generating both of them in the first level of the script
3) Some strong assumption about ordering are made elsewhere (ie first masters and then workers). That's reasonable when dealing with a single list, but please let's document it to avoid missing it in future

agent/03_agent_deployment.sh
```diff
@@ -134,10 +208,40 @@ function attach_fleeting_iso() {
...
+
+if [[ "$IP_STACK" = "v4" ]]; then
+    external_subnet=$EXTERNAL_SUBNET_V4
+else
+    external_subnet=$EXTERNAL_SUBNET_V6
+fi
+
+# Use an IP address outside of the range that is set for DHCP, just for sanity checks.
+# It shouldn't matter as the dhcp addresses will not be picked up when using static IPs
+base_ip=40
```
andfasano, line 229: Current dhcp range ends up to 60, just to avoid any potentital conflict it's better to stay out of it

andfasano: Additionally, see also here: https://github.com/metal3-io/metal3-dev-env/blob/681b4d26ab5f90b0d08679df36b82ebb74eca993/lib/network.sh#L133-L134

agent/03_agent_deployment.sh
bfournie: Good points @andfasano. I changed to use globals and reworked it to use one function to set everything up to ensure consistency.

agent/03_agent_deployment.sh
bfournie: Changed to start IPs at 80 to avoid any overlap with DHCP.

agent/03_agent_deployment.sh
```diff
@@ -107,17 +134,64 @@ stringData:
...
+    num_ips=${#ip_list[@]}
+
+    # Create a yaml for each host in nmstateconfig.yaml
+    for (( i=0; i<$num_ips; i++ ))
+    do
+        cat >> "${FLEETING_MANIFESTS_PATH}/nmstateconfig.yaml" << EOF
+apiVersion: agent-install.openshift.io/v1beta1
+kind: NMStateConfig
+metadata:
+  name: mynmstateconfig
```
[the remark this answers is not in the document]
bfournie: Thanks @rwsu, yes good point. I changed to use the hostname in the manifest that is generated at the same time as the IPs and Macs (see comment above).

agent/03_agent_deployment.sh
```diff
@@ -107,17 +150,61 @@ stringData:
...
 }
 
 function generate_fleeting_iso() {
     export REPO_PATH=${WORKING_DIR}
+
     sync_repo_and_patch fleeting https://github.com/openshift-agent-team/fleeting ${FLEETING_PR}
 
-    generate_fleeting_manifests ${FLEETING_PATH}
-    
+    generate_fleeting_manifests ${FLEETING_PATH} ${FLEETING_NODES_IPS}
```
andfasano, line 205: nit: this could be simplified, but we could do it into another pr

bfournie: Good catch, updated.

andfasano:
ipv6 failures are not related to this patch

PR #1382, opened 2022-04-20
Agent - use static IPs for all hosts by default

bfournie:
Now that https://github.com/openshift-agent-team/fleeting/pull/25 has merged can configure static IPs for all hosts. Will leave the flag so we can test a scenario with static IP on Node 0 and dynamic IPs on other hosts.


andfasano:
ipv6 failure not related to the current patch

PR #1383, opened 2022-04-22
Include NUM_WORKERS in agent based installs

rwsu:
AgentClusterInstall.Spec.provisionRequirements.workerAgents should
reflect the value set in NUM_WORKERS.

This allows high availability clusters to be installed through
the agent based installer flow.


pawanpinjarkar (approved):
/lgtm

PR #1386, opened 2022-04-29
Agent: use installer branch

zaneb:
Use the master-agent-installer branch of the openshift/installer repo
instead of the openshift-agent-team/fleeting prototype to do agent-based
installation.

### DIFF
--- a/agent/03_agent_deployment.sh
+++ b/agent/03_agent_deployment.sh
@@ -198,15 +198,47 @@ EOF
     set -x
 }
 
+function sync_installer() {
+    local DEST="${OPENSHIFT_INSTALL_PATH}"
+    if [ ! -d "${DEST}" ]; then
+        echo "Syncing $DEST"
+
+        mkdir -p "${DEST}"
+        git clone https://github.com/openshift/installer.git "${DEST}"
+
+        pushd "${DEST}"
+
+        DEFAULT_BRANCH=agent-installer
+
+        git am --abort || true
+        git checkout "${DEFAULT_BRANCH}"
+        git fetch origin
+        git rebase "origin/${DEFAULT_BRANCH}"
+    else
+        pushd "${DEST}"
+    fi
+
+    # If set, use the specified PR number
+    if [ -n "${INSTALLER_PR}" ]; then
+      echo "Fetching PR ${INSTALLER_PR}"
+      git fetch origin "pull/${INSTALLER_PR}/head"
+      git checkout FETCH_HEAD
+    fi
+
+    popd
+}
+
+function build_installer() {
+    sync_installer
+
+    pushd "${OPENSHIFT_INSTALL_PATH}"
+    ./hack/build.sh
+    popd
+}
+
 function generate_fleeting_iso() {
-    export REPO_PATH=${WORKING_DIR}
-
-    sync_repo_and_patch fleeting https://github.com/openshift-agent-team/fleeting ${FLEETING_PR}
-
-    generate_fleeting_manifests
-
-    pushd ${FLEETING_PATH}
-    make iso 
+    pushd "${WORKING_DIR}"
+    "${OPENSHIFT_INSTALL_PATH}/bin/openshift-install" agent create image
     popd
 }
 
@@ -230,9 +262,9 @@ get_static_ips_and_macs
 
 set_api_and_ingress_vip
 
+build_installer
+generate_fleeting_manifests
 generate_fleeting_iso
 
 attach_fleeting_iso master $NUM_MASTERS
 attach_fleeting_iso worker $NUM_WORKERS
-
-
--- a/agent/common.sh
+++ b/agent/common.sh
@@ -1,8 +1,8 @@
 #!/usr/bin/env bash
 set -euxo pipefail
 
-export FLEETING_PATH=${FLEETING_PATH:-$WORKING_DIR/fleeting}
-export FLEETING_ISO=${FLEETING_ISO:-$FLEETING_PATH/output/fleeting.iso}
-export FLEETING_MANIFESTS_PATH="${FLEETING_PATH}/manifests"
-export FLEETING_PR=${FLEETING_PR:-}
+export FLEETING_ISO="${WORKING_DIR}/output/fleeting.iso"
+export FLEETING_MANIFESTS_PATH="${WORKING_DIR}/manifests"
+
+export INSTALLER_PR="${INSTALLER_PR:-}"
 export FLEETING_STATIC_IP_NODE0_ONLY=${FLEETING_STATIC_IP_NODE0_ONLY:-"false"}
--- a/config_example.sh
+++ b/config_example.sh
@@ -317,13 +317,10 @@ set -x
 ## Agent Deployment
 ##
 
-# To use a local checkout of the fleeting repo, instead of fetching the remote
-# main branch
-# export FLEETING_PATH=~/go/src/github.com/openshift-agent-team/fleeting
-
-# When set, the changes associated to the specified pull request (for the fleeting repo)
-# are fetched, instead of using the main branch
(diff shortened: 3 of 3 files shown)


bfournie:
Was able to do a complete install using:
*export INSTALLER_PR=5852*

config_example.sh should be updated to replace:
# export FLEETING_PR=12
with
# export INSTALLER_PR=xx

We should eventually replace all "FLEETING" strings but can do that as a follow on

rwsu (commented):
I was able to deploy a cluster with the installer PR. I think config_example.sh should be updated too because there are some references to fleeting there.

agent/03_agent_deployment.sh
```diff
@@ -198,15 +198,46 @@ EOF
...
+        if [ -n "${INSTALLER_PR}" ]; then
+          echo "Fetching PR ${INSTALLER_PR}"
+          git fetch origin "pull/${INSTALLER_PR}/head:pr${INSTALLER_PR}"
+          git checkout "pr${INSTALLER_PR}"
+        fi
+
+        popd
+    fi
+}
+
```
andfasano, line 230: dev-scripts already has some support for building and using the installer, I think we should try to reuse / adapt the current features as much as possible to avoid duplications. 
See [KNI_INSTALL_FROM_GIT](https://github.com/openshift-metal3/dev-scripts/blob/41b9cfb172af92788ab9c3395b2d60ae90f886f0/docs/developing-installer.md).
The [clone_installer()](https://github.com/openshift-metal3/dev-scripts/blob/41b9cfb172af92788ab9c3395b2d60ae90f886f0/ocp_install_env.sh#L77) and [build_installer()](https://github.com/openshift-metal3/dev-scripts/blob/41b9cfb172af92788ab9c3395b2d60ae90f886f0/ocp_install_env.sh#L84) functions could be reused / refactored to checkout our current development branch (instead of the master)

agent/common.sh
```diff
@@ -1,8 +1,7 @@
 #!/usr/bin/env bash
 set -euxo pipefail
 
-export FLEETING_PATH=${FLEETING_PATH:-$WORKING_DIR/fleeting}
-export FLEETING_ISO=${FLEETING_ISO:-$FLEETING_PATH/output/fleeting.iso}
-export FLEETING_MANIFESTS_PATH="${FLEETING_PATH}/manifests"
-export FLEETING_PR=${FLEETING_PR:-}
+export INSTALLER_PATH="${OPENSHIFT_INSTALL_PATH:-${HOME}/go/src/github.com/openshift/installer}"
+export FLEETING_ISO=${FLEETING_ISO:-$INSTALLER_PATH/output/fleeting.iso}
+export FLEETING_MANIFESTS_PATH="${INSTALLER_PATH}/manifests"
```
andfasano, line 6: Even though initially we'll be forced to build always the installer from the src, I think it would be preferrable to copy the manifests/artifacts in the `WORKING_DIR` and continue to use the one found there. Probably we could start creating an `agent` folder within the `WORKING_DIR` to store such artifacts

andfasano: Another point: maybe it could be useful (even though with some additional maintenance effort) to maintain the previous fleeting support for a while, until the installer code will be consolidated, instead of throwing it away immediately?

agent/03_agent_deployment.sh
zaneb: That's where I copied-and-pasted the code from, but:

- `clone_installer` clones to the wrong place (path is hardcoded instead of using `$OPENSHIFT_INSTALL_PATH`) and then, worse, checks out the wrong branch (`main` or `master`). You used to be able to specify the branch but, ahem, _somebody_ [repurposed](https://github.com/openshift-metal3/dev-scripts/commit/ecf0b3807344dffe1a47e1e21d5f9d3ed9a1ba81#diff-35a54395325de0227bbdd613dbf61c3265b5f6ab205cf32348e3538399cf1b9eR276-R280) that argument :stuck_out_tongue_winking_eye:
- `build_installer` mostly does stuff we don't need. Once you take that away there's only one line left (running `hack/build.sh`), so I just inlined it.

andfasano: And that's why I suggested to refactor the code (especially for those parts that are in common for both the flows), since the requirements changed over the time - and now there's a need to use a different branch, so `sync_repo_and_patch` could be enhanced to support it, maybe through an env var given that it's not a very common request.

And currently all the functions for manipulating the installer are defined in [ocp_install_env.sh](https://github.com/openshift-metal3/dev-scripts/blob/master/ocp_install_env.sh), and in future there could be other bits that could be useful for our flows.

agent/03_agent_deployment.sh
```diff
@@ -198,15 +198,47 @@ EOF
     set -x
 }
 
-function generate_fleeting_iso() {
-    export REPO_PATH=${WORKING_DIR}
+function sync_installer() {
```
andfasano, line 201: A possibility to avoid the duplication of `sync_repo_and_patch` could be to change https://github.com/openshift-metal3/dev-scripts/blob/894ff9dfa867fd9583557f020347d4800f9d2d64/utils.sh#L269 into 

```
DEFAULT_BRANCH=${DEFAULT_BRANCH:-$(git remote show origin | sed -n '/HEAD branch/s/.*: //p')}
```

And this would allow to override it in the `sync_installer()` function, ie: 

```
DEFAULT_BRANCH=agent-installer sync_repo_and_patch installer https://github.com/openshift/installer.git ${INSTALLER_PR}
```

agent/03_agent_deployment.sh
```diff
@@ -198,15 +198,47 @@ EOF
...
+        git clone https://github.com/openshift/installer.git "${DEST}"
+
+        pushd "${DEST}"
 
-    sync_repo_and_patch fleeting https://github.com/openshift-agent-team/fleeting ${FLEETING_PR}
+        DEFAULT_BRANCH=agent-installer
 
-    generate_fleeting_manifests
+        git am --abort || true
+        git checkout "${DEFAULT_BRANCH}"
```
zaneb, line 214: I'm undecided about the right thing to do here. Currently it only checks out the branch the first time, upon cloning.
The `sync_repo_and_patch` code always checks out the branch and rebases it on the upstream. So that would mean you can't test a change locally on a branch other than `agent-installer`.
It seems to me that you sometimes want to automatically rebase, but sometimes don't. I'm not sure how to determine which is which.

zaneb: The would appear to combine the worst of both worlds - you can't work on any local branch except `agent-installer`, and you have to manually update it.

andfasano: When working on a local branch the test on  `OPENSHIFT_INSTALL_PATH` should avoid (re)checking out the branch or the PR (as it happens in the build_installer() function, even though the current function does not manage the additional PR param) - since the `sync_repo_and_patch` should be required only when a local checkout is not available - thus allowing the build installer to build whatever it's found in the `OPENSHIFT_INSTALL_PATH` - while here I just noticed that (if I'm not wrong with the tabs alignment) the PR checkout is always executed and maybe this caused some confusion (at least for me).

Considering also that for the CI we'll not need to manage the code at all (but the installer will have to be extracted from the ephemeral release payload), it could be worth evaluating to reuse in some way the whole [03_build_installer.sh](https://github.com/openshift-metal3/dev-scripts/blob/master/03_build_installer.sh) as it is by adjusting a couple of integration points (namely, the checkout of the `agent-installer` branch instead of the `main`, and probably the `TAGS` when building it). In this way we could continue to use the already existing `KNI_INSTALL_FROM_GIT` feature when developing locallly (with the possibility to override `OPENSHIFT_INSTALL_PATH` when required).

pawanpinjarkar:
Tested the changes and works fine.

PR #1387, opened 2022-05-03
AGENT-217: Use `AGENT_E2E_TEST_SCENARIO` env var for the Agent based installer.

pawanpinjarkar:
Users who want to deploy a cluster with the agent-based installer are required now to just define one env var `AGENT_E2E_TEST_SCENARIO` in the `config_user.sh` file. The value of `AGENT_E2E_TEST_SCENARIO` can be set as the scenario one wants to deploy, e.g. SNO_IPV4, SNOIPV6, HA_IPV4, etc.
The underlying code then sets the lower-level details such as number of masters, number of workers, memory, disk, CPU, etc.

To test the changes for IPV4 SNO deployment,
- Set `export AGENT_E2E_TEST_SCENARIO= SNO_IPV4` in  `config_<user>.sh` file.

`AGENT_E2E_TEST_SCENARIO is not set at all or is set to empty string`
> AGENT_E2E_TEST_SCENARIO is missing or empty. Did you forget to set the AGENT_E2E_TEST_SCENARIO env var in the config_<USER>.sh file? Supported values: COMPACT_IPV4, COMPACT_IPV6, HA_IPV4, HA_IPV6, SNO_IPV4, SNO_IPV6.

`AGENT_E2E_TEST_SCENARIO is set with a wrong value`
> Invalid value for AGENT_E2E_TEST_SCENARIO. Supported values: COMPACT_IPV4, COMPACT_IPV6, HA_IPV4, HA_IPV6, SNO_IPV4, SNO_IPV6.


pawanpinjarkar:
cc @andfasano @bfournie

bfournie:
If ingressVIP is the only field that is different when NUM_MASTERS=1 it would be more maintainable (although perhaps less readable) to put the conditional only around that field, something like:

   cat > "${FLEETING_MANIFESTS_PATH}/agent-cluster-install.yaml" << EOF
apiVersion: extensions.hive.openshift.io/v1beta1
kind: AgentClusterInstall
metadata:
  name: test-agent-cluster-install
  namespace: cluster0
spec:
  apiVIP: ${API_VIP}
EOF
  if [[ "$NUM_MASTERS" > "1" ]]; then
    cat >> "${FLEETING_MANIFESTS_PATH}/agent-cluster-install.yaml" << EOF
  ingressVIP: ${INGRESS_VIP}
EOF
  fi
    cat >> "${FLEETING_MANIFESTS_PATH}/agent-cluster-install.yaml" << EOF
<rest of data here>
EOF

If we anticipate other fields would also need to be different, than perhaps the readability would rule and the approach you have would be better.

andfasano:
@pawanpinjarkar did you maybe already plan to add support for different configuration scenarios directly within this PR? I realized that it'd would be useful at this point to have a single config var to be defined by the user, ie something like `AGENT_CONFIG=sno`, `AGENT_CONFIG=ha` or `AGENT_CONFIG=compact`. Apart simplifying the setting of the lower level vars such as `NUM_MASTERS` or `NUM_WORKERS`, the `AGENT_CONFIG` variable could be used as well in the code to drive the decisions where needed, as in this case - with the additional benefit of expressing more clearly the intention, thus with some improved readability

pawanpinjarkar:
> 

Thank you for your review. Yes, it makes things easy to read and configure if we just use `AGENT_CONFIG` env to describe the scenario one wants to create( SNO, Compact, HA, etc). Once `AGENT_CONFIG` is set for any of these scenarios, in `agent/common.sh`, we can define underlying requirements such as disk count, a number of master nodes, worker nodes, CPUs, memory, etc.

I think we can also define if the deployment is IPV4 or IPV6 with values as `SNO_IPV4` , `SNO_IPV6`instead of just `SNO`

pawanpinjarkar:
> > 
> 
> Thank you for your review. Yes, it makes things easy to read and configure if we just use `AGENT_CONFIG` env to describe the scenario one wants to create( SNO, Compact, HA, etc). Once `AGENT_CONFIG` is set for any of these scenarios, in `agent/common.sh`, we can define underlying requirements such as disk count, a number of master nodes, worker nodes, CPUs, memory, etc.
> 
> I think we can also define if the deployment is IPV4 or IPV6 with values as `SNO_IPV4` , `SNO_IPV6`instead of just `SNO`

Maybe IPV4 or IPV6 needs to set at dev-scripts level and we can just use scenarios like `SNO`

andfasano:
> > > 
> > 
> > 
> > Thank you for your review. Yes, it makes things easy to read and configure if we just use `AGENT_CONFIG` env to describe the scenario one wants to create( SNO, Compact, HA, etc). Once `AGENT_CONFIG` is set for any of these scenarios, in `agent/common.sh`, we can define underlying requirements such as disk count, a number of master nodes, worker nodes, CPUs, memory, etc.
> > I think we can also define if the deployment is IPV4 or IPV6 with values as `SNO_IPV4` , `SNO_IPV6`instead of just `SNO`
> 
> Maybe IPV4 or IPV6 needs to set at dev-scripts level and we can just use scenarios like `SNO`

I guess it's more convenient to specify the network stack as part of the scenario name, as it's a fundamental dimension of the scenario matrix, so sounds fine to me use to vars such as `SNO_IPV4`, `SNO_IPV6` or `SNO_DUALSTACK`. Note that @lranjbar started documenting the underlying required vars in https://github.com/openshift-metal3/dev-scripts/pull/1388

rwsu (commented):
I wasn't able deploy the SNO_IPV4 config. I think it is missing the logic to remove ingress_vip and possibly api_vip from AgentClusterInstall for SNO.

rwsu:
Should we add AGENT_CONFIG to config_example.sh? A list of available options and explaining what they are would be useful.

pawanpinjarkar:
> I wasn't able deploy the SNO_IPV4 config. I think it is missing the logic to remove ingress_vip and possibly api_vip from AgentClusterInstall for SNO.

Yes, I forgot to push that change.

agent/01_agent_install_requirements.sh
```diff
@@ -0,0 +1,28 @@
+# #!/usr/bin/env bash
```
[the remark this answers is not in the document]
pawanpinjarkar: Done

pawanpinjarkar:
/ depends on https://github.com/openshift-metal3/dev-scripts/pull/1392 and https://github.com/openshift-metal3/dev-scripts/pull/1393

agent/01_agent_install_requirements.sh
```diff
@@ -0,0 +1,28 @@
...
+        echo "Using CONFIG ${XDG_CONFIG_HOME}/dev-scripts/config" 1>&2
+        CONFIG="${XDG_CONFIG_HOME}/dev-scripts/config"
+    else
+        error "Please run with a configuration environment set."
+        error "eg CONFIG=config_example.sh ./01_all_in_one.sh"
+        exit 1
+    fi
+fi
+source $CONFIG
+
```
rwsu, line 24: I suggest having a more informative error when AGENT_CONFIG is not set. This is what I'm seeing atm:
```
./agent/01_agent_install_requirements.sh: line 25: AGENT_CONFIG: unbound variable
make: *** [Makefile:30: agent_requirements] Error 1
```

agent/01_agent_install_requirements.sh
```diff
@@ -0,0 +1,28 @@
...
+if [ "${AGENT_CONFIG}" != "COMPACT_IPV4" ] && [ "${AGENT_CONFIG}" != "COMPACT_IPV6" ] && [ "${AGENT_CONFIG}" != "HA_IPV4" ] && [ "${AGENT_CONFIG}" != "HA_IPV6" ] && [ "${AGENT_CONFIG}" != "SNO_IPV4" ] && [ "${AGENT_CONFIG}" != "SNO_IPV6" ] ; then
+    echo "Invalid value: AGENT_CONFIG. Supported values: COMPACT_IPV4, COMPACT_IPV6, HA_IPV4, HA_IPV6, SNO_IPV4, SNO_IPV6."
```
rwsu, line 26:
```suggestion
    echo "Invalid value for AGENT_CONFIG. Supported values: COMPACT_IPV4, COMPACT_IPV6, HA_IPV4, HA_IPV6, SNO_IPV4, SNO_IPV6."
```

agent/01_agent_install_requirements.sh
```diff
@@ -0,0 +1,28 @@
...
+    elif [[ -f "${XDG_CONFIG_HOME}/dev-scripts/config" ]]; then
+        echo "Using CONFIG ${XDG_CONFIG_HOME}/dev-scripts/config" 1>&2
+        CONFIG="${XDG_CONFIG_HOME}/dev-scripts/config"
+    else
+        error "Please run with a configuration environment set."
+        error "eg CONFIG=config_example.sh ./01_all_in_one.sh"
+        exit 1
+    fi
+fi
+source $CONFIG
```
rwsu, line 23: A similar code block exist in common.sh. I think we can either turn it into a function and reuse the code block that way.

Or remove these lines entirely. Update the Makefile with requirement listed ahead of agent_requirements in the "agent" target. And removing requirements from the agent_requirements target.:
```

agent: requirements agent_requirements configure agent_build_installer agent_configure agent_create_cluster

agent_requirements:
	./agent/01_agent_install_requirements.sh
```

config_example.sh
```diff
@@ -319,3 +319,11 @@ set -x
...
 # Set whether static IPs will be used for all nodes or only Node0
 # export AGENT_STATIC_IP_NODE0_ONLY="true"
+
+# Set a single config variable to create a cluster for the different scenarios 
+# i.e. Single Node Openshift(SNO), Highly Available (HA) or Compact cluster.
+# The only supported values for AGENT_CONFIG are COMPACT_IPV4, COMPACT_IPV6, HA_IPV4, HA_IPV6, SNO_IPV4 and SNO_IPV6.
+# When set, the code internally sets other low level details such as disk size, memory, number of masters and workers,
+# cpu and ip stack.
+# This config variable is used only by the agent based installer.
+# export AGENT_CONFIG=SNO_IPV4
```
rwsu, line 329: Should there be a default value for AGENT_CONFIG?

rwsu (changes requested):
Looking good. I was able to deploy a SNO cluster.

common.sh
```diff
@@ -269,6 +269,47 @@ export EXTRA_WORKER_MEMORY=${EXTRA_WORKER_MEMORY:-${WORKER_MEMORY}}
...
+        ;;
+    "SNO_IPV4" | "SNO_IPV6" )
+        export NUM_MASTERS=1
+        export MASTER_VCPU=8
+        export MASTER_DISK=120
+        export MASTER_MEMORY=32768
+        export NUM_WORKERS=0
+        ;;
+esac
+
```
andfasano, line 312: I'd suggest to evaluate this case only when `AGENT_CONFIG` is not empty. In this way you could add also reasonable defaults that are valid for all the configuration (when agent config is defined) and leave to the case the additional specific settings for every distinct configuration.

Forr example, for setting the ip stack you could use something like: 
```
IP_STACK=${AGENT_CONFIG##*_IP}
```

config_example.sh
andfasano: Agree, there should be an empty default in common.sh

agent/04_agent_configure.sh
```diff
@@ -83,8 +83,14 @@ metadata:
   name: test-agent-cluster-install
   namespace: cluster0
 spec:
+EOF
+if [[ "$NUM_MASTERS" > "1" ]]; then
```
andfasano, line 87: Maybe it would be slightly more clear if the condition was set againt `AGENT_CONFIG` (ie when it's not compact)

agent/01_agent_install_requirements.sh
andfasano: Couldn't be just the default case for the below switch (evaluated just when `AGENT_CONFIG` is not empty)?

agent/01_agent_install_requirements.sh
andfasano: I guess that if we keep `AGENT_CONFIG` in common.sh, evaluated only when defined (see comments below), we could avoid at all this part

agent/04_agent_configure.sh
mhanss: I would suggest to put the condition around the `set_api_and_ingress_vip` function and set the empty value for api_vip and egress_vip when the case is SNO.
For Example:
```
if [[ $AGENT_CONFIG  =~ "SNO" ]]; then
   echo "Setting empty value for api_vip and ingress_vip as these are not required for SNO cluster"
   API_VIP=""
   INGRESS_VIP=""
else
   set_api_and_ingress_vip 
fi
```

config_example.sh
pawanpinjarkar: Done

zaneb:
Shouldn't we set a default value? It seems like this would break existing configs.

common.sh
```diff
@@ -57,6 +57,8 @@ if [ -z "${CONFIG:-}" ]; then
 fi
 source $CONFIG
 
+export AGENT_CONFIG=${AGENT_CONFIG:-}
```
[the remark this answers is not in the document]
pawanpinjarkar: I think the earlier changes I had is the better way to go i.e. https://github.com/openshift-metal3/dev-scripts/pull/1387#discussion_r875273860 I would like to go with those changes and make the suggestions made by @rwsu WDYT @andfasano That way, it becomes easy to read the targets and do not break any non agent scenario.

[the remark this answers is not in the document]
pawanpinjarkar: i.e. either have a `agent_pre_configure` step or `agent_requirements` as was in my earlier changes.

agent/04_agent_configure.sh
```diff
@@ -83,8 +83,14 @@ metadata:
   name: test-agent-cluster-install
   namespace: cluster0
 spec:
+EOF
+if [[ "$AGENT_CONFIG" == "COMPACT_IPV4" || "$AGENT_CONFIG" == "COMPACT_IPV6" || "$AGENT_CONFIG" == "HA_IPV4" || "$AGENT_CONFIG" == "HA_IPV6" ]]; then
```
mhanss, line 87: We can put below single condition to check whether it is SNO or not, this will help to avoid adding multiple conditions in future.
```suggestion
if [[ ! "$AGENT_CONFIG" =~ "SNO" ]]; then
```

agent/04_agent_configure.sh
```diff
@@ -221,7 +227,9 @@ sudo yum install -y nmstate
 
 get_static_ips_and_macs
 
-set_api_and_ingress_vip
+if [[ "$AGENT_CONFIG" == "COMPACT_IPV4" || "$AGENT_CONFIG" == "COMPACT_IPV6" || "$AGENT_CONFIG" == "HA_IPV4" || "$AGENT_CONFIG" == "HA_IPV6" ]]; then
```
mhanss, line 230: Please check https://github.com/openshift-metal3/dev-scripts/pull/1387#discussion_r888699614

pawanpinjarkar:
This PR was originally meant to simply support SNO cluster creation with agent installer but later the changes to just one config env var were also included. The changes just for SNO creation are tracked from a [separate PR ](https://github.com/openshift-metal3/dev-scripts/pull/1409)while this PR will only focus on using one config variable.

agent/04_agent_configure.sh
pawanpinjarkar: I would like to keep the check based on NUM_MASTERS as already exists today. Moreover, this check will fail when AGENT_CONFIG ( aka E2E_TEST_SCENARIO) is incorrectly set to just HA

common.sh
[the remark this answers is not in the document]
pawanpinjarkar: Discussed with @andfasano over slack and we agreed to some changes. This PR has been updated accordingly.

pawanpinjarkar:
This PR has gone through a log of reviews and changes with different ideas. I have finalized the approach as below -

1. Use env var `E2E_TEST_SCENARIO` instead of `AGENT_CONFIG`
2. We only set an empty default for this env var because we want users to explicitly know what cluster they are deploying. 
3. Users will have to mandatorily define the env var `E2E_TEST_SCENARIO` in the config_<USER>.sh file.
4. If `E2E_TEST_SCENARIO` is not set at all or is set to an empty string, users are presented with an error, and further processing is stopped.
`E2E_TEST_SCENARIO is missing or empty. Did you forget to set the E2E_TEST_SCENARIO env var in the config_.sh file? Supported values: COMPACT_IPV4, COMPACT_IPV6, HA_IPV4, HA_IPV6, SNO_IPV4, SNO_IPV6.`
5. If `E2E_TEST_SCENARIO` is set with a wrong value, users are presented with an error and further processing is stopped.
`Invalid value for E2E_TEST_SCENARIO. Supported values: COMPACT_IPV4, COMPACT_IPV6, HA_IPV4, HA_IPV6, SNO_IPV4, SNO_IPV6.`

@andfasano @rwsu @lranjbar

common.sh
```diff
@@ -269,6 +271,40 @@ export EXTRA_WORKER_MEMORY=${EXTRA_WORKER_MEMORY:-${WORKER_MEMORY}}
 export EXTRA_WORKER_DISK=${EXTRA_WORKER_DISK:-${WORKER_DISK}}
 export EXTRA_WORKER_VCPU=${EXTRA_WORKER_VCPU:-${WORKER_VCPU}}
 
+if [[ ! -z ${E2E_TEST_SCENARIO} ]]; then
+  export IP_STACK=${E2E_TEST_SCENARIO##*_IP}
+
```
andfasano, line 276: minor optimization: scenario base name could be extracted by using 

```SCENARIO=${E2E_TEST_SCENARIO%%_*}```

and then it could be used to simplify the switch cases

File: config_example.sh [hunk 12]
```diff
@@ -596,9 +596,10 @@ set -x
...
-# export MASTER_VCPU=8
-# export MASTER_MEMORY=32768
-# export MASTER_DISK=120
+# Set a single config variable E2E_TEST_SCENARIO to create a cluster for the different scenarios 
+# i.e. Single Node Openshift(SNO), Highly Available (HA) or Compact cluster.
+# The only supported values for E2E_TEST_SCENARIO are COMPACT_IPV4, COMPACT_IPV6, HA_IPV4, HA_IPV6, SNO_IPV4 and SNO_IPV6.
+# When set, the code internally sets other low level details such as disk size, memory, number of masters and workers,
+# cpu and ip stack.
+# This config variable is used only by the agent based installer and is required.
+export E2E_TEST_SCENARIO=HA_IPV4
```
andfasano, line 605:
```suggestion
# export E2E_TEST_SCENARIO=HA_IPV4
```

agent/01_agent_requirements.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+set -o pipefail
+
+SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )/.." && pwd )"
+
+source $SCRIPTDIR/common.sh
+
+if [[ -z ${E2E_TEST_SCENARIO} ]]; then
```
andfasano, line 9: For consistency with the other steps, it would be better to invoke the `early_deploy_validation` func at the beginning

agent/03_agent_build_installer.sh
```diff
@@ -25,4 +25,4 @@ source $SCRIPTDIR/03_build_installer.sh
 # Copy install binary if built from src
 if [ ! -z "$KNI_INSTALL_FROM_GIT" -a -f "$OPENSHIFT_INSTALL_PATH/bin/openshift-install" ]; then 
     cp "$OPENSHIFT_INSTALL_PATH/bin/openshift-install" "$OCP_DIR"
-fi
\ No newline at end of file
+fi
```
andfasano, line 28: Is it just a newline?

common.sh
pawanpinjarkar: Done

config_example.sh
pawanpinjarkar: Typo. Done.

agent/01_agent_requirements.sh
pawanpinjarkar: Done

agent/03_agent_build_installer.sh
pawanpinjarkar: My editor auto-added it.

PR #1389, opened 2022-05-06
network: add support to dualstack

dougsland:
The dualstack introduces apiVIPs and ingressVIPs and deprecate
apiVIP and ingressVIP fields. This change just adapt the script.

Signed-off-by: Douglas Schilling Landgraf <dougsland@redhat.com>

### DIFF
--- a/network.sh
+++ b/network.sh
@@ -23,6 +23,9 @@ function wrap_if_ipv6(){
     echo "$1"
 }
 
+export STRINGS_SEPARATOR=","
+export PATH_CONF_DNSMASQ="/etc/NetworkManager/dnsmasq.d/openshift-${CLUSTER_NAME}.conf"
+
 export IP_STACK=${IP_STACK:-"v6"}
 export HOST_IP_STACK=${HOST_IP_STACK:-${IP_STACK}}
 
@@ -193,26 +196,88 @@ if [ -n "${NETWORK_CONFIG_FOLDER:-}" ]; then
   NETWORK_CONFIG_FOLDER="$(readlink -m $NETWORK_CONFIG_FOLDER)"
 fi
 
-function set_api_and_ingress_vip() {
+function is_varset_add_comma() {
+    # Arguments:
+    #     First argument: the var to be checked if is already set
+    #     Second argument: new value to be assigned to the var
+    #
+    # Description:
+    #     Adds comma as separator in case var is already set. Useful for multiple API_VIP or INGRESS_VIP
+    #     Example: 192.168.111.5,fd2e:6f44:5dd8:c956::14
+    #
+    # Returns:
+    #     Print the string using comma
+
+    # if string is *NOT* NULL/empty
+    if [[ -n "${1}" ]]; then
+        echo "$1,$2"
+    else
+        echo "$2"
+    fi
+}
+
+function get_vips() {
+    # Arguments:
+    #     None
+    #
+    # Description:
+    #     Gets the INGRESS VIP and API VIP addresses (ipv4 and ipv6)
+    #
+    # Returns:
+    #     None
+    #
+    if [[ -n "${EXTERNAL_SUBNET_V4}" ]]; then
+        API_VIP=$(dig +noall +answer "api.${CLUSTER_DOMAIN}" @$(network_ip ${BAREMETAL_NETWORK_NAME}) | awk '{print $NF}')
+        INGRESS_VIP=$(nth_ip $EXTERNAL_SUBNET_V4 4)
+    fi
+
+    if [[ -n "${EXTERNAL_SUBNET_V6}" ]]; then
+        API_VIP=$(is_varset_add_comma "${API_VIP}" $(dig -t AAAA +noall +answer "api.${CLUSTER_DOMAIN}" @$(network_ip ${BAREMETAL_NETWORK_NAME}) | awk '{print $NF}'))
+        INGRESS_VIP=$(is_varset_add_comma "${INGRESS_VIP}" $(nth_ip $EXTERNAL_SUBNET_V6 4))
+    fi
+}
+
+
+function add_dnsmasq_multi_entry() {
+    # Arguments:
+    #     First argument: the type of entry to be added in openshift-${CLUSTER_NAME}.conf
+    #     Types available: apivip OR ingressvip
+    #
+    #     Second argument: The list (or single entry) of apivip or ingressvip
+    #
+    # Description:
+    #     Add entries into openshift-${CLUSTERNAME}.conf for dnsmasq
+    #
+    # Returns:
+    #     None
+    for i in ${2//${STRINGS_SEPARATOR}/ }; do
+        if [ "${1}" = "apivip" ] ; then
+            echo "address=/api.${CLUSTER_DOMAIN}/${i}" | sudo tee "${PATH_CONF_DNSMASQ}"
+        fi
+
+        if [ "${1}" = "ingressvip" ] ; then
+            echo "address=/.apps.${CLUSTER_DOMAIN}/${i}" | sudo tee -a "${PATH_CONF_DNSMASQ}"
+        fi
+    done
+}
+
+set_api_and_ingress_vip() {
   # NOTE: This is equivalent to the external API DNS record pointing the API to the API VIP
   if [ "$MANAGE_BR_BRIDGE" == "y" ] ; then
-      if [[ -z "${EXTERNAL_SUBNET_V4}" ]]; then
-          API_VIP=$(dig -t AAAA +noall +answer "api.${CLUSTER_DOMAIN}" @$(network_ip ${BAREMETAL_NETWORK_NAME}) | awk '{print $NF}')
-          INGRESS_VIP=$(nth_ip $EXTERNAL_SUBNET_V6 4)
-      else
-          API_VIP=$(dig +noall +answer "api.${CLUSTER_DOMAIN}" @$(network_ip ${BAREMETAL_NETWORK_NAME}) | awk '{print $NF}')
-          INGRESS_VIP=$(nth_ip $EXTERNAL_SUBNET_V4 4)
-      fi
-      echo "address=/api.${CLUSTER_DOMAIN}/${API_VIP}" | sudo tee /etc/NetworkManager/dnsmasq.d/openshift-${CLUSTER_NAME}.conf
-      echo "address=/.apps.${CLUSTER_DOMAIN}/${INGRESS_VIP}" | sudo tee -a /etc/NetworkManager/dnsmasq.d/openshift-${CLUSTER_NAME}.conf
-      echo "listen-address=::1" | sudo tee -a /etc/NetworkManager/dnsmasq.d/openshift-${CLUSTER_NAME}.conf
+      get_vips
+
+      add_dnsmasq_multi_entry "apivip" "${API_VIP}"
+      add_dnsmasq_multi_entry "ingressvip" "${INGRESS_VIP}"
+
+      echo "listen-address=::1" | sudo tee -a "${PATH_CONF_DNSMASQ}"
 
       # Risk reduction for CVE-2020-25684, CVE-2020-25685, and CVE-2020-25686
       # See: https://access.redhat.com/security/vulnerabilities/RHSB-2021-001
-      echo "cache-size=0" | sudo tee -a /etc/NetworkManager/dnsmasq.d/openshift-${CLUSTER_NAME}.conf
+      echo "cache-size=0" | sudo tee -a "${PATH_CONF_DNSMASQ}"
 
       sudo systemctl reload NetworkManager
   else
+      # Specific for users *NOT* using devscript with KVM (virsh) for deploy. (Reads: baremetal)
       if [[ -z "${EXTERNAL_SUBNET_V4}" ]]; then
           API_VIP=$(dig -t AAAA +noall +answer "api.${CLUSTER_DOMAIN}"  | awk '{print $NF}')
       else
--- a/ocp_install_env.sh
+++ b/ocp_install_env.sh
@@ -122,6 +122,58 @@ EOF
   fi
 }
 
+function renderVIPs() {
+    # Arguments:
+    #     First argument: field name
+    #     Second argument: value for the field
+    #
+    # Description:
+    #     This function helps to write in a YAML format multiple resources. (i.e: apiVIPs and ingressVIPs)
+    #     Example: apiVIPs, "192.168.11.5, fd2e:6f44:5dd8:c956::15"
+    #
+    # Returns:
+    #     YAML formatted resource
+    STRINGS_SEPARATOR=","
+    resource="${2}";
+
+    echo "    ${1}"
+    for data in ${resource//${STRINGS_SEPARATOR}/ }; do
+        echo "    - ${data}"
+    done
+}
+
+function setVIPs() {
+    # Arguments:
+    #     The type of VIP: apivips OR ingressvips
+    #
+    # Description:
+    #     apiVIP and ingressVIP both has been DEPRECATED in 4.12 in favor of apiVIPs and ingressVIPs.
+    #     This functions helps to write the new apiVIPs/ingressVIPs format or set the old fields.
+    #
+    # Returns:
+    #     The YAML formatted APIVIP or INGRESSVIP (supports new and old format)
+    case "${1}" in
+    "apivips")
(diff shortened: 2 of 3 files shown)


dougsland:
Tested using @creydr patch https://github.com/openshift/installer/pull/5798 and seems to work.

ocp_install_env.sh
```diff
@@ -130,6 +130,22 @@ EOF
...
+  # The char , is used for multiple vips which is supported only in dualstack mode
+  if [[ "${API_VIP}" == *","* ]]; then
+cat <<EOF
+    apiVIPs: ${API_VIP}
+    ingressVIPs: ${INGRESS_VIP}
+EOF
+  else
+cat <<EOF
+    apiVIP: ${API_VIP}
+    ingressVIP: ${INGRESS_VIP}
```
creydr, line 144: When https://github.com/openshift/installer/pull/5798 got merged, you can also specify the single vip in the plural field. (-> no need for `apiVIP` and `ingressVIP` anymore). E.g.:
```
    apiVIPs:
    - apiVIP
    ingressVIPs:
    - ingressVIP
```
This allows you to have the same rendering for one and multiple vips, probably too.

ocp_install_env.sh
```diff
@@ -130,6 +130,22 @@ EOF
   fi
 }
 
+function apiVIP_ingressVIP_dualstack() {
+  # dual stack will be added 4.12, let's keep compat mode
+  # The char , is used for multiple vips which is supported only in dualstack mode
+  if [[ "${API_VIP}" == *","* ]]; then
+cat <<EOF
+    apiVIPs: ${API_VIP}
+    ingressVIPs: ${INGRESS_VIP}
```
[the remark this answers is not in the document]
cybertron: Yeah, this won't do what we need. The comma format isn't valid for YAML.

ocp_install_env.sh
cybertron: We still need support for older releases in dev-scripts so we need to have a way to use the old field names.

However, I'd prefer to look at the openshift release number to determine which we need. Even in the case of a single VIP in 4.12, we should still be using the new fields (the current implementation would always use the old fields for single VIP deployments, IIUC). At some point the deprecated ones will go away and we'd have to revisit this anyway. Better to do it now when it's fresh in our minds.

ocp_install_env.sh
```diff
@@ -130,6 +130,22 @@ EOF
   fi
 }
 
+function apiVIP_ingressVIP_dualstack() {
```
cybertron, line 133: Nit: this isn't specific to dual stack. This function will be called for all deployment types.

ocp_install_env.sh
```diff
@@ -130,6 +130,22 @@ EOF
...
+  # dual stack will be added 4.12, let's keep compat mode
+  # The char , is used for multiple vips which is supported only in dualstack mode
+  if [[ "${API_VIP}" == *","* ]]; then
```
cybertron, line 136: This makes a couple of assumptions I'm not sure are valid:

- That if there are multiple API VIPs there will always be multiple ingress VIPs
- That we only want to use the new fields if multiple VIPs are specified

I think this needs to be based on the openshift version like it is for dnsvip, and the logic should handle a mismatch in the number of VIPs between ingress and API. I'm not sure the latter is strictly necessary, but there's no inherent reason you couldn't have two API VIPs and only one ingress (or vice versa).

cybertron (changes requested):

dougsland:
/unhold looks like it's working

Here some quick demo out of devscripts for testing:
```
#!/bin/bash

function set_multiple_resources() {
    # This function helps to write in a YAML format multiple resources.
    # Example: apiVIPs and ingressVIPs
    resource="${2}";
    echo "    ${1}"
    for data in ${resource//,/ }; do
        echo "    - ${data}"
    done
}

function openshift_version() {
    return 4
}

function set_VIPs() {
    # apiVIP and ingressVIP both has been DEPRECATED in 4.12 in favor of apiVIPs and ingressVIPs.
    # This functions helps to write the new apiVIPs/ingressVIPs format or set the old fields.
    case "${1}" in

    apivips)
        #if printf '%s\n4.12\n' "$(openshift_version)" | sort -V -C; then
            if [[ "${API_VIP}" == *","* ]]; then
                set_multiple_resources "apiVIPs:" "${API_VIP}"
                return
            fi
        #else
# fallback mode for users not using the new apiVIPs schema
cat <<EOF
    apiVIP: ${API_VIP}
EOF
        #fi
    ;;
    ingressvips)
        #if printf '%s\n4.12\n' "$(openshift_version)" | sort -V -C; then
            if [[ "${INGRESS_VIP}" == *","* ]]; then
                set_multiple_resources "ingressVIPs:" "${INGRESS_VIP}"
                return
            fi
        #else
# fallback mode for users not using the new igressVIPs schema
cat <<EOF
     ingressVIP: ${INGRESS_VIP}
EOF
        #fi
    ;;
    esac
}


API_VIP="192.168.2.3"
INGRESS_VIP="192.168.1.1, 192.168.2.3"
set_VIPs apivips
set_VIPs ingressvips

```

ocp_install_env.sh
```diff
@@ -130,6 +130,54 @@ EOF
...
+        else
+# fallback mode for users not using the new apiVIPs schema
+cat <<EOF
+    apiVIP: ${API_VIP}
+EOF
+        fi
+    ;;
+    "ingressvips")
+        if printf '%s\n4.12\n' "$(openshift_version)" | sort -V -C; then
+            if [[ "${INGRESS_VIP}" == *","* ]]; then
```
creydr, line 167: same as above

creydr (changes requested):

ocp_install_env.sh
```diff
@@ -130,6 +130,54 @@ EOF
...
+    done
+}
+
+function set_VIPs() {
+    # apiVIP and ingressVIP both has been DEPRECATED in 4.12 in favor of apiVIPs and ingressVIPs.
+    # This functions helps to write the new apiVIPs/ingressVIPs format or set the old fields.
+    # 
+    # Args:
+    #   1 - resource name, example: apiVIPs or ingressVIPs
+    #   2 - data
```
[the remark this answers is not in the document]
dougsland: sure

ocp_install_env.sh
```diff
@@ -130,6 +130,54 @@ EOF
...
+    case "${1}" in
+
+    "apivips")
+        if printf '%s\n4.12\n' "$(openshift_version)" | sort -V -C; then
+            if [[ "${API_VIP}" == *","* ]]; then
```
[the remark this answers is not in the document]
creydr: But isn't the fallback not for the version check?

ocp_install_env.sh
```diff
@@ -130,6 +130,65 @@ EOF
...
+function renderVIPs() {
+    # This function helps to write in a YAML format multiple resources.
+    # Example: apiVIPs and ingressVIPs
+    resource="${2}";
+
+    # case statement to make sure we are compatible with single string for apiVIP and ingressVIP
+    case "${1}" in
+
+    "apiVIP:")
+         echo "apivip: $(adjustVIPs ${resource} {resource})"
```
creydr, line 149: needs to be intended (same level in yaml as `apiVIPs` - as you did in line 159)

ocp_install_env.sh
```diff
@@ -130,6 +130,65 @@ EOF
...
+         return
+    ;;
+    "ingressVIP:")
+         echo "ingressvip: $(adjustVIPs ${resource})"
```
creydr, line 153: same as above

creydr (changes requested):
In case of OCP version <= 4.11 and having multiple VIPs defined in API_VIP this would result in picking only the first one. IMHO we should render the full string (instead of only the first VIP) and then later let the validation fail on it as the user entered garbage (multiple VIPs not supported in <=4.11).
And I am wondering if this could be written without playing around with `IFS`, but this might be a personal thing :D

dougsland:
> In case of OCP version <= 4.11 and having multiple VIPs defined in API_VIP this would result in picking only the first one. IMHO we should render the full string (instead of only the first VIP) and then later let the validation fail on it as the user entered garbage (multiple VIPs not supported in <=4.11). And I am wondering if this could be written without playing around with `IFS`, but this might be a personal thing :D

All good,  adjusted the script.

ocp_install_env.sh
```diff
@@ -130,6 +130,48 @@ EOF
...
+    case "${1}" in
+    "apivips")
+        if printf '4.12\n%s\n' "$(openshift_version)" | sort -V -C; then \
+            # OCP version is equals or newer as 4.12 and supports the new VIPs fields
+            renderVIPs "apiVIPs:" "${API_VIP}"
+        else
+            # OCP version is older as 4.12 and does not support the new VIPs fields
+cat <<EOF
+    apiVIP: ${API_VIP}
+EOF
```
creydr, line 157: nit: as you need `echo` in renderVIPs too, you could use `echo "    apiVIP: ${API_VIP}"` here to... Probably makes it clearer to read here.

ocp_install_env.sh
```diff
@@ -130,6 +130,48 @@ EOF
...
+    ;;
+    "ingressvips")
+        if printf '4.12\n%s\n' "$(openshift_version)" | sort -V -C; then \
+            # OCP version is equals or newer as 4.12 and supports the new VIPs fields
+            renderVIPs "ingressVIPs:" "${INGRESS_VIP}"
+        else
+            # OCP version is older as 4.12 and does not support the new VIPs fields
+cat <<EOF
+    ingressVIP: ${INGRESS_VIP}
+EOF
```
creydr, line 168: same as above

creydr:
Seems lgtm from my side. @cybertron could you check on this too?

vpickard:
@cybertron Can you please give ok-to-test on this PR? Thanks

cybertron (commented):
Looks like a good start. In my local tests the new logic seems to do the right thing.

We do need a followup, either in this PR or a separate one, to actually populate both VIPs when dual stack is in use. I think that's going to require the following things:

- Adding a second DNS entry for the api in vm_setup_vars.yaml (this may need to be conditionalized so we don't do it in single-stack deployments)
- Adding a dual stack conditional in network.sh that looks up the new api DNS entry and adds it to $API_VIP.
- Ditto for ingress, except it's a simple variable assignment there
- Add secondary entries to the NM dnsmasq config in network.sh

I'm not sure how much the unmanaged br bridge case gets used these days, so we can probably leave that side of the logic alone. If someone needs dual stack vip support there we can work with them to do that, but in the meantime I'm not sure we have a good way to test it anyway.

zhaozhanqi:
@cybertron can we merge this PR?

cybertron:
The current version of this patch is only a partial implementation. We still need to actually configure both VIPs when dev-scripts is run in dual stack mode. Right now you have to hack in overrides for the VIP variables.

mkowalski:
Drive-by review, but is there a chance of e.g. providing a sample of `config.sh` with a correct syntax to use dual-stack VIPs? It would really help other teams that use dev-scripts and don't have internal knowledge of the implementation

cybertron:
> Drive-by review, but is there a chance of e.g. providing a sample of `config.sh` with a correct syntax to use dual-stack VIPs? It would really help other teams that use dev-scripts and don't have internal knowledge of the implementation

The intent is that when this is complete you'll get dual VIPs any time you set IP_STACK=v4v6. The VIP variables aren't actually exposed as an interface so I don't think it's possible to set them manually.

dougsland:
tested locally, `ipv4 and dualstack` everything `working`.
@zhaozhanqi  @cybertron @creydr  ready for review and tests.

Thanks!

dougsland:
Please see the commit message for details about this feature enabler and testing.

dougsland:
Tested in the following scenarios and everything worked:
```
OPENSHIFT 4.11
    - v4   - [x]
    - v4v6 - [x]
    - v6   - [x]

OPENSHIFT 4.12
    - v4   - [x]
    - v4v6 - [x]
    - v6   - [x]
```

dougsland:
I don't believe this is related to the patch:
```
ERROR: After October 2020 you may experience errors when installing or updating packages. This is because pip will change the way that it resolves dependency conflicts.\n\nWe recommend you use --use-feature=2020-resolver to test your packages with the new resolver before it becomes the default.\n\nrequests-oauthlib 1.3.1 requires oauthlib>=3.0.0, but you'll have oauthlib 2.1.0 which is incompatible.\nflask-oauthlib 0.9.6 requires requests-oauthlib<1.2.0,>=0.6.2, but you'll have requests-oauthlib 1.3.1 which is incompatible.\n",
```

cybertron (changes requested):
Mostly just a few nitpicks, but I think we should fix the redundant version logic because that's going to be confusing. One of those if branches is dead code.

network.sh
```diff
@@ -193,26 +196,89 @@ if [ -n "${NETWORK_CONFIG_FOLDER:-}" ]; then
...
+
+    # if string is *NOT* NULL/empty
+    if [[ -n "${1}" ]]; then
+        echo "$1,$2"
+    else
+        echo "$2"
+    fi
+}
+
+function dig_ip() {
```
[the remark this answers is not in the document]
dougsland: /me nods. Probably good idea to rename the function.

network.sh
```diff
@@ -193,26 +196,89 @@ if [ -n "${NETWORK_CONFIG_FOLDER:-}" ]; then
...
+    #     Types available: apivip OR ingressvip
+    #
+    #     Second argument: The list (or single entry) of apivip or ingressvip
+    #
+    # Description:
+    #     Add entries into openshift-${CLUSTERNAME}.conf for dnsmasq
+    #
+    # Returns:
+    #     None
+    for i in $(echo "${2}" | sed "s/${STRINGS_SEPARATOR}/ /g")
```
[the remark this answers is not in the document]
dougsland: sure

ocp_install_env.sh
```diff
@@ -122,6 +122,58 @@ EOF
   fi
 }
 
+function renderVIPs() {
+    # Arguments:
+    #     First argument: type of data
```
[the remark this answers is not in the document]
dougsland: done

cybertron:
Waiting on ci, but this lgtm now.

dougsland:
> Waiting on ci, but this lgtm now.

I am also waiting all my local tests finish here, so far, so good.

dougsland:
CI error not related to patch:
```
    "requirements": null,
    "state": "present",
    "stderr": "WARNING: Running pip install with root privileges is generally not a good idea. Try `pip3 install --user` instead.\nERROR: After October 2020 you may experience errors when installing or updating packages. This is because pip will change the way that it resolves dependency conflicts.\n\nWe recommend you use --use-feature=2020-resolver to test your packages with the new resolver before it becomes the default.\n\nrequests-oauthlib 1.3.1 requires oauthlib>=3.0.0, but you'll have oauthlib 2.1.0 which is incompatible.\nflask-oauthlib 0.9.6 requires requests-oauthlib<1.2.0,>=0.6.2, but you'll have requests-oauthlib 1.3.1 which is incompatible.\n",
    "stderr_lines": [
        "WARNING: Running pip install with root privileges is generally not a good idea. Try `pip3 install --user` instead.",
        "ERROR: After October 2020 you may experience errors when installing or updating packages. This is because pip will change the way that it resolves dependency conflicts.",
        "",
        "We recommend you use --use-feature=2020-resolver to test your packages with the new resolver before it becomes the default.",
        "",
        "requests-oauthlib 1.3.1 requires oauthlib>=3.0.0, but you'll have oauthlib 2.1.0 which is incompatible.",
        "flask-oauthlib 0.9.6 requires requests-oauthlib<1.2.0,>=0.6.2, but you'll have requests-oauthlib 1.3.1 which is incompatible."
```

dougsland:
all local tests worked with the following config:

```
OPENSHIFT 4.11
    - v4   - [x]
    - v4v6 - [x]
    - v6   - [x]

OPENSHIFT 4.12
    - v4   - [x]
    - v4v6 - [x]
    - v6   - [x]
```

creydr (commented):
/lgtm

cybertron:
> /lgtm
> 
> Tested with `IP_STACK=v4v6`: 4.12: worked 4.11: Got error message, as expected with invalid install config (2 vips in apiVIP / ingressVIP field)

/hold

We can't break 4.11 dual stack with this patch. It looks like we need to make the version check smart enough to only include the first VIP on older releases.

dougsland:
already merged in another PR.

PR #1392, opened 2022-05-11
AGENT-164: Customize openshift installer checkout, build and extraction

andfasano:
This patch modifies the openshift installer functions so that it will be possible to reuse the `03_build_installer.sh` also for non-baremetal flows (it will be used by the agent flow in a subsequent patch).

In addition, it allows the `sync_repo_and_patch` to use a different branch (also this feature will be required by the agent flow, as currently is being developed in its own installer branch)

### DIFF
--- a/utils.sh
+++ b/utils.sh
@@ -266,16 +266,19 @@ function sync_repo_and_patch {
 
     pushd $DEST
 
-    DEFAULT_BRANCH=$(git remote show origin | sed -n '/HEAD branch/s/.*: //p')
+    REPO_NAME=$(basename "$1" .git | tr "a-z" "A-Z" | tr "-" "_")
+    REPO_BRANCH_VAR="${REPO_NAME}_REPO_BRANCH"
+    REPO_BRANCH=${!REPO_BRANCH_VAR:-$(git remote show origin | sed -n '/HEAD branch/s/.*: //p')}
 
     git am --abort || true
-    git checkout ${DEFAULT_BRANCH}
+    git checkout ${REPO_BRANCH}
     git fetch origin
-    git rebase origin/${DEFAULT_BRANCH}
+    git rebase origin/${REPO_BRANCH}
 
     # If set, use the specified PR number
-    if test "$#" -gt "2" ; then
-      REPO_PR=$3
+    REPO_PR_VAR="${REPO_NAME}_REPO_PR"
+    REPO_PR=${!REPO_PR_VAR:-}
+    if [[ ! -z "${REPO_PR}" ]] ; then
       echo "Fetching PR ${REPO_PR}"
       git fetch origin pull/${REPO_PR}/head:pr${REPO_PR}
       git checkout pr${REPO_PR}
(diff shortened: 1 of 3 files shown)


utils.sh
```diff
@@ -267,11 +267,12 @@ function sync_repo_and_patch {
     pushd $DEST
 
     DEFAULT_BRANCH=$(git remote show origin | sed -n '/HEAD branch/s/.*: //p')
+    REPO_BRANCH=${REPO_BRANCH:-$DEFAULT_BRANCH}
```
hardys, line 270: This assumes we can set REPO_BRANCH globally, which I don't think will always work - e.g we use this function to clone metal3-dev-env, hive and installer repos, where the required branch may be different?

Instead I think we need to support per-repo variables e.g similar to the suggestion [here](https://github.com/openshift-metal3/dev-scripts/pull/1372#discussion_r838318920)

Then we could do e.g `INSTALLER_REPO_BRANCH` `METAL3_DEV_ENV_REPO_BRANCH` etc

common.sh
```diff
@@ -187,8 +187,18 @@ if [ "${UPSTREAM_IRONIC:-false}" != "false" ] ; then
     export BAREMETAL_OPERATOR_LOCAL_IMAGE=${BAREMETAL_OPERATOR_LOCAL_IMAGE:-"quay.io/metal3-io/baremetal-operator"}
 fi
 
+# Default the build tags used when building from the source the OpenShift installer
+export OPENSHIFT_INSTALLER_BUILD_TAGS=${OPENSHIFT_INSTALLER_BUILD_TAGS:-"libvirt baremetal"}
+
+# Default installer command extracted
+export OPENSHIFT_INSTALLER_CMD=${OPENSHIFT_INSTALLER_CMD:-"openshift-baremetal-install"}
```
hardys, line 194: I wonder if these really need to be common - `build_installer` is only used in `03_build_installer.sh` so I wonder if it would be simpler/clearer if we moved the function into `03_build_installer.sh` and defined the build variables (and defaults) in that script directly?

hardys: Probably the same for `clone_installer` and `OPENSHIFT_INSTALLER_PR` - not necessarily blocking on that refactor, but it seems like this might be a good time to do it?

andfasano: Yep, I've moved everything but `OPENSHIFT_INSTALLER_CMD`, since `OPENSHIFT_INSTALLER` seems to be more widely used and complicated to move

pawanpinjarkar:
Tested with https://github.com/openshift-metal3/dev-scripts/pull/1393 and a compact cluster installation seems to be working.

andfasano:
https://github.com/openshift/installer/pull/5909 landed

/retest-required

hardys:
/approve
/override ci/prow/e2e-metal-ipi-proxy-ipv6
/override ci/prow/e2e-metal-ipi-proxy-ipv4

These jobs are failing, investigation/fixes ongoing but unrelated to this PR

PR #1364, opened 2022-03-11
Update assisted operator index image

carbonin:
The ocpmetal image is no longer being maintained or updated.
All assisted images should now be pulled from edge-infrastructure.

cc @dtantsur

PR #1394, opened 2022-05-11
Update ASSISTED_OPERATOR_INDEX to point to an updated index

javipolo:
The current default index for assisted-operator is too old and not updated anymore, so we should use a different one


mkowalski:
Duplicate of https://github.com/openshift-metal3/dev-scripts/pull/1364

javipolo:
Duplicate of https://github.com/openshift-metal3/dev-scripts/pull/1364

PR #1398, opened 2022-05-17
AGENT-51: Update config_example.sh to be more documented

lranjbar:
This updates the formatting of config_example.sh. I tried to keep the content the same as much as possible. I have made some Ansible that will generate a config_$USER.sh for you and this is what the generated config_$USER.sh looks like.

@zaneb @pawanpinjarkar @andfasano @rwsu @bfournie

### DIFF
--- a/config_example.sh
+++ b/config_example.sh
@@ -1,287 +1,564 @@
 #!/bin/bash
 
-# You can get this token from https://console-openshift-console.apps.ci.l2s4.p1.openshiftapps.com/ by
-# clicking on your name in the top right corner and coping the login
+################################################################################
+## Development Settings
+##
+
+# CI_TOKEN - **REQUIRED**
+# You can get this token from https://console-openshift-console.apps.ci.l2s4.p1.openshiftapps.com/
+# by clicking on your name in the top right corner and coping the login
 # command (the token is part of the command)
 set +x
 export CI_TOKEN=''
 set -x
 
+# CI_SERVER -
 # When testing e.g cluster-bot builds, you must change CI_TOKEN to one
 # from a different cluster, in which case this should be updated
+#
 #export CI_SERVER=api.ci.l2s4.p1.openshiftapps.com
 
-# Select a different release stream from which to pull the latest image, if the
-# image name is not specified
-#export OPENSHIFT_RELEASE_STREAM=4.8
+# WORKING_DIR -
+# By default, dev-scripts' working directory is set to `/opt/dev-scripts`.
+# Make sure that the filesystem has at least 80GB of free space: `df -h /`.
+# Alternatively you may have a large `/home` filesystem, in which case you can
+# `export WORKING_DIR=/home/dev-scripts` and the scripts will create this directory
+# with appropriate permissions. In the event you create this directory manually it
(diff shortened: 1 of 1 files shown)


config_example.sh
```diff
@@ -1,287 +1,743 @@
 #!/bin/bash
 
+####################################################################################################
+## Development Settings
+##
+
+################################################################################
```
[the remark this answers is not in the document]
rwsu: I concur. I find them a bit distracting.

config_example.sh
```diff
@@ -1,287 +1,743 @@
...
 # You can get this token from https://console-openshift-console.apps.ci.l2s4.p1.openshiftapps.com/ by
 # clicking on your name in the top right corner and coping the login
 # command (the token is part of the command)
 set +x
 export CI_TOKEN=''
 set -x
+################################################################################
 
+################################################################################
+# CI_SERVER -
```
rwsu, line 18: I like the variable name followed by description format change.

config_example.sh
```diff
@@ -1,287 +1,743 @@
...
+#
+# metallb container location (optional)
+#export METALLB_IMAGE_BASE=
+#export METALLB_IMAGE_TAG=
+#
+################################################################################
 
-# Provisioning interface on the helper ndoe
+####################################################################################################
+## Network Settings
```
rwsu, line 262: The subsections are a nice way to see the different types of configuration options.

rwsu (commented):
I like the change overall. Like Zane suggested I would suggest removing the long "###" bars for variables and leave them for the sections.

zaneb:
/lgtm
Thanks Lisa!

rwsu (commented):
/lgtm

Nice improvement.

config_example.sh
```diff
@@ -1,287 +1,545 @@
...
+# DNS_VIP (Defaults to v6)
+#
+# v4 Default:
 #export DNS_VIP="11.0.0.2"
+#
+# v6 Default:
+#export DNS_VIP="fd2e:6f44:5dd8:c956::2"
 
-# Network type
+# Network type (Defaults to v6)
```
hardys, line 242: The `defaults to v6` here and elsewhere is slightly misleading - the values for `NETWORK_TYPE`, and subnet/VIP values are [defaulted depending on the value](https://github.com/openshift-metal3/dev-scripts/blob/master/network.sh#L104..L130) of `IP_STACK`

hardys, line 242: The `defaults to v6` here and elsewhere is slightly misleading - the values for `NETWORK_TYPE`, and subnet/VIP values are [defaulted depending on the value](https://github.com/openshift-metal3/dev-scripts/blob/master/network.sh#L104..L130) of `IP_STACK`

hardys:
One small nit mentioned which would be nice to fix, otherwise lgtm, thanks!

hardys:
One small nit mentioned which would be nice to fix, otherwise lgtm, thanks!

config_example.sh
lranjbar: That's true these values are based off `IP_STACK` and `HOST_IP_STACK`. 🤔 I will update the wording.

zaneb:
@lranjbar stop checking GitHub notifications when you are on holiday.

/approve

PR #1402, opened 2022-05-26
rhcos.sh: support FCOS artifacts

vrutkovs:
For OKD IPI we use FCOS as a base OS. However its artifacts are .xz packed.
This ensures both RHCOS .gz and FCOS .xz are supported

### DIFF
--- a/rhcos.sh
+++ b/rhcos.sh
@@ -1,19 +1,20 @@
 if $OPENSHIFT_INSTALLER coreos print-stream-json >/dev/null 2>&1; then
     $OPENSHIFT_INSTALLER coreos print-stream-json > $OCP_DIR/rhcos.json
-    export MACHINE_OS_INSTALLER_IMAGE_URL=$(jq -r .architectures.$(uname -m).artifacts.openstack.formats[\"qcow2.gz\"].disk.location $OCP_DIR/rhcos.json)
-    export MACHINE_OS_INSTALLER_IMAGE_SHA256=$(jq -r .architectures.$(uname -m).artifacts.openstack.formats[\"qcow2.gz\"].disk[\"sha256\"] $OCP_DIR/rhcos.json)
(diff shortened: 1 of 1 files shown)


vrutkovs:
/cc @elfosardo 
/cc @dtantsur 

This is required to test OKD installs via BM IPI

rhcos.sh
```diff
@@ -1,19 +1,20 @@
 if $OPENSHIFT_INSTALLER coreos print-stream-json >/dev/null 2>&1; then
     $OPENSHIFT_INSTALLER coreos print-stream-json > $OCP_DIR/rhcos.json
-    export MACHINE_OS_INSTALLER_IMAGE_URL=$(jq -r .architectures.$(uname -m).artifacts.openstack.formats[\"qcow2.gz\"].disk.location $OCP_DIR/rhcos.json)
-    export MACHINE_OS_INSTALLER_IMAGE_SHA256=$(jq -r .architectures.$(uname -m).artifacts.openstack.formats[\"qcow2.gz\"].disk[\"sha256\"] $OCP_DIR/rhcos.json)
+    TOP_LEVEL_FORMAT="$(jq -r '.architectures.x86_64.artifacts.openstack.formats | keys[]' $OCP_DIR/rhcos.json | head -n1)"
+    export MACHINE_OS_INSTALLER_IMAGE_URL=$(jq -r .architectures.$(uname -m).artifacts.openstack.formats[\"${TOP_LEVEL_FORMAT}\"].disk.location $OCP_DIR/rhcos.json)
```
dtantsur, line 4: Given that you're using the first key from `formats`, cannot you just pick the first value?

vrutkovs: CI is using older  `jq` which doesn't support indexes (attempted in https://github.com/openshift-metal3/dev-scripts/commit/cf60a53b2c4ba513b7d782f386bc8cb7684134cd), so I went with uglier-but-works way

paul-maidment:
/lgtm
nice removal of hard wiring!

mkowalski (approved):

PR #1408, opened 2022-05-31
Agent: use OpenShift version extracted from the release payload

andfasano:
Configure the `ClusterImageSet` manifest using the version fetched from the downloaded release payload

### DIFF
--- a/agent/04_agent_configure.sh
+++ b/agent/04_agent_configure.sh
@@ -10,6 +10,7 @@ source $SCRIPTDIR/network.sh
 source $SCRIPTDIR/utils.sh
 source $SCRIPTDIR/validation.sh
 source $SCRIPTDIR/agent/common.sh
+source $SCRIPTDIR/ocp_install_env.sh
 
 early_deploy_validation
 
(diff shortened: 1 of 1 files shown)


zaneb:
I don't think anything looks at the name (even in principle they only have to match), but this seems fine.

zaneb:
/hold
This will break my hacky workaround to fix CI in openshift/installer#5954, so we shouldn't merge it until CI is fixed properly in https://github.com/openshift/installer/pull/5949.

zaneb:
/hold cancel
I tested and it appears nothing looks at the name at all as of now.

agent/04_agent_configure.sh
```diff
@@ -64,6 +64,9 @@ function generate_cluster_manifests() {
   # To be change in ${OCP_DIR}/cluster-manifests
   MANIFESTS_PATH="${OCP_DIR}/manifests"
 
+  # Fetch current OpenShift version from the release payload
+  VERSION="$(openshift_version ${OCP_DIR})"
```
rwsu, line 68: I'm seeing an error in my local test of "make agent":
```
++(./agent/04_agent_configure.sh:68): generate_cluster_manifests(): openshift_version ocp/ostest
./agent/04_agent_configure.sh: line 68: openshift_version: command not found

```
I think this needs to be added to the beginning of the file to include the openshift_version function:
`source $SCRIPTDIR/ocp_install_env.sh`

rwsu (changes requested):

agent/04_agent_configure.sh
andfasano: Thanks @rwsu

PR #1410, opened 2022-06-03
AGENT-165: revert PR1404 to re-enable release image hard-coding

rwsu:
The installer code has reverted back to hard-coding in:
https://github.com/openshift/installer/pull/5954/

Revert dev-scripts back to hard-coding so that the release
image versions are consistent. Reverts #1404.

Without revert, local agent deployments are failing because
the ClusterImageSet is set with spec.releaseImage:

registry.ci.openshift.org/ocp/release:4.11.0-0.nightly-2022-06-01-200905

while the assisted-service RELEASE_IMAGES env var has:

quay.io/openshift-release-dev/ocp-release:4.10.0-rc.1-x86_64

The mismatch causes create-cluster-and-infraenv.service to fail.

### DIFF
--- a/agent/04_agent_configure.sh
+++ b/agent/04_agent_configure.sh
@@ -87,7 +87,7 @@ spec:
   clusterDeploymentRef:
     name: ${CLUSTER_NAME}
   imageSetRef:
-    name: openshift-${OPENSHIFT_RELEASE_STREAM}
+    name: openshift-v4.10.0
   networking:
     clusterNetwork:
     - cidr: ${CLUSTER_NETWORK}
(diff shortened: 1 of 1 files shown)


lranjbar:
/lgtm

It seems we are going to need to make a task to unravel this issue after our initial builds are complete.

andfasano:
/hold

The current problem should be fixed by https://github.com/openshift/installer/pull/5949

zaneb:
I'd prefer not to revert this, because any revert of https://github.com/openshift/installer/pull/5954 can't be tested in CI without this. (Even before openshift/installer#5954 it was working fine locally, so it's really in CI that we have to test it.)

rwsu:
Closing. Issue fixed by https://github.com/openshift/installer/pull/5949

PR #1411, opened 2022-06-08
Agent: remove workarounds for old file locations

zaneb:
Since openshift/installer#5975:
ZTP manifests are now read from cluster-manifests.
The agent.iso output is now in the work directory directly.


rwsu (commented):
/lgtm

PR #1414, opened 2022-06-22
Ensure bootstrap_ip doesn't fail

derekhiggins:
Force the command to get the bootstrap ip to succeed.

If running a DHCP server outside of dev-scripts
(with MANAGE_BR_BRIDGE=n) then the net-dhcp-leases command
will return nothing, because we have "set -o pipefail" this
causes the whole script to stall.

### DIFF
--- a/utils.sh
+++ b/utils.sh
@@ -552,7 +552,7 @@ function bootstrap_ip {
                       | grep "${pref_ip}" \
                       | tail -n1 \
                       | awk '{print $5}' \
-                      | sed -e 's/\(.*\)\/.*/\1/'
+                      | sed -e 's/\(.*\)\/.*/\1/' || true
   else
     echo "Unable to retrieve bootstrap IP with infinite leases enabled." 1>&2
   fi


utils.sh
```diff
@@ -552,7 +552,7 @@ function bootstrap_ip {
                       | grep "${pref_ip}" \
                       | tail -n1 \
                       | awk '{print $5}' \
-                      | sed -e 's/\(.*\)\/.*/\1/'
+                      | sed -e 's/\(.*\)\/.*/\1/' || true
```
cybertron, line 555: If we're doing this we could arguably remove the DHCP_LEASE_EXPIRY check too. We may actually have a new enough libvirt that we don't need to skip this anymore.

cybertron (commented):
/lgtm

One possible additional change inline, but I could push that as a followup too.

utils.sh
hardys: Sounds like we could handle that as a followup since it's not strictly related to this change?

PR #1417, opened 2022-06-23
AGENT-40: Use output kubeconfig to check for cluster complete

zaneb:


agent/05_agent_create_cluster.sh
```diff
@@ -54,7 +54,7 @@ function wait_for_cluster_ready() {
   sleep 5m
```
andfasano, line 54: I think that also this block could be removed, given that we don't have to wait anymore for the kubeconfig availabiity

zaneb: I haven't tried but if we start oc before the bootstrap control plane is up, won't it fail?

andfasano: Yes, you're right, we could keep the above block as a poor-man wait-for until the installer command will be ready

PR #1419, opened 2022-07-01
Power BM hosts down

derekhiggins:
BM hosts need to be powered down so and old
cluster doesn't clash with the new one being
installed

Also change the config example of "NODES_PLATFORM"
to match what is in the README.

### DIFF
--- a/02_configure_host.sh
+++ b/02_configure_host.sh
@@ -36,6 +36,24 @@ if [ ! -z "${VM_NODES_FILE}" ]; then
   exit 1
 fi
 
+# If deploying to a real baremetal environment, we need to make
+# sure all nodes are powered down
+if [[ "${NODES_PLATFORM}" == "baremetal" ]] && [ -n "$NODES_FILE" ] ; then
+    set +x
+    cat $NODES_FILE | jq -r '.nodes[].driver_info | ( .address + " " + .username + " " + .password ) ' |
+    while read ADDRESS USER PASSWORD ; do
+        if [[ $ADDRESS =~ ^ipmi://([^:]*)(:([0-9]+))?$ ]] ; then
+            IPMIIP=${BASH_REMATCH[1]}
+            IPMIPORT=${BASH_REMATCH[3]:-623}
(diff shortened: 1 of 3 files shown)


02_configure_host.sh
```diff
@@ -36,6 +36,23 @@ if [ ! -z "${VM_NODES_FILE}" ]; then
...
 
+# If deploying to a real baremetal environment, we need to make
+# sure all nodes are powered down
+if [[ "${NODES_PLATFORM}" == "baremetal" ]] && [ -n "$NODES_FILE" ] ; then
+    set +x
+    cat $NODES_FILE | jq -r '.nodes[].driver_info | ( .address + " " + .username + " " + .password ) ' |
+    while read ADDRESS USER PASSWORD ; do
+        if [[ $ADDRESS =~ ^ipmi://(.*)$ ]] ; then
+            IPMIIP=${BASH_REMATCH[1]}
+            ipmitool -I lanplus -H $IPMIIP -U $USER -P $PASSWORD power off || echo "WARNING: Skipping power down of $ADDRESS"
```
andfasano, line 47: Maybe in case of error it could be useful to print also the error code

derekhiggins: done

02_configure_host.sh
```diff
@@ -36,6 +36,23 @@ if [ ! -z "${VM_NODES_FILE}" ]; then
...
 fi
 
+# If deploying to a real baremetal environment, we need to make
+# sure all nodes are powered down
+if [[ "${NODES_PLATFORM}" == "baremetal" ]] && [ -n "$NODES_FILE" ] ; then
+    set +x
+    cat $NODES_FILE | jq -r '.nodes[].driver_info | ( .address + " " + .username + " " + .password ) ' |
+    while read ADDRESS USER PASSWORD ; do
+        if [[ $ADDRESS =~ ^ipmi://(.*)$ ]] ; then
+            IPMIIP=${BASH_REMATCH[1]}
```
dtantsur, line 46: This may contain a port, I think

derekhiggins: thanks, fixed

PR #1424, opened 2022-07-08
AGENT-287: For agent mirror testing force disconnect from masters

bfournie:
This simulates a disconnect from the masters when using the agent with mirroring. It sets a bogus /etc/hosts entry so that the local mirror must be used.  This is currently needed since we must have a valid connection to create the mirror and create the iso so we can't break the connection until after that's done.

### DIFF
--- a/agent/05_agent_create_cluster.sh
+++ b/agent/05_agent_create_cluster.sh
@@ -39,6 +39,32 @@ function attach_agent_iso() {
     done
 }
 
+function force_mirror_disconnect() {
+
+  # Set a bogus entry in /etc/hosts on all masters to force the local mirror to be used
+  node0_name=$(printf ${MASTER_HOSTNAME_FORMAT} 0)
+  node0_ip=$(sudo virsh net-dumpxml ostestbm | xmllint --xpath "string(//dns[*]/host/hostname[. = '${node0_name}']/../@ip)" -)
+  ssh_opts=(-o 'StrictHostKeyChecking=no' -q core@${node0_ip})
+
+  for (( n=0; n<${NUM_MASTERS}; n++ ))
+  do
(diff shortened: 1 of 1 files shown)


pawanpinjarkar (commented):
/lgtm

Can you also mention the Jira number in the commit message and PR title?

agent/05_agent_create_cluster.sh
```diff
@@ -39,6 +39,31 @@ function attach_agent_iso() {
...
+     ssh_opts=(-o 'StrictHostKeyChecking=no' -q core@${node_ip})
+
+     until ssh "${ssh_opts[@]}" "[[ -f /etc/hosts ]]"
+     do
+       echo "Waiting for $node_name to set remote host disconnect "
+       sleep 30s;
+     done
+
+     # Set a bogus entry in /etc/hosts to break remote access
+     ssh "${ssh_opts[@]}" "echo '125.12.15.15 quay.io' | sudo tee -a /etc/hosts "
```
rwsu, line 62: The mirroring command I see from dev-scripts is:
`oc adm release mirror --insecure=true -a /opt/dev-scripts/pull_secret.json --from registry.ci.openshift.org/ocp/release:4.11.0-0.nightly-2022-07-13-131410 --to-release-image virthost.ostest.test.metalkube.org:5000/localimages/local-release-image:4.11.0-0.nightly-2022-07-13-131410 --to virthost.ostest.test.metalkube.org:5000/localimages/local-release-image 
`
Shouldn't there be an entry for registry.ci.openshift.org?

bfournie: Yes there should be an entry for registry also, added it.

rwsu (commented):
/lgtm

PR #1425, opened 2022-07-13
AGENT-283: Remove unwanted ClusterDeployment.Spec fields

pawanpinjarkar:
Remove the unwanted fields from the ClusterDeployment.Spec.

### DIFF
--- a/agent/04_agent_configure.sh
+++ b/agent/04_agent_configure.sh
@@ -128,13 +128,6 @@ spec:
     name: test-agent-cluster-install
     version: v1beta1
   clusterName: ${CLUSTER_NAME}
-  controlPlaneConfig:
-    servingCertificates: {}
-  platform:
-    agentBareMetal:
-      agentSelector:
-        matchLabels:
-          bla: aaa
   pullSecretRef:
     name: pull-secret
 EOF


agent/04_agent_configure.sh
```diff
@@ -122,21 +122,7 @@ metadata:
   namespace: cluster0
 spec:
   baseDomain: ${BASE_DOMAIN}
-  clusterInstallRef:
```
zaneb, line 125: This is definitely used: https://github.com/openshift/assisted-service/blob/da8164f04040feae88de132de4cfe02447129dde/internal/controller/controllers/clusterdeployments_controller.go#L140-L146

agent/04_agent_configure.sh
```diff
@@ -122,21 +122,7 @@ metadata:
...
-    group: extensions.hive.openshift.io
-    kind: AgentClusterInstall
-    name: test-agent-cluster-install
-    version: v1beta1
   clusterName: ${CLUSTER_NAME}
-  controlPlaneConfig:
-    servingCertificates: {}
-  platform:
```
zaneb, line 133: Platform is used in the Hive controller (not sure how this interacts with the MCE controller?): https://github.com/openshift/hive/blob/master/pkg/controller/clusterdeployment/clusterdeployment_controller.go
I think we should probably set it to `none: {}` until further notice.

zaneb: actually no, let's remove it for now.

rwsu (approved):
/approve

PR #1426, opened 2022-07-18
Install the yq util system-wise

elfosardo:
This is to avoid the system not finding the yq binary when using
recent pip versions.

### DIFF
--- a/01_install_requirements.sh
+++ b/01_install_requirements.sh
@@ -100,7 +100,7 @@ fi
 
 # We use yq in a few places for processing YAML but it isn't packaged
 # for CentOS/RHEL so we have to install from pip.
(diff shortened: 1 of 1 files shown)


01_install_requirements.sh
```diff
@@ -100,7 +100,7 @@ fi
 
 # We use yq in a few places for processing YAML but it isn't packaged
 # for CentOS/RHEL so we have to install from pip.
-pip3 install --user 'yq>=2.10.0'
+pip3 install 'yq>=2.10.0'
```
derekhiggins, line 103: Will this now need to be run with sudo ?

elfosardo:
/retest
gateway issues lol

PR #1430, opened 2022-07-25
AGENT-174: Use wait-for commands to wait for bootstrap

zaneb:
Use the new wait-for validations-ok (AGENT-172) and wait-for
bootstrap-complete (AGENT-174) to determine when the bootstrap is
complete. This should give us more debug information than we have had
hitherto.


agent/05_agent_create_cluster.sh
```diff
@@ -71,13 +67,13 @@ function wait_for_cluster_ready() {
...
 
-  until ssh "${ssh_opts[@]}" "[[ -f /var/lib/kubelet/kubeconfig ]]" 
-  do 
-    echo "Waiting for bootstrap... "
-    sleep 1m; 
-  done
-
-  sleep 5m
+  local openshift_install="$(realpath "${OCP_DIR}/openshift-install")"
+  if ! "${openshift_install}" --dir="${OCP_DIR}" --log-level=debug agent wait-for validations-ok; then
```
andfasano, line 71: we don't have an explicit command for the validations the check it's embedded directly within the `bootstrap-complete` one

PR #1432, opened 2022-07-27
Fail if not CentOS Stream 8

elfosardo:
We don't support CentOS 8 since a while, fail if it's not CentOS Stream 8


elfosardo:
/hold
this needs https://github.com/openshift-metal3/dev-scripts/pull/1348 plus a change in the CI config

elfosardo:
we also need https://github.com/openshift/release/pull/31636

01_install_requirements.sh
```diff
@@ -50,9 +50,10 @@ case $DISTRO in
     # install network-scripts package to be able to use legacy network commands
     sudo dnf install -y network-scripts
     if [[ $DISTRO == "centos8" ]]; then
-      echo "CentOS is not supported anymore. Please switch to CentOS Stream / RHEL / Rocky Linux"
-    fi
-    if [[ $DISTRO == "centos8" || $DISTRO == "almalinux8" || $DISTRO == "rocky8" ]]; then
```
cybertron, line 55: I think we lost the rocky and alma part of this conditional.

cybertron (changes requested):

01_install_requirements.sh
elfosardo: yep, lost it during the rebase.... thanks!

cybertron (commented):
/lgtm

PR #1435, opened 2022-08-10
AGENT-316: Add support for oc mirror

bfournie:
Add a new variable - OC_MIRROR to mirror images using the `oc mirror` command.  If OC_MIRROR is not set then the mirroring will be done using 'oc adm release mirror'.  In addition a new variable REGISTRY_BACKEND is used to determine hw the registry is set, either the legacy way by creating a podman registry, or via the mirror-registry command.

Note that OC_MIRROR is currently only supported for the Agent implementation.

### DIFF
--- a/agent/01_agent_requirements.sh
+++ b/agent/01_agent_requirements.sh
@@ -4,12 +4,50 @@ set -o pipefail
 
 SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )/.." && pwd )"
 
-source $SCRIPTDIR/validation.sh
 source $SCRIPTDIR/common.sh
+source $SCRIPTDIR/agent/common.sh
+source $SCRIPTDIR/utils.sh
+source $SCRIPTDIR/validation.sh
 
 early_deploy_validation
 
 if [[ -z ${AGENT_E2E_TEST_SCENARIO} ]]; then
     printf "\nAGENT_E2E_TEST_SCENARIO is missing or empty. Did you forget to set the AGENT_E2E_TEST_SCENARIO env var in the config_<USER>.sh file?"
     invalidAgentValue
+fi
+
+if [[ ${REGISTRY_BACKEND} = "quay" ]]; then
+
+   mkdir -p ${WORKING_DIR}/mirror-registry
+   pushd ${WORKING_DIR}/mirror-registry
+   # run the exec in this dir as execution-environment.tar is also needed
+   mirror_registry_file=mirror-registry.tar.gz
+   mirror_registry_exec=${mirror_registry_file%%.*}
+   if [[ ! -f "./${mirror_registry_exec}" ]]; then
+      curl -O -L https://developers.redhat.com/content-gateway/rest/mirror/pub/openshift-v4/clients/mirror-registry/latest/${mirror_registry_file}
+      tar xzf ${mirror_registry_file}
+      chmod +x ${mirror_registry_exec}
+      rm -f ${mirror_registry_file}
+   fi
+   popd
+
+fi
+
+if [[ ! -z "${OC_MIRROR}" ]] && [[ "${OC_MIRROR}" == true ]]; then
+
+   oc_mirror_file=oc-mirror.tar.gz
+   oc_mirror_exec=${oc_mirror_file%%.*}
+   if [[ ! -f "/usr/local/bin/${oc_mirror_exec}" ]]; then
+      curl -O -L https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp/stable/${oc_mirror_file}
+      tar xzf ${oc_mirror_file}
+      chmod +x ${oc_mirror_exec}
+      sudo mv -f ${oc_mirror_exec} /usr/local/bin
+      rm -f ${oc_mirror_file}
+   fi
+
+   # set up the channel using the most recent candidate release
+   release_candidate=`oc mirror list releases --channel=candidate-${OPENSHIFT_RELEASE_STREAM} | tail -1`
+   export OPENSHIFT_RELEASE_TAG="${release_candidate}-x86_64"
+   export OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE="${LOCAL_REGISTRY_DNS_NAME}:${LOCAL_REGISTRY_PORT}/${MIRROR_IMAGE_URL_SUFFIX}:${OPENSHIFT_RELEASE_TAG}"
+
 fi
--- a/agent/04_agent_configure.sh
+++ b/agent/04_agent_configure.sh
@@ -11,6 +11,7 @@ source $SCRIPTDIR/utils.sh
 source $SCRIPTDIR/validation.sh
 source $SCRIPTDIR/agent/common.sh
 source $SCRIPTDIR/ocp_install_env.sh
+source $SCRIPTDIR/oc_mirror.sh
 
 early_deploy_validation
 
@@ -272,16 +273,18 @@ spec:
 EOF
 
 if [ ! -z "${MIRROR_IMAGES}" ]; then
-# TODO - get the mirror registry info from output of 'oc adm release mirror'
 
-    cat > "${MIRROR_PATH}/registries.conf" << EOF
+   # Set up registries.conf and ca-bundle.crt for mirroring
+   if [[ -z "${OC_MIRROR}" ]] ; then
+
+      cat > "${MIRROR_PATH}/registries.conf" << EOF
 [[registry]]
 prefix = ""
 location = "registry.ci.openshift.org/ocp/release"
 mirror-by-digest-only = false
 
 [[registry.mirror]]
-location = "${LOCAL_REGISTRY_DNS_NAME}:${LOCAL_REGISTRY_PORT}/localimages/local-release-image"
+location = "${LOCAL_REGISTRY_DNS_NAME}:${LOCAL_REGISTRY_PORT}/${MIRROR_IMAGE_URL_SUFFIX}"
 
 [[registry]]
 prefix = ""
@@ -289,11 +292,43 @@ location = "quay.io/openshift-release-dev/ocp-v4.0-art-dev"
 mirror-by-digest-only = false
 
 [[registry.mirror]]
-location = "${LOCAL_REGISTRY_DNS_NAME}:${LOCAL_REGISTRY_PORT}/localimages/local-release-image"
+location = "${LOCAL_REGISTRY_DNS_NAME}:${LOCAL_REGISTRY_PORT}/${MIRROR_IMAGE_URL_SUFFIX}"
 EOF
+   else
 
-cp $REGISTRY_DIR/certs/$REGISTRY_CRT ${MIRROR_PATH}/ca-bundle.crt
+      cat > "${MIRROR_PATH}/registries.conf" << EOF
+[[registry]]
+prefix = ""
+location = "quay.io/openshift-release-dev/ocp-release"
+mirror-by-digest-only = false
 
+[[registry.mirror]]
+location = "${LOCAL_REGISTRY_DNS_NAME}:${LOCAL_REGISTRY_PORT}/openshift/release-images"
+
+[[registry]]
+prefix = ""
+location = "quay.io/openshift-release-dev/ocp-v4.0-art-dev"
+mirror-by-digest-only = false
+
+[[registry.mirror]]
+location = "${LOCAL_REGISTRY_DNS_NAME}:${LOCAL_REGISTRY_PORT}/openshift/release"
+
+[[registry]]
+prefix = ""
+location = "registry.redhat.io/ubi8"
+mirror-by-digest-only = false
+
+[[registry.mirror]]
+location = "${LOCAL_REGISTRY_DNS_NAME}:${LOCAL_REGISTRY_PORT}/ubi8"
+EOF
+   fi
+
+   # Store the certs for registry
+   if [[ "${REGISTRY_BACKEND}" = "podman" ]]; then
+      cp $REGISTRY_DIR/certs/$REGISTRY_CRT ${MIRROR_PATH}/ca-bundle.crt
+   else
+      cp ${WORKING_DIR}/quay-install/quay-rootCA/rootCA.pem ${MIRROR_PATH}/ca-bundle.crt
+   fi
 fi
 
     cat > "${MANIFESTS_PATH}/infraenv.yaml" << EOF
@@ -481,9 +516,9 @@ get_static_ips_and_macs
 
 if [ ! -z "${MIRROR_IMAGES}" ]; then
 
-  setup_local_registry
+     setup_local_registry
 
-  setup_release_mirror
+     setup_release_mirror
 
 fi
 
--- a/agent/common.sh
+++ b/agent/common.sh
@@ -11,8 +11,14 @@ export CLUSTER_HOST_PREFIX_V4="24"
 export EXTERNAL_SUBNET_V6="fd2e:6f44:5dd8:c956::/64"
 
 if [ -n "$MIRROR_IMAGES" ]; then
-    # We're going to be using a locally modified release image
-    export OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE="${LOCAL_REGISTRY_DNS_NAME}:${LOCAL_REGISTRY_PORT}/localimages/local-release-image:${OPENSHIFT_RELEASE_TAG}"
+
+    export MIRROR_IMAGE_URL_SUFFIX="localimages/local-release-image"
+
+    if [ "${OC_MIRROR}" == true ]; then
+        export MIRROR_IMAGE_URL_SUFFIX="openshift/release-images"
+    fi
+
+    export OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE="${LOCAL_REGISTRY_DNS_NAME}:${LOCAL_REGISTRY_PORT}/${MIRROR_IMAGE_URL_SUFFIX}:${OPENSHIFT_RELEASE_TAG}"
 fi
 
 function getReleaseImage() {
--- a/common.sh
+++ b/common.sh
@@ -74,6 +74,12 @@ export SSH_PUB_KEY="${SSH_PUB_KEY:-$(cat $HOME/.ssh/id_rsa.pub)}"
 # mirror images for installation in restricted network
 export MIRROR_IMAGES=${MIRROR_IMAGES:-}
 
+# use oc-mirror command to mirror images
+export OC_MIRROR=${OC_MIRROR:-}
+
+# file containing auths for oc-mirror
+export DOCKER_CONFIG_FILE=${DOCKER_CONFIG_FILE:-$HOME/.docker/config.json}
+
 # Setup up a local proxy for installation
 export INSTALLER_PROXY=${INSTALLER_PROXY:-}
 export INSTALLER_PROXY_PORT=${INSTALLER_PROXY_PORT:-8215}
@@ -98,6 +104,10 @@ export REGISTRY_PASS=${REGISTRY_PASS:-ocp-pass}
 export REGISTRY_DIR=${REGISTRY_DIR:-$WORKING_DIR/registry}
 export REGISTRY_CREDS=${REGISTRY_CREDS:-$HOME/private-mirror-${CLUSTER_NAME}.json}
 export REGISTRY_CRT=registry.2.crt
+export REGISTRY_BACKEND=${REGISTRY_BACKEND:-"podman"}
+if [[ ${REGISTRY_BACKEND} = "quay" ]]; then
+    export LOCAL_REGISTRY_PORT=8443
+fi
 
 # Set this variable to build the installer from source
 export KNI_INSTALL_FROM_GIT=${KNI_INSTALL_FROM_GIT:-}
@@ -203,9 +213,14 @@ if env | grep -q "_LOCAL_IMAGE=" ; then
     export MIRROR_IMAGES=true
 fi
 
-if [ -n "$MIRROR_IMAGES" ]; then
-    # We're going to be using a locally modified release image
-    export OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE="${LOCAL_REGISTRY_DNS_NAME}:${LOCAL_REGISTRY_PORT}/localimages/local-release-image:latest"
+if [[ ! -z "${OC_MIRROR}" ]] && [[ "${OC_MIRROR}" == "true" ]]; then
+    export MIRROR_IMAGES=true
+
+    if [[ -x "/usr/local/bin/oc-mirror" ]]; then
+       # set up the channel using the most recent candidate release
+       release_candidate=`oc-mirror list releases --channel=candidate-${OPENSHIFT_RELEASE_STREAM} | tail -1`
+       export OPENSHIFT_RELEASE_TAG="${release_candidate}-x86_64"
+    fi
 fi
 
 # Set variables
@@ -428,3 +443,8 @@ if [[ ! -z ${AGENT_E2E_TEST_SCENARIO} ]]; then
   fi
 fi
 
+# For IPv6 (default case) mirror images are used since quay doesn't support IPv6
+if [[ -n "$MIRROR_IMAGES" || -z "$IP_STACK" || "$IP_STACK" = "v6" ]]; then
+   # We're going to be using a locally modified release image
+   export OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE="${LOCAL_REGISTRY_DNS_NAME}:${LOCAL_REGISTRY_PORT}/localimages/local-release-image:latest"
+fi
--- a/config_example.sh
+++ b/config_example.sh
@@ -513,6 +513,15 @@ set -x
 #
 #export ENABLE_LOCAL_REGISTRY=true
 
+# REGISTRY_BACKEND
+# Define the local registry to use when ENABLE_LOCAL_REGISTRY=true
+# The default (legacy) registry is started by podman
+# The other available backend is quay which is installed by the
+# mirror-registry application
+# Default: podman
+# Choices: podman, quay
+# export REGISTRY_BACKEND=podman
+
 # LOCAL_REGISTRY_DNS_NAME -
 # Local image registry DNS name.
 # Default: "virthost.$CLUSTER_NAME.test.metalkube.org"
@@ -549,6 +558,19 @@ set -x
 # Default: true
 #
 #export MIRROR_IMAGES=true
+
+# OC_MIRROR -
+# When MIRROR_IMAGES is true, use the 'mirror-registry' command to create the
+# registry and the 'oc mirror' command to do the mirroring.
+# Note, this is currently only supported for the Agent implementation.
+# Default: false
+#export OC_MIRROR=false
+
+# When using OC_MIRROR the auths for the mirror will be added to DOCKER_CONFIG_FILE
+# or an UNAUTHORIZED error will result when attempting to use it. An example entry
+# in this file is:
+# "virthost.ostest.test.metalkube.org:8443": { "auth": "c3RhY2s6cXVheXBhc3N3b3Jk" },
+#export DOCKER_CONFIG_FILE=$HOME/.docker/config.json
 
 # MIRROR_OLM -
 # Comma-separated list of OLM operators to mirror into the local registry. This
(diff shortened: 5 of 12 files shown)


agent/01_agent_requirements.sh
```diff
@@ -1,15 +1,44 @@
 #!/bin/bash
 
-set -o pipefail
+# set -o pipefail
```
elfosardo, line 3: why removing the pipefail?
I would also add unbound variables check if possible

agent/01_agent_requirements.sh
```diff
@@ -1,15 +1,44 @@
...
+   mkdir -p ${WORKING_DIR}/mirror-registry
+   pushd ${WORKING_DIR}/mirror-registry
+   # run the exec in this dir as execution-environment.tar is also needed
+   mirror_registry_file=mirror-registry.tar.gz
+   mirror_registry_exec=${mirror_registry_file%%.*}
+   if [[ ! -f "./${mirror_registry_exec}" ]]; then
+      curl -O -L https://developers.redhat.com/content-gateway/rest/mirror/pub/openshift-v4/clients/mirror-registry/latest/${mirror_registry_file}
+      tar xzf ${mirror_registry_file}
+      chmod +x ${mirror_registry_exec}
+      rm ${mirror_registry_file}
```
elfosardo, line 31: maybe force the removal?

agent/01_agent_requirements.sh
```diff
@@ -1,15 +1,44 @@
...
+   popd
+
+   oc_mirror_file=oc-mirror.tar.gz
+   oc_mirror_exec=${oc_mirror_file%%.*}
+   if [[ ! -f "/usr/local/bin/${oc_mirror_exec}" ]]; then
+      curl -O -L https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp/stable/${oc_mirror_file}
+      tar xzf ${oc_mirror_file}
+      chmod +x ${oc_mirror_exec}
+      sudo mv -f ${oc_mirror_exec} /usr/local/bin
+      rm ${oc_mirror_file}
```
elfosardo, line 42: ditto

oc_mirror.sh
```diff
@@ -0,0 +1,71 @@
+#!/usr/bin/env bash
+set -x
```
elfosardo, line 2: maybe add at  least -e ?

elfosardo (changes requested):

agent/01_agent_requirements.sh
bfournie: I forget why that was done for debug, fixed.

agent/01_agent_requirements.sh
bfournie: Added the force

agent/01_agent_requirements.sh
bfournie: Added the force

oc_mirror.sh
bfournie: Added

rwsu (changes requested):
I hit this error:
```
+(/home/rwsu/go/src/github.com/openshift-metal3/dev-scripts/oc_mirror.sh:71): publish_mirror_registry(): oc mirror --dest-skip-tls --config /tmp/imageset--3jybQ2i4qI docker://virthost.ostest.test.metalkube.org:8443 --verbose 10
Checking push permissions for virthost.ostest.test.metalkube.org:8443
Using image virthost.ostest.test.metalkube.org:8443/oc-mirror to check permissions
error: config GVK not recognized: mirror.openshift.io/v1alpha1, Kind=ImageSetConfiguration
```

oc_mirror.sh
```diff
@@ -0,0 +1,71 @@
...
+  ocp:
+    channels:
+    - name: candidate-${OPENSHIFT_RELEASE_STREAM}
+  additionalImages:
+  - name: registry.redhat.io/ubi8/ubi:latest
+EOF
+
+   # TODO - print message on failure that password for must be added to .docker/config.json
+   # The failure may have an error such as "UNAUTHORIZED: access to the requested resource is not authorized"
+   oc mirror --dest-skip-tls --config ${tmpimageset} docker://${LOCAL_REGISTRY_DNS_NAME}:${LOCAL_REGISTRY_PORT} --log-level debug
```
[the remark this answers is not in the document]
bfournie: Again this might be an oc version issue, as I see it as an option with 4.12
```
$ oc mirror --help | grep log-level
      --log-level string    Log level (e.g. "debug | info | warn | error") (default "info")

```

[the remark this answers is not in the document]
bfournie: It turns out I was using an older version of oc-mirror which is why I didn't see that the options changed. I've updated to add a oc_mirror_cleanup.sh to remove oc-mirror and ensure the latest will be used on the next run.

bfournie:
> > I hit this error:
> > ```
> > +(/home/rwsu/go/src/github.com/openshift-metal3/dev-scripts/oc_mirror.sh:71): publish_mirror_registry(): oc mirror --dest-skip-tls --config /tmp/imageset--3jybQ2i4qI docker://virthost.ostest.test.metalkube.org:8443 --verbose 10
> > Checking push permissions for virthost.ostest.test.metalkube.org:8443
> > Using image virthost.ostest.test.metalkube.org:8443/oc-mirror to check permissions
> > error: config GVK not recognized: mirror.openshift.io/v1alpha1, Kind=ImageSetConfiguration
> > ```
> 
> Can you get the 'oc version'? I'm not seeing this with 4.11
> 
> ```
> $ oc version
> Client Version: 4.11.0-0.nightly-2022-09-01-204001
> ```

I fixed this to use alpah2, I wasn't seeing the problem as I had an older version of oc-mirror.

rwsu (changes requested):

oc_mirror.sh
```diff
@@ -0,0 +1,71 @@
...
+    skipTLS: true
+mirror:
+  ocp:
+    channels:
+    - name: candidate-${OPENSHIFT_RELEASE_STREAM}
+  additionalImages:
+  - name: registry.redhat.io/ubi8/ubi:latest
+EOF
+
+   # TODO - print message on failure that password for must be added to .docker/config.json
```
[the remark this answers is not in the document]
rwsu: I think we should improve the user experience here. When a user enables OC_MIRROR=true and sees an UNAUTHORIZED.. error, they will not have any idea of how to fix it unless they happen to come across this part of the code. 

I think additional documentation should be added to config_example.sh for OC_MIRROR to explain what needs to be added to .docker/config.json. Since the password is hard-coded in agent/common.sh, config_example.sh should explain that this should be added to .docker/config.json:

`"virthost.ostest.test.metalkube.org:8443": { "auth": "c3RhY2s6cXVheXBhc3N3b3Jk" },`

I think the secrets for registry.redhat.io and quay.io also need to be added to .docker/config.json, and this should be explained in config_example.sh for OC_MIRROR. When they weren't present, I was seeing these errors:

```
+(/home/rwsu/go/src/github.com/openshift-metal3/dev-scripts/oc_mirror.sh:59): publish_mirror_registry(): cat
+(/home/rwsu/go/src/github.com/openshift-metal3/dev-scripts/oc_mirror.sh:78): publish_mirror_registry(): oc mirror --dest-skip-tls --config /tmp/imageset--Blj8LVZGWj docker://virthost.ostest.test.metalkube.org:8443
Checking push permissions for virthost.ostest.test.metalkube.org:8443
Creating directory: oc-mirror-workspace/src/publish
Creating directory: oc-mirror-workspace/src/v2
Creating directory: oc-mirror-workspace/src/charts
Creating directory: oc-mirror-workspace/src/release-signatures
No metadata detected, creating new workspace
level=info msg=trying next host error=failed to authorize: failed to fetch anonymous token: unexpected status: 401 Unauthorized host=registry.redhat.io
error: failed to authorize: failed to fetch anonymous token: unexpected status: 401 Unauthorized
```

`error: unable to retrieve source image quay.io/openshift-release-dev/ocp-v4.0-art-dev manifest sha256:6b357378e32d5f07537b6a657f53b332ac89c7e3800ca4281d59afadaf7cbff3: unauthorized: access to the requested resource is not authorized
error: unable to retrieve source image quay.io/openshift-release-dev/ocp-v4.0-art-dev manifest sha256:9a6435cb4f9394d7f10ae6d25478093e0bb58ee6188c942808d8ecc91c6118a3: unauthorized: access to the requested resource is not authorized
`

agent/common.sh
```diff
@@ -6,9 +6,31 @@ export AGENT_STATIC_IP_NODE0_ONLY=${AGENT_STATIC_IP_NODE0_ONLY:-"false"}
 # Override command name in case of extraction
 export OPENSHIFT_INSTALLER_CMD="openshift-install"
 
+export OC_MIRROR=false
```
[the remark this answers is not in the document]
bfournie: Removed

oc_mirror.sh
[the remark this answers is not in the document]
bfournie: Good points. I've updated config_example.sh to document how to update .docker/config.json.

As we talked about, there is an RFE to pass the auth into the oc-mirror command https://issues.redhat.com/browse/OCPBUGS-241. If that is implemented we can remove this note as we won't need to updated config.json

rwsu (commented):
/lgtm

rwsu (approved):

agent/01_agent_requirements.sh
```diff
@@ -6,10 +6,41 @@ SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )/.." && pwd )"
 
 source $SCRIPTDIR/validation.sh
 source $SCRIPTDIR/common.sh
-
-early_deploy_validation
```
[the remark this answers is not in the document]
bfournie: No that was intentional. As its done by `01_install_requirements.sh`, which gets invoked we shouldn't also need to do it here.  But mainly it was introducing a problem:
++(/home/stack/dev-scripts-bfournie/validation.sh:86): early_deploy_validation(): mktemp --tmpdir test-token--XXXXXXXXXX
+(/home/stack/dev-scripts-bfournie/validation.sh:86): early_deploy_validation(): _test_token=/tmp/test-token--N9YPbUz7Ul
/home/stack/dev-scripts-bfournie/validation.sh: line 87: _tmpfiles: unbound variable

oc_mirror.sh
```diff
@@ -0,0 +1,82 @@
...
+    }
+  }
+}
+EOF
+
+   jq -s '.[0] * .[1]' ${tmpauthfile} ${PULL_SECRET_FILE} > ${tmppullsecret}
+   cp ${tmppullsecret} ${PULL_SECRET_FILE}
+
+   # Note that this must also be manually added to .docker/config.json, we don't currently do this in dev-scripts to avoid writing
+   # to user directory.
```
andfasano, line 31: Couldn't be created a backup file before modifying it (with a log message)? It's a little bit annoying the manual modification

config_example.sh
```diff
@@ -550,6 +550,23 @@ set -x
...
+# When using OC_MIRROR the auths for the mirror must be added to .docker/config.json
+# or an UNAUTHORIZED error will result when attempting to use it. An example entry
+# in this file is:
+# "virthost.ostest.test.metalkube.org:8443": { "auth": "c3RhY2s6cXVheXBhc3N3b3Jk" },
+# where the password used for the mirror is generated by
+# 'echo -n "${QUAY_REGISTRY_USER}:${QUAY_REGISTRY_PASS}" | base64 -w0`'
+#
+# In addition, the authorization for quay.io and registry.redhat.io must be added
+# to .docker/config.json, the same content that is in $PULL_SECRET_FILE.
+#export OC_MIRROR=false
```
andfasano, line 568: What happens if this config var is enabled in a non-agent flow?

bfournie: Its a noop - we don't use if for anything outside of the agent

oc_mirror.sh
bfournie: I'm still kind of resistant about modifying a file in the user home dir, I just added a check to error out if it wasn't defined in that file. But if people feel strongly about changing this file I can add it the same way we modify the pull secret file (which isn't in the home dir).
Note there is a RFE to pass the auth file into oc-mirror https://issues.redhat.com/browse/OCPBUGS-241, this would allow us to use a file not in the user home dir, it looks like there is resistance to relying on this file.

bfournie: OK I ended up adding the virthost auth to $HOME/.docker/config.json and adding an env setting for this file and allowing it be changed.  Hopefully this will help reduce the manual changes and confusion around this file.

config_example.sh
andfasano: It doesn't seem fully correct. An external dev-scripts user, by reading the doc will have not clue about it. 
Also, by setting the `MIRROR_IMAGES`, another local registry will be anyway setup during the configure [step](https://github.com/openshift-metal3/dev-scripts/blob/9f4c324a56d8b43fbb9be01921144f6ccf37e7cd/02_configure_host.sh#L71-L73) in the agent flow, thus resulting in two local registries, ie: 
```
CONTAINER ID  IMAGE                                         COMMAND               CREATED            STATUS                PORTS                   NAMES
33070dd9195c  quay.io/libpod/registry:2.8                   /etc/docker/regis...  About an hour ago  Up About an hour ago                          registry
...
09c4bbd671d8  registry.redhat.io/quay/quay-rhel8:v3.7.8     registry              59 minutes ago     Up 59 minutes ago     0.0.0.0:8443->8443/tcp  quay-app
```
I'd still think that a `battery swap` approach for the registry implementation (default or quay) would be preferred in general and less confusing from my point of view. If not something to be addressed in this PR, I'd at least make it clear that's something temporarly specific for the agent flow (thus the config name)

agent/01_agent_requirements.sh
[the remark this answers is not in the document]
bfournie: OK the issue was with the source agent/common.sh. I changed the 'set' to fix the problem and can now include early_deploy_validation

agent/04_agent_configure.sh
```diff
@@ -143,28 +144,61 @@ spec:
...
-location = "${LOCAL_REGISTRY_DNS_NAME}:${LOCAL_REGISTRY_PORT}/localimages/local-release-image"
+location = "${LOCAL_REGISTRY_DNS_NAME}:${LOCAL_REGISTRY_PORT}/${MIRROR_IMAGE_URL_SUFFIX}"
 EOF
 
-cp $REGISTRY_DIR/certs/$REGISTRY_CRT ${MIRROR_PATH}/ca-bundle.crt
+   cp $REGISTRY_DIR/certs/$REGISTRY_CRT ${MIRROR_PATH}/ca-bundle.crt
+
+   else
+
+      cat > "${MIRROR_PATH}/registries.conf" << EOF
```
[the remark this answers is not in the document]
bfournie: That's due to the auths needed in ~/.docker/config.json. I've updated it to include the same auths as PULL_SECRET_FILE.

However earlier there was an issue in that 4.12 didn't have stable_ or candidated_ channels set up, not sure if those have been set up yet.

config_example.sh
bfournie: I added a check to not call setup_local_registry() if OC_MIRROR is set. The legacy registry might still be set up from a previous run as I don't think it gets cleaned up, but fixing that is out of scope.

It it would be clearer as I can set MIRROR_IMAGES when OC_MIRROR is set which is what we do with IPv6.

common.sh
```diff
@@ -208,6 +214,12 @@ if [ -n "$MIRROR_IMAGES" ]; then
     export OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE="${LOCAL_REGISTRY_DNS_NAME}:${LOCAL_REGISTRY_PORT}/localimages/local-release-image:latest"
 fi
 
+if [[ ! -z "${OC_MIRROR}" ]] && [[ "${OC_MIRROR}" == "true" ]]; then
```
andfasano, line 217: Looks like this block should go before the previous one, ie in a similar fashion to when "_LOCAL_IMAGE" is defined (which implies activating `MIRROR_IMAGES` as well), otherwise `OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE` will not be set

bfournie: Yep, good catch. Fixed.

oc_mirror.sh
```diff
@@ -0,0 +1,83 @@
...
+mirror:
+  platform:
+    channels:
+    - name: candidate-${OPENSHIFT_RELEASE_STREAM}
+      type: ocp
+  additionalImages:
+  - name: registry.redhat.io/ubi8/ubi:latest
+EOF
+
+   oc mirror --dest-skip-tls --config ${tmpimageset} docker://${LOCAL_REGISTRY_DNS_NAME}:${LOCAL_REGISTRY_PORT}
```
andfasano, line 81: IIUC all the executions of `oc mirror` are performed in the current directory, while the working dir should be set to `${WORKING_DIR}` (preferably a sub folder) (I see the `oc-mirror.log` and `oc-mirror-workspace` folder being created in the curr dir, which was something not expected)

oc_mirror_cleanup.sh
```diff
@@ -0,0 +1,24 @@
...
+   sudo ./mirror-registry uninstall --quayRoot ${WORKING_DIR}/oc-mirror/ -v --autoApprove
+   popd
+fi
+
+if [[ -f "/usr/local/bin/oc-mirror" ]]; then
+  sudo rm "/usr/local/bin/oc-mirror"
+fi
+
+if [ -d "./oc-mirror-workspace" ]; then
+  rm -rf "./oc-mirror-workspace"
```
andfasano, line 17: And ofc also this one should point to the curr dir

agent/01_agent_requirements.sh
```diff
@@ -6,10 +6,43 @@ SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )/.." && pwd )"
...
+   if [[ ! -f "/usr/local/bin/${oc_mirror_exec}" ]]; then
+      curl -O -L https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp/stable/${oc_mirror_file}
+      tar xzf ${oc_mirror_file}
+      chmod +x ${oc_mirror_exec}
+      sudo mv -f ${oc_mirror_exec} /usr/local/bin
+      rm -f ${oc_mirror_file}
+   fi
+
+   # set up the channel using the most recent candidate release
+   release_candidate=`oc mirror list releases --channel=candidate-${OPENSHIFT_RELEASE_STREAM} | tail -1`
```
andfasano, line 44: As discussed, it looks like this one is ignoring `OPENSHIFT_RELEASE_IMAGE` when set, thus resulting in a different version installed.
Additionally, if possible it would be preferrable to set `OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE` in one of the existing spots to minimize the points where it gets touched.
Even better, but very likely to be addressed in another PR, to avoid using that var for disconnected setups

bfournie: > As discussed, it looks like this one is ignoring `OPENSHIFT_RELEASE_IMAGE` when set, thus resulting in a different version installed. Additionally, if possible it would be preferrable to set `OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE` in one of the existing spots to minimize the points where it gets touched. Even better, but very likely to be addressed in another PR, to avoid using that var for disconnected setups

The reason it's using `4.11.6` when `OPENSHIFT_RELEASE_STREAM=4.12` is because for oc-mirror the ImageSetConfiguration requires a "candidate" channel and there are currently no "candidate" releases for 4.12, the most recent candidate release is `4.11.6`

andfasano: Ok, how can then be used `oc mirror` with a custom provided version, since that's the case we'll have also in the CI?

lranjbar:
Working on this PR: we found out that `oc mirror` does not support CI nightly releases at this point.

The work for AGENT-331 is based off this branch so I would like to get this merged. Right now I'm going to create PRs for the job but have it disabled for now. Once there is a 4.12 candidate we will be able to turn on the CI job.

In the meantime we can use this work to test locally. I created AGENT-368 as a follow up on the tech debt. @andfasano 😃

lranjbar:
Here are the results of testing this PR:
- oc mirror is working for the IPV4 scenario and the IPV6 and IPV4/IPV6 scenarios hit an issue that is fixed in 4.12
- the legacy mirror scripts are working as expected

image

To avoid this PR going stale we can merge these and some more testing after 4.12 is released

agent/04_agent_configure.sh
```diff
@@ -272,28 +273,62 @@ spec:
 EOF
 
 if [ ! -z "${MIRROR_IMAGES}" ]; then
-# TODO - get the mirror registry info from output of 'oc adm release mirror'
 
-    cat > "${MIRROR_PATH}/registries.conf" << EOF
+   # Set up registries.conf and ca-bundle.crt for mirroring
+   if [[ -z "${OC_MIRROR}" ]] || [[ ! "${OC_MIRROR}" == true ]]; then
```
zaneb, line 278: nit: I don't think you need both of these checks, although from my (suppressed) memory of autotools there were certain flavors of UNIX in the 1980s that did :laughing:

agent/common.sh
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-set -euxo pipefail
+set -o pipefail
```
zaneb, line 2: This change appears to disable logging, as well as allow undefined variables and not exit on error. I don't think we want that. Was this accidentally left over from some debugging?

oc_mirror.sh
```diff
@@ -0,0 +1,90 @@
+#!/usr/bin/env bash
+set -ex
```
zaneb, line 2: Would prefer to see `-u` and `-o pipefail` also set. It's too easy to misspell the name of an environment variable.

oc_mirror_cleanup.sh
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env bash
+set -x
```
zaneb, line 2: Would also prefer to see `set -euxo pipefail` here.

bfournie: Fixed

oc_mirror.sh
bfournie: Fixed

agent/common.sh
bfournie: Yeah now I remember, I changed that because there was a bug in including validation.sh that I thought I'd fix later. Thanks for pointing it out. Fixed.

agent/04_agent_configure.sh
bfournie: Fixed

agent/common.sh
```diff
@@ -11,8 +11,14 @@ export CLUSTER_HOST_PREFIX_V4="24"
 export EXTERNAL_SUBNET_V6="fd2e:6f44:5dd8:c956::/64"
 
 if [ -n "$MIRROR_IMAGES" ]; then
-    # We're going to be using a locally modified release image
-    export OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE="${LOCAL_REGISTRY_DNS_NAME}:${LOCAL_REGISTRY_PORT}/localimages/local-release-image:${OPENSHIFT_RELEASE_TAG}"
+
+    export MIRROR_IMAGE_URL_SUFFIX="localimages/local-release-image"
+
+    if [ "${OC_MIRROR}" == true ]; then
+        export MIRROR_IMAGE_URL_SUFFIX="openshift/release-images"
```
andfasano, line 18: any specific reason to change the target repo? Keeping the default one will make  easier to use this code in future also for the standard flow

agent/common.sh
```diff
@@ -11,8 +11,14 @@ export CLUSTER_HOST_PREFIX_V4="24"
...
+    fi
+
+    export OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE="${LOCAL_REGISTRY_DNS_NAME}:${LOCAL_REGISTRY_PORT}/${MIRROR_IMAGE_URL_SUFFIX}:${OPENSHIFT_RELEASE_TAG}"
```
andfasano, line 21: This second override here it's not clear. I would consider the previous one just a temporary workaround until it will possible to specify a given version/pullspec (otherwise this won't work on CI, especially for disconnected builds), but wouldn't see the need to repeat it here

common.sh
```diff
@@ -98,6 +104,10 @@ export REGISTRY_PASS=${REGISTRY_PASS:-ocp-pass}
 export REGISTRY_DIR=${REGISTRY_DIR:-$WORKING_DIR/registry}
 export REGISTRY_CREDS=${REGISTRY_CREDS:-$HOME/private-mirror-${CLUSTER_NAME}.json}
 export REGISTRY_CRT=registry.2.crt
+export REGISTRY_BACKEND=${REGISTRY_BACKEND:-"podman"}
+if [[ ${REGISTRY_BACKEND} = "quay" ]]; then
+    export LOCAL_REGISTRY_PORT=8443
```
andfasano, line 109: Why not using the usual 5000 port? The backend may change, buy the way the service is exposed should remain the same

common.sh
```diff
@@ -203,9 +213,14 @@ if env | grep -q "_LOCAL_IMAGE=" ; then
...
-    # We're going to be using a locally modified release image
-    export OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE="${LOCAL_REGISTRY_DNS_NAME}:${LOCAL_REGISTRY_PORT}/localimages/local-release-image:latest"
+if [[ ! -z "${OC_MIRROR}" ]] && [[ "${OC_MIRROR}" == "true" ]]; then
+    export MIRROR_IMAGES=true
+
+    if [[ -x "/usr/local/bin/oc-mirror" ]]; then
+       # set up the channel using the most recent candidate release
+       release_candidate=`oc-mirror list releases --channel=candidate-${OPENSHIFT_RELEASE_STREAM} | tail -1`
+       export OPENSHIFT_RELEASE_TAG="${release_candidate}-x86_64"
+    fi
```
andfasano, line 223: Could also this one be removed and kept just the one define in 01_agent_requirements?

network.sh
```diff
@@ -119,7 +119,6 @@ elif [[ "$IP_STACK" = "v6" ]]; then
   export SERVICE_SUBNET_V6=${SERVICE_SUBNET_V6:-"fd02::/112"}
   export NETWORK_TYPE=${NETWORK_TYPE:-"OVNKubernetes"}
   export MIRROR_IMAGES=true
-  export OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE="${LOCAL_REGISTRY_DNS_NAME}:${LOCAL_REGISTRY_PORT}/localimages/local-release-image:latest"
```
andfasano, line 122: As mentioned above, it seems to me simpler to keep the target repo name the same one, indepedently by the selected backend. The quay one will not support ipv6, so the execution should just fail in that case.

oc_mirror.sh
```diff
@@ -0,0 +1,90 @@
+#!/usr/bin/env bash
+set -euxo pipefail
+
+# Utility functions to create a local mirror registry using 'mirror-registry' and mirror
+# a release using 'oc mirror'
+
+function add_auth_to_pull_secret() {
```
andfasano, line 7: On a second pass, I guess that the pull-secret prepared by dev-script in the WORKING_DIR should already be fine (it already contains the registry), so it could be copied directly as ./docker/config.json file

agent/common.sh
bfournie: > any specific reason to change the target repo? Keeping the default one will make easier to use this code in future also for the standard flow

Running the `oc mirror` command as in oc_mirror.sh will result in an ICSP using `openshift/release-images` which is why the target has to be set like that.
  repositoryDigestMirrors:
  - mirrors:
    - virthost.ostest.test.metalkube.org:8443/openshift/release
    source: quay.io/openshift-release-dev/ocp-v4.0-art-dev
  - mirrors:
    - virthost.ostest.test.metalkube.org:8443/openshift/release-images
    source: quay.io/openshift-release-dev/ocp-release

agent/common.sh
bfournie: > This second override here it's not clear. I would consider the previous one just a temporary workaround until it will possible to specify a given version/pullspec (otherwise this won't work on CI, especially for disconnected builds), but wouldn't see the need to repeat it here

Yeah, may need to be able to clean this up so OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE is just set for OC_MIRROR and we use the one defined in common.sh in all other cases, although I'm not sure about the concern with CI as its set to the same value.

common.sh
bfournie: We could use the same port although having different ports has been useful in testing to easily know which backend is being used. I'd prefer to keep them different so its obvious.

common.sh
bfournie: Good catch, yes we should have that only defined on one place, will fix.

network.sh
bfournie: See above, regarding the output of the oc-mirror command in the ICSP.

oc_mirror.sh
bfournie: I'd have to test that but I think it was the pull secret in $WORKING_DIR was missing
```
    "virthost.ostest.test.metalkube.org:8443": {
      "auth": "<auth value>"
    },
```

PR #1437, opened 2022-08-18
billi release script

andfasano:
Script to prepare a BILLI release from a specific git commit.

### DIFF
--- a/agent/release/billi-release.sh
+++ b/agent/release/billi-release.sh
@@ -1,0 +1,91 @@
+#!/usr/bin/env bash
+
+function verify_if_release_image_exists() {
+    local release_image=$1
+    if [ ! "$(podman manifest inspect ${release_image})" ]; then
+        echo "Cannot download image ${release_image}"
+    fi
+}
+
+function clone_agent_repo() {
+    
+    local commit=$1
+
+    if [[ ! -d "installer" ]]; then 
+        git clone https://github.com/openshift/installer.git
+    fi
+    
+    if [[ ! -z "${commit}" ]]; then
+        echo "Building from commit $commit"
+    fi
+
+    pushd installer
+    git checkout $commit
+    popd
+}
+
+function build_agent_installer() {
+    pushd installer
+    hack/build.sh
+    popd
+}
+
+function patch_openshift_install_release_version() {
+    local version=$1
+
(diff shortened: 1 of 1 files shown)


agent/release/billi-release.sh
```diff
@@ -0,0 +1,94 @@
...
+    echo
+    echo "Important note: the specified <release version> must match the one pointed by <release image>"
+    echo
+    echo "example: ./billi-release.sh 38e719a 4.11.0 quay.io/openshift-release-dev/ocp-release@sha256:300bce8246cf880e792e106607925de0a404484637627edf5f517375517d54a4"
+
+    exit 1
+fi
+
+commit=$1
+release_version=$2
```
rwsu, line 86: I think this can be simplified. The release version can be determined from the release image using:

oc adm release info -o template --template '{{.metadata.version}}' --insecure=true quay.io/openshift-release-dev/ocp-release@sha256:300bce8246cf880e792e106607925de0a404484637627edf5f517375517d54a4

rwsu (changes requested):

agent/release/billi-release.sh
andfasano: Thanks @rwsu it seems really a good idea!

agent/release/billi-release.sh
```diff
@@ -0,0 +1,91 @@
+#!/bin/bash
```
elfosardo, line 1: probably better use #!/usr/bin/env bash

agent/release/billi-release.sh
```diff
@@ -0,0 +1,91 @@
...
+
+    if [[ ! -d "installer" ]]; then 
+        git clone https://github.com/openshift/installer.git
+    fi
+    
+    if [[ ! -z "${commit}" ]]; then
+        echo "Building from commit $commit"
+    fi
+
+    cd installer
```
elfosardo, line 22: popd / pushd instead ?

agent/release/billi-release.sh
```diff
@@ -0,0 +1,91 @@
...
+    git checkout $commit
+    cd ..
+}
+
+function build_agent_installer() {
+    cd installer
```
elfosardo, line 28: ditto

agent/release/billi-release.sh
zaneb: I like this but we should probably also document how to log in to the registry.

PR #1438, opened 2022-08-18
Migrate network config to NetworkManager keyfiles

elfosardo:
this takes back work done in https://github.com/openshift-metal3/dev-scripts/pull/1290 but using keyfiles instead of direct nmcli commands

### DIFF
--- a/01_install_requirements.sh
+++ b/01_install_requirements.sh
@@ -47,8 +47,6 @@ fi
 # vm-setup/install-package-playbook.yml
 case $DISTRO in
   "centos8"|"rhel8"|"almalinux8"|"rocky8")
-    # install network-scripts package to be able to use legacy network commands
-    sudo dnf install -y network-scripts
     if [[ $DISTRO == "centos8" ]] && [[ "$NAME" != *"Stream"* ]]; then
         echo "CentOS is not supported, please switch to CentOS Stream / RHEL / Rocky / Alma"
         exit 1
--- a/02_configure_host.sh
+++ b/02_configure_host.sh
@@ -131,7 +131,7 @@ if [ ${NUM_EXTRA_WORKERS} -ne 0 ]; then
   jq "{nodes: .nodes[-${NUM_EXTRA_WORKERS}:]}" ${ORIG_NODES_FILE} | tee ${EXTRA_NODES_FILE}
 fi
 
-ZONE="\nZONE=libvirt"
+ZONE="libvirt"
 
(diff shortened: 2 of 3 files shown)


elfosardo:
interesting, this works perfectly locally

elfosardo:
this has been tested on CentOS Stream 8 and 9 successfully
we're using Rocky 8.5 in CI which makes me think that's why it's failing here

cybertron:
That's weird. This is working for me locally in Alma 8.6. :-/

elfosardo:
> That's weird. This is working for me locally in Alma 8.6. :-/

@cybertron in CI we pin Rocky 8.5 and it doesn't like this :/

cybertron:
Could we just unpin now that we have https://github.com/openshift-metal3/dev-scripts/commit/522d3c853a36c44125ea165cf8e633c5e88d7693 and don't need to pin in order for ipv6 to work? At least I think that's why we were pinned to 8.5.

elfosardo:
/retest
may need to change a couple of things, let's rerun failed jobs to confirm

elfosardo:
this is really unsettling:
+(./02_configure_host.sh:286): main(): sudo nmcli con show
NAME             UUID                                  TYPE      DEVICE     
cluster6bm       c4e94d83-6a93-499e-bf6a-628529b2466c  bridge    cluster6bm 
cluster6pr       3b2093a6-0149-3a8b-9ef6-4a775a94cb4c  bridge    cluster6pr 
virbr0           eaff8f5b-c6e0-4c96-a768-36d5f43310ff  bridge    virbr0     
cluster6bm-eth2  2150b483-d02f-40e2-ad74-c64d81bbe8ee  ethernet  eth2       
eth3             59cc9865-b3bd-33ef-86c5-0478d8bbc11f  ethernet  eth3    

but the error is:
"level=fatal msg=failed to fetch Terraform Variables: failed to fetch dependency of "Terraform Variables": failed to generate asset "Platform Provisioning Check": platform.baremetal.provisioningBridge: Invalid value: "cluster6pr": could not find interface "cluster6pr", valid interfaces are cluster6bm, lo"

iurygregory:
@elfosardo just wondering if we still need this , should we trigger a retest in the jobs?

elfosardo:
> @elfosardo just wondering if we still need this , should we trigger a retest in the jobs?

we definitely need this
please do not trigger a retest, it will still fail
troubleshooting the issue is taking some time

PR #1440, opened 2022-08-23
Adjust for image-customization-controller image rename

dtantsur:
The payload image is called machine-image-customization-controller.


dtantsur:

Overriding this job since the code is not v6-specific, another v6 job has passed and the code is probably only used locally anyway.

PR #1441, opened 2022-08-24
Fix support for Rocky and Alma Linux

elfosardo:


elfosardo:
/retest
I believe the serial job is still foobar

cybertron:
/lgtm

Oops, I probably missed this because all of the necessary packages were already installed on my host.

PR #1442, opened 2022-08-26
Bug 2087096: Add a dummy interface to IPv6 bridge

bfournie:
In order to ensure that the IPv6 bridge interface comes up add a dummy interface. This is needed for versions of NetworkManager after 1.32.


derekhiggins:
/lgtm
thanks bob

PR #1443, opened 2022-08-29
Fix another missing Alma/Rocky reference

cybertron:
Not sure how this got missed, but I guess that's the danger of not
CI'ing with all of the supported platforms.


elfosardo:
@cybertron we should override ci/prow/e2e-metal-ipi that has been renamed to ci/prow/e2e-metal-ipi-bm   or just merge this

PR #1444, opened 2022-08-30
AGENT-239: Use IPv6 addresses in nmstate config

bfournie:
For agent IPv6 testing set up the nmstate config to use IPv6 addressing.

I've tested this with both IPv6 and IPv4 using 4.12 and https://github.com/openshift/installer/pull/6249

### DIFF
--- a/agent/05_agent_create_cluster.sh
+++ b/agent/05_agent_create_cluster.sh
@@ -67,6 +67,10 @@ function force_mirror_disconnect() {
 }
 
 function enable_assisted_service_ui() {
+  if [[ "${IP_STACK}" = "v6" ]]; then
+       echo "In a disconnected environment the assisted-installer GUI cannot be enabled"
+       return
+  fi
   node0_ip=$(get_node0_ip)
   ssh_opts=(-o 'StrictHostKeyChecking=no' -o 'UserKnownHostsFile=/dev/null' -q core@${node0_ip})
 
--- a/agent/04_agent_configure.sh
+++ b/agent/04_agent_configure.sh
@@ -207,6 +207,14 @@ EOF
 
     num_ips=${#AGENT_NODES_IPS[@]}
 
+    if [[ "$IP_STACK" = "v4" ]]; then
+       interface_type="ipv4"
+       route_dest="0.0.0.0/0"
+    else
+       interface_type="ipv6"
+       route_dest="::/0"
+    fi
+
     # Create a yaml for each host in nmstateconfig.yaml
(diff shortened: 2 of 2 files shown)


karmab:
@zaneb @pawanpinjarkar can you review?

rwsu:
Using the other installer IPv6 PRs, I was able to install a compact_ipv6 cluster.

/lgtm

agent/05_agent_create_cluster.sh
```diff
@@ -67,6 +67,10 @@ function force_mirror_disconnect() {
 }
 
 function enable_assisted_service_ui() {
+  if [ ! -z "${MIRROR_IMAGES}" ]; then
```
andfasano, line 70: This potentially could be enabled also for non-disconnected environment. I guess using IP_STACK would be a better match in our case

bfournie: OC_MIRROR does imply disconnected, if we use IP_STACK we could still have a disconnected env with IPv4 which would fail when using GUI.

andfasano: Checking IP_STACK it's also the approach take in other part of the CI to verify if the current environment is disconnected or not, ie https://github.com/openshift/release/blob/0e6a79426cd6aeebc497d5d79fd8029aabd74058/ci-operator/step-registry/baremetalds/e2e/test/baremetalds-e2e-test-commands.sh#L31-L40

bfournie: Changed to use IP_STACK

andfasano: May be worth using the inverted logic to include also the dualstack case (v4v6), which is considered disconnected as well as the v6 (so that essentially the GUI will be enabled only for v4 case)?

bfournie: No, v4v6 doesn't necessarily require a disconnected install - see https://github.com/openshift-metal3/dev-scripts/blob/master/network.sh#L121

I do think we're getting away from the purpose of this PR which is not related to the setting of the GUI and I'd recommend we handle that in a separate PR.

andfasano: I'm not fully sure about that, but agree that this increment is good enough to proceed and the dualstack case could be eventually fixed in a later PR if required

bfournie: OK thanks. Yes I'm working on changes for v4v6 now and I'll make sure to test this particular case with IP_STACK=v4v6

rwsu (commented):
/lgtm

PR #1445, opened 2022-08-31
AGENT-325: Set AgentClusterInstall networkType using IP_STACK

rwsu:
networkType is set to OpenShiftSDN if IP_STACK == v4

networkType is set to OVNKubernetes if IP_STACK == v4v6|v6

### DIFF
--- a/agent/04_agent_configure.sh
+++ b/agent/04_agent_configure.sh
@@ -110,6 +110,7 @@ cat >> "${MANIFESTS_PATH}/agent-cluster-install.yaml" << EOF
       hostPrefix: ${CLUSTER_HOST_PREFIX}
     serviceNetwork:
     - ${SERVICE_NETWORK}
+    networkType: ${NETWORK_TYPE}
   provisionRequirements:
     controlPlaneAgents: ${NUM_MASTERS}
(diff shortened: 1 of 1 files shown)


agent/04_agent_configure.sh
```diff
@@ -110,6 +110,20 @@ cat >> "${MANIFESTS_PATH}/agent-cluster-install.yaml" << EOF
...
+    networkType: OVNKubernetes
+EOF
+    ;;
+  *)
+cat >> "${MANIFESTS_PATH}/agent-cluster-install.yaml" << EOF
+    networkType: OpenShiftSDN
+EOF
+    ;;
+esac
+cat >> "${MANIFESTS_PATH}/agent-cluster-install.yaml" << EOF
```
andfasano, line 126: @rwsu what about setting the field just as in https://github.com/openshift-metal3/dev-scripts/blob/522d3c853a36c44125ea165cf8e633c5e88d7693/ocp_install_env.sh#L218? `NETWORK_TYPE` should already contain the right value based on IP_STACK

rwsu: Ah. Even better. I'll make that change. Thanks!

PR #1447, opened 2022-09-07
Handle error when creating dummy interface

bfournie:
In certain network configurations the 'ip link' commands can fail. We have handling for the other cases when the command is used but not for when creating the new dummy interface for IPv6. Add handling for that case.


dtantsur:
/lgtm

I wonder if we should delete the old interface for people who reuse their box.

bfournie:
> /lgtm
> 
> I wonder if we should delete the old interface for people who reuse their box.

The old interface won't get created because of the name length problem,

dtantsur:
> The old interface won't get created because of the name length problem

I mean, on systems that do not have the length issue.

PR #1448, opened 2022-09-08
AGENT-329: add mce deployment

andfasano:
This patch introduces a new specific agent config variable to deploy MCE in a connected environment

### DIFF
--- a/agent/05_agent_create_cluster.sh
+++ b/agent/05_agent_create_cluster.sh
@@ -9,6 +9,7 @@ source $SCRIPTDIR/common.sh
 source $SCRIPTDIR/network.sh
 source $SCRIPTDIR/utils.sh
 source $SCRIPTDIR/validation.sh
+source $SCRIPTDIR/ocp_install_env.sh
 source $SCRIPTDIR/agent/common.sh
 
 early_deploy_validation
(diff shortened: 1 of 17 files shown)


lranjbar:
Looks good so far we will want to make the release images a variable/template later on.

agent/05_agent_create_cluster.sh
```diff
@@ -96,6 +97,51 @@ function wait_for_cluster_ready() {
...
+  apply_manifest "$mceManifests/agent_mce_1_02_mce.yaml"
+
+  wait_for_crd "agentserviceconfigs.agent-install.openshift.io"
+  apply_manifest "$mceManifests/agent_mce_1_03_agentserviceconfig.yaml"
+
+  wait_for_crd "clusterimagesets.hive.openshift.io"
+  apply_manifest "$mceManifests/agent_mce_1_04_clusterimageset.yaml"
+
+  apply_manifest "$mceManifests/agent_mce_1_05_autoimport.yaml"
+  oc wait -n multicluster-engine managedclusters local-cluster --for condition=ManagedClusterJoined=True --timeout 10m || exit 1
```
lranjbar, line 132: In my experience the `wait_for_crd` just slows down the job. We can keep it to see the impact. Since Openshift is fully running by this point you should be able to batch create all of these and then check for the volume and cluster. It will take it some time to reconcile but it should reconcile.

andfasano: Right, I left them (not batched) mostly for the sake of clarity and just to be sure that the crds where available - they should be already present as you pointed out, so the overall crds checks should be pretty quick

rwsu (commented):
I was able to see the mce deployed on a compact cluster.

/lgtm

lranjbar:
Unfortunately looks like there is a merge conflict but I was also successful in deploying MCE with this branch.

PR #1449, opened 2022-09-09
AGENT-245: DHCP only agent deployment

pawanpinjarkar:
This PR supports DHCP agent deployment with just agent-config.yaml and install-config.yaml. Set only `rendezvousIP` in the agent-config.yaml ( no hosts at all or no hosts with any network config). Devscripts provides no nmstateconfig.yaml

To create a DHCP agent deployment, `export AGENT_E2E_TEST_SCENARIO=HA_IPV4_DHCP` in config_\<user\>.sh.

Also, this is a change for 4.12 so `export OPENSHIFT_RELEASE_STREAM=4.12`

To test, run `make agent`

### DIFF
--- a/agent/04_agent_configure.sh
+++ b/agent/04_agent_configure.sh
@@ -298,6 +298,7 @@ metadata:
   namespace: ${CLUSTER_NAMESPACE}
 rendezvousIP: ${AGENT_NODES_IPS[0]}
 EOF
+    set +x
     pull_secret=$(cat $PULL_SECRET_FILE)
     cat > "${MANIFESTS_PATH}/install-config.yaml" << EOF
 apiVersion: v1
@@ -382,5 +383,6 @@ if [[ $NETWORKING_MODE == "DHCP" ]]; then
   generate_install_config_agent_config
 else
   generate_cluster_manifests
-  generate_extra_cluster_manifests
 fi
+
+generate_extra_cluster_manifests
--- a/agent/04_agent_configure.sh
+++ b/agent/04_agent_configure.sh
@@ -325,7 +325,7 @@ networking:
   clusterNetwork:
   - cidr: ${cluster_network}
     hostPrefix: ${cluster_host_prefix}
-  networkType: OpenShiftSDN
+  networkType: ${NETWORK_TYPE}
   machineNetwork:
   - cidr: ${machine_network}
   serviceNetwork: 
--- a/agent/04_agent_configure.sh
+++ b/agent/04_agent_configure.sh
@@ -284,13 +284,11 @@ EOF
   fi
 }
 
-function generate_install_config_agent_config() {
+function generate_agent_config() {
 
   MANIFESTS_PATH="${OCP_DIR}"
   mkdir -p ${MANIFESTS_PATH}
   
-  setNetworkingVars
-
     cat > "${MANIFESTS_PATH}/agent-config.yaml" << EOF
 apiVersion: v1alpha1
 metadata:
@@ -298,8 +296,17 @@ metadata:
   namespace: ${CLUSTER_NAMESPACE}
 rendezvousIP: ${AGENT_NODES_IPS[0]}
 EOF
-    set +x
-    pull_secret=$(cat $PULL_SECRET_FILE)
+}
+
+function generate_install_config() {
+
+  MANIFESTS_PATH="${OCP_DIR}"
+  mkdir -p ${MANIFESTS_PATH}
+  
+  setNetworkingVars
+
+  set +x
+  pull_secret=$(cat $PULL_SECRET_FILE)
     cat > "${MANIFESTS_PATH}/install-config.yaml" << EOF
 apiVersion: v1
 baseDomain: ${BASE_DOMAIN}
@@ -380,7 +387,8 @@ fi
   fi
 
 if [[ $NETWORKING_MODE == "DHCP" ]]; then
-  generate_install_config_agent_config
+  generate_agent_config
+  generate_install_config
 else
   generate_cluster_manifests
 fi
--- a/agent/04_agent_configure.sh
+++ b/agent/04_agent_configure.sh
@@ -14,6 +14,7 @@ source $SCRIPTDIR/ocp_install_env.sh
 
 early_deploy_validation
 
+CLUSTER_NAMESPACE=cluster0
 
 function get_static_ips_and_macs() {
 
@@ -33,8 +34,12 @@ function get_static_ips_and_macs() {
         external_subnet=$EXTERNAL_SUBNET_V6
     fi
 
-    # Set outside the range used for dhcp
-    base_ip=80
+    if [[ $NETWORKING_MODE == "DHCP" ]]; then
+      base_ip=20
+    else
+      # Set outside the range used for dhcp
+      base_ip=80
+    fi
 
     for (( i=0; i<${static_ips}; i++ ))
     do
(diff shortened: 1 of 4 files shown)


lranjbar (commented):
Why do we have to set these values again?

agent/04_agent_configure.sh
```diff
@@ -60,6 +65,20 @@ function get_static_ips_and_macs() {
...
+    SERVICE_NETWORK=${SERVICE_SUBNET_V4}
+    MACHINE_NETWORK=${EXTERNAL_SUBNET_V4}
+    CLUSTER_HOST_PREFIX=${CLUSTER_HOST_PREFIX_V4}
+  elif [[ "$IP_STACK" = "v6" ]]; then
+    CLUSTER_NETWORK=${CLUSTER_SUBNET_V6}
+    SERVICE_NETWORK=${SERVICE_SUBNET_V6}
+    MACHINE_NETWORK=${EXTERNAL_SUBNET_V6}
+    CLUSTER_HOST_PREFIX=${CLUSTER_HOST_PREFIX_V6}
+  fi
+}
```
lranjbar, line 80: Why do we have to set these values again here?

agent/04_agent_configure.sh
```diff
@@ -14,6 +14,7 @@ source $SCRIPTDIR/ocp_install_env.sh
 
 early_deploy_validation
 
+CLUSTER_NAMESPACE=cluster0
```
lranjbar, line 17: This should be moved to environment variable instead of normal variable so it can be changed in Openshift CI.

lranjbar (commented):
Why do we have to set these values again?

agent/04_agent_configure.sh
pawanpinjarkar: ok, so we can set it always irrespective of networking mode. 
https://github.com/pawanpinjarkar/dev-scripts/blob/c1716abad59ec5e2684677410d3129a38bdfc83b/agent/04_agent_configure.sh#L379-L383

pawanpinjarkar: Maybe it was added initially to help with better readability?

agent/04_agent_configure.sh
```diff
@@ -288,8 +372,12 @@ if [ ! -z "${MIRROR_IMAGES}" ]; then
...
 
-if [[ "${NUM_MASTERS}" > "1" ]]; then
-  set_api_and_ingress_vip
-fi
+ if [[ "${NUM_MASTERS}" > "1" ]]; then
+    set_api_and_ingress_vip
+  fi
 
-generate_cluster_manifests
+if [[ $NETWORKING_MODE == "DHCP" ]]; then
```
rwsu, line 379: I don't think using DHCP to differentiate between agent-config and ZTP manifests is entirely correct. It is also possible to use agent-config with static IP addresses. Also possible would be to ZTP manifests in DHCP mode by not providing nmstateconfigs or nmstateconfigs configured to use DHCP if that is an option. I think we can refactor it in a future update. It would be better to have a choice of which type of manifests to use: install+agent config or ZTP manifests. And then for each manifest type, have static and DHCP options.

agent/04_agent_configure.sh
lranjbar: Sorry if I wasn't clear in my first comment. This is duplicate code and should be removed. Network settings are already set up in [network.sh](https://github.com/openshift-metal3/dev-scripts/blob/master/network.sh#L86-L134)

bfournie: There are needed to set up the subnets to use in the manifests based on the IP_STACK and the settings in network.sh, but these really should be lowercase as they're not env variables and not used outside of this file

agent/04_agent_configure.sh
pawanpinjarkar: Agree, to refactor if we add more scenarios in dev-scripts as mentioned above.

andfasano: I'd prefer to refactor it now as it seems somehow misleading,even though I won't block on this. It'd seem more natural to have a sort of `export AGENT_MANIFESTS={$AGENT_MANIFESTS:-installer}` config var default so that by default we generate installer config files (alternatively, it could be se to `ztp`)

config_example.sh
```diff
@@ -622,11 +622,23 @@ set -x
...
+# - DHCP_HA_IPV6 
+# - SNO_IPV4
+# - SNO_IPV6
+# - DHCP_SNO_IPV4
+# - DHCP_SNO_IPV6
 # When set, the code internally sets other low level details such as disk size, memory, number of masters and workers,
 # cpu and ip stack.
 # This config variable is used only by the agent based installer and is required.
-# export AGENT_E2E_TEST_SCENARIO=HA_IPV4
+# export AGENT_E2E_TEST_SCENARIO=DHCP_COMPACT_IPV4
```
andfasano, line 641: In general I like the idea of extending the scenario naming convention to include other salient factors, but I think that topology type should come first (ie `COMPACT_IPV4_DHCP` or `SNO_IPV4_DHCP`)

agent/04_agent_configure.sh
```diff
@@ -290,9 +374,13 @@ if [ ! -z "${MIRROR_IMAGES}" ]; then
...
+    set_api_and_ingress_vip
+  fi
 
-generate_cluster_manifests
-generate_extra_cluster_manifests
+if [[ $NETWORKING_MODE == "DHCP" ]]; then
+  generate_install_config_agent_config
+else
+  generate_cluster_manifests
+  generate_extra_cluster_manifests
```
andfasano, line 385: The extra manifests function should be applied in any case, as it exercises the functionality of deploying the custom manifests

agent/04_agent_configure.sh
```diff
@@ -273,6 +284,79 @@ EOF
...
+  setNetworkingVars
+
+    cat > "${MANIFESTS_PATH}/agent-config.yaml" << EOF
+apiVersion: v1alpha1
+metadata:
+  name: ${CLUSTER_NAME}
+  namespace: ${CLUSTER_NAMESPACE}
+rendezvousIP: ${AGENT_NODES_IPS[0]}
+EOF
+    pull_secret=$(cat $PULL_SECRET_FILE)
```
andfasano, line 301: Here it's required a `set +x` to avoid leaking the pull secret in the standard output (even better at the beginning of the function)

config_example.sh
pawanpinjarkar: If using, `AGENT_MANIFESTS` then we don't need to set like `AGENT_E2E_TEST_SCENARIO=DHCP_COMPACT_IPV4`

andfasano: If not using the `AGENT_MANIFESTS` then I'd keep a consistent naming convention (by adding the `_DHCP` suffix at the end of `AGENT_E2E_TEST_SCENARIO`). This would also make the variable more readable and minimize the impacts when the `AGENT_MANIFESTS` will be introduced

common.sh
```diff
@@ -354,16 +359,33 @@ fi
...
 
 if [[ ! -z ${AGENT_E2E_TEST_SCENARIO} ]]; then
-  export IP_STACK=$(echo ${AGENT_E2E_TEST_SCENARIO##*_IP} | tr V v)
-  SCENARIO=${AGENT_E2E_TEST_SCENARIO%%_*}
+  IFS='_'
+  read -a arr <<<"$AGENT_E2E_TEST_SCENARIO"
+  delimiterCount=$(echo "$AGENT_E2E_TEST_SCENARIO" | tr -cd '_' | wc -c)
+
+  SCENARIO=${arr[0]}
+  export IP_STACK=${arr[1]}
```
andfasano, line 381: Accepted values for `IP_STACK` are `v4`, `v6` or `v4v6`, so it seems that `IP` must be stripped off from `arr[1]`, ie: ```export IP_STACK=${arr[1]##*IP}``` (in the previous implementation there was a similar string manipulation)

andfasano:
I've been able to successfully deploy a cluster dhcp based

/approve

PR #1453, opened 2022-09-14
AGENT-158: Add support for testing agent with dual-stack

bfournie:
Add dual-stack testing. The nmstate config now takes both v4 and v6 interfaces in dual-stack. In addition the agent-cluster-install.yaml takes the v4 and v6 subnets in ClusterNetwork, ServiceNetwork, and the now MachineNetwork.

Tested with COMPACT_IPV4, COMPACT_IPv6, and COMPACT_IPV4V6.

### DIFF
--- a/agent/04_agent_configure.sh
+++ b/agent/04_agent_configure.sh
@@ -16,9 +16,97 @@ early_deploy_validation
 
 CLUSTER_NAMESPACE=${CLUSTER_NAMESPACE:-"cluster0"}
 
+function get_nmstate_interface_block {
+
+index="$1"
+
+if [[ "$IP_STACK" = "v4" ]]; then
+  echo "ipv4:
+          enabled: true
+          address:
+            - ip: ${AGENT_NODES_IPS[index]}
+              prefix-length: ${CLUSTER_HOST_PREFIX_V4}
+          dhcp: false"
+elif [[ "$IP_STACK" = "v6" ]]; then
+  echo "ipv6:
+          enabled: true
+          address:
+            - ip: ${AGENT_NODES_IPSV6[index]}
+              prefix-length: ${CLUSTER_HOST_PREFIX_V6}
+          dhcp: false"
+else
+       # v4v6
+  echo "ipv4:
+          enabled: true
+          address:
+            - ip: ${AGENT_NODES_IPS[index]}
+              prefix-length: ${CLUSTER_HOST_PREFIX_V4}
+          dhcp: false
+        ipv6:
+          enabled: true
+          address:
+            - ip: ${AGENT_NODES_IPSV6[index]}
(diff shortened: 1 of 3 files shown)


agent/04_agent_configure.sh
```diff
@@ -27,19 +104,21 @@ function get_static_ips_and_macs() {
...
     base_ip=80
 
     for (( i=0; i<${static_ips}; i++ ))
     do
         ip=${base_ip}+${i}
-        AGENT_NODES_IPS+=($(nth_ip ${external_subnet} ${ip}))
+        if [[ "$IP_STACK" = "v4" ]]; then
+            AGENT_NODES_IPS+=($(nth_ip ${EXTERNAL_SUBNET_V4} ${ip}))
+        elif [[ "$IP_STACK" = "v6" ]]; then
+            AGENT_NODES_IPS+=($(nth_ip ${EXTERNAL_SUBNET_V6} ${ip}))
```
andfasano, line 116: For consistency here I'd expect `AGENT_NODES_IPSV6`

agent/04_agent_configure.sh
```diff
@@ -27,19 +104,21 @@ function get_static_ips_and_macs() {
...
+        else
+            # For v4v6 an additional address will be added for v6
+            AGENT_NODES_IPS+=($(nth_ip ${EXTERNAL_SUBNET_V4} ${ip}))
+            AGENT_NODES_IPSV6+=($(nth_ip $EXTERNAL_SUBNET_V6 ${ip}))
```
andfasano, line 120: Q: Would be possible to keep the dual stack case like the ipv6, in order not to differentiate between `AGENT_NODES_IPS` and `AGENT_NODES_IPSV6`?

agent/04_agent_configure.sh
bfournie: Changed to make this more consistent.

andfasano:
I think it requires https://github.com/openshift/installer/pull/6364 to be tested properly?

bfournie:
> I think it requires [openshift/installer#6364](https://github.com/openshift/installer/pull/6364) to be tested properly?

No, it does not as we are using 
the ZTP manifests in dev-scripts and therefore we set up the MachineNetwork here. That installer PR just adds the MachineNetwork when the agent-cluster-install.yaml is generated from install-config

andfasano:
I've tested the PR with the compact scenario and it worked as expected, with all the nodes having an ipv4 and ipv6 address (and reachable through them).
There is still a minor non-blocking problem, as the `ostestbm` network dhcp section gets configured with unused addresses (so slightly more confusing due the presence of both ipv4 and ipv6 addresses), but it could be fixed in another PR.

/approve

PR #1456, opened 2022-09-15
[WIP] Test waiting twice to double the timeout before giving up

andfasano:
/draft

Just for a quick test, not be merged


zaneb:
This succeeded on the first attempt in both cases. This suggests that the problem is with something (assisted-service?) in 4.12, since the dev-scripts jobs run against 4.11.
/close

PR #1418, opened 2022-06-24
Add interface for custom networkConfig

cybertron:
This provides a way to specify a custom networkConfig for each node
in a deployment. The user sets NETWORK_CONFIG_FOLDER to a path that
contains YAML files named for each node. The contents of these files
are included in the install-config definition for the appropriate
node.

PR #1457, opened 2022-09-21
OPNET-156: Add static and bond network configs

cybertron:
These are some example configs that can be used with the NETWORK_CONFIG_FOLDER option added in #1418. In addition to making it easy for new users to test the networkConfig feature, this will also make it easier to run such configurations in CI.

To use, just set the variable to the appropriate path. For example:

export NETWORK_CONFIG_FOLDER=network-configs/static


cybertron:
/approve

None of the files in this PR are used in ci yet so the failing job is definitely not related.

PR #1458, opened 2022-10-03
AGENT-331: Add support for MCE disconnected scenario

lranjbar:
These changes mirror MCE using `oc mirror` and updates registries.conf accordingly.

This PR is based off #1435 and is verified to work with the IPv4 disconnected scenario.

AGENT-331

### DIFF
--- a/agent/04_agent_configure.sh
+++ b/agent/04_agent_configure.sh
@@ -272,56 +272,9 @@ spec:
   releaseImage: $(getReleaseImage)
 EOF
 
-if [ ! -z "${MIRROR_IMAGES}" ]; then
-
+if [[ ! -z "${MIRROR_IMAGES}" ]] || [[ ! -z "${OC_MIRROR}" ]]; then
    # Set up registries.conf and ca-bundle.crt for mirroring
-   if [[ -z "${OC_MIRROR}" ]] ; then
-
-      cat > "${MIRROR_PATH}/registries.conf" << EOF
-[[registry]]
-prefix = ""
-location = "registry.ci.openshift.org/ocp/release"
-mirror-by-digest-only = false
-
-[[registry.mirror]]
-location = "${LOCAL_REGISTRY_DNS_NAME}:${LOCAL_REGISTRY_PORT}/${MIRROR_IMAGE_URL_SUFFIX}"
-
-[[registry]]
-prefix = ""
-location = "quay.io/openshift-release-dev/ocp-v4.0-art-dev"
-mirror-by-digest-only = false
-
-[[registry.mirror]]
-location = "${LOCAL_REGISTRY_DNS_NAME}:${LOCAL_REGISTRY_PORT}/${MIRROR_IMAGE_URL_SUFFIX}"
-EOF
-   else
-
-      cat > "${MIRROR_PATH}/registries.conf" << EOF
-[[registry]]
-prefix = ""
-location = "quay.io/openshift-release-dev/ocp-release"
-mirror-by-digest-only = false
-
-[[registry.mirror]]
-location = "${LOCAL_REGISTRY_DNS_NAME}:${LOCAL_REGISTRY_PORT}/openshift/release-images"
-
-[[registry]]
-prefix = ""
-location = "quay.io/openshift-release-dev/ocp-v4.0-art-dev"
-mirror-by-digest-only = false
-
-[[registry.mirror]]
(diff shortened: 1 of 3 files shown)


agent/04_agent_configure.sh
```diff
@@ -522,6 +504,10 @@ if [ ! -z "${MIRROR_IMAGES}" ]; then
 
 fi
 
+if [ "${OC_MIRROR}" == "true " && "${MIRROR_MCE}" == "true " ]; then
```
celebdor, line 507: https://www.shellcheck.net/wiki/SC2107

Since we are using bash, either

```suggestion
if [[ "${OC_MIRROR}" == "true " && "${MIRROR_MCE}" == "true " ]]; then
```

or

```suggestion
if [ "${OC_MIRROR}" == "true " ] && [ "${MIRROR_MCE}" == "true " ]; then
```

celebdor (changes requested):

agent/04_agent_configure.sh
```diff
@@ -522,6 +504,10 @@ if [ ! -z "${MIRROR_IMAGES}" ]; then
 
 fi
 
+if [ "${OC_MIRROR}" == "true " ] && [  "${MIRROR_MCE}" == "true " ]; then
```
andfasano, line 507: Right now disconnected scenario is identified by the config variable `MIRROR_IMAGES`, so the decision of mirroring (or not) the dependencies required by MCE should be made primarily using such var.
After that, the decision on which approach to be used for performing the mirror could be done by evaluating `OC_MIRROR`. Eventually, if we don't want to support the legacy way of mirroring the build then should fail in that case.

andfasano, line 507: Guess it was a typo: `MIRROR_MCE` --> `AGENT_DEPLOY_MCE`

agent/assets/agent-create-registries-conf-playbook.yaml
```diff
@@ -0,0 +1,19 @@
+- name: Create registries.conf for Agent Installer
+  hosts: localhost
+  collections:
+   - community.general
+  gather_facts: no
+  vars:
+    - mirror_images: "{{ lookup('env', 'MIRROR_IMAGES') }}"
+    - oc_mirror: "{{ lookup('env', 'OC_MIRROR') }}"
+    - mirror_path: "{{ lookup('env', 'MIRROR_PATH') }}"
```
andfasano, line 9: Looks like that `MIRROR_PATH` needs to be exported

agent/04_agent_configure.sh
bfournie: I believe that mce requires the OC_MIRROR implementation as it allows an ImageSet to be build with the proper operators.

agent/04_agent_configure.sh
```diff
@@ -272,56 +272,9 @@ spec:
...
-[[registry]]
-prefix = ""
-location = "registry.redhat.io/ubi8"
-mirror-by-digest-only = false
-
-[[registry.mirror]]
-location = "${LOCAL_REGISTRY_DNS_NAME}:${LOCAL_REGISTRY_PORT}/ubi8"
-EOF
-   fi
+  ansible-playbook "${SCRIPTDIR}/agent/assets/agent-create-registries-conf-playbook.yaml"
```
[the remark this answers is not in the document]
lranjbar: Yeah I think we could start to move the other files into templates and playbooks. 😄  I wanted to simplify the if statements we had here.

agent/04_agent_configure.sh
lranjbar: I'll make a follow up PR to try to refactor to the following:

```
export MIRROR_IMAGES=true
export MIRROR_OLM=multicluster-engine,local-storage-operator
export OC_MIRROR=true
export AGENT_DEPLOY_MCE=true
```

agent/assets/registries_conf.j2
```diff
@@ -0,0 +1,69 @@
+{# base oc mirror registries.conf #}
```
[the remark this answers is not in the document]
lranjbar: This would used in the disconnected ZTP scenario the most. Maybe I can move it there. It would also be needed if an install-config leaves out the ICSP section too.

agent/assets/agent-create-registries-conf-playbook.yaml
lranjbar: Will address in a follow up PR

PR #1459, opened 2022-10-04
Bump OPENSHIFT_RELEASE_STREAM to 4.12

derekhiggins:
This changes the default development version to 4.12

PR #1462, opened 2022-10-10
Set default release to 4.12

cybertron:
4.12 is the current development version and some of the default branches in other repos don't work correctly with 4.11, which causes wasted time debugging version mismatches.


dougsland:
huge +1

dougsland:
looks like this one became a dup of https://github.com/openshift-metal3/dev-scripts/pull/1459

cybertron:
Ah, awesome. This is already done. :-)

PR #1463, opened 2022-10-10
Set default release to 4.13

cybertron:


cybertron:
/hold

This shouldn't go in until 4.12 branches are cut, but since that's happening in the relatively near future I thought I'd get it proposed.

cybertron:
This was done elsewhere.

PR #1466, opened 2022-10-18
For agent, if RELEASE_IMAGE is a digest don't set override

bfournie:
if the RELEASE_IMAGE is a digest it should be used as is since registries.conf will have mirror-by-digest-only=true.


bfournie:
Included in https://github.com/openshift-metal3/dev-scripts/commit/9f804d2332a6d75737fab92017def419c8f7d18e
/close

PR #1468, opened 2022-10-20
Avoid unbound variable with default IP_STACK

cybertron:
We have a default set for IP_STACK, but it isn't taking effect soon enough for common.sh. Make the reference -u safe so we don't fail on default configs.


bfournie:
Thanks Ben
/lgtm

cybertron:
/approve

Given the triviality of the change I think it's safe enough to self-approve.

PR #1469, opened 2022-10-21
Check for existing cluster before deployment

cybertron:
One thing I've noticed is that a lot of new dev-scripts users don't realize they need to run make clean between deployments. This adds a check for the ocp/<cluster name> directory before deployment and stops if it already exists. That directory is created pretty early in the process and removed by make clean so it should be a good indicator of whether a deployment has already been run. It's also specific to a given cluster so it shouldn't affect environments that are running multiple clusters of different names.

### DIFF
--- a/sanitychecks.sh
+++ b/sanitychecks.sh
@@ -5,6 +5,13 @@ set -euxo pipefail
 # The minimum amount of space required for a default installation, expressed in GB
 MIN_SPACE_REQUIRED=${MIN_SPACE_REQUIRED:=80}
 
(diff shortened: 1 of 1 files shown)


cybertron:
/approve

The bond job failed well after this new code would have run and it's a pretty minimal change so I think we can merge it now.

cybertron:

That job is perma-failing and almost certainly wasn't broken by this change.

cybertron:

Not sure why this is still blocking...

PR #1471, opened 2022-10-24
AGENT-392: Add support for install-config.yaml/agent-config.yaml

bfournie:
Add support for install-config.yaml and agent-config.yaml for agent deployments and use it by default. ZTP manifests can still be used by setting the new 'AGENT_USE_ZTP_MANIFESTS' to true.  This uses ansible and jinja to create the yaml files files based on https://github.com/openshift-metal3/dev-scripts/pull/1458.

### DIFF
--- a/agent/04_agent_configure.sh
+++ b/agent/04_agent_configure.sh
@@ -412,23 +412,20 @@ function generate_install_agent_config() {
   mkdir -p ${INSTALL_CONFIG_PATH}
 
   # set arrays as strings to pass in env
-  nodes_macs=$(printf '%s#' "${AGENT_NODES_MACS[@]}")
-  export AGENT_NODES_MACS_STR=${nodes_macs::-1}
-  nodes_ips=$(printf '%s#' "${AGENT_NODES_IPS[@]}")
+  nodes_ips=$(printf '%s,' "${AGENT_NODES_IPS[@]}")
   export AGENT_NODES_IPS_STR=${nodes_ips::-1}
-  nodes_ipsv6=$(printf '%s#' "${AGENT_NODES_IPSV6[@]}")
+  nodes_ipsv6=$(printf '%s,' "${AGENT_NODES_IPSV6[@]}")
   export AGENT_NODES_IPSV6_STR=${nodes_ipsv6::-1}
-  nodes_hostnames=$(printf '%s#' "${AGENT_NODES_HOSTNAMES[@]}")
+  nodes_macs=$(printf '%s,' "${AGENT_NODES_MACS[@]}")
+  export AGENT_NODES_MACS_STR=${nodes_macs::-1}
+  nodes_hostnames=$(printf '%s,' "${AGENT_NODES_HOSTNAMES[@]}")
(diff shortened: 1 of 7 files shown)


agent/04_agent_configure.sh
```diff
@@ -511,11 +430,48 @@ if [[ "${NUM_MASTERS}" > "1" ]]; then
...
+  mkdir -p ${MIRROR_PATH}
+fi
+
+if [[ ${AGENT_USE_ZTP_MANIFESTS} == true ]]; then
+
+   generate_cluster_manifests
+
+ else
+
+  INSTALL_CONFIG_PATH="${OCP_DIR}"
```
andfasano, line 447: nit: would be nice to have a function like `generate_install_agent_config` for this block

agent/assets/installconfig/install-agent-config-playbook.yaml
```diff
@@ -0,0 +1,53 @@
...
+    - networking_mode: "{{ lookup('env', 'NETWORKING_MODE') }}"
+    - ssh_pub_key: "{{ lookup('env', 'SSH_PUB_KEY') }}"
+    - pull_secret: "{{ lookup('env', 'PULL_SECRET_FILE') }}"
+    - pull_secret_contents: "{{ lookup('file', pull_secret) | to_json }}"
+    - mirror_images: "{{ lookup('env', 'MIRROR_IMAGES') }}"
+    - mirror_path: "{{ lookup('env', 'MIRROR_PATH') }}"
+    - mirror_command: "{{ lookup('env', 'MIRROR_COMMAND') }}"
+    - local_image_url_suffix: "{{ lookup('env', 'LOCAL_IMAGE_URL_SUFFIX') }}"
+    - local_registry_dns_name: "{{ lookup('env', 'LOCAL_REGISTRY_DNS_NAME') }}"
+    - local_registry_port: "{{ lookup('env', 'LOCAL_REGISTRY_PORT') }}"
```
andfasano, line 39: Given that as a first step to replace the bash config files generation with an ansible playbook this approach looks fine, I'm wondering if we could reduce the number of input vars required and "push" some of them within the playbook (like for example the `AGENT_NODES_IPS` or, more in general, some of the logic within `get_static_ips_and_macs`).

agent/04_agent_configure.sh
bfournie: Good point, added

agent/assets/installconfig/install-agent-config-playbook.yaml
bfournie: Yes I'd like to clean up the vars but now these are still used in 04_agent_configure.sh to set up the ZTP manifests. The next step will be to move the ZTP manifest setup to ansible also, at which point we need to make these settings common, via a vars/main.yaml file. So I'd like to push that off to phase2.

lranjbar: I agree that we should look at cleaning up the variables in a follow up PR. Maybe once we have more of these files moved into ansible and out of the bash script.

lranjbar (approved):

andfasano:
The current serial-ipv4 failure appears to be not related to the current patch

PR #1472, opened 2022-10-25
agent: new step to assembly an ephemeral release payload

andfasano:
This patch adds a new step just for the agent flow (a simplified version of the [04_setup_ironic.sh](https://github.com/openshift-metal3/dev-scripts/blob/master/04_setup_ironic.sh)), so that it would be possible to package a local ehemeral release, comprised of one (or more) locally rebuilt images.
This feature could be useful when testing multiple changes impacting different release images at once; for the agent flow it becomes useful also when testing changes in the openshift-installer required by the assisted-service.


agent/04_agent_prepare_release.sh
```diff
@@ -0,0 +1,96 @@
...
+# export INSTALLER_LOCAL_REPO=~/go/src/github.com/openshift/installer
+# export INSTALLER_DOCKERFILE=images/installer/Dockerfile.ci
+
+early_deploy_validation
+write_pull_secret
+
+# Release mirroring could be required by the subsequent steps
+# even if the current one will be skipped
+if [[ ! -z "${MIRROR_IMAGES}" ]]; then
+   setup_release_mirror
```
bfournie, line 32: I would expect. particularly for REGISTRY_BACKEND=quay that would need to call `setup_local_registry` first. I think for podman backend that it will be called in 02_configure_host.sh

agent/05_agent_configure.sh
```diff
@@ -497,11 +498,6 @@ get_static_ips_and_macs
 
 
 if [[ ! -z "${MIRROR_IMAGES}" ]]; then
-
-    setup_local_registry
```
bfournie, line 501: See comment in 04_agent_prepare_release about calling setup_local_registry

agent/04_agent_prepare_release.sh
andfasano: The local registry is a sort of global shared resource that may be needed by different steps in the workflow. I think we should try to keep having, also for the sake of simplicity, an (early) single point of instantiation for the local registry regardless of the underlying backend (and the `02_configure_host.sh` seems a good fit).
Also mirroring the release it's something that could be required by different steps, so that's why I moved it here up in the chain

bfournie:
I've been able to successfully test both the installer and assisted-service repos with this PR, the config-<user> I used includes this, for example:
```
export INSTALLER_REPO_PR=6530
export AGENT_INSTALLER_API_SERVER_LOCAL_REPO=/home/stack/assisted-service
export AGENT_INSTALLER_API_SERVER_DOCKERFILE=Dockerfile.assisted-service.ocp
export MIRROR_IMAGES=true
```
The only change that's needed is for setup_local_registry https://github.com/openshift-metal3/dev-scripts/blob/master/02_configure_host.sh#L71-L73 to also be called when using the quay registry backend.

avishayt:
Add some docs to agent/README.md?

agent/04_agent_prepare_release.sh
```diff
@@ -0,0 +1,96 @@
...
+fi
+
+function build_local_release() {
+    # Sanity checks
+    if [ -z "${MIRROR_IMAGES}" ] ; then 
+        echo "Please set MIRROR_IMAGES to rebuild a local release"
+        exit 1
+    fi
+
+    if [[ ! "$OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE" =~ "$LOCAL_REGISTRY_DNS_NAME" ]] ; then
```
bfournie, line 42: I wonder why this check is necessary. Doesn't it conflict with this setting https://github.com/openshift-metal3/dev-scripts/blob/master/common.sh#L206 ?

andfasano: It should not. It's a sanity check to ensure that the subsequent [push operation](https://github.com/openshift-metal3/dev-scripts/pull/1472/files/e1b28eea8bb8be3d1aee47602d39c07c8daafee3#diff-a53878775eb60ffffce541ffc1a8f37dcb1aeb39b6526e93b733a33fd510c0b4R87) will point to the local registry

bfournie:
/lgtm

I've been testing with this fix, its been very useful to test changes to the assisted-service and installer at the same time.

agent/04_agent_prepare_release.sh
```diff
@@ -0,0 +1,96 @@
...
+SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )/.." && pwd )"
+
+LOGDIR=${SCRIPTDIR}/logs
+source $SCRIPTDIR/logging.sh
+source $SCRIPTDIR/common.sh
+source $SCRIPTDIR/network.sh
+source $SCRIPTDIR/utils.sh
+source $SCRIPTDIR/validation.sh
+source $SCRIPTDIR/agent/common.sh
+source $SCRIPTDIR/ocp_install_env.sh
```
bfournie, line 13: Also need to include here:
source $SCRIPTDIR/oc_mirror.sh

To handle the case where:
export MIRROR_COMMAND=oc-mirror

Otherwise get a failure:
/home/stack/dev-scripts-latest/utils.sh: line 403: setup_oc_mirror: command not found

bfournie:
Confirmed both with using local repos and with using upstream, with multiple config scenarios.

/lgtm

02_configure_host.sh
```diff
@@ -68,7 +68,7 @@ fi
 
 export ANSIBLE_FORCE_COLOR=true
 
-if use_podman_registry; then
+if [[ ! -z "${MIRROR_IMAGES}" || $(env | grep "_LOCAL_IMAGE=")  || ! -z "${ENABLE_CBO_TEST:-}" || ! -z "${ENABLE_LOCAL_REGISTRY}" ]]; then
```
bfournie, line 71: With this patch when I set
export REGISTRY_BACKEND=quay
I got the error below.

Can you add to this file:
`source oc_mirror.sh`

+(./02_configure_host.sh:72): main(): setup_local_registry
+(utils.sh:534): setup_local_registry(): [[ quay = \p\o\d\m\a\n ]]
+(utils.sh:537): setup_local_registry(): setup_quay_mirror_registry
utils.sh: line 537: setup_quay_mirror_registry: command not found
+(utils.sh:1): setup_local_registry(): removetmp
+(utils.sh:668): removetmp(): '[' -n ' /tmp/test-token--hLZwRQT5DH' ']'
+(utils.sh:668): removetmp(): rm -rf /tmp/test-token--hLZwRQT5DH
make: *** [Makefile:39: configure] Error 127

bfournie:
Including oc_mirror.sh with its settings exposed that some uninitialized env variables since oc_mirror.sh uses:
`set -euxo pipefail`

common.sh should include the following:
```
export ENABLE_LOCAL_REGISTRY=${ENABLE_LOCAL_REGISTRY:-}

export DISABLE_MULTICAST=${DISABLE_MULTICAST:-false}

```

common.sh
```diff
@@ -356,6 +356,9 @@ if [[ $DISTRO == "almalinux8" || $DISTRO == "rocky8" ]]; then
     ALMA_PYTHON_OVERRIDE="-e ansible_python_interpreter=/usr/libexec/platform-python"
 fi
 
+# Defaults the DISABLE_MULTICAST variable
+export DISABLE_MULTICAST=false
+
```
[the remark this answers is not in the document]
andfasano: Thanks for spotting it @bfournie!

utils.sh
```diff
@@ -674,15 +674,4 @@ function auth_template_and_removetmp(){
     removetmp
 }
 
-use_podman_registry() {
```
bfournie, line 677: Maybe I'm missing something but isn't this function used twice in 02_configure_host.sh
https://github.com/openshift-metal3/dev-scripts/blob/master/02_configure_host.sh#L71 AND
https://github.com/openshift-metal3/dev-scripts/blob/master/02_configure_host.sh#L285

And its use was only removed at line 71 so the function can't be removed yet (or the other use needs to be changed).

Not sure if that is causing any problems although seeing a lot of strange test failures.

bfournie:
In addition to the [issue](https://github.com/openshift-metal3/dev-scripts/pull/1472#discussion_r1042742620 ) with missing use_podman_registry() I found another issue with the registries not being set up properly when using
`export MIRROR_COMMAND=oc-mirror`

I've made these changes to get it to work. I could add these in in a follow-on PR if you want to merge this one first
```
$ git diff oc_mirror.sh
diff --git a/oc_mirror.sh b/oc_mirror.sh
index 304c587..ae4f5a3 100755
--- a/oc_mirror.sh
+++ b/oc_mirror.sh
@@ -10,9 +10,6 @@ function add_auth_to_pull_secret() {
 
    tmpauthfile=$(mktemp --tmpdir "quayauth--XXXXXXXXXX")
    _tmpfiles="$_tmpfiles $tmpauthfile"
-   tmppullsecret=$(mktemp --tmpdir "pullsecret--XXXXXXXXXX")
-   _tmpfiles="$_tmpfiles $tmppullsecret"
-
 
    cat > "${tmpauthfile}" << EOF
 {
@@ -24,9 +21,7 @@ function add_auth_to_pull_secret() {
 }
 EOF
 
-   jq -s '.[0] * .[1]' ${tmpauthfile} ${PULL_SECRET_FILE} > ${tmppullsecret}
-   cp ${tmppullsecret} ${PULL_SECRET_FILE}
-
+   cp ${tmpauthfile} ${REGISTRY_CREDS}
 }
```
 ```
$ git diff utils.sh
diff --git a/utils.sh b/utils.sh
index 7fc8a44..e3fa312 100755
--- a/utils.sh
+++ b/utils.sh
@@ -685,4 +685,15 @@ use_podman_registry() {
   return 1
 }
 
+use_quay_registry() {
+  # return true if using the local registry and the backend is podman
+  if [[ ! -z "${MIRROR_IMAGES}" || $(env | grep "_LOCAL_IMAGE=")  || ! -z "${ENABLE_CBO_TEST:-}" || ! -z "${ENABLE_LOCAL_REGISTRY}" ]]; then
+     if [[ "${REGISTRY_BACKEND}" = "quay" ]]; then
+        return 0
+     fi
+  fi
+
+  return 1
+}
```
```
$ git diff 02_configure_host.sh
diff --git a/02_configure_host.sh b/02_configure_host.sh
index 4d3d38f..31714b3 100755
--- a/02_configure_host.sh
+++ b/02_configure_host.sh
@@ -279,15 +280,15 @@ fi
 sudo sed -i "/${LOCAL_REGISTRY_DNS_NAME}/d" /etc/hosts
 echo "${PROVISIONING_HOST_EXTERNAL_IP} ${LOCAL_REGISTRY_DNS_NAME}" | sudo tee -a /etc/hosts
 
-# Remove any previous file, or podman login panics when reading the
-# blank authfile with a "assignment to entry in nil map" error
-rm -f ${REGISTRY_CREDS}
 if use_podman_registry; then
+    # Remove any previous file, or podman login panics when reading the
+    # blank authfile with a "assignment to entry in nil map" error
+    rm -f ${REGISTRY_CREDS}
     # create authfile for local registry
     sudo podman login --authfile ${REGISTRY_CREDS} \
         -u ${REGISTRY_USER} -p ${REGISTRY_PASS} \
         ${LOCAL_REGISTRY_DNS_NAME}:${LOCAL_REGISTRY_PORT}
-else
+elif ! use_quay_registry; then
```

02_configure_host.sh
andfasano: Thanks for detecting it @bfournie! I will for the moment revert back to the usage of the podman registry, which in any case will be sufficient for covering the local builds

PR #1474, opened 2022-11-03
Add temporary support for mce installation

mhanss:
LSO has not been published to the 4.12 redhat-operators catalog, so it cannot be installed on OpenShift 4.12. Until this is resolved, installing LSO with 4.11 catalog as redhat-operators-v4-11.

### DIFF
--- a/agent/04_agent_configure.sh
+++ b/agent/04_agent_configure.sh
@@ -461,7 +461,11 @@ EOF
 function oc_mirror_mce() {
    tmpimageset=$(mktemp --tmpdir "mceimageset--XXXXXXXXXX")
    _tmpfiles="$_tmpfiles $tmpimageset"
-
+   # TODO: Remove this once LSO is published to the 4.12 catalog.
+   local version="$(openshift_version ${OCP_DIR})"
+   if [[ ( -n "${AGENT_DEPLOY_MCE}" ) && ( ${version} == "4.12" || ${version} == "4.13" ) ]]; then
+     local OPENSHIFT_RELEASE_STREAM="4.11"
+   fi
    cat > "${tmpimageset}" << EOF
 ---
 apiVersion: mirror.openshift.io/v1alpha2
--- a/agent/05_agent_create_cluster.sh
+++ b/agent/05_agent_create_cluster.sh
@@ -102,9 +102,81 @@ function wait_for_cluster_ready() {
   echo "Cluster is ready!"
 }
 
+function wait_for_operator() {
+    subscription="$1"
+    namespace="${2:-}"
+    echo "Waiting for operator ${subscription} to get installed on namespace ${namespace}..."
+
+    for _ in $(seq 1 60); do
+        csv=$(oc -n "${namespace}" get subscription "${subscription}" -o jsonpath='{.status.installedCSV}' || true)
+        if [[ -n "${csv}" ]]; then
+            if [[ "$(oc -n "${namespace}" get csv "${csv}" -o jsonpath='{.status.phase}')" == "Succeeded" ]]; then
+                echo "ClusterServiceVersion (${csv}) is ready"
+                return 0
+            fi
+        fi
+
+        sleep 10
+    done
+
+    echo "Timed out waiting for csv to become ready!"
+    return 1
+}
+
+function install_lso_operator_latest_available() {
+  local version="$(openshift_version ${OCP_DIR})"
+  local operator_version="4.11"
+  local catalog_source_name=redhat-operators-v${operator_version/./-}
+
+  if [[ ( -n "${AGENT_DEPLOY_MCE}" ) && ( ${version} == "4.12" || ${version} == "4.13" ) ]]; then
+    cat > "${mceManifests}/catalog_source_lso_operator.yaml" << EOF
+kind: CatalogSource
+apiVersion: operators.coreos.com/v1alpha1
+metadata:
+  name: ${catalog_source_name}
+  namespace: openshift-marketplace
+spec:
+  displayName: Red Hat Operators v${operator_version}
+  image: registry.redhat.io/redhat/redhat-operator-index:v${operator_version}
+  priority: -100
+  publisher: Red Hat
+  sourceType: grpc
+  updateStrategy:
+    registryPoll:
+      interval: 10m0s
+EOF
+    apply_manifest "$mceManifests/catalog_source_lso_operator.yaml"
+    oc delete subscription local-storage-operator -n openshift-local-storage
+    cat > "${mceManifests}/lso_operator.yaml" << EOF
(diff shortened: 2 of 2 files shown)


agent/04_agent_configure.sh
```diff
@@ -461,7 +461,10 @@ EOF
 function oc_mirror_mce() {
    tmpimageset=$(mktemp --tmpdir "mceimageset--XXXXXXXXXX")
    _tmpfiles="$_tmpfiles $tmpimageset"
-
+   # TODO: Remove this once LSO is published to the 4.12 catalog.
+   if [[ ${OPENSHIFT_RELEASE_STREAM} == "4.12" || ${OPENSHIFT_RELEASE_STREAM} == "4.13" ]]; then
```
bfournie, line 465: Need to add another conditional so this is only changed when using MCE. Also it would be useful to log a warning message so its obvious that the release stream being used is not the one the user set
if [ ! -z "${AGENT_DEPLOY_MCE}" ]; then

agent/05_agent_create_cluster.sh
```diff
@@ -102,9 +102,60 @@ function wait_for_cluster_ready() {
...
+    done
+
+    echo "Timed out waiting for csv to become ready!"
+    return 1
+}
+
+function create_lso_operator_v4_11() {
+  if [[ ${OPENSHIFT_RELEASE_STREAM} == "4.12" || ${OPENSHIFT_RELEASE_STREAM} == "4.13" ]]; then
+    oc delete subscription local-storage-operator -n openshift-local-storage
+    cat > "${mceManifests}/lso_operator_4_11.yaml" << EOF
```
bfournie, line 129: I'd remove the 4_11 from the name of this file so it be made more generic if you need to this same mechanism for future releases

agent/05_agent_create_cluster.sh
```diff
@@ -102,9 +102,60 @@ function wait_for_cluster_ready() {
...
+metadata:
+ name: local-storage-operator
+ namespace: openshift-local-storage
+spec:
+ installPlanApproval: Automatic
+ name: local-storage-operator
+ source: redhat-operators-v4-11
+ sourceNamespace: openshift-marketplace
+EOF
+    apply_manifest "$mceManifests/lso_operator_4_11.yaml"
```
bfournie, line 150: also remove 4_11 name here

agent/05_agent_create_cluster.sh
```diff
@@ -102,9 +102,60 @@ function wait_for_cluster_ready() {
...
+---
+apiVersion: operators.coreos.com/v1alpha1
+kind: Subscription
+metadata:
+ name: local-storage-operator
+ namespace: openshift-local-storage
+spec:
+ installPlanApproval: Automatic
+ name: local-storage-operator
+ source: redhat-operators-v4-11
```
bfournie, line 147: should use a variable for "4-11" so can use for future releases

agent/05_agent_create_cluster.sh
```diff
@@ -102,9 +102,60 @@ function wait_for_cluster_ready() {
...
+    apply_manifest "$mceManifests/lso_operator_4_11.yaml"
+    wait_for_operator "local-storage-operator" "openshift-local-storage"
+  fi
+}
+
 function mce_prepare_postinstallation_manifests() {
   local mceManifests=$1
-
+  # TODO: Remove this once LSO is published to the 4.12 catalog.
+  create_lso_operator_v4_11
```
bfournie, line 158: If we think this same issue can happen in the future, e.g. when 4.13 is being released and its operator is not released yet we should make this generic, for example, call it "create_lp_operator_latest_available' and use a variable to contain the latest - 4.11 in this case

agent/04_agent_configure.sh
mhanss: Condition added!

agent/05_agent_create_cluster.sh
mhanss: Changes have been implemented!

lranjbar (changes requested):

lranjbar:
Update: After talking about this with the team we will first take this approach and move this job to a weekly with a follow up approach to try to solving the underlying problem.

agent/05_agent_create_cluster.sh
```diff
@@ -102,9 +102,81 @@ function wait_for_cluster_ready() {
...
+  displayName: Red Hat Operators v${operator_version}
+  image: registry.redhat.io/redhat/redhat-operator-index:v${operator_version}
+  priority: -100
+  publisher: Red Hat
+  sourceType: grpc
+  updateStrategy:
+    registryPoll:
+      interval: 10m0s
+EOF
+    apply_manifest "$mceManifests/catalog_source_lso_operator.yaml"
```
andfasano, line 148: Any specific reason for not having it as a manifest asset directly under `/agent/asset/mce` folder? Otherwise, also for consistency with the other manifests, I'd prefer to have it there.
I guess it could also be added directly as an extra manifest.

agent/05_agent_create_cluster.sh
```diff
@@ -102,9 +102,81 @@ function wait_for_cluster_ready() {
...
+---
+apiVersion: operators.coreos.com/v1alpha1
+kind: Subscription
+metadata:
+ name: local-storage-operator
+ namespace: openshift-local-storage
+spec:
+ installPlanApproval: Automatic
+ name: local-storage-operator
+ source: ${catalog_source_name}
```
andfasano, line 168: I guess this could be added directly in the existing manifest, in this way it seems it could be possible to simplify the code

andfasano (commented):
Thanks @mhanss for your contribution. In general, I think it would be simpler to directly hard-code the "working" configuration in the existing manifests - so that we'll have the minimum impact possible on the existing code.
Also, it would be simpler in future to throw away such temporary change once the catalog will be working.

agent/05_agent_create_cluster.sh
celebdor: I think putting it in /agent/asset/mce makes more sense

mhanss: @andfasano @celebdor As per the current implementation, we only need to remove the code from the current file but when we will move the files to the `mce` directory then we need to remove code/files from multiple places when the operator would be available to the 4.12 catalog. Please let me know your thoughts, I am okay with both approaches.

andfasano: @bfournie already started an activity to move the templating via Ansible, so it'd be better if we could stay on that side for the new developments (and keep away from bash). Wouldn't be sufficient just to define the source conditionally in this case?

bfournie:
Closing this as its no longer needed.
/close

PR #1476, opened 2022-11-04
Fix Assisted Service cleanup script

mkowalski:
This PR fixes the script responsible for cleaning up the Assisted
Service deployment. It previously relied on the variable
`NODES_PLATFORM` that has a value of `libvirt` and is unrelated
to how the Assisted Service is deployed. With this change the
Make target `assisted_deployment_cleanup` works correctly again.

This PR changes the generic `clean` target as Assisted cleanup is
redundant at this point as underlying OCP cluster has already been
deleted.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -50,7 +50,7 @@ ocp_run:
 gather:
 	./must_gather.sh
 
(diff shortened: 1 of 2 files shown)


cybertron:
/lgtm
/retest

Looks like some sort of infra failure on all the jobs.

elfosardo:
@mkowalski this looks stuck, you want to try and rebase it or we can close it?

PR #1477, opened 2022-11-08
WIP webui: Allow embedding a webui service

celebdor:
dev-scripts currently supports running webui on node0 by sshing into it and using podman to run a webui image. This commits adds the ability to embed the logic to run the webui in the image itself, so it will run as soon as possible.

There should be some follow-up work to embed the image itself into the ISO to allow for disconnected. Until then, one could put the image reference to point to dev-scripts registry.


bfournie:
Closing this as its 2 years old. Let's reopen if needed in future.
/close

PR #1478, opened 2022-11-15
AGENT-429: Dynamically generate mirror configuration for agent

bfournie:
The mirror registries configuration for the agent is currently hardcoded. This is incorrect when a particular release image digest is defined, for example here is the ICSP for recent 4.12 image when using:
OPENSHIFT_RELEASE_IMAGE=registry.ci.openshift.org/ocp/release@sha256:508e696f24fc844cbbf9599ee6516c2f76d032f363a041df0441e3629ce64807

imageContentSources:
- mirrors:
  - virthost.ostest.test.metalkube.org:5000/localimages/local-release-image 
 source: registry.ci.openshift.org/ocp/4.12-2022-11-11-131611
- mirrors:
  - virthost.ostest.test.metalkube.org:5000/localimages/local-release-image 
 source: registry.ci.openshift.org/ocp/release

The mirror registries should be extracted from the log files when the 'oc adm release mirror' or 'oc mirror' commands are run. This will be correct for all images and is easier to maintain.

For the ZTP manifests an extra conversion is needed to put it in the expected format for registries.conf.

### DIFF
--- a/agent/assets/installconfig/install-config_yaml.j2
+++ b/agent/assets/installconfig/install-config_yaml.j2
@@ -70,18 +70,8 @@ pullSecret: {{ pull_secret_contents }}
 sshKey: {{ ssh_pub_key }} 
 {% if mirror_images == "true" %}
 imageContentSources:
-{% if mirror_command == "oc-mirror" %}
-- mirrors:
-    - "{{ local_registry_dns_name }}:{{ local_registry_port }}/openshift/release-images"
-  source: "quay.io/openshift-release-dev/ocp-release"
-- mirrors:
-    - "{{ local_registry_dns_name }}:{{ local_registry_port }}/openshift/release"
-  source: "quay.io/openshift-release-dev/ocp-v4.0-art-dev"
-- mirrors:
-    - "{{ local_registry_dns_name }}:{{ local_registry_port }}/ubi8"
-  source: "registry.redhat.io/ubi8"
-{% if agent_deploy_mce == "true" %}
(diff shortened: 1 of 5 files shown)


agent/assets/installconfig/install-config_yaml.j2
```diff
@@ -70,18 +70,9 @@ pullSecret: {{ pull_secret_contents }}
 sshKey: {{ ssh_pub_key }} 
 {% if mirror_images == "true" %}
 imageContentSources:
+{{ image_content_sources }}
 {% if mirror_command == "oc-mirror" %}
```
pawanpinjarkar, line 74: The two `if`s can be combined into one. Otherwise, LGTM.

agent/assets/ztp/registries_conf.j2
```diff
@@ -1,28 +1,7 @@
+{% if mirror_images == "true" %}
 {# base oc mirror registries.conf #}
+{{ registries }}
 {% if mirror_command == "oc-mirror" %}
```
pawanpinjarkar, line 4: The two `if`s can be combined into one. Otherwise, LGTM.

agent/assets/installconfig/install-config_yaml.j2
bfournie: Fixed

agent/assets/ztp/registries_conf.j2
bfournie: Fixed

PR #1479, opened 2022-11-20
AGENT-435: Create ZTP manifest files using Ansible

bfournie:
Move the creation of the ZTP manifest files to Ansible, similar to what is currently done for registries.conf and install/agent-config.yaml. A common Ansible role has been added to share vars and templates between the ZTP manifests and the install/agent-config.yaml.

A follow-on PR will add a role for mce and move files under there. In addition a follow-on PR will move the setting of variables currently done in 04_agent_configure.sh to Ansible.


lranjbar:
Hey everyone so when I started the ansible thing and added the unit test capability I made an ansible collection out of the /agent folder. Digging through this.. [actually matches the collection folder structure.](https://docs.ansible.com/ansible/latest/dev_guide/developing_collections_structure.html#collection-structure)

So no changes are needed. :) 

/lgtm

PR #1484, opened 2022-11-29
Better default for jobs names

derekhiggins:
Using the search string "nightly-<version>" on job names ensures we don't get incorrect hits for upgrade jobs (upgrading from <version> )


derekhiggins:
/override 
CI jobs don't touch this script

PR #1486, opened 2022-12-02
AGENT-350: Gather agent data with 'make agent_gather'

pawanpinjarkar:
A new target `make agent_gather` allows the user to gather useful troubleshooting data when things go wrong even before the API services are available.

Signed-off-by: Pawan Pinjarkar [ppinjark@redhat.com](mailto:ppinjark@redhat.com)

### DIFF
--- a/agent/04_agent_configure.sh
+++ b/agent/04_agent_configure.sh
@@ -268,6 +268,7 @@ else
     ip=${AGENT_NODES_IPSV6[0]}
   fi
   configure_dnsmasq ${ip} ""
+  echo "${ip}" >> ${OCP_DIR}/node0-ip
 fi
 
 generate_cluster_manifests
--- a/agent/04_agent_configure.sh
+++ b/agent/04_agent_configure.sh
@@ -268,7 +268,7 @@ else
     ip=${AGENT_NODES_IPSV6[0]}
   fi
   configure_dnsmasq ${ip} ""
-  echo "${ip}" >> ${OCP_DIR}/node0-ip
+  echo "${ip}" > ${OCP_DIR}/node0-ip
 fi
 
 generate_cluster_manifests
--- a/agent/gather.sh
+++ b/agent/gather.sh
@@ -5,21 +5,15 @@ SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )/.." && pwd )"
 
 source $SCRIPTDIR/agent/04_agent_configure.sh
 
-get_static_ips_and_macs
-
-if [[ "$IP_STACK" = "v4" ]]; then
-    ip=${AGENT_NODES_IPS[0]}
-  else
-    ip=${AGENT_NODES_IPSV6[0]}
-  fi
+ip=$(head -n 1 ${OCP_DIR}/node0-ip)
 
 ssh -o 'ConnectTimeout=30' -o 'StrictHostKeyChecking=no' -o 'UserKnownHostsFile=/dev/null' core@${ip} << EOF
 # check is agent-gather command is present
 if ! command -v /usr/local/bin/agent-gather &> /dev/null
 then
-    echo "/usr/local/bin/agent-gather could not be found."
-    exit
+    echo "Skipping gathering agent logs, agent-gather script not present."
+    exit 0
 fi
 
-sudo /usr/local/bin/agent-gather -O | tar -xJvf -
+sudo /usr/local/bin/agent-gather
 EOF
--- a/agent/gather.sh
+++ b/agent/gather.sh
@@ -1,10 +1,11 @@
 #!/usr/bin/env bash
+# shellcheck source=/dev/null
 set -euxo pipefail
 
 SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )/.." && pwd )"
 
-source $SCRIPTDIR/agent/04_agent_configure.sh
+source "$SCRIPTDIR"/agent/04_agent_configure.sh
 
-ip=$(<${OCP_DIR}/node0-ip)
+ip=$(<"${OCP_DIR}"/node0-ip)
 
-ssh -o 'ConnectTimeout=30' -o 'StrictHostKeyChecking=no' -o 'UserKnownHostsFile=/dev/null' core@${ip} if ! command -v /usr/local/bin/agent-gather '>&' /dev/null ';' then echo "Skipping gathering agent logs, agent-gather script not present." else sudo /usr/local/bin/agent-gather -o ';' fi >agent-gather.tar.xz
+ssh -o 'ConnectTimeout=30' -o 'StrictHostKeyChecking=no' -o 'UserKnownHostsFile=/dev/null' core@"${ip}" if ! command -v /usr/local/bin/agent-gather '>&' /dev/null ';' then echo "Skipping gathering agent logs, agent-gather script not present." else sudo /usr/local/bin/agent-gather -o ';' fi >agent-gather.tar.xz
--- a/agent/gather.sh
+++ b/agent/gather.sh
@@ -1,0 +1,25 @@
+#!/usr/bin/env bash
+set -euxo pipefail
+
+SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )/.." && pwd )"
+
+source $SCRIPTDIR/agent/04_agent_configure.sh
+
+get_static_ips_and_macs
+
+if [[ "$IP_STACK" = "v4" ]]; then
+    ip=${AGENT_NODES_IPS[0]}
+  else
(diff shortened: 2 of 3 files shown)


agent/gather.sh
```diff
@@ -0,0 +1,25 @@
...
+    ip=${AGENT_NODES_IPS[0]}
+  else
+    ip=${AGENT_NODES_IPSV6[0]}
+  fi
+
+ssh -o 'ConnectTimeout=30' -o 'StrictHostKeyChecking=no' -o 'UserKnownHostsFile=/dev/null' core@${ip} << EOF
+# check is agent-gather command is present
+if ! command -v /usr/local/bin/agent-gather &> /dev/null
+then
+    echo "/usr/local/bin/agent-gather could not be found."
```
andfasano, line 20: The absence of the agent-gather script it's something potentially expected, so I'd rephrase to message to something like "Skipping gathering agent logs, script not present" and using an explicit `exit 0` to manifest that's not a issue

agent/gather.sh
```diff
@@ -0,0 +1,25 @@
...
+
+source $SCRIPTDIR/agent/04_agent_configure.sh
+
+get_static_ips_and_macs
+
+if [[ "$IP_STACK" = "v4" ]]; then
+    ip=${AGENT_NODES_IPS[0]}
+  else
+    ip=${AGENT_NODES_IPSV6[0]}
+  fi
```
andfasano, line 14: This approach may certainly work, but I see a potential fragility, either because the `get_static_ips_and_macs` performs other actions either because it could change in future and have a impact here.
Since by design we decided to use the first IP calculated as the node0 ip (as correctly also adopted here), why don't we simply serialize that ip in a `node0-ip` file [when it's defined](https://github.com/openshift-metal3/dev-scripts/blob/ac7ac11880a0fe76a510fe1504ea125eca34cad4/agent/04_agent_configure.sh#L49), so that it could be reused here?

agent/gather.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+set -euxo pipefail
+
+ip=$(<node0-ip)
```
pawanpinjarkar, line 4: This needs a fix :man_facepalming:

agent/gather.sh
```diff
@@ -0,0 +1,15 @@
...
+
+ssh -o 'ConnectTimeout=30' -o 'StrictHostKeyChecking=no' -o 'UserKnownHostsFile=/dev/null' core@${ip} << EOF
+# check is agent-gather command is present
+if ! command -v /usr/local/bin/agent-gather &> /dev/null
+then
+    echo "Skipping gathering agent logs, agent-gather script not present."
+    exit 0
+fi
+
+sudo /usr/local/bin/agent-gather -O | tar -xJvf -
```
zaneb, line 14: Why are we extracting the files here instead of just providing the tarball as a CI artifact, as I believe we do with the must-gather?

pawanpinjarkar: That's just an oversight.

agent/gather.sh
```diff
@@ -0,0 +1,19 @@
...
+
+ssh -o 'ConnectTimeout=30' -o 'StrictHostKeyChecking=no' -o 'UserKnownHostsFile=/dev/null' core@${ip} << EOF
+# check is agent-gather command is present
+if ! command -v /usr/local/bin/agent-gather &> /dev/null
+then
+    echo "Skipping gathering agent logs, agent-gather script not present."
+    exit 0
+fi
+
+sudo /usr/local/bin/agent-gather
```
zaneb, line 18: We still need to pass the `-O` option, otherwise the tarball will get created on the VM, not on the hypervisor.

agent/gather.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
+set -euxo pipefail
+
+SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )/.." && pwd )"
+
+source $SCRIPTDIR/agent/04_agent_configure.sh
+
+ip=$(head -n 1 ${OCP_DIR}/node0-ip)
+
+ssh -o 'ConnectTimeout=30' -o 'StrictHostKeyChecking=no' -o 'UserKnownHostsFile=/dev/null' core@${ip} << EOF
```
zaneb, line 10: We'll need to redirect the stdout (since we will pass the `-O` flag to agent-gather, see below) to a tarfile.

I tried this out and it turns out that passing the command to ssh on stdin (using `<<EOF`) means this is treated as an interactive shell, and that means the motd is prepended to the output. To avoid that, you have to write the command to be executed as CLI arguments.
It is just possible to write this same code as one massive command line:

```
ssh -o 'ConnectTimeout=30' -o 'StrictHostKeyChecking=no' -O 'UserKnownHostsFile=/dev/null' core@${ip} if ! command -v /usr/local/bin/agent-gather '>&' /dev/null ';' then echo "Skipping..." '>&2;' else sudo /usr/local/bin/agent-gather -o ';' fi >agent-gather.tar.xz
```

 but it might be better to split up the logic into 2 ssh calls, and have the if statement in the script here rather than the remote one.
The exit code if the command does not exist will be 127, so in theory we could exploit that to do only one ssh call.

agent/gather.sh
```diff
@@ -0,0 +1,19 @@
...
+# check is agent-gather command is present
+if ! command -v /usr/local/bin/agent-gather &> /dev/null
+then
+    echo "Skipping gathering agent logs, agent-gather script not present."
```
zaneb, line 14: Another problem with running this on the VM is that when we redirect the output, this text will end up in the tarfile. This can be avoided by adding `>&2` (and that's probably worth doing anyway), although it won't be required if we move this code into the local script as suggested above.

agent/04_agent_configure.sh
```diff
@@ -268,6 +268,7 @@ else
     ip=${AGENT_NODES_IPSV6[0]}
   fi
   configure_dnsmasq ${ip} ""
+  echo "${ip}" >> ${OCP_DIR}/node0-ip
```
zaneb, line 271: Any reason we are appending (`>>`) instead of replacing (`>`) here?

pawanpinjarkar: Agree `>` is better here else I have to make sure only the first item from the node0-ip file is used because of the unnecessary appending with `>>` and had to do `ip=$(head -n 1 ${OCP_DIR}/node0-ip)`. With `>` we can rely on `ip=$(<${OCP_DIR}/node0-ip)`. Thank you.

dougsland:
I would also recommend run `spellcheck` tool to your scripts. Usually helps bring fixes and standard to the code too.

agent/gather.sh
```diff
@@ -0,0 +1,11 @@
...
+# shellcheck source=/dev/null
+set -euxo pipefail
+
+SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )/.." && pwd )"
+
+source "$SCRIPTDIR"/agent/04_agent_configure.sh
+
+ip=$(<"${OCP_DIR}"/node0-ip)
+
+ssh -o 'ConnectTimeout=30' -o 'StrictHostKeyChecking=no' -o 'UserKnownHostsFile=/dev/null' core@"${ip}" if ! command -v /usr/local/bin/agent-gather '>&' /dev/null ';' then echo "Skipping gathering agent logs, agent-gather script not present." else sudo /usr/local/bin/agent-gather -o ';' fi >agent-gather.tar.xz
```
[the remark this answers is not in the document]
pawanpinjarkar: I thought the capital O was a typo and changed it :smile:

agent/gather.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
+# shellcheck source=/dev/null
+set -euxo pipefail
+
+SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )/.." && pwd )"
+
+source "$SCRIPTDIR"/agent/04_agent_configure.sh
```
andfasano, line 7: Sourcing a step script is usually not recommended and to be avoided, since it could run unwated pieces of code.  In this case it doesn't even seem to be necessary, a `source $SCRIPTDIR/common.sh` should be sufficient to access the OCP_DIR

agent/04_agent_configure.sh
```diff
@@ -268,6 +268,7 @@ else
     ip=${AGENT_NODES_IPSV6[0]}
   fi
   configure_dnsmasq ${ip} ""
+  echo "${ip}" > ${OCP_DIR}/node0-ip
```
zaneb, line 271: This is in the section that only runs for SNO clusters, so that's likely why the compact jobs are failing.

pawanpinjarkar: With https://github.com/openshift-metal3/dev-scripts/pull/1488 we don't need this now.

PR #1487, opened 2022-12-05
Add padding to pass rates script output

honza:
Sorry, my OCD just could not


honza:
cc @derekhiggins

dougsland:
looks sane to me.

scripts/periodics_pass_rates.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
```
zaneb, line 1: Since we're doing OCD, this file definitely contains bash-specific syntax like `$()` :slightly_smiling_face:

honza: It's not OCD; it literally didn't work on my machine :slightly_smiling_face:.  I'll change it to `#!/usr/bin/env bash`, would that be satisfactory to you?

PR #1488, opened 2022-12-08
AGENT-350: Run agent_gather on the hosts

pawanpinjarkar:
Signed-off-by: Pawan Pinjarkar <ppinjark@redhat.com>

### DIFF
--- a/agent/04_agent_configure.sh
+++ b/agent/04_agent_configure.sh
@@ -268,7 +268,6 @@ else
     ip=${AGENT_NODES_IPSV6[0]}
   fi
   configure_dnsmasq ${ip} ""
-  echo "${ip}" > ${OCP_DIR}/node0-ip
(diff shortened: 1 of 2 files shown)


pawanpinjarkar:
depends on https://github.com/openshift/installer/pull/6683

zaneb:
> depends on [openshift/installer#6683](https://github.com/openshift/installer/pull/6683)

And also a backport before we can use it in the CI jobs.

The more I think about it, the more I think the right thing for CI specifically is to run agent-gather on _all_ of the hosts. We may need e.g. the log from a host that is not node0 to debug something.

pawanpinjarkar:
> openshift/installer#6683

As this PR is now to run agent gather on all the hosts, it does not depend on https://github.com/openshift/installer/pull/6683

PR #1489, opened 2022-12-09
Treat mirror_images as a boolean in Ansible roles

zaneb:
The environment variable MIRROR_IMAGES is only checked for being empty or non-empty in bash (i.e. both "false" and "true" evaluate as true).

In the Ansible roles, mirror_images was explicitly compared to "true", which led to inconsistent behaviour. To restore a semblance of sanity, make mirror_images a boolean inside Ansible, and convert MIRROR_IMAGES using its normal bash definition at the source.


bfournie:
/lgtm 
Tested both with MIRROR_IMAGES set and unset
Will approve once e2e test passes

PR #1490, opened 2022-12-12
Use V4 addresses in case of dual stack. Use hostnames instead of host ip

pawanpinjarkar:
Signed-off-by: Pawan Pinjarkar <ppinjark@redhat.com>

### DIFF
--- a/agent/04_agent_configure.sh
+++ b/agent/04_agent_configure.sh
@@ -17,12 +17,11 @@ early_deploy_validation
 
 export CLUSTER_NAMESPACE=${CLUSTER_NAMESPACE:-"cluster0"}
 
-function add_host_entry {
+function add_ip_host_entry {
     ip=${1}
     hostname=${2}
 
-    echo "${hostname}">>"${OCP_DIR}"/hostname
-    echo "${ip} ${hostname}" | sudo tee -a /etc/hosts
+    echo "${ip} ${hostname}">>"${OCP_DIR}"/hosts
 }
 
 function add_dns_entry {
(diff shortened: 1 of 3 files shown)


agent/04_agent_configure.sh
```diff
@@ -17,9 +17,16 @@ early_deploy_validation
 
 export CLUSTER_NAMESPACE=${CLUSTER_NAMESPACE:-"cluster0"}
 
+function add_host_entry {
+    ip=${1}
+    hostname=${2}
+
+    echo "${hostname}">>"${OCP_DIR}"/hostname
+    echo "${ip} ${hostname}" | sudo tee -a /etc/hosts
```
bfournie, line 25: It looks like this is going to appending to /etc/hosts each time its called.  The entry should be removed after running agent-gather.sh

agent/04_agent_configure.sh
```diff
@@ -63,14 +70,17 @@ function get_static_ips_and_macs() {
         if [[ "$IP_STACK" = "v4" ]]; then
             AGENT_NODES_IPS+=($(nth_ip ${EXTERNAL_SUBNET_V4} ${ip}))
 	    add_dns_entry ${AGENT_NODES_IPS[i]} ${AGENT_NODES_HOSTNAMES[i]}
+      add_host_entry ${AGENT_NODES_IPS[i]} ${AGENT_NODES_HOSTNAMES[i]}
```
bfournie, line 73: Indentation problem here and in the calls below

agent/gather.sh
```diff
@@ -6,15 +6,15 @@ SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )/.." && pwd )"
...
 
-for ip in $(cat "${OCP_DIR}"/hostip); 
+for host in $(cat "${OCP_DIR}"/hostname); 
 do 
-    echo "Trying to gather agent logs on host ${ip}"
-    if ssh -o 'ConnectTimeout=30' -o 'StrictHostKeyChecking=no' -o 'UserKnownHostsFile=/dev/null' core@"${ip}" agent-gather -O >agent-gather-"${ip}".tar.xz; then
-    echo "Agent logs saved to agent-gather-"${ip}".tar.xz" >&2
+    echo "Trying to gather agent logs on host ${host}"
+    if ssh -o 'ConnectTimeout=30' -o 'StrictHostKeyChecking=no' -o 'UserKnownHostsFile=/dev/null' core@"${host}" agent-gather -O >agent-gather-"${host}".tar.xz; then
+    echo "Agent logs saved to agent-gather-"${host}".tar.xz" >&2
```
bfournie, line 13: Indentation problem

agent/gather.sh
```diff
@@ -6,15 +6,16 @@ SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )/.." && pwd )"
...
+        echo "Agent logs saved to agent-gather-"${host}".tar.xz" >&2
     else
         if [ $? == 127 ]; then
-            echo "Skipping gathering agent logs, agent-gather script not present on host ${ip}." >&2
+            echo "Skipping gathering agent logs, agent-gather script not present on host ${host}." >&2
         fi
-        rm agent-gather-"${ip}".tar.xz
+        rm agent-gather-"${host}".tar.xz
     fi
+    sudo sed -i "/${host}/d" /etc/hosts
```
bfournie, line 20: Thinking about this some more, we should probably remove this in agent/cleanup.sh since there is no guarantee that gather will be run after these host entries are added.

pawanpinjarkar: right

agent/04_agent_configure.sh
```diff
@@ -17,9 +17,16 @@ early_deploy_validation
 
 export CLUSTER_NAMESPACE=${CLUSTER_NAMESPACE:-"cluster0"}
 
+function add_host_entry {
+    ip=${1}
+    hostname=${2}
+
+    echo "${hostname}">>"${OCP_DIR}"/hostname
```
bfournie, line 24: I think the OCP_DIR has already been removed this point https://github.com/openshift-metal3/dev-scripts/blob/master/ocp_cleanup.sh#L19

Instead of using a file, should be able to get the list of hostnames that were added to dns using, for example:
`$ sudo virsh net-dumpxml ostestbm | xmllint --xpath "//dns/host/hostname" -`

agent/cleanup.sh
```diff
@@ -14,3 +14,8 @@ early_cleanup_validation
 
 rm -rf "${OCP_DIR}/manifests"
 rm -rf "${OCP_DIR}/output"
+
+for host in $(cat "${OCP_DIR}"/hostname);
```
bfournie, line 18: I think the OCP_DIR has already been removed this point https://github.com/openshift-metal3/dev-scripts/blob/master/ocp_cleanup.sh#L19

Instead of using a file, should be able to get the list of hostnames that were added to dns using, for example:
$ sudo virsh net-dumpxml ostestbm | xmllint --xpath "//dns/host/hostname" -

andfasano: Hostnames should be always available from the network definition

pawanpinjarkar: Save ip and hostname in "ip\<SPACE\>hostname" format
in "${OCP_DIR}"/hosts file. Then, read and parse
this data, use ip just to ssh into the host
and use the hostname in agent gather tar filename.
No need to save into /etc/hosts just to be able to resolve the hostname and also then no need to worry about cleaning.

bfournie: You can remove this file from the PR (agent/cleanup.sh)

zaneb:
https://github.com/openshift/release/pull/34861 should fix the tests, but it's a mystery why this hasn't been a problem with all of the agent jobs on other repos?

PR #1496, opened 2022-12-20
AGENT-499: Fixes for auth when using quay registry

bfournie:
This is a follow-up to https://github.com/openshift-metal3/dev-scripts/pull/1472 to get the auth set up properly when using the quay registry backend.

### DIFF
--- a/02_configure_host.sh
+++ b/02_configure_host.sh
@@ -69,7 +69,7 @@ fi
 
 export ANSIBLE_FORCE_COLOR=true
 
-if use_podman_registry; then
+if use_registry ""; then
     setup_local_registry
 fi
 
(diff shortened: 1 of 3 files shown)


bfournie:
To test this use
```
export MIRROR_IMAGES=true
export REGISTRY_BACKEND=quay
```

Without this fix it will fail with an auth error

utils.sh
```diff
@@ -685,4 +685,15 @@ use_podman_registry() {
   return 1
 }
 
+use_quay_registry() {
+  # return true if using the local registry and the backend is podman
```
andfasano, line 689: The comment seems to be wrong.
Minor: it'd be nice to remove the duplication and have just a single function, ie `use_registry "podman"` or `use_registry "quay"`. The argument could be optionally checked if not empty, so that the same function could be reused in `02_configure_host.sh` line 72, to remove also that duplication

02_configure_host.sh
```diff
@@ -280,15 +280,15 @@ fi
...
+    # blank authfile with a "assignment to entry in nil map" error
+    rm -f ${REGISTRY_CREDS}
     # create authfile for local registry
     sudo podman login --authfile ${REGISTRY_CREDS} \
         -u ${REGISTRY_USER} -p ${REGISTRY_PASS} \
         ${LOCAL_REGISTRY_DNS_NAME}:${LOCAL_REGISTRY_PORT}
-else
+elif ! use_quay_registry; then
     # Create a blank authfile in order to have something valid when we read it in 04_setup_ironic.sh
     echo '{}' | sudo dd of=${REGISTRY_CREDS}
```
andfasano, line 293: Q: is that something still valid/required? cc @derekhiggins

utils.sh
```diff
@@ -685,4 +685,15 @@ use_podman_registry() {
...
+  if [[ ! -z "${MIRROR_IMAGES}" || $(env | grep "_LOCAL_IMAGE=")  || ! -z "${ENABLE_CBO_TEST:-}" || ! -z "${ENABLE_LOCAL_REGISTRY}" ]]; then
+     if [[ "${REGISTRY_BACKEND}" = "quay" ]]; then
```
rwsu, line 691:
```suggestion
     if [[ "${REGISTRY_BACKEND}" == "quay" ]]; then
```

utils.sh
bfournie: Good idea. Done.

utils.sh
bfournie: Fixed.

rwsu (commented):
/lgtm

PR #1497, opened 2022-12-21
AGENT-500: Add configurable step to oc-mirror to generate tarfile

bfournie:
Add the ability to use oc-mirror to create a mirror registry by generating a tarfile first and then publishing the tarfile to the registry. The can be used to test a common use case for disconnected installs.

### DIFF
--- a/oc_mirror.sh
+++ b/oc_mirror.sh
@@ -84,10 +84,12 @@ function set_full_imageset() {
    # Create imageset with all images from releaseImage
    imageset=$1
 
+   latest_release=$(oc-mirror list releases --channel candidate-${OPENSHIFT_RELEASE_STREAM}| tail -n1)
+
    cat > "${imageset}" << EOF
 apiVersion: mirror.openshift.io/v1alpha2
 kind: ImageSetConfiguration
-archiveSize: 24
+archiveSize: 16
 storageConfig:
   local:
     path: metadata
@@ -97,6 +99,8 @@ mirror:
       - "amd64"
     channels:
       - name: candidate-${OPENSHIFT_RELEASE_STREAM}
+        minVersion: $latest_release
+        maxVersion: $latest_release
         type: ocp
   additionalImages:
   - name: registry.redhat.io/ubi8/ubi:latest
--- a/oc_mirror.sh
+++ b/oc_mirror.sh
@@ -49,14 +49,11 @@ function setup_quay_mirror_registry() {
    popd
 }
 
-# Mirror the upstream channel directly to the local registry
-function mirror_to_mirror_publish() {
+function create_registry_imageset() {
 
-   # Create imageset containing the local URL and the OCP release to mirror
-   tmpimageset=$(mktemp --tmpdir "imageset--XXXXXXXXXX")
-   _tmpfiles="$_tmpfiles $tmpimageset"
+   imageset=$1
 
-   cat > "${tmpimageset}" << EOF
+   cat > "${imageset}" << EOF
 apiVersion: mirror.openshift.io/v1alpha2
 kind: ImageSetConfiguration
 archiveSize: 4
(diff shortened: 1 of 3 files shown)


oc_mirror.sh
```diff
@@ -81,3 +78,74 @@ EOF
...
+
+   # Create imageset with all images from releaseImage
+   imageset=$1
+
+   latest_release=$(oc-mirror list releases --channel candidate-${OPENSHIFT_RELEASE_STREAM}| tail -n1)
+
+   cat > "${imageset}" << EOF
+apiVersion: mirror.openshift.io/v1alpha2
+kind: ImageSetConfiguration
+archiveSize: 16
```
rwsu, line 92: I would suggest adding a comment to explain why the archiveSize is set to 16. I think you mentioned otherwise the archive file would be split into multiple files.

oc_mirror.sh
```diff
@@ -49,11 +49,8 @@ function setup_quay_mirror_registry() {
    popd
 }
 
-# Set up a mirror using the 'oc mirror' command
-# The backend registry can be either 'podman' or 'quay'
-function setup_oc_mirror() {
-
-   update_docker_config
+# Mirror the upstream channel directly to the local registry
+function mirror_to_mirror_publish() {
```
andfasano, line 53: Probably in future it could be interesting also to mirror OLM components

oc_mirror.sh
```diff
@@ -81,3 +78,74 @@ EOF
...
+apiVersion: mirror.openshift.io/v1alpha2
+kind: ImageSetConfiguration
+archiveSize: 16
+storageConfig:
+  local:
+    path: metadata
+mirror:
+  platform:
+    architectures:
+      - "amd64"
```
andfasano, line 99: Q: Any specific reason to specify just here the architecture?

oc_mirror.sh
```diff
@@ -81,3 +78,74 @@ EOF
    popd
 
 }
+
+function set_full_imageset() {
```
andfasano, line 82: nit: maybe this name it's a little bit misleading, this function just generate a manifest file for mirroring to the local storage

oc_mirror.sh
```diff
@@ -81,3 +78,74 @@ EOF
...
+        maxVersion: $latest_release
+        type: ocp
+  additionalImages:
+  - name: registry.redhat.io/ubi8/ubi:latest
+EOF
+
+}
+
+# Use the oc-mirror command to generate a tar file of the release image
+function mirror_to_file() {
```
andfasano, line 112: I'd probably merge this function with the previous one

oc_mirror.sh
bfournie: Done

oc_mirror.sh
bfournie: Added to other imageset

oc_mirror.sh
bfournie: Yes I think we can easily do that here in a separate commit.

oc_mirror.sh
bfournie: Its split this way to allow the file to be generated separate from the publish so we could push the file to a node and publish it to the registry there.

oc_mirror.sh
bfournie: Yes agree. Changed the names of these functions.

rwsu (commented):
/lgtm

bfournie:
Both e2e-metal-ipi-serial-ipv4  and e2e-metal-ipi-bm have passed multiple times on this PR after the last commit as can be seen in the PR history, however each time a retest is run invariably one of them fails due a random issue. The failure is completely unrelated to this PR which has added a code which is not even exercised. Attempting to override after the last failure
/override e2e-metal-ipi-serial-ipv4

PR #1460, opened 2022-10-05
Mirror cleanup and enhancements

bfournie:
Various cleanup and changes based on review and discussion

PR #1498, opened 2022-12-21
Move registry cleanup back to realclean

cybertron:
Removing the registry in regular clean does bad things to the developer workflow because while iterating on changes the entire release has to be re-mirrored each time. Those who want to wipe their entire environment between runs have the realclean option available.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -56,7 +56,7 @@ ocp_run:
 gather:
 	./must_gather.sh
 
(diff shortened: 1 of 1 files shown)


cybertron:
/cc @bfournie 

This undoes one of your changes from #1460 so if there was a reason you wanted that behavior let me know. I really don't. :-)

bfournie:
Thanks @cybertron, yeah I changed this when I added a new backend registry for quay (via the mirror-registry) command as they couldn't coexist, but its fine to move this back to realclean. I can add a note that when using  REGISTRY_BACKEND=quay the docker registry needs to be removed first.

/lgtm

PR #1500, opened 2023-01-02
Use ansible 6

elfosardo:
update with metal3-dev-env hash


elfosardo:
/hold
we need to verify dualstack job

PR #1507, opened 2023-01-17
AGENT-509: Warn if podman registry in use when configured for quay

bfournie:
If the quay registry backend is configured but the podman registry is still set up, the "mirror-registry" command will fail and it will be difficult to debug. Exit the command with instructions of how to remove the podman registry.

### DIFF
--- a/oc_mirror.sh
+++ b/oc_mirror.sh
@@ -34,6 +34,11 @@ function update_docker_config() {
 
 function setup_quay_mirror_registry() {
 
+   if sudo podman container exists registry; then
+     echo "The podman registry is currently running and will cause a conflict with quay registry. Run \"registry_cleanup.sh\" to remove podman registry."
+     exit 1
+   fi
+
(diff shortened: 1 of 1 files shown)


oc_mirror.sh
```diff
@@ -34,6 +34,11 @@ function update_docker_config() {
 
 function setup_quay_mirror_registry() {
 
+   if sudo podman container exists registry; then
+     echo "The podman registry is currently running and will cause a conflict with quay registry. Run \"registry_cleanup.sh\" to remove podman registry."
```
andfasano, line 38: Is there any specific scenario that will make it happen? I'd expect that a `make clean` could be sufficient in such case

bfournie: Yes the reason this now happens is that registry_cleanup.sh was moved out of `make clean` and into `make realclean` to avoid having to rebuild the registry on each run - https://github.com/openshift-metal3/dev-scripts/pull/1498.  Which works fine until switch registry backends, which should not happen that often.

PR #1508, opened 2023-01-25
Fix unbound variable in agent/common.sh

cybertron:
We need to source network.sh before referencing IP_STACK or it may be unset in the default case.


cybertron:
Note that this is currently causing make clean to fail in ipv6 deployments unless you explicitly set IP_STACK.

PR #1509, opened 2023-01-26
WIP: Add support for doing mirroring from host

bfournie:
Adds a new env variable - AGENT_REGISTRY_ON_NODE. When set to `true` and MIRROR_IMAGES is also `true`, instead of creating a registry and mirroring images on the dev-scripts machine, the images will be mirrored to a tarfile, the tarfile transferred to a host, and the host will run a registry to serve the images. Eventually these tools and tarfile will be part of the ISO and this transfer will not be needed,

This is strictly a POC for AGENT-262. It currently only works for the bootstrap images and requires a new agent installer service to start the registry and push the image tarfile to it, see https://github.com/openshift/installer/pull/6806. It also uses the libpod registry strictly for test (this is what dev-scripts users). This will be replaced with docker-registry.


bfournie:
Support for this has been added in the openshift appliance, closing.
/close

PR #1510, opened 2023-01-26
OPNET-200: Add support for ipv6-primary dual stack

cybertron:
Adds a new value "v6v4" for IP_STACK to allow deployment of clusters with ipv6 as the primary address family.

### DIFF
--- a/config_example.sh
+++ b/config_example.sh
@@ -193,9 +193,10 @@ set -x
 # single-stack cluster on dual-stack hosts.
 
 # IP_STACK -
-# IP stack for the cluster.
+# IP stack for the cluster. "v4v6" is ipv4-primary dual stack, "v6v4" is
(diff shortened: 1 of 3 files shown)


cybertron:
/hold

Needs to be documented. Pushing the functional changes so we can collaborate on this feature.

cybertron:
I believe the dual stack job is failing on a known SDN bug with dual stack. It looks like the cluster deployed fine though so it's unlikely to be related to this. I'm about to push a new revision with docs for the new value so it will reset the jobs anyway.

cybertron:
/hold cancel

Docs added. Note that this doesn't currently work due to lack of support in the rest of OpenShift, but we need this to develop and test that support anyway.

cybertron:
/retest

Looks like dual stack is passing again. Let's see how it does with this.

PR #1513, opened 2023-02-03
Dockerfile.ci: add ipcalc

EmilienM:
ipcalc is used by some CI jobs and it'll be nice to have it in the CI
image directly.


cybertron:
/ok-to-test

Are we using this from a container image in the current jobs then?

EmilienM:
I found another way to get ipcalc outside of this image. We can close it.

/close

PR #1514, opened 2023-02-14
remove restart network service step

zhaozhanqi:
seems no need restart network service when INT_IF is specified and ipv6 enabled .  After remove this line.  I can deploy ipv6 dual stack with specified INT_IF


dtantsur:
/approve

The CI seems happy, so am I. Thanks.

PR #1515, opened 2023-02-22
Update dev-script CI image

derekhiggins:
We temporarily need to support both RHEL 7 and 8
as we can't change the ci-operator image build to switch
to rhel a new image in lockstep. Once the switch over is done
we can remove the RHEL7 support.

This should allow use to eventually remove the NSS_WRAPPER_PASSWD
hack in baremetalds-packet-setup-commands.sh

### DIFF
--- a/images/Dockerfile.ci
+++ b/images/Dockerfile.ci
@@ -1,9 +1,13 @@
 # This Dockerfile builds the image used by the e2e-metal-ipi test steps in the OpenShift CI.
 # For more details about the test see https://steps.svc.ci.openshift.org/job/openshift-baremetal-operator-master-e2e-metal-ipi
 
-FROM registry.ci.openshift.org/openshift/release:golang-1.14
+FROM registry.ci.openshift.org/openshift/release:rhel-8-release-golang-1.19-openshift-4.13
+
 ENV HOME /output
-RUN INSTALL_PKGS="ansible python-pip nss_wrapper" && \
+
+# For RHEL7
+RUN if [ ! -e /usr/bin/dnf ]; then \
+    INSTALL_PKGS="ansible python-pip nss_wrapper" && \
     yum install -y $INSTALL_PKGS && \
     pip install packet-python && \
     ansible-galaxy collection install "community.general:4.8.1" && \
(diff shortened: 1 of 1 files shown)


images/Dockerfile.ci
```diff
@@ -1,16 +1,12 @@
 # This Dockerfile builds the image used by the e2e-metal-ipi test steps in the OpenShift CI.
 # For more details about the test see https://steps.svc.ci.openshift.org/job/openshift-baremetal-operator-master-e2e-metal-ipi
 
-FROM registry.ci.openshift.org/openshift/release:golang-1.14
+FROM registry.ci.openshift.org/openshift/release:rhel-8-release-golang-1.19-openshift-4.13
 ENV HOME /output
 RUN INSTALL_PKGS="ansible python-pip nss_wrapper" && \
     yum install -y $INSTALL_PKGS && \
```
elfosardo, line 7: probably worth to change to dnf since we're here ?

derekhiggins: yup, can do

elfosardo:
@derekhiggins the image specified in FROM is being replaced with what I think is src or root
Replaced Dockerfile FROM image registry.ci.openshift.org/openshift/release:rhel-8-release-golang-1.19-openshift-4.13

you need to check the ci-operator config to make this change

derekhiggins:
> @derekhiggins the image specified in FROM is being replaced with what I think is src or root Replaced Dockerfile FROM image registry.ci.openshift.org/openshift/release:rhel-8-release-golang-1.19-openshift-4.13
> 
> you need to check the ci-operator config to make this change

I don't think I can change it as this Dockerfile wont build on RHEL8, so I've updated the file to support both RHEL 7 + 8

once I chance the ci-operator config I'll remove the RHEL7 support

images/Dockerfile.ci
```diff
@@ -1,9 +1,13 @@
...
 
-FROM registry.ci.openshift.org/openshift/release:golang-1.14
+FROM registry.ci.openshift.org/openshift/release:rhel-8-release-golang-1.19-openshift-4.13
+
 ENV HOME /output
-RUN INSTALL_PKGS="ansible python-pip nss_wrapper" && \
+
+# For RHEL7
+RUN if [ ! -e /usr/bin/dnf ]; then \
+    INSTALL_PKGS="ansible python-pip nss_wrapper" && \
```
dtantsur, line 10: nit: variable not needed

PR #1521, opened 2023-03-28
Use 100 GB for agent default disk size

bfournie:
The H/W validation minimum was changed to be consistent with docs in https://github.com/openshift/installer/pull/7025. Set the disk size for agent disk to be minimum.


zaneb:
The requirement change has been backported all the way to 4.12, so this should work now.
/approve

PR #1516, opened 2023-03-03
[WIP] Agent tui e2e tests

andfasano:
work in progress

This is a sample, uncompleted script showing how to interact with the libivirt console for testing the agent-tui automatically

PR #1524, opened 2023-04-19
AGENT-600: Add e2e test to exercise agent-tui check and fix

rwsu:
Adds a new configuration option AGENT_E2E_TEST_TUI_BAD_DNS to enable a test scenario where the DNS is disabled on the hosts by configuring the host to use an incorrect IP address for DNS.

This causes the agent-tui to reach the state where it shows the release image cannot be pulled and waits for a user to fix it the issue through the console.

The test then invokes a series of console key presses to simulate a user fixing the issue through the console. The key presses should fix the network configuration and allow the automated flow to continue.

This test will allow us to catch the case where the agent-tui fails to run but automated flow is still able to continue to success. If the agent-tui fails to run now and if the automated flow is able to still continue, it will not succeed because the broken DNS will not allow the release image to be reached.


rwsu:
This PR incorporated @andfasano's POC https://github.com/openshift-metal3/dev-scripts/pull/1516. Thank you for working out how to simulate the key presses through the console.

agent/06_agent_create_cluster.sh
```diff
@@ -171,6 +171,31 @@ fi
...
+  echo "Running test scenario: bad DNS record(s) in agent-config.yaml"
+  echo "Waiting for 5 mins to arrive at agent-tui check screen"
+  sleep 300
+
+  # Take screenshots of console before fixing DNS. The screenshot may help us see if
+  # agent-tui has reached the expected failure state.
+  for (( n=0; n<${NUM_MASTERS}; n++ ))
+  do
+    name=${CLUSTER_NAME}_master_${n}
+    sudo virsh screenshot $name "${OCP_DIR}/${name}_console_screenshot_before_dns_fix.ppm"
```
andfasano, line 185: Nice addition @rwsu !

agent/05_agent_configure.sh
```diff
@@ -224,10 +224,6 @@ function generate_cluster_manifests() {
      export INGRESS_VIP=${INGRESS_VIPS%${VIPS_SEPARATOR}*}
   fi
 
-  if [[ "$IP_STACK" = "v4v6" ]]; then
-     export PROVISIONING_HOST_EXTERNAL_IP_DUALSTACK=$(nth_ip $EXTERNAL_SUBNET_V6 1)
```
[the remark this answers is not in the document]
rwsu: Done. I've changed the AGENT_E2E_TEST_TUI_BAD_DNS test scenario to only change master-0's DNS IP address.

common.sh
```diff
@@ -471,3 +471,5 @@ if [[ -n "$MIRROR_IMAGES" || -z "${IP_STACK:-}" || "$IP_STACK" = "v6" ]]; then
        export OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE="${LOCAL_REGISTRY_DNS_NAME}:${LOCAL_REGISTRY_PORT}/${LOCAL_IMAGE_URL_SUFFIX}:latest"
    fi
 fi
+
+export AGENT_E2E_TEST_TUI_BAD_DNS=${AGENT_E2E_TEST_TUI_BAD_DNS:-}
```
andfasano, line 475: I'm wondering if we could find a better name for this.  I see some conflict with `AGENT_E2E_TEST_SCENARIO` config var, which indicates a different setup (the topology to be deployed), while here we're really naming a specific e2e test. I do not exclude the possibility to have some additional tests in future, so maybe it could be worth spending some time to find a proper prefix and/or naming convention.

agent/05_agent_configure.sh
```diff
@@ -277,4 +273,24 @@ fi
...
+  # DNS server IP address
+  if [[ "$IP_STACK" = "v4" ]]; then
+    # from 192.168.111.1 to 192.168.111.2
+    yq -i -y '.hosts[0].networkConfig["dns-resolver"].config.server[0] = "192.168.111.2"' "${OCP_DIR}/agent-config.yaml"
+  fi
+  if [[ "$IP_STACK" = "v6" ]]; then
+    # from fd2e:6f44:5dd8:c956::1 to fd2e:6f44:5dd8:c956::2
+    yq -i -y '.hosts[0].networkConfig["dns-resolver"].config.server[0] = "fd2e:6f44:5dd8:c956::2"' "${OCP_DIR}/agent-config.yaml"
+  fi
+  if [[ "$IP_STACK" = "v4v6" ]]; then
```
andfasano, line 287: Really appreciated the intention to cover all the networking flavors, but do we plan to have CI jobs for all of them while running this test? Otherwise, it could be limited just to one networking type, if it helps for the simplification.

agent/05_agent_configure.sh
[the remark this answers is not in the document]
bfournie: For the Ansible templates, the networking must be set up in the agent scripts and then the values are applied to the manifests via ansible, that seems to consistent. Unless there is another place where the manifests are modified that can be moved to Ansible.

agent/05_agent_configure.sh
rwsu: I was going to only add the test for IPv4 only. So yeah, let's remove support for other networking type.

agent/05_agent_configure.sh
```diff
@@ -277,4 +273,24 @@ fi
 
 generate_cluster_manifests
 
+if [ "$AGENT_E2E_TEST_TUI_BAD_DNS" = "true" ]; then
+  # Create a bad dns configuration by changing master-0's
+  # DNS server IP address
+  if [[ "$IP_STACK" = "v4" ]]; then
+    # from 192.168.111.1 to 192.168.111.2
+    yq -i -y '.hosts[0].networkConfig["dns-resolver"].config.server[0] = "192.168.111.2"' "${OCP_DIR}/agent-config.yaml"
```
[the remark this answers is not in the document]
rwsu: Of course, will use a value from a different subnet.

agent/05_agent_configure.sh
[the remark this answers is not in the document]
rwsu: Ah, I cannot read. I will add a ansible tasks to modify the manifests.

common.sh
rwsu: Yes, the name can be confusing. How about AGENT_TEST_CASES?

common.sh
```diff
@@ -471,3 +471,5 @@ if [[ -n "$MIRROR_IMAGES" || -z "${IP_STACK:-}" || "$IP_STACK" = "v6" ]]; then
        export OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE="${LOCAL_REGISTRY_DNS_NAME}:${LOCAL_REGISTRY_PORT}/${LOCAL_IMAGE_URL_SUFFIX}:latest"
    fi
 fi
+
+export AGENT_TEST_CASESS=${AGENT_TEST_CASES:-}
```
bfournie, line 475: Typo `AGENT_TEST_CASESS`

agent/roles/manifests/tasks/main.yml
```diff
@@ -18,6 +18,10 @@
   import_tasks: install-config.yml
   when: agent_use_ztp_manifests == 'false'
 
+- name: Patch agent-config manifest to test agent-tui
+  ansible.builtin.command: yq -i -y '.hosts[0].networkConfig["dns-resolver"].config.server[0] = "192.168.123.1"' "{{ install_path }}/agent-config.yaml"
```
andfasano, line 22: This works, thus I was expecting a change on the same template mechanism adopted for generating the config file. IINW the dns section is configured [here](https://github.com/openshift-metal3/dev-scripts/blob/122d238613e66e040ef049a609e9450ee84cb868/agent/roles/manifests/templates/agent-config_yaml.j2#L24), and the `provisioning_host_external_ip` is currently defined in [vars/main.yml](https://github.com/openshift-metal3/dev-scripts/blob/122d238613e66e040ef049a609e9450ee84cb868/agent/roles/manifests/vars/main.yml#L30).
Maybe a new `dns_resolver_ip` var could be added (and used for the templating), conditionally defined based on the `AGENT_TEST_CASES`, ie something like:

```
dns_resolver_ip: "{{ '192.168.123.1' if lookup('env', 'AGENT_TEST_CASES') == 'bad_dns' else provisioning_host_external_ip }}"
```
(not tested, but in the past I've used a similar approach)?

bfournie: Yes, I think @andfasano suggestion is good. For consistency/readability you could also add a new var in [here](https://github.com/openshift-metal3/dev-scripts/blob/master/agent/roles/manifests/vars/main.yml#L7) such as
agent_test_cases: lookup('env', 'AGENT_TEST_CASES')

and then use agent_test_cases == 'bad_dns' in the snippet above

rwsu: I've replaced the builtin command with conditionals in the template.

common.sh
rwsu: Good catch.

andfasano:
/approve

Great job @rwsu it works like a charm! Definitely a must-have addition for the agent-installer-utils repo :)

agent/e2e/agent-tui/test-fix-wrong-dns.sh
```diff
@@ -0,0 +1,130 @@
...
+pressEnter "Select '<Configure network>' button"
+pressEnter "Select 'Edit a connection'"
+pressEnter "Select 'enp2s0'"
+
+pressTab "Goto DNS field" 10
+pressBackspace "Cleanup DNS previous value" 15
+pressKeys "Insert correct DNS address" "${GOOD_DNS_IP}"
+pressTab "Goto <OK> button" 14
+
+# if [ $IP_STACK = "v6" ]; then
```
bfournie, line 92: Will support eventually be added for IPv6?

PR #1525, opened 2023-04-28
Agent-613: Support imageDigestSources for oc adm release mirror

bfournie:
In 4.14 the `oc adm release mirror` command has been updated to output ImageDigestSources instead of ImageContentSources. Update the agent script to handle both.

See https://github.com/openshift/oc/pull/1341#issuecomment-1498475901 for the slightly different output.

Also there is a minor fix for inadvertent hardcoding of working dir.

### DIFF
--- a/agent/05_agent_configure.sh
+++ b/agent/05_agent_configure.sh
@@ -176,9 +176,11 @@ function get_mirror_info {
     tmpmirrorinfo=$(mktemp --tmpdir "mirror--XXXXXXXXXX")
     _tmpfiles="$_tmpfiles $tmpmirrorinfo"
     if [[ ${MIRROR_COMMAND} == "oc-adm" ]]; then
-       sed -n '/imageContentSources/,/^ *$/p' ${MIRROR_LOG_FILE} | tail -n+2 > ${tmpmirrorinfo}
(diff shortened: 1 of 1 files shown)


agent/05_agent_configure.sh
```diff
@@ -176,7 +176,7 @@ function get_mirror_info {
     tmpmirrorinfo=$(mktemp --tmpdir "mirror--XXXXXXXXXX")
     _tmpfiles="$_tmpfiles $tmpmirrorinfo"
     if [[ ${MIRROR_COMMAND} == "oc-adm" ]]; then
-       sed -n '/imageContentSources/,/^ *$/p' ${MIRROR_LOG_FILE} | tail -n+2 > ${tmpmirrorinfo}
+       sed -n -E '/imageContentSources|imageDigestSources/,/^ *$/p' ${MIRROR_LOG_FILE} | tail -n+2 > ${tmpmirrorinfo}
```
andfasano, line 179: From a maintenance point of view maybe it could be slightly more clear to have something like this, with a proper comment explaining why we're fetching one field instead of another:

```
if printf '4.14\n%s\n' "$(openshift_version)" | sort -V -C; then
...
```

PR #1526, opened 2023-05-04
AGENT-603: Support 'create pxe-files' command for ABI

pawanpinjarkar:
Generates pxe assets by running `./openshift-install agent create pxe-files` command. 
Set `AGENT_E2E_TEST_BOOT_MODE` to `PXE` along with `AGENT_E2E_TEST_SCENARIO` to `SNO_IPV4`, for example, to create an SNO cluster to be booted with PXE. The `AGENT_E2E_TEST_BOOT_MODE` can be set as `PXE` or `ISO`. `ISO` is the default boot mode.

To test, set `export INSTALLER_REPO_PR=7102` in config_<user\>.sh

### DIFF
--- a/agent/06_agent_create_cluster.sh
+++ b/agent/06_agent_create_cluster.sh
@@ -173,33 +173,36 @@ openshift_install="$(realpath "${OCP_DIR}/openshift-install")"
 
 if [[ "${BOOT_MODE}" == "PXE" ]]; then
   create_pxe_files ${asset_dir} ${openshift_install}
-else
+fi
+
+if [[ "${BOOT_MODE}" == "ISO" ]]; then
   create_image ${asset_dir} ${openshift_install}
-  if [[ "${AGENT_DISABLE_AUTOMATED:-}" == "true" ]]; then
-    disable_automated_installation
-  fi
+fi
 
-  attach_agent_iso master $NUM_MASTERS
-  attach_agent_iso worker $NUM_WORKERS
+if [[ "${AGENT_DISABLE_AUTOMATED:-}" == "true" ]]; then
+  disable_automated_installation
+fi
 
-  if [ ! -z "${MIRROR_IMAGES}" ]; then
-    force_mirror_disconnect
-  fi
+attach_agent_iso master $NUM_MASTERS
+attach_agent_iso worker $NUM_WORKERS
 
-  if [ ! -z "${AGENT_ENABLE_GUI:-}" ]; then
-    enable_assisted_service_ui
-  fi
+if [ ! -z "${MIRROR_IMAGES}" ]; then
+  force_mirror_disconnect
+fi
 
-  wait_for_cluster_ready
+if [ ! -z "${AGENT_ENABLE_GUI:-}" ]; then
+  enable_assisted_service_ui
+fi
 
-  # Temporary fix for the CI. To be removed once we'll 
-  # be able to generate the cluster credentials
-  if [ ! -f "${OCP_DIR}/auth/kubeadmin-password" ]; then
-      oc patch --kubeconfig="${OCP_DIR}/auth/kubeconfig" secret -n kube-system kubeadmin --type json -p '[{"op": "replace", "path": "/data/kubeadmin", "value": "'"$(openssl rand -base64 18 | tr -d '\n' | tee "${OCP_DIR}/auth/kubeadmin-password" | htpasswd -nBi -C 10 "" | tr -d ':\n' | sed -e 's/\$2y\$/$2a$/' | base64 -w 0 -)"'"}]'
-  fi
+wait_for_cluster_ready
 
-  if [ ! -z "${AGENT_DEPLOY_MCE}" ]; then
-    mce_complete_deployment
-  fi
+# Temporary fix for the CI. To be removed once we'll 
+# be able to generate the cluster credentials
+if [ ! -f "${OCP_DIR}/auth/kubeadmin-password" ]; then
+    oc patch --kubeconfig="${OCP_DIR}/auth/kubeconfig" secret -n kube-system kubeadmin --type json -p '[{"op": "replace", "path": "/data/kubeadmin", "value": "'"$(openssl rand -base64 18 | tr -d '\n' | tee "${OCP_DIR}/auth/kubeadmin-password" | htpasswd -nBi -C 10 "" | tr -d ':\n' | sed -e 's/\$2y\$/$2a$/' | base64 -w 0 -)"'"}]'
+fi
+
+if [ ! -z "${AGENT_DEPLOY_MCE}" ]; then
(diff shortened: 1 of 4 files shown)


agent/06_agent_create_cluster.sh
```diff
@@ -163,31 +169,35 @@ function mce_complete_deployment() {
...
+else
+  create_image
+  if [[ "${AGENT_DISABLE_AUTOMATED:-}" == "true" ]]; then
+    disable_automated_installation
+  fi
 
-attach_agent_iso master $NUM_MASTERS
-attach_agent_iso worker $NUM_WORKERS
+  attach_agent_iso master $NUM_MASTERS
+  attach_agent_iso worker $NUM_WORKERS
```
bfournie, line 181: Wondering why these additional spaces were added here and below

bfournie:
Are you able to do a successful install with PXE?

pawanpinjarkar:
> Are you able to do a successful install with PXE?

This PR and the issue AGENT-603 if only to be able to run the command and creaet pxe assets in dev-scripts.. Actual install with PXE is handled in https://issues.redhat.com/browse/AGENT-604

agent/06_agent_create_cluster.sh
pawanpinjarkar: Just because of indentation as the create_image and rest of the instructions are in the else block.

agent/06_agent_create_cluster.sh
```diff
@@ -14,6 +14,12 @@ source $SCRIPTDIR/agent/common.sh
 
 early_deploy_validation
 
+function create_pxe_files() {
+    local asset_dir="${1:-${OCP_DIR}}"
+    local openshift_install="$(realpath "${OCP_DIR}/openshift-install")"
+    "${openshift_install}" --dir="${asset_dir}" --log-level=debug agent create pxe-files
```
andfasano, line 20: nit: there's some duplication here that could be removed by making parametric the agent subcommand params

common.sh
```diff
@@ -390,12 +391,27 @@ if [[ ! -z ${AGENT_E2E_TEST_SCENARIO} ]]; then
   export IP_STACK=$(echo ${arr[1]##*IP} | tr V v)
   
   if [[ "$delimiterCount" == "2" ]]; then
-    export NETWORKING_MODE=${arr[2]}
-    if [[ $NETWORKING_MODE != "DHCP" ]]; then
+    val=${arr[2]}
+    if [[ "${val}" == "DHCP" ]]; then
+      export NETWORKING_MODE=$val
+    elif [[ "${val}" == "PXE" ]]; then
```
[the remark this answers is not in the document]
pawanpinjarkar: I think we want to use/redefine the values to `SNO_IPV4_DHCP_ISO` instead of the current `SNO_IPV4_DHCP` when creating the ISO. This way, it will be easy to parse the `BOOT_MODE` to either `ISO` or `PXE` and will also help in readability.

e.g. `SNO_IPV4_DHCP_ISO`, `SNO_IPV4_DHCP_PXE`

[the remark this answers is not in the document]
andfasano: That is going to break all the CI jobs, if we don't have it defaulted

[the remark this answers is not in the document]
pawanpinjarkar: It defaults to an empty string [here](https://github.com/openshift-metal3/dev-scripts/pull/1526/files#diff-a045819fa24dfbeb74522ddaf6eac77b5a4d39762752db45251e6101cd5c987dR379) but I agree on an explicit value as `ISO` will be helpful.

config_example.sh
```diff
@@ -654,19 +654,33 @@ set -x
...
-# - COMPACT_IPV6_DHCP
-# - HA_IPV4
-# - HA_IPV6
-# - HA_IPV4_DHCP
-# - HA_IPV6 _DHCP
-# - SNO_IPV4
-# - SNO_IPV6
-# - SNO_IPV4_DHCP
-# - SNO_IPV6_DHCP
+# - COMPACT_IPV4_ISO
```
rwsu, line 660: I wonder if boot mode should be included in AGENT_E2E_TEST_SCENARIO. The more options we include in AGENT_E2E_TEST_SCENARIO the longer this list becomes. What do people think about having BOOT_MODE separate from AGENT_E2E_TEST_SCENARIO?

pawanpinjarkar: That's a good suggestion!

pawanpinjarkar: I think if we are going to implement this, then the env var name should be `AGENT_E2E_TEST_BOOT_MODE`

andfasano: That sounds fine to me. I think that, in general, we could distinguish two broader set of configuration variables, primary and secondary ones.
Primary config vars are core traits, and they could be an aggregate var (like for example the topology), and they definited inline. Secondary variables capture maybe some less frequently used config traits, and could be defined separately.

config_example.sh
```diff
@@ -683,3 +685,9 @@ set -x
 # deployment systemd services of the Agent based installation. This is
 # particularly useful for WebUI development.
 # export AGENT_DISABLE_AUTOMATED=false
+
+# When set, the code internally sets the boot mode for the agents.
+# This config variable is used only by the agent based installer and is optional.
+# The default value for AGENT_E2E_TEST_BOOT_MODE is 'ISO'.
+# When set to 'PXE', the AGENT_E2E_TEST_SCENARIO can only be set to IPV4 scenarios
```
rwsu, line 692: I'm curious, why does IPv6 and dual stack not work?

rwsu (commented):
Have a question about why IPv6 and dual stack doesn't work with PXE. Otherwise
/lgtm

config_example.sh
andfasano: The limitation is reported by [libvirt documentation](https://libvirt.org/formatnetwork.html) (see `bootp` and `tftp` elements). There's no explicit reasoning for that, I guess (but I could be wrong) it was due the additional work required, and the fact IPv4 was a more common scenario

pawanpinjarkar: Also, PXE is nothing but the same ignition packed as in ISO but with different packaging and we have dev-scripts support for IPV6 and Dual stack for ISO ( and in CI) which means we in a way are testing the cluster creation via another way. There is no difference in the internal contents of PXE ( initd to be specific) compared to ISO, both have the same ignition data for the cluster. So given the limitation of libvirt, what @andfasano mentioned above, it is enough to just test if we can boot using PXE and IPV4. Once the installer branch opens, QE is going to manually test PXE with IPV6 in their lab to be sure. @

rwsu: Thanks. The libvirt documentation says tftp isn't supported for IPv6. It maybe worth adding that in the comments in case someone else asks in the future.

rwsu (commented):
/lgtm

PR #1527, opened 2023-05-07
metallb: Override external frr image

liornoy:
This commit adds the --external_frr_image flag
into metallb `inv e2etest` command, thus
overrides the external containers` frr image.
This is required as we're facing e2e test failures with  FRR's BFD + echo test case, and to prevent
it we want to align the FRR version between
the speakers and the external container.


liornoy:
wrong place. sorry.

PR #1528, opened 2023-05-17
AGENT-604: support agent iPXE booting

andfasano:
This patch adds the possibility to test ABI via iPXE booting, to verify that the PXE artifacts generated with the `agent create pxe-files` command can be successfully used. IPv4/IPv6 supported.

Requires https://github.com/openshift-metal3/dev-scripts/pull/1526, https://github.com/openshift/installer/pull/7185

### DIFF
--- a/agent/06_agent_create_cluster.sh
+++ b/agent/06_agent_create_cluster.sh
@@ -168,23 +168,85 @@ function mce_complete_deployment() {
   mce_prepare_postinstallation_manifests ${mceManifests}
   mce_apply_postinstallation_manifests ${mceManifests}
 }
+
+# Generate the PXE artifacts
+function create_pxe_files() {
+    local asset_dir=${1}
+    local openshift_install=${2}
+    "${openshift_install}" --dir="${asset_dir}" --log-level=debug agent create pxe-files
+}
+
+# Setup the environment to allow iPXE booting, by reusing libvirt native features
+# to configure dnsmaq tftp server and pxe boot file
+function setup_pxe_server() {
+    local PXE_SERVER_DIR=${WORKING_DIR}/pxe
+    mkdir -p ${PXE_SERVER_DIR}
+
+    # Create the iPXE boot file
+    local PXE_SERVER_PORT=8089
+    local PXE_SERVER_URL=http://$(wrap_if_ipv6 ${PROVISIONING_HOST_EXTERNAL_IP}):${PXE_SERVER_PORT}
+    local PXE_BOOT_FILE=boot.ipxe
+
+    sudo cat > ${PXE_SERVER_DIR}/${PXE_BOOT_FILE} << EOF
+#!ipxe
+kernel ${PXE_SERVER_URL}/agent-vmlinuz.x86_64 initrd=main coreos.live.rootfs_url=${PXE_SERVER_URL}/agent-rootfs.x86_64.img ignition.firstboot ignition.platform.id=metal
+initrd --name main ${PXE_SERVER_URL}/agent-initrd.x86_64.img
+boot
+EOF
+
+    # Configure the DHCP options for PXE, based on the network type
+    sudo virsh net-dumpxml ${BAREMETAL_NETWORK_NAME} > ${WORKING_DIR}/${BAREMETAL_NETWORK_NAME}
+
+    local DHCP_PXE_OPTS="dhcp-boot=${PXE_SERVER_URL}/${PXE_BOOT_FILE}"
+    if [[ "${IP_STACK}" = "v6" ]]; then
+      DHCP_PXE_OPTS="dhcp-option=option6:bootfile-url,${PXE_SERVER_URL}/${PXE_BOOT_FILE}"
+    fi
+    sudo sed -i "/<\/dnsmasq:options>/i   <dnsmasq:option value='${DHCP_PXE_OPTS}'/>" ${WORKING_DIR}/${BAREMETAL_NETWORK_NAME}
+    
+    sudo virsh net-define ${WORKING_DIR}/${BAREMETAL_NETWORK_NAME}
+    sudo virsh net-destroy ${BAREMETAL_NETWORK_NAME}
+    sudo virsh net-start ${BAREMETAL_NETWORK_NAME}
+
+    # Copy the generated PXE artifacts in the tftp server location
+    cp ${SCRIPTDIR}/${OCP_DIR}/pxe/* ${PXE_SERVER_DIR}
+
+    # Run a local http server to provide all the necessary PXE artifacts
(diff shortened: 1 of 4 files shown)


agent/06_agent_create_cluster.sh
```diff
@@ -163,13 +163,74 @@ function mce_complete_deployment() {
...
+    mkdir -p ${PXE_SERVER_DIR}
+
+    # Create the iPXE boot file
+    local BASEURL=tftp://${PROVISIONING_HOST_EXTERNAL_IP}
+    sudo cat > ${PXE_SERVER_DIR}/boot.ipxe << EOF
+#!ipxe
+kernel ${BASEURL}/agent-vmlinuz.x86_64 initrd=main coreos.live.rootfs_url=${BASEURL}/agent-rootfs.x86_64.img ignition.firstboot ignition.platform.id=metal
+initrd --name main ${BASEURL}/agent-initrd.x86_64.img
+boot
+EOF
```
pawanpinjarkar, line 187: Let's add the `ipxeBaseURL` in the agent-config.yaml [here](https://github.com/openshift-metal3/dev-scripts/blob/master/agent/roles/manifests/templates/agent-config_yaml.j2#L11) similar to [here](https://github.com/openshift/installer/pull/6723/files#diff-6adc75282b02154f0ac7fb942201d1f45026288a8e905cfb4cc9acd4cfb8cadfR37). The installer PR  to read the new field and generate the iPXE script is ready and is awaiting the merge https://github.com/openshift/installer/pull/6723 This way dev-scripts can also test the installer generating the iPXE script file and it can then be simply used here instead of creating it.

agent/06_agent_create_cluster.sh
```diff
@@ -163,13 +163,74 @@ function mce_complete_deployment() {
...
+
+# Setup the environment to allow iPXE booting, by reusing libvirt native features
+# to configure dnsmaq tftp server and pxe boot file. Note that this is supported
+# only for IPv4.
+function setup_pxe_server() {
+    local PXE_SERVER_DIR=${WORKING_DIR}/pxe
+    mkdir -p ${PXE_SERVER_DIR}
+
+    # Create the iPXE boot file
+    local BASEURL=tftp://${PROVISIONING_HOST_EXTERNAL_IP}
```
rwsu, line 181: Does the tftp server run internally in libvirt?

rwsu (commented):
I have one question about the location of the tftp server. Otherwise, the pxe boot worked for me and I was able to install a cluster.

agent/06_agent_create_cluster.sh
andfasano: Yes, it is provided by the dnsmasq instance used by libvirt (for each network)

agent/06_agent_create_cluster.sh
andfasano: Thanks @pawanpinjarkar it's a good idea we could follow up as soon as #6273 lands

agent/06_agent_create_cluster.sh
```diff
@@ -163,13 +163,74 @@ function mce_complete_deployment() {
...
+    sudo cat > ${PXE_SERVER_DIR}/boot.ipxe << EOF
+#!ipxe
+kernel ${BASEURL}/agent-vmlinuz.x86_64 initrd=main coreos.live.rootfs_url=${BASEURL}/agent-rootfs.x86_64.img ignition.firstboot ignition.platform.id=metal
```
zaneb, line 184: When using the `coreos-installer iso extract pxe` command to extract the pxe artifacts from `agent.x86_64.iso`, it produces files named `agent.x86_64-vmlinuz`, `agent.x86_64-rootfs.img`, `agent.x86_64-initrd.img`.
@pawanpinjarkar could we maybe rename the files output from the `agent create pxe-files` command to match? That will make the transition easier for users who are starting with the coreos-installer workaround on 4.12 & 4.13.

agent/06_agent_create_cluster.sh
```diff
@@ -163,13 +163,74 @@ function mce_complete_deployment() {
...
+    # Configure the baremetal network for PXE 
+    sudo virsh net-dumpxml ${BAREMETAL_NETWORK_NAME} > ${WORKING_DIR}/${BAREMETAL_NETWORK_NAME}
+    sudo sed -i "/<\/dhcp>/i       <bootp file='boot.ipxe'/>" ${WORKING_DIR}/${BAREMETAL_NETWORK_NAME}
+    sudo sed -i "/<\/ip>/i     <tftp root='${PXE_SERVER_DIR}'/>" ${WORKING_DIR}/${BAREMETAL_NETWORK_NAME}
+
+    sudo virsh net-define ${WORKING_DIR}/${BAREMETAL_NETWORK_NAME}
+    sudo virsh net-destroy ${BAREMETAL_NETWORK_NAME}
+    sudo virsh net-start ${BAREMETAL_NETWORK_NAME}
+
+    # Copy the generated PXE artifacts in the tftp server location
```
zaneb, line 198: If there's a way we can use HTTP here that would be preferable. It's what our users will actually do, and will probably be more reliable for CI as well. The main reason for using iPXE is to avoid TFTP as much as possible.

agent/06_agent_create_cluster.sh
pawanpinjarkar: FIxed via https://github.com/openshift/installer/pull/7185/commits/ba82593d8cba14c6877b7686d446cf0467065935

agent/06_agent_create_cluster.sh
andfasano: Ok, found a way that looks simple enough and use http, kept in a separate commit for better comparison. Local tests worked fine.

andfasano: Switch over also to direct use of dnsmasq options via xml, which seems to be a more promising approach to support in future also ipv6 (yet to be tested), but at least for the ipv4 case seems equal to the usage of `bootp` element

andfasano: So, it looks like that dnsmasq option approach worked well with ipv6, so added the support also for that

zaneb:
I haven't tested again but this looks good.
Are we going to merge Pawan's PXE PR first and then rebase this iPXE one?

andfasano:
> I haven't tested again but this looks good. Are we going to merge Pawan's PXE PR first and then rebase this iPXE one?

Yes, this one https://github.com/openshift-metal3/dev-scripts/pull/1526

pawanpinjarkar:
This PR should be tested against the installer PR https://github.com/openshift/installer/pull/7102. I tested the recent changes with SNO IPV6 and seems fine.

/lgtm

PR #1532, opened 2023-06-06
AGENT-604: open pxe server port

andfasano:
Port 8089 is currently used by the PXE server launched on the provisioning host (when the `AGENT_E2E_TEST_BOOT_MODE` is set to `PXE` ), and it needs to be accessible from the guests


pawanpinjarkar (approved):
/lgtm

andfasano:
Required by https://github.com/openshift/release/pull/39907

PR #1533, opened 2023-06-08
AGENT-616: Create unconfigured ignition and config-image and mount to VM

bfournie:
This provides a test for both https://issues.redhat.com/browse/AGENT-558 and https://issues.redhat.com/browse/AGENT-559. It runs the installer commands to create an unconfigured ignition and to create a config-image. It creates an ISO from the unconfigured ignition suitable for booting the VM. It also attaches the config-image to the VM which will be loaded when detected during system boot.

Also adds a new config setting `AGENT_APPLIANCE_HOTPLUG` that when set will configure the hosts using the unconfigured ignition and then pend for user input before creating the config-image and mounting it to the host. This emulates a situation where a user boots the host first before adding the config-image and continuing the installation.

### DIFF
--- a/agent/06_agent_create_cluster.sh
+++ b/agent/06_agent_create_cluster.sh
@@ -23,7 +23,63 @@ function create_pxe_files() {
 function create_image() {
     local asset_dir=${1}
     local openshift_install=${2}
+
+    if [ "${AGENT_USE_APPLIANCE_MODEL}" == true ]; then
+      create_factory_image
+    else
+      create_automated_image
+    fi
+}
+
+function create_automated_image() {
     "${openshift_install}" --dir="${asset_dir}" --log-level=debug agent create image
+}
+
+function create_factory_image() {
+    config_image_drive="sdd"
+
+    # The command to create the config-image must be run out of a separate asset directory using same assets
+    mkdir -p ${config_image_dir}
+    if [[ ${AGENT_USE_ZTP_MANIFESTS} == true ]]; then
+       mkdir -p ${config_image_dir}/cluster-manifests
+       cp ${asset_dir}/cluster-manifests/*.yaml ${config_image_dir}/cluster-manifests/
+    else
+       cp ${asset_dir}/*.yaml ${config_image_dir}
+    fi
+
+    # Create the unconfigured ignition and include it in an ISO
+    "${openshift_install}" --dir="${asset_dir}" --log-level=debug agent create cluster-manifests
+
+    # Remove any static networking configuration from unconfigured so that config-image sets it
+    rm "${asset_dir}/cluster-manifests/nmstateconfig.yaml"
+    rm "${asset_dir}/.openshift_install_state.json"
+
+    "${openshift_install}" --dir="${asset_dir}" --log-level=debug agent create unconfigured-ignition
+    base_iso_url=$(oc adm release info --registry-config "$PULL_SECRET_FILE" --image-for=machine-os-images --insecure=true $OPENSHIFT_RELEASE_IMAGE)
+    oc image extract --path /coreos/coreos-$(uname -m).iso:$HOME/.cache/agent/image_cache --registry-config "$PULL_SECRET_FILE" --confirm $base_iso_url
+    local agent_iso_abs_path="$(realpath "${OCP_DIR}")"
+    podman run --privileged --rm -v /run/udev:/run/udev -v "${agent_iso_abs_path}:${agent_iso_abs_path}" -v "$HOME/.cache/agent/image_cache/:$HOME/.cache/agent/image_cache/" quay.io/coreos/coreos-installer:release iso ignition embed -f -i "${agent_iso_abs_path}/unconfigured-agent.ign" -o "${agent_iso_abs_path}/agent.iso" $HOME/.cache/agent/image_cache/coreos-$(uname -m).iso
+
+    if [ "${AGENT_APPLIANCE_HOTPLUG}" != true ]; then
+        create_config_image
+    fi
+}
+
+function create_config_image() {
+
+    "${openshift_install}" --log-level=debug --dir="${config_image_dir}" agent create config-image
+
+    # Copy the auth files to OCP_DIR so wait-for command can access it
+    cp -r ${config_image_dir}/auth ${asset_dir}
+}
+
+function set_device_config_image() {
+
+    for (( n=0; n<${2}; n++ ))
+    do
+        name=${CLUSTER_NAME}_${1}_${n}
+        sudo virsh change-media --domain ${name} --path ${config_image_drive} --source "${PWD}/${config_image_dir}/agentconfig.noarch.iso" --live --update
+    done
 }
 
 function attach_agent_iso() {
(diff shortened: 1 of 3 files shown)


agent/06_agent_create_cluster.sh
```diff
@@ -255,6 +311,21 @@ if [ ! -z "${AGENT_ENABLE_GUI:-}" ]; then
   enable_assisted_service_ui
 fi
 
+if [[ "${AGENT_USE_APPLIANCE_MODEL}" == true ]] && [[ "${AGENT_APPLIANCE_HOTPLUG}" == true ]]; then
+
+    # Wait for user input before creating config-image and mounting it to continue installation
+    set +x
+    config_image_msg="An unconfigured ISO has been installed on the hosts. Press any key to build a config-image and mount it to continue installation."
```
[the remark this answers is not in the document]
bfournie: It was mainly added for the demo - to show what the system looks like with an unconfigured ignition and then simulate a hotplug on-demand. And yes, if AGENT_APPLIANCE_HOTPLUG is not set or false it won't wait for user input.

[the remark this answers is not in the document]
bfournie: Also, even though AGENT_APPLIANCE_HOTPLUG was added for a demo I think its useful to keep for manual testing to interrogate the before/after of loading the config-image. Its just not an option for automatic or CI testing.

agent/06_agent_create_cluster.sh
```diff
@@ -23,9 +23,55 @@ function create_pxe_files() {
...
+
+function create_automated_image() {
     "${openshift_install}" --dir="${asset_dir}" --log-level=debug agent create image
 }
 
+function create_factory_image() {
+    config_image_drive="sdd"
+
+    # The command to create the config-image must be run out of a separate asset directory using same assets
+    mkdir -p ${config_image_dir}
```
danielerez, line 42: When config-image is used, 'config_image_dir' should be used also in wait_for_cluster_ready func, right?
I.e. instead of '--dir="${OCP_DIR}"'

agent/06_agent_create_cluster.sh
```diff
@@ -23,9 +23,55 @@ function create_pxe_files() {
...
+        create_config_image
+    fi
+}
+
+function create_config_image() {
+
+    "${openshift_install}" --log-level=debug --dir="${config_image_dir}" agent create config-image
+
+    # Copy the auth files to OCP_DIR so wait-for command can access it
+    cp -r ${config_image_dir}/auth ${asset_dir}
```
danielerez, line 63: Isn't it cleaner to just change the dir in wait_for_cluster_ready func.
I.e. in config-image flow use --dir="${config_image_dir}" instead of --dir="${OCP_DIR}"

agent/06_agent_create_cluster.sh
```diff
@@ -23,9 +23,55 @@ function create_pxe_files() {
...
+    # Create the unconfigured ingition and include it in an ISO
+    "${openshift_install}" --dir="${asset_dir}" --log-level=debug agent create cluster-manifests
+    "${openshift_install}" --dir="${asset_dir}" --log-level=debug agent create unconfigured-ignition
+    base_iso_url=$(oc adm release info --registry-config "$PULL_SECRET_FILE" --image-for=machine-os-images --insecure=true $OPENSHIFT_RELEASE_IMAGE)
+    oc image extract --path /coreos/coreos-x86_64.iso:$HOME/.cache/agent/image_cache --registry-config "$PULL_SECRET_FILE" --confirm $base_iso_url
+    local agent_iso_abs_path="$(realpath "${OCP_DIR}")"
+    podman run --privileged --rm -v /run/udev:/run/udev -v "${agent_iso_abs_path}:${agent_iso_abs_path}" -v "$HOME/.cache/agent/image_cache/:$HOME/.cache/agent/image_cache/" quay.io/coreos/coreos-installer:release iso ignition embed -f -i "${agent_iso_abs_path}/unconfigured-agent.ign" -o "${agent_iso_abs_path}/agent.iso" $HOME/.cache/agent/image_cache/coreos-x86_64.iso
+
+    if [ "${AGENT_APPLIANCE_HOTPLUG}" != true ]; then
+        create_config_image
```
danielerez, line 54: Isn't it too late here? iiuc, the VMs are already running after step 02: https://github.com/openshift-metal3/dev-scripts//blob/d65d61800da90a3b51cf5fb0171a1f3571693984/02_configure_host.sh#L125

bfournie: No the domain isn't started until this command is run
`sudo virt-xml ${name} --edit target=sdc --disk="boot_order=2" --start`

agent/06_agent_create_cluster.sh
bfournie: Good point. Changed

agent/06_agent_create_cluster.sh
```diff
@@ -243,6 +303,21 @@ if [ ! -z "${AGENT_ENABLE_GUI:-}" ]; then
   enable_assisted_service_ui
 fi
 
+if [[ "${AGENT_USE_APPLIANCE_MODEL}" == true ]] && [[ "${AGENT_APPLIANCE_HOTPLUG}" == true ]]; then
```
rwsu, line 306: Is there any difference between having the config image already attached to the host when the host first boots, versus mounting it say 2 minutes after the host has booted?

bfournie: Functionally there is no difference, the contents will be copied when the config-image can be read. This flag is mainly useful to allow checking the contents of the unconfigured ignition prior to copying the config-image. I don't expect this flag will be used with automated testing, but its useful for demos and manual testing.

rwsu (commented):
/lgtm

agent/06_agent_create_cluster.sh
```diff
@@ -23,9 +23,55 @@ function create_pxe_files() {
...
 
+function create_factory_image() {
+    config_image_drive="sdd"
+
+    # The command to create the config-image must be run out of a separate asset directory using same assets
+    mkdir -p ${config_image_dir}
+    cp ${asset_dir}/*.yaml ${config_image_dir}
+
+    # Create the unconfigured ingition and include it in an ISO
+    "${openshift_install}" --dir="${asset_dir}" --log-level=debug agent create cluster-manifests
```
rwsu, line 46: After the "create cluster-manifests" line, could we add these two lines?
```
   rm "${asset_dir}/cluster-manifests/nmstateconfig.yaml"
   rm "${asset_dir}/.openshift_install_state.json"
```
This will ensure that the nmstateconfig is not in the unconfigured-ignition. Would also make break ssh and the network until the config image contents is copied and applied to the host. If the cluster installation succeeds then we know that the static network config has been applied correctly by load-config-iso.sh in https://github.com/openshift/installer/pull/7200.

bfournie: Yes, good suggestion, I will make that change. Note that until the config-image is applied it is possible to ssh to 192.168.11.20 - the address picked up via DHCP.

rwsu (commented):
/lgtm

agent/06_agent_create_cluster.sh
```diff
@@ -23,9 +23,65 @@ function create_pxe_files() {
...
+    # Create the unconfigured ignition and include it in an ISO
+    "${openshift_install}" --dir="${asset_dir}" --log-level=debug agent create cluster-manifests
+
+    # Remove any static networking configuration from unconfigured so that config-image sets it
+    rm "${asset_dir}/cluster-manifests/nmstateconfig.yaml"
+    rm "${asset_dir}/.openshift_install_state.json"
+
+    "${openshift_install}" --dir="${asset_dir}" --log-level=debug agent create unconfigured-ignition
+    base_iso_url=$(oc adm release info --registry-config "$PULL_SECRET_FILE" --image-for=machine-os-images --insecure=true $OPENSHIFT_RELEASE_IMAGE)
+    oc image extract --path /coreos/coreos-x86_64.iso:$HOME/.cache/agent/image_cache --registry-config "$PULL_SECRET_FILE" --confirm $base_iso_url
```
andfasano, line 59: This could break for ARM deployments (arch hard-coded to x86_64)?

bfournie: Good catch. Changed to use uname

rwsu (commented):
/lgtm

PR #1534, opened 2023-06-15
OCPBUGS-15056: Use v4 provisioning network for v4-primary dual stack

cybertron:
Under some circumstances, using an ipv6 provisioning network in a v4-primary dual stack env is causing BMO to bring nodes up with only a v6 address. This is a problem because in v4-primary envs we only configure the local registry to be accessible on v4.

However, since there's technically no reason you couldn't run this way if you really wanted, let's also add a v6 DNS entry for the virthost so the registry is also accessible on v6 in dual stack clusters.

This is a belt-and-suspenders approach to the fix. Either of these changes on its own would fix the immediate problem, but doing both gives us arguably more sane defaults and allows more combinations of options to work in dev-scripts.


cybertron:
Holding off on running ci for the time being until I've had a chance to test some of these combinations locally.

cybertron:
While I'm still having problems deploying dual stack with redfish-virtualmedia in some environments, those environments were broken before this patch too and it at least gets them closer to working. I think we can move forward with this.

PR #1535, opened 2023-06-22
Add new param bootstrapExternalStaticDNS

sgoveas:
Add new install-config param 'bootstrapExternalStaticDNS' with value 192.168.111.1 for cluster with static IP address

https://github.com/openshift/installer/pull/6585

### DIFF
--- a/ocp_install_env.sh
+++ b/ocp_install_env.sh
@@ -85,7 +85,7 @@ function baremetal_network_configuration() {
 cat <<EOF
     provisioningNetwork: "${PROVISIONING_NETWORK_PROFILE}"
 EOF
-  if printf '%s\n4.6\n' "$(openshift_version)" | sort -V -C; then
+    if printf '%s\n4.6\n' "$(openshift_version)" | sort -V -C; then
 cat <<EOF
     provisioningHostIP: "${CLUSTER_PROVISIONING_IP}"
     bootstrapProvisioningIP: "${BOOTSTRAP_PROVISIONING_IP}"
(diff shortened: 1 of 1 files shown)


ocp_install_env.sh
```diff
@@ -108,6 +108,7 @@ EOF
 cat <<EOF
     bootstrapExternalStaticIP: "${BOOTSTRAP_IP}"
     bootstrapExternalStaticGateway: "${PROVISIONING_HOST_EXTERNAL_IP}"
+    bootstrapExternalStaticDNS: "192.168.111.1"
```
dtantsur, line 111: This has to be derived from some variables, not hardcoded (see other parameters).

sgoveas: done

elfosardo:
the error is not related to this change
giving it another try

ocp_install_env.sh
```diff
@@ -109,6 +109,11 @@ cat <<EOF
     bootstrapExternalStaticIP: "${BOOTSTRAP_IP}"
     bootstrapExternalStaticGateway: "${PROVISIONING_HOST_EXTERNAL_IP}"
 EOF
+    if [[ $(echo "$(openshift_version) >= 4.14" | bc -l) -eq 1 ]]; then
```
derekhiggins, line 112: This fails for versions less then 4.10
```
[derekh@laptop ~]$ echo "4.10 >= 4.14" | bc -l
0
[derekh@laptop ~]$ echo "4.9 >= 4.14" | bc -l
1

```

derekhiggins:
thanks, would you mind squashing the 3 commits into 1

PR #1541, opened 2023-06-27
AGENT-650: Add none platform support for compact and HA clusters

rwsu:
New PLATFORM_TYPE configuration variable to enable none platform.

When PLATFORM_TYPE=none, the agent deployment scripts deploy haproxy to serve as the external load balancer and DNS entries for api, api-int, *.apps are updated to point to haproxy.


network.sh
```diff
@@ -259,6 +259,12 @@ function add_dnsmasq_multi_entry() {
     #     None
     for i in ${2//${VIPS_SEPARATOR}/ }; do
         if [ "${1}" = "apivip" ] ; then
+            if [[ "$IP_STACK" == "v4" || "$IP_STACK" == "v4v6" ]]; then
+               echo "server=/${CLUSTER_DOMAIN}/$(nth_ip $EXTERNAL_SUBNET_V4 1)" | sudo tee -a "${PATH_CONF_DNSMASQ}"
+            fi
+            if [[ "$IP_STACK" == "v6" ]]; then
+               echo "server=/${CLUSTER_DOMAIN}/$(nth_ip $EXTERNAL_SUBNET_V6 1)" | sudo tee -a "${PATH_CONF_DNSMASQ}"
+            fi
```
andfasano, line 267: This change does not seem required to me (and maybe also potentially harmful for other dev-scripts flows, non-agent related). If that is required by haproxy, I think it'd be a simpler approach to just enrich the host DNS (defined by `$PATH_CONF_DNSMASQ`) with just the nodes name (in addition to api and api-int)

agent/cleanup.sh
```diff
@@ -21,3 +21,5 @@ case "${AGENT_E2E_TEST_BOOT_MODE}" in
     rm -rf ${WORKING_DIR}/pxe
     ;;
 esac
+
+sudo systemctl stop haproxy
```
andfasano, line 25: This requires an `|| true` to avoid job failure in case `haproxy` wasn't started

agent/05_agent_configure.sh
```diff
@@ -245,7 +245,104 @@ function generate_cluster_manifests() {
   ansible-playbook -vvv \
           -e install_path=${SCRIPTDIR}/${INSTALL_CONFIG_PATH} \
           "${SCRIPTDIR}/agent/create-manifests-playbook.yaml"
+}
+
+function enable_load_balancer() {
+  if [[ "${PLATFORM_TYPE}" == "none" && "${NUM_MASTERS}" == "3" ]]; then
```
andfasano, line 251: Any specific reason to limit just to the compact case?

agent/common.sh
```diff
@@ -7,6 +7,8 @@ export AGENT_STATIC_IP_NODE0_ONLY=${AGENT_STATIC_IP_NODE0_ONLY:-"false"}
 
 export AGENT_USE_ZTP_MANIFESTS=${AGENT_USE_ZTP_MANIFESTS:-"false"}
 
+export PLATFORM_TYPE=${PLATFORM_TYPE:-""}
```
andfasano, line 10: The name of the config var seems pretty generic, being a specif agent one I'd suggest `AGENT_PLATFORM_TYPE`. Also, it would need to be documented in config_example.sh

agent/05_agent_configure.sh
```diff
@@ -245,7 +245,104 @@ function generate_cluster_manifests() {
...
+    # setup haproxy as the load balancer
+    if [[ "${IP_STACK}" == "v6" ]]; then
+      # The "wildcard" is different depending on IP stack.
+      # See http://docs.haproxy.org/1.6/configuration.html#4.2-bind
+      export HAPROXY_WILDCARD="[::]"
+    else
+      export HAPROXY_WILDCARD="*"
+    fi
+
+    sudo dnf install -y haproxy
```
andfasano, line 261: For transient services like this, my personal preference would be to use a container if possible, to limit the impacts on the host system

agent/cleanup.sh
andfasano: Also, by using the following approach, `/etc/haproxy/haproxy.cfg` should be deleted

agent/05_agent_configure.sh
```diff
@@ -245,7 +245,104 @@ function generate_cluster_manifests() {
...
+  stats show-node
+  stats show-desc Stats for ocp4 cluster
+  stats auth admin:ocp4
+  stats uri /stats
+listen api-server-6443
+  bind ${HAPROXY_WILDCARD}:6443
+  mode tcp
+  server master-0 master-0.${CLUSTER_DOMAIN}:6443 check inter 1s
+  server master-1 master-1.${CLUSTER_DOMAIN}:6443 check inter 1s
+  server master-2 master-2.${CLUSTER_DOMAIN}:6443 check inter 1s
```
andfasano, line 301: It feels like an iteration would be required here (and below), based on NUM_MASTERS and NUM_WORKERS

vm_setup_vars.yml
```diff
@@ -37,14 +37,20 @@ baremetal_network_cidr_v6: "{{ lookup('env', 'EXTERNAL_SUBNET_V6') }}"
...
+dns_extrahosts: 
+  - ip: "{{ baremetal_network_cidr | nthhost(1) if lookup('env', 'PLATFORM_TYPE') == 'none' else baremetal_network_cidr | nthhost(5) }}"
     hostnames:
       - "api"
+  - ip: "{{ baremetal_network_cidr | nthhost(1) if lookup('env', 'PLATFORM_TYPE') == 'none' else baremetal_network_cidr | nthhost(5) }}"
+    hostnames:
+      - "api-int"
+  - ip: "{{ baremetal_network_cidr | nthhost(1) if lookup('env', 'PLATFORM_TYPE') == 'none' else baremetal_network_cidr | nthhost(4) }}"
+    hostnames:
+      - "*.apps"
```
andfasano, line 53: See previous comment

network.sh
rwsu: Yes, enriching the host DNS in $PATH_CONF_DNSMASQ would also work. Will update.

agent/cleanup.sh
rwsu: Thanks. Will update.

agent/05_agent_configure.sh
rwsu: I was just going off of the HA and compact scenarios which are always limited to 3 masters. I'll change it since it is possible for NUM_MASTERS to be any positive number.

agent/05_agent_configure.sh
```diff
@@ -245,7 +245,104 @@ function generate_cluster_manifests() {
...
+  stats hide-version
+  stats refresh 30s
+  stats show-node
+  stats show-desc Stats for ocp4 cluster
+  stats auth admin:ocp4
+  stats uri /stats
+listen api-server-6443
+  bind ${HAPROXY_WILDCARD}:6443
+  mode tcp
+  server master-0 master-0.${CLUSTER_DOMAIN}:6443 check inter 1s
```
andfasano, line 299: Btw, by looking at the [haproxy docs](https://www.haproxy.com/documentation/hapee/latest/configuration/config-sections/backend/), it looks like also IPs could be used (and that sounds as a good match for using `AGENT_NODES_IPS` or `AGENT_NODES_IPS_IPV6`

vm_setup_vars.yml
rwsu: I think the concern is this change will affect non-agent dev-script users. Will move adding dns entries inside the agent scripts.

pawanpinjarkar:
@rwsu I tested this PR with installer PR 7236 for the IPV4 Compact cluster with the below settings but the Bootstrap failed to complete. Is this tested with a successful cluster installation?
```
export AGENT_PLATFORM_TYPE=none
export INSTALLER_REPO_PR=7236
```

rwsu:
> @rwsu I tested this PR with installer PR 7236 for the IPV4 Compact cluster with the below settings but the Bootstrap failed to complete. Is this tested with a successful cluster installation?
> 
> ```
> export AGENT_PLATFORM_TYPE=none
> export AGENT_E2E_TEST_SCENARIO=COMPACT_IPV4
> export KNI_INSTALL_FROM_GIT=true
> export INSTALLER_REPO_PR=7236
> ```

I tested with both COMPACT_IPV4 and HA_IPV4. I just tested again using your settings and the cluster installed. Could you post more details about why the bootstrap failed in your environment? Thanks.

agent/05_agent_configure.sh
```diff
@@ -264,7 +371,20 @@ if [[ ! -z "${MIRROR_IMAGES}" ]]; then
 fi
 
 if [[ "${NUM_MASTERS}" > "1" ]]; then
-   set_api_and_ingress_vip
+  if [[ "${AGENT_PLATFORM_TYPE}" == "none" ]]; then
+    # for platform "none" both API and INGRESS point to the same
+    # load balancer IP address
+    get_vips
+    if [[ "$IP_STACK" = "v6" ]]; then
+      export load_balancer_ip=$(nth_ip $EXTERNAL_SUBNET_V6 1)
```
andfasano, line 379: `PROVISIONING_HOST_EXTERNAL_IP` could be used as well, will allow to simplify further

agent/docs/none-platform.md
```diff
@@ -0,0 +1,119 @@
...
+    <dnsmasq:option value='cache-size=0'/>
+  </dnsmasq:options>
+</network>
+````
+
+'haproxy' is the load balancer deployed for enabling the None platform with the agent-based installer. The haproxy service
+is configured and enabled by the enable_load_balancer() function.
+
+'haproxy' runs on the hypervisor host and is accessed by the hosts forming the OpenShift cluster through
+the 'ostestbm' network through that network's .1 or ::1 IP address. Its configuration file path is /etc/haproxy/haproxy.cfg.
```
bfournie, line 79: Nit - instead of "through the 'ostestbm' network through that network's..." can just use
"through the ostestbm network's"

agent/roles/manifests/templates/install-config_yaml.j2
```diff
@@ -44,7 +44,7 @@ networking:
 {% endif %}
   networkType: {{ network_type }} 
 platform:
-{% if num_masters == "1" %}
+{% if (num_masters == "1") or (platform_type == "none") %}
```
bfournie, line 47: When using SNO could automatically set PLATFORM_TYPE=none in common.sh which would simplify this check.

config_example.sh
```diff
@@ -708,3 +708,9 @@ set -x
 # This test case is only supported when IP_STACK=v4.
 #
 # export AGENT_TEST_CASES='bad_dns'
+
+# Sets the install-config.yaml's platform type.
+# When undefined, 'baremetal' is assumed. This is the default.
+# When set to 'none', userManagedNetworking is set to true in AgetnClusterInstall and
+# haproxy is deployed as the load balancer.
+# export AGENT_PLATFORM_TYPE=none
```
bfournie, line 716: Can automatically set this to none when AGENT_E2E_TEST_SCENARIO is SNO.

config_example.sh
```diff
@@ -708,3 +708,9 @@ set -x
 # This test case is only supported when IP_STACK=v4.
 #
 # export AGENT_TEST_CASES='bad_dns'
+
+# Sets the install-config.yaml's platform type.
+# When undefined, 'baremetal' is assumed. This is the default.
```
bfournie, line 713: Would it make sense to set the default value to 'baremetal' then?

agent/05_agent_configure.sh
```diff
@@ -264,7 +371,20 @@ if [[ ! -z "${MIRROR_IMAGES}" ]]; then
 fi
 
 if [[ "${NUM_MASTERS}" > "1" ]]; then
-   set_api_and_ingress_vip
+  if [[ "${AGENT_PLATFORM_TYPE}" == "none" ]]; then
+    # for platform "none" both API and INGRESS point to the same
+    # load balancer IP address
+    get_vips
```
bfournie, line 377: Is it necessary to call get_vips when using SNO (NUM_MASTERS=1)?

pawanpinjarkar:
> > @rwsu I tested this PR with installer PR 7236 for the IPV4 Compact cluster with the below settings but the Bootstrap failed to complete. Is this tested with a successful cluster installation?
> > ```
> > export AGENT_PLATFORM_TYPE=none
> > export AGENT_E2E_TEST_SCENARIO=COMPACT_IPV4
> > export KNI_INSTALL_FROM_GIT=true
> > export INSTALLER_REPO_PR=7236
> > ```
> 
> @pawanpinjarkar I tested with both COMPACT_IPV4 and HA_IPV4. I just tested again using your settings and the cluster installed. Could you post more details about why the bootstrap failed in your environment? Thanks.

I am still unable to test the successful cluster deployment. Discussed the details on slack with @rwsu

agent/05_agent_configure.sh
```diff
@@ -245,7 +249,110 @@ function generate_cluster_manifests() {
...
+}
+
+function add_haproxy_server_lines() {
+  num_servers=${1}
+  type=${2}
+  port=${3}
 
+  for (( n=0; n<${num_servers}; n++ ))
+  do
+    cat << EOF >> haproxy.cfg
```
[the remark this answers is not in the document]
rwsu: Ok. Will use $WORKING_DIR.

agent/05_agent_configure.sh
rwsu: Thanks, that does simplify it more.

agent/docs/none-platform.md
rwsu: Thanks Bob. That wasn't pretty.

agent/05_agent_configure.sh
rwsu: SNO would go to the else clause, so we are ok here.

rwsu:
> > > @rwsu I tested this PR with installer PR 7236 for the IPV4 Compact cluster with the below settings but the Bootstrap failed to complete. Is this tested with a successful cluster installation?
> > > ```
> > > export AGENT_PLATFORM_TYPE=none
> > > export AGENT_E2E_TEST_SCENARIO=COMPACT_IPV4
> > > export KNI_INSTALL_FROM_GIT=true
> > > export INSTALLER_REPO_PR=7236
> > > ```
> > 
> > 
> > @pawanpinjarkar I tested with both COMPACT_IPV4 and HA_IPV4. I just tested again using your settings and the cluster installed. Could you post more details about why the bootstrap failed in your environment? Thanks.
> 
> I am still unable to test the successful cluster deployment. Discussed the details on slack with @rwsu

Hi @pawanpinjarkar. I restested this PR in a new VM and found the missing ports that needed to be open in libvirt so that haproxy could be reached. When you have a chance, can you try again?

config_example.sh
rwsu: Yes, that would make sense.

agent/05_agent_configure.sh
```diff
@@ -245,7 +249,112 @@ function generate_cluster_manifests() {
...
+
+function add_haproxy_server_lines() {
+  num_servers=${1}
+  type=${2}
+  port=${3}
 
+  for (( n=0; n<${num_servers}; n++ ))
+  do
+    cat << EOF >> ${WORKING_DIR}/haproxy.cfg
+  server ${type}-$n ${type}-$n.${CLUSTER_DOMAIN}:${port} check inter 1s
```
andfasano, line 262: (weird, I'm pretty sure I commented this line but wasn't able to find my comment anymore :confused: , so re-adding it)

Unless required for other reasons, it's possible to use directly the ip of the instance here instead of the dns name: 
```suggestion
  server ${type}-$n ${AGENT_NODES_IPS[i]}:${port} check inter 1s
```

(or `AGENT_NODES_IPSV6` depending on the network stack) That will allow to remove the block https://github.com/openshift-metal3/dev-scripts/pull/1541/files#diff-4111b804ca04676258ffb172752880ebb5b54f9fa5031d5e620d8193e8dd86d8R35-R38, as a further simplification

agent/05_agent_configure.sh
```diff
@@ -245,7 +249,112 @@ function generate_cluster_manifests() {
...
+  bind ${HAPROXY_WILDCARD}:80
+  mode tcp
+  balance source
+EOF
+      add_haproxy_server_lines $NUM_MASTERS "master" "80"
+    fi
+
+    sudo firewall-cmd --zone=libvirt --add-port=22623/tcp
+    sudo firewall-cmd --zone=libvirt --add-port=6443/tcp
+    sudo firewall-cmd --zone=libvirt --add-port=443/tcp
```
andfasano, line 349: In case of a new firewall port to be opened, it's recommeded to add it here https://github.com/openshift-metal3/dev-scripts/blob/d06b36e96f37feed07fa213207eb471bcff62e8e/02_configure_host.sh#L258 in the step dedicated to setup the firewall (my expectation would be just to find the haproxy port added, the others should be already present)

agent/cleanup.sh
```diff
@@ -21,3 +21,9 @@ case "${AGENT_E2E_TEST_BOOT_MODE}" in
     rm -rf ${WORKING_DIR}/pxe
     ;;
 esac
+
+sudo podman rm -f extlb || true
+sudo rm ${WORKING_DIR}/haproxy.* || true
+sudo firewall-cmd --zone=libvirt --remove-port=22623/tcp
```
andfasano, line 27: See previous comment

pawanpinjarkar:
> > > > @rwsu I tested this PR with installer PR 7236 for the IPV4 Compact cluster with the below settings but the Bootstrap failed to complete. Is this tested with a successful cluster installation?
> > > > ```
> > > > export AGENT_PLATFORM_TYPE=none
> > > > export AGENT_E2E_TEST_SCENARIO=COMPACT_IPV4
> > > > export KNI_INSTALL_FROM_GIT=true
> > > > export INSTALLER_REPO_PR=7236
> > > > ```
> > > 
> > > 
> > > @pawanpinjarkar I tested with both COMPACT_IPV4 and HA_IPV4. I just tested again using your settings and the cluster installed. Could you post more details about why the bootstrap failed in your environment? Thanks.
> > 
> > 
> > I am still unable to test the successful cluster deployment. Discussed the details on slack with @rwsu
> 
> Hi @pawanpinjarkar. I restested this PR in a new VM and found the missing ports that needed to be open in libvirt so that haproxy could be reached. When you have a chance, can you try again?

With the latest changes, I am able to install the cluster successfully.

agent/05_agent_configure.sh
rwsu: I remember seeing it. I think it got lost after one of the git pushes. Thanks for adding back the comment. I will add this simplification.

andfasano:
Looks like the ipv6 failure was a genuine one? 

```
...
level=fatal msg=failed to fetch Agent Installer ISO: failed to load asset "AgentClusterInstall Config": failed to unmarshal cluster-manifests/agent-cluster-install.yaml: error unmarshaling JSON: while decoding JSON: json: unknown field "userManagedNetworking"
```

common.sh
```diff
@@ -384,6 +384,12 @@ export AGENT_E2E_TEST_BOOT_MODE=${AGENT_E2E_TEST_BOOT_MODE:-"ISO"}
 # to open a firewall port
 export AGENT_PXE_SERVER_PORT=${AGENT_PXE_SERVER_PORT:-8089}
 
+# Ports that need to be open for platform "none". Hosts inside the libvirt firewalld zone
+# need to communicate with haproxy deployed on the libvirt host.
+export AGENT_MCS_PORT=22623
+export AGENT_KUBE_API_PORT=6443
+export AGENT_INGRESS_ROUTER_PORT=443
```
andfasano, line 391: What's the reason to open ports 443/6443 (at least 6443 shouldn't be already accessible)?

common.sh
```diff
@@ -384,6 +384,12 @@ export AGENT_E2E_TEST_BOOT_MODE=${AGENT_E2E_TEST_BOOT_MODE:-"ISO"}
 # to open a firewall port
 export AGENT_PXE_SERVER_PORT=${AGENT_PXE_SERVER_PORT:-8089}
 
+# Ports that need to be open for platform "none". Hosts inside the libvirt firewalld zone
+# need to communicate with haproxy deployed on the libvirt host.
+export AGENT_MCS_PORT=22623
```
andfasano, line 389: nit: not sure about the word `MCS`, I'd prefer a more speaking one

agent/05_agent_configure.sh
```diff
@@ -245,7 +245,122 @@ function generate_cluster_manifests() {
...
+  stats auth admin:ocp4
+  stats uri /stats
+listen api-server-6443
+  bind ${HAPROXY_WILDCARD}:6443
+  mode tcp
+EOF
+    add_haproxy_server_lines $NUM_MASTERS "master" "6443"
 
+    cat << EOF >> ${WORKING_DIR}/haproxy.cfg
+listen machine-config-server-22623
```
andfasano, line 316:
```suggestion
listen machine-config-server-${AGENT_MCS_PORT}
```

(and also the other occurrences(

agent/docs/none-platform.md
```diff
@@ -0,0 +1,107 @@
...
+
+With None platform, the user is responsible for configuring the DNS entries for
+the Kubernetes API, OpenShift application wildcard, and the names of the control plane
+and compute machines. Users are also responsible for providing a load balancer
+infrastructure for the API and application ingress. 
+
+In the context of dev-scripts, None platform requires these DNS records to be 
+present in the 'ostestbm' network.
+* api.<cluster_name>.<base_domain> 
+* api-int.<cluster_name>.<base_domain> 'new'
```
andfasano, line 20: (IIRC except for SNO)

agent/docs/none-platform.md
```diff
@@ -0,0 +1,107 @@
...
+* *.apps.<cluster_name>.<base_domain>. 'new'
+
+All three names point to the load balancer's IP address. 
+
+The DNS records are add to the ostestbm libvirt network by the enable_load_balancer()
```
andfasano, line 25:
```suggestion
The DNS records are added to the ostestbm libvirt network by the enable_load_balancer() 
```

agent/docs/none-platform.md
```diff
@@ -0,0 +1,107 @@
...
+
+The OpenShift services configured to use the external load balancer are:
+* Kubernetes api port 6443
+* Machine config server port 22623 - serves the ignition configs
+* Ingress router ports 443 and 80
+
+These ports are opened in the libvirt firewalld domain so that the VMs inside the domain can communicate with haproxy deployed
+on the hypervisor host.
+
+In '/etc/NetworkManager/dnsmasq.d/openshift-ostest.conf', api and .app addresses are updated to point to the load balancer IP.
```
andfasano, line 91: Guess this could be removed

agent/05_agent_configure.sh
rwsu: The port numbers aren't meant to be changed. Let me remove the environment variables and just list the port numbers in 02_configure_host.sh.

common.sh
rwsu: Going to remove these environment variable as the port numbers cannot be changed.

common.sh
rwsu: 6443 is not already accessible. If the port is not open, and after the 2 control plane nodes reboot after fetching their ignition, they will get stuck, do nothing, and block the installation from progressing. The reason is they cannot reach the kube-api running on the bootstrap node unless port 6443 is opened. The kube-api requests has to go through the external load balancer which lives outside libvirt firewalld zone.

If port 443 is not opened, one of the problems is it blocks the authentication operator.
```
[rwsu@nesting-vm dev-scripts]$ oc get co
NAME                                       VERSION       AVAILABLE   PROGRESSING   DEGRADED   SINCE   MESSAGE
authentication                             4.14.0-ec.3   False       True          False      7m37s   OAuthServerRouteEndpointAccessibleControllerAvailable: Get "https://oauth-openshift.apps.ostest.test.metalkube.org/healthz": dial tcp 192.168.111.1:443: connect: connection refused...
console                                    4.14.0-ec.3   False       True          False      19m     DeploymentAvailable: 0 replicas available for console deployme
nt...
ingress                                    4.14.0-ec.3   True        False         True       11m     The "default" ingress controller reports Degraded=True: Degrad
edConditions: One or more other status conditions indicate a degraded state: CanaryChecksSucceeding=False (CanaryChecksRepetitiveFailures: Canary route checks for t
he default ingress controller are failing)
```

agent/docs/none-platform.md
rwsu: This is still required. Without it, doing this will fail:
```
[rwsu@ dev-scripts]export KUBECONFIG=./ocp/ostest/auth/kubeconfig
[rwsu@ dev-scripts]$ oc get co
Unable to connect to the server: dial tcp 192.168.111.5:6443: connect: no route to host
```

02_configure_host.sh
```diff
@@ -255,7 +255,7 @@ ANSIBLE_FORCE_COLOR=true ansible-playbook \
     -e "{use_firewalld: True}" \
     -e "provisioning_interface=$PROVISIONING_NETWORK_NAME" \
     -e "baremetal_interface=$BAREMETAL_NETWORK_NAME" \
-    -e "{provisioning_host_ports: [80, ${LOCAL_REGISTRY_PORT}, 8000, ${INSTALLER_PROXY_PORT}, ${AGENT_PXE_SERVER_PORT}]}" \
+    -e "{provisioning_host_ports: [80, ${LOCAL_REGISTRY_PORT}, 8000, ${INSTALLER_PROXY_PORT}, ${AGENT_PXE_SERVER_PORT}, 22623, 6443, 443]}" \
```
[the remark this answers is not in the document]
andfasano: Update: it turned out that I had a polluted environment with ports 443/6443 already opened, and on a subsequent check it appears that currently all the ports defined in `provisioning_host_ports` are permanent, as they are not cleaned up. So as discussed in this case it'd be better not to use `provisioning_host_ports` to define the ports required to be opened for `none` scenario

PR #1543, opened 2023-06-28
Power down redfish bm nodes before provisioning

derekhiggins:


cybertron:
Redfish is our default now, right? So the passing ci jobs mean this is working?

derekhiggins:
> Redfish is our default now, right? So the passing ci jobs mean this is working?

It isn't used in CI because its wrapped in
> if [[ "${NODES_PLATFORM}" == "baremetal" ]] && [ -n "$NODES_FILE" ] ; then

in CI NODES_PLATFORM == libvirt 

but I have been using this in a baremetal env and all is working

derekhiggins:

log gather failure, unrelated and this PR isn't tested in the job

PR #1548, opened 2023-07-07
Use a v4 provisioning address when v4 is primary in dualstack

dtantsur:
Currently, dualstack is broken because a v6 provisioning node cannot
reach quay or the v4-only local mirror.


mkowalski (approved):
Technically speaking it's not because "quay doesn't support IPv6" (because it supports it already for some time now) but because internal Red Hat IPv6 network doesn't have outside-world connectivity, i.e.

```
[root@localhost ~]# ping6 google.ch
PING google.ch(bj-in-f94.1e100.net (2607:f8b0:4004:c0b::5e)) 56 data bytes
^C
--- google.ch ping statistics ---
3 packets transmitted, 0 received, 100% packet loss, time 2024ms
```

It doesn't affect the implementation, but the side-effect is that if you use dev-scripts dual-stack (or even single-stack v6) on a host that has proper IPv6 connectivity, you don't need the `oc-mirror` dance anymore

Anyway for our use we still should mirror whenever IPv6 is primary network

dtantsur:
Thank you Mat!

/retest

The CI should be better now (except for dualstack).

PR #1551, opened 2023-07-16
Fix DHCP network mode on agent-config.yaml

danielerez:
NETWORKING_MODE env var supports 'DHCP' value in upper-case. Hence, in order to support this network mode in agent-config template, changed to upper-case 'DHCP'.


bfournie:
Nice catch!

/approve
/lgtm

PR #1552, opened 2023-07-17
Fix failure if nmstateconfig.yaml doesn't exist

bfournie:
Fix script failure if the file doesn't exist.


pawanpinjarkar (commented):
/lgtm

PR #1553, opened 2023-07-21
Agent: Move userManagedNetworking under networking

zaneb:
This data ended up one level too high, which broke SNO CI.


andfasano:
/lgtm

Wondering how the original patch passed for the SNO job

zaneb:
> Wondering how the original patch passed for the SNO job

It [didn't](https://prow.ci.openshift.org/pr-history/?org=openshift-metal3&repo=dev-scripts&pr=1541). It failed to get a lease from Equinix so the error wasn't visible, and the job isn't required so it merged anyway.

PR #1554, opened 2023-07-24
Agent: Remove bogus override of EXTERNAL_SUBNET_V6

zaneb:
Not all scripts in a agent-based install source agent/common.sh - specifically, the configure step 02_configure_host.sh does not, and this is the step that sets up the DHCP config.

The result of overriding EXTERNAL_SUBNET_V6 in agent/common.sh then is that we get inconsistent values in different parts of dev-scripts. The DHCP server is set up with the default value from network.sh, while the install-config gets the overridden value.

The two values differ only in their prefix length (120 vs. 64). Unfortunately, assisted refuses to accept that hosts are in the machine network if the prefix length of the network they are in does not match exactly the prefix length of the machine network specified in the install-config. When this happens, assisted rejects the hosts with the cryptic error "No connectivity to the majority of hosts in the cluster." (see https://issues.redhat.com/browse/OCPBUGS-7928).

As a result, testing was broken on multinode DHCP scenarios with IPv6, including the standard HA_IPV4V6_DHCP scenario we use for e2e-agent-ha-dualstack in many repositories.

The original reason for this was that the mask length of the clusterNetwork was used for the static IP addresses on the machineNetwork. By no longer conflating those two, we eliminate the need to force the two networks to have the same subnet mask by the use of these overrides.


zaneb:
Note that we started seeing failures now because #1551 made DHCP scenarios actually use DHCP. Prior to that they were silently using static IPs.

bfournie:
This fix causes the dual-stack job to fail without DHCP, i.e. `COMPACT_IPV4V6` with the same cryptic `No connectivity to the majority of hosts `error.  That must have been why this code was added with dual-stack support in https://github.com/openshift-metal3/dev-scripts/commit/25f40208cf3b54a8da45f0b19846b37b3697966f, the DHCP issue prevented it from catching DHCP failures are that time.

I think we need to leave it in in this case, e.g.

```
if [[ $NETWORKING_MODE != "DHCP" ]] && [[ $IP_STACK == 'v4v6' ]]; then
   # Override to consistently use the proper subnet when using dual-stack
   export CLUSTER_HOST_PREFIX_V4="24"
   if [[ "$IP_STACK" != "v4" ]]; then
      export EXTERNAL_SUBNET_V6="fd2e:6f44:5dd8:c956::/64"
   fi
fi
```

zaneb:
Thanks for testing that out @bfournie! I found the root cause and I think we should be good to go now (I tested with `HA_IPV4V6`).

rwsu (commented):
/lgtm

PR #1557, opened 2023-07-30
OCPBUGS-16912: Agent use nmstateconfig for DHCP tests

bfournie:
The Agent IPV6_DCHCP test fails due to the address assigned via DHCPv6 not matching the rendezvousIP. The reason is that by default NetworkManager sends DHCPv6 Solicitations with a Client ID of DUID-UID (prefix=00:04, see https://datatracker.ietf.org/doc/html/rfc6355) containing a machine specific value. When using nmstateconfig for DHCP, the NetworkManager keyfiles are setup for each interface with dhcp-duid=ll (prefix=00:03, see https://datatracker.ietf.org/doc/html/rfc3315#section-9.4) which means that the DHCPv6 Solicitation is based on the mac address. This allows the DHCPv6 server to assign a deterministic address which can be set to match the rendezvousIP, e.g
https://github.com/metal3-io/metal3-dev-env/blob/main/vm-setup/roles/libvirt/templates/network.xml.j2#L93

Using the nmstateconfig is also recommended for IPv4 and is consistent with configurations we have seen from customers. For both cases the nmstate configs are used - https://nmstate.io/examples.html#dynamic-ip-configuration.

https://github.com/openshift-metal3/dev-scripts/pull/1555 is also needed and included here.

### DIFF
--- a/agent/roles/manifests/templates/nmstateconfig_yaml.j2
+++ b/agent/roles/manifests/templates/nmstateconfig_yaml.j2
@@ -17,19 +17,19 @@ spec:
     {{ net.interfaces(macs[loop.index0]) }}
 {% if ip_stack == "v4" %}
         ipv4:
-          {{ net.ip(ips[loop.index0], cluster_host_prefix_v4) }}
+          {{ net.ip(networking_mode, ips[loop.index0], cluster_host_prefix_v4) }}
     {{ net.dns(provisioning_host_external_ip) }}
     {{ net.route("0.0.0.0/0", provisioning_host_external_ip) }}
 {% elif ip_stack == "v6" %}
         ipv6:
-          {{ net.ip(ipsv6[loop.index0], cluster_host_prefix_v6) }}
+          {{ net.ip(networking_mode, ipsv6[loop.index0], cluster_host_prefix_v6) }}
     {{ net.dns(provisioning_host_external_ip) }}
     {{ net.route("::/0", provisioning_host_external_ip) }}
 {% else %}
         ipv4:
-          {{ net.ip(ips[loop.index0], cluster_host_prefix_v4) }}
+          {{ net.ip(networking_mode, ips[loop.index0], cluster_host_prefix_v4) }}
         ipv6:
-          {{ net.ip(ipsv6[loop.index0], cluster_host_prefix_v6) }}
+          {{ net.ip(networking_mode, ipsv6[loop.index0], cluster_host_prefix_v6) }}
     {{ net.dns_dualstack(provisioning_host_external_ip, provisioning_host_external_ip_dualstack) }}
     {{ net.route_dualstack(provisioning_host_external_ip, provisioning_host_external_ip_dualstack) }}
 {% endif %}
--- a/agent/roles/manifests/templates/nmstateconfig_yaml.j2
+++ b/agent/roles/manifests/templates/nmstateconfig_yaml.j2
@@ -17,19 +17,19 @@ spec:
     {{ net.interfaces(macs[loop.index0]) }}
 {% if ip_stack == "v4" %}
         ipv4:
-          {{ net.ip(networking_mode, ips[loop.index0], cluster_host_prefix_v4) }}
+          {{ net.ip(networking_mode, ips[loop.index0], external_subnet_v4_prefixlen) }}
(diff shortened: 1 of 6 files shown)


agent/roles/manifests/templates/nmstateconfig_yaml.j2
```diff
@@ -17,19 +17,19 @@ spec:
     {{ net.interfaces(macs[loop.index0]) }}
 {% if ip_stack == "v4" %}
         ipv4:
-          {{ net.ip(ips[loop.index0], cluster_host_prefix_v4) }}
```
zaneb, line 20: It would be good to have the change that is cleaning up omissions from #1554 in a separate commit.

bfournie: Yes I will push a separate commit for that

rwsu:
Hi @bfournie. Can you help me understand how this works? The ostestbm network has the DHCP host-id to host-ip mappings. For example:
````
  <ip family='ipv6' address='fd2e:6f44:5dd8:c956::1' prefix='120'>
    <dhcp>
      <range start='fd2e:6f44:5dd8:c956::14' end='fd2e:6f44:5dd8:c956::3c'/>
      <host id='00:03:00:01:00:04:da:4e:c7:c1' name='master-0' ip='fd2e:6f44:5dd8:c956::14'>
        <lease expiry='60' unit='minutes'/>
     </host>
````
The MAC address is embedded in the host-id. Is the prefix "00:03:00:01:" ignored when matching an interface's MAC address to the nmstate configs because dhcp-duid=ll is used?

zaneb:
https://datatracker.ietf.org/doc/html/rfc8415#section-11.4

bfournie:
> Hi @bfournie. Can you help me understand how this works? The ostestbm network has the DHCP host-id to host-ip mappings. For example:
> 
> ```
>   <ip family='ipv6' address='fd2e:6f44:5dd8:c956::1' prefix='120'>
>     <dhcp>
>       <range start='fd2e:6f44:5dd8:c956::14' end='fd2e:6f44:5dd8:c956::3c'/>
>       <host id='00:03:00:01:00:04:da:4e:c7:c1' name='master-0' ip='fd2e:6f44:5dd8:c956::14'>
>         <lease expiry='60' unit='minutes'/>
>      </host>
> ```
> 
> The MAC address is embedded in the host-id. Is the prefix "00:03:00:01:" ignored when matching an interface's MAC address to the nmstate configs because dhcp-duid=ll is used?

When `duid-ll` is set the host will send a Client ID containing the "00:03:00:01" prefix and the mac suffix that will match what was added by dev-scripts here https://github.com/metal3-io/metal3-dev-env/blob/main/vm-setup/roles/libvirt/templates/network.xml.j2#L93. 
For example the host will send:
`dnsmasq-dhcp[1318507]: DHCPSOLICIT(ostestbm) 00:03:00:01:00:f3:f4:3e:09:88`

When `duid-ll` is not set, by default NetworkManager will send a Client ID with a DUID-UID, which is generated on the host and can't be determined, so there's no way to match. For example the host still send:
`dnsmasq-dhcp[1318507]: DHCPSOLICIT(ostestbm) 00:04:90:2a:c0:06:7b:79:85:30:88:7c:e7:81:27:2a:d4:09`

rwsu:
/lgtm
Thanks for the explanation. I had a trouble understanding what the prefix means. I understand now that 00:03 means to use Link Layer DUID and 00:01 after that means ethernet hardware type, then followed by MAC address or link-layer address.

zaneb:
Is this change still necessary now that we have openshift/installer#7384?
I worry that making this change could potentially cover up regressions in future.

bfournie:
> Is this change still necessary now that we have [openshift/installer#7384](https://github.com/openshift/installer/pull/7384)? I worry that making this change could potentially cover up regressions in future.

I agree its no longer required for DHCP IPv6 after the installer change, but I think its useful to have the nmstate config for v4 and v6 since it seems to be common customer configuration.

How about if I add a flag to enable NMSTATE_CONFIG_FOR_DHCP which would be disabled by default so we'd keep the current behavior but could easily test it out?

rwsu (commented):
/lgtm

PR #1561, opened 2023-08-07
appliance: ensure config-image is created once

danielerez:
The config-image should be created only once (as the ISO is shared by all machines).
Hence, moved the config-image creation logic outside of the 'attach_appliance_diskimage' func.
I.e. Since this func is called twice (both for masters and workers).

### DIFF
--- a/agent/06_agent_create_cluster.sh
+++ b/agent/06_agent_create_cluster.sh
@@ -123,11 +123,6 @@ function attach_appliance_diskimage() {
     local config_image_drive="sdd"
     local appliance_disk_image="${OCP_DIR}/appliance.raw"
 
-    # Create the config ISO
-    mkdir -p ${config_image_dir}
-    cp ${asset_dir}/*.yaml ${config_image_dir}
-    create_config_image
-
     for (( n=0; n<${2}; n++ ))
     do
         name=${CLUSTER_NAME}_${1}_${n}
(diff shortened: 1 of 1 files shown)


agent/06_agent_create_cluster.sh
```diff
@@ -360,6 +355,11 @@ case "${AGENT_E2E_TEST_BOOT_MODE}" in
     ;;
 
   "DISKIMAGE" )
+    # Create the config ISO
+    mkdir -p ${config_image_dir}
+    cp ${asset_dir}/*.yaml ${config_image_dir}
+    create_config_image
```
andfasano, line 361: Could make sense to move these commands directly within the `create_config_image`? They anyhow prepare the input ingredients for the subsequent `create config-image` installer command

danielerez: It's actually intentionally separated, see logic for it in 'create_factory_image': https://github.com/openshift-metal3/dev-scripts//blob/4f8e2a93864a5ee6d613174413741812dc4e20d8/agent/06_agent_create_cluster.sh#L43

andfasano:
/lgtm

As a side note, I think it'd be useful to have an appliance job, even optional, in dev-scripts openshift/release CI, so that it could be possible to verify/tests those PRs directly in the CI (without the need to run them locally) cc @bfournie @rwsu

PR #1562, opened 2023-08-09
Removing sourcing of the non-step network.sh script

andfasano:
Sourcing the non-step script `network.sh` within the non-step script `agent/common.sh` introduces a not explicit dependency that may produce unwanted side-effects.

In particular, since the `agent/common.sh` is source into the `agent/01_agent_requirements.sh` _before_ the generic `/01_install_requirements.sh`, all the functions in `network.sh` needing a specific software requirement - like ansible for [nth_ip](https://github.com/openshift-metal3/dev-scripts/blob/f4c1080455a3d1928b6ee675ec379de31ad703e5/network.sh#L7) - are going to fail.

The blocking issue was initially discovered by @danielerez while trying to activate the `INSTALLER_PROXY` config var, which requires the [PROVISIONING_HOST_EXTERNAL_IP](https://github.com/openshift-metal3/dev-scripts/blob/f4c1080455a3d1928b6ee675ec379de31ad703e5/network.sh#L154-L158) that may need the `nth_ip` function)

### DIFF
--- a/agent/common.sh
+++ b/agent/common.sh
@@ -1,8 +1,6 @@
 #!/usr/bin/env bash
 set -euxo pipefail
 
-source network.sh
-
 export AGENT_STATIC_IP_NODE0_ONLY=${AGENT_STATIC_IP_NODE0_ONLY:-"false"}
 
 export AGENT_USE_ZTP_MANIFESTS=${AGENT_USE_ZTP_MANIFESTS:-"false"}


andfasano:
cc @cybertron @elfosardo for additional review, as not sure about the original issue

agent/common.sh
```diff
@@ -1,8 +1,6 @@
 #!/usr/bin/env bash
 set -euxo pipefail
 
-source network.sh
```
bfournie, line 4: That source was added in https://github.com/openshift-metal3/dev-scripts/pull/1508, is it no longer necessary?

andfasano: Waiting for a reply from @cybertron (or @elfosardo), as it doesn't seem necessary. And in case it was, it'd be better in any case to source the file here (but eventually in a step script)

bfournie: Might be worth testing this case per the comment in #1508 
>> Note that this is currently causing make clean to fail in ipv6 deployments unless you explicitly set IP_STACK.

andfasano: Thanks @bfournie good idea

cybertron:
/lgtm

Looks fine. The problematic IP_STACK reference was removed in https://github.com/openshift-metal3/dev-scripts/commit/8fd09eb3e6102e78e229ad077452562f0ea8590c

PR #1563, opened 2023-08-09
utils: fix mechanism for mirroring OLM operators

mkowalski:


mkowalski:
Tested manually using disconnected IPv6 setup

```
# oc get catalogsource -A
NAMESPACE               NAME                     DISPLAY                                                                                    TYPE   PUBLISHER   AGE
openshift-marketplace   mirror-catalog-for-olm   Mirror index for OLM packages from registry.redhat.io/redhat/redhat-operator-index:v4.13   grpc   Local       11m
```

```
# oc get catalogsource mirror-catalog-for-olm -n openshift-marketplace -o yaml
apiVersion: operators.coreos.com/v1alpha1
kind: CatalogSource
[...]
spec:
  displayName: Mirror index for OLM packages from registry.redhat.io/redhat/redhat-operator-index:v4.13
  grpcPodConfig:
    securityContextConfig: restricted
  image: virthost.ostest.test.metalkube.org:5000/olm-index/redhat-operator-index:v4.13
  publisher: Local
  sourceType: grpc
  updateStrategy:
    registryPoll:
      interval: 30m
status:
  connectionState:
    address: mirror-catalog-for-olm.openshift-marketplace.svc:50051
    lastConnect: "2023-08-09T14:42:11Z"
    lastObservedState: READY
  registryService:
    createdAt: "2023-08-09T14:39:28Z"
    port: "50051"
    protocol: grpc
    serviceName: mirror-catalog-for-olm
    serviceNamespace: openshift-marketplace
```

```
# oc get clusterserviceversion -n openshift-nmstate
NAME                                              DISPLAY                       VERSION               REPLACES   PHASE
kubernetes-nmstate-operator.4.13.0-202307242035   Kubernetes NMState Operator   4.13.0-202307242035              Succeeded
```

elfosardo (commented):
/lgtm

config_example.sh
```diff
@@ -593,7 +593,13 @@ set -x
 # Note - MIRROR_OLM is currently not supported when MIRROR_COMMAND=oc-mirror, it
```
cybertron, line 593: Should we remove/change this note?

06_create_cluster.sh
```diff
@@ -64,6 +64,14 @@ if [[ -n "${APPLY_EXTRA_WORKERS}" ]]; then
     fi
 fi
 
+# Mirrored OLM operators need to have respective ICSP and CatalogSources applied. This can be done
```
cybertron, line 67: This works, but the other option is to copy the manifests into $OCP_DIR/manifests so they're included in the deployment by the installer. That's how the old version handled this (line 138 and 139 of the 04 script).

mkowalski: Oh okay, I need to try it out then. I just didn't know about this "auto-apply mechanism" so I took the manual path

config_example.sh
mkowalski: By the way I tested and `MIRROR_COMMAND=oc-mirror` works **only** for deployments via Agent Installer. It's another story that at some point we need to move everything to use `oc-mirror`, but now even if we wanted, we can't run our flow with it

cybertron: Ah, fair enough. I had it in my head that we needed to change commands too, but I guess that is not the case.

cybertron:
/lgtm

The previous version of this worked for me locally, and I believe the only change in this one is how we deploy the manifests.

config_example.sh
mkowalski: The default for `MIRROR_COMMAND` is `oc-adm` so as long as this doesn't change, we are safe

cybertron:
/approve

Riccardo previously lgtm'd this too and it doesn't affect any of the existing ci jobs on this repo so this didn't break any of the failing optional jobs.

PR #1564, opened 2023-08-09
AGENT-428: Add support for bonding with unique macs

bfournie:
Add a new variable - BOND_CONFIG, that defines the bonding mode to be used with nmstate configuration when creating 2 nics in a bond. Use the existing variable - BOND_PRIMARY_INTERFACE to indicate that a bond should be created, a random mac will be generated for the 2nd interface so that the two bond interfaces have unique macs.

 Co-Authored-By: Richard Su <rwsu@redhat.com>

### DIFF
--- a/agent/05_agent_configure.sh
+++ b/agent/05_agent_configure.sh
@@ -83,8 +83,15 @@ function get_static_ips_and_macs() {
             add_dns_entry ${AGENT_NODES_IPSV6[i]} ${AGENT_NODES_HOSTNAMES[i]}
         fi
 
-        # Get the generated mac addresses
+        # Get the generated mac addresses or, in the case of a bond, the configured ones
         AGENT_NODES_MACS+=($(sudo virsh dumpxml $cluster_name | xmllint --xpath "string(//interface[descendant::source[@bridge = '${BAREMETAL_NETWORK_NAME}']]/mac/@address)" -))
+        if [[ ${BOND_CONFIG} != "none" ]]; then
+           mac1=${AGENT_NODES_MACS[-1]}
+           # increment to get next mac
+           next=$(printf "%02x" $((${mac1##*:} + 1)))
+           mac2="${mac1%:*}:$next"
+           AGENT_NODES_MACS+=(${mac2})
+        fi
     done
 }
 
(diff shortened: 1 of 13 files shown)


agent/05_agent_configure.sh
```diff
@@ -407,6 +428,15 @@ else
   configure_dnsmasq ${ip} ""
 fi
 
+# Note that BOND_PRIMARY_INTERFACE creates the same mac on both ports so can't be used here
```
cybertron, line 431: I would love to apply this logic to the existing feature. While the duplicate macs work, it's obviously less than ideal. Would you have any interest in just updating BOND_PRIMARY_INTERFACE to use this method?

bfournie: > I would love to apply this logic to the existing feature. While the duplicate macs work, it's obviously less than ideal. Would you have any interest in just updating BOND_PRIMARY_INTERFACE to use this method?

@cybertron Yes I'd love to make this common and just use one variable. Let me try moving it out of agent. I would eventually need some help to test it in an non-agent setup.

One thing regarding the naming - since the env variable can be set the bonding mode I'm not sure if `BOND_PRIMARY_INTERFACE` is the right name for the setting.

bfournie: @cybertron take a look at the latest changes and see if that makes sense,

cybertron:
Hmm, I tried this locally but I think it may break the bond job because it changes the MAC from what dnsmasq expects. We could get around that with static IPs, but could we avoid the problem by reusing the MAC for the first interface and only generating a new one for the second?

bfournie:
> > Hmm, I tried this locally but I think it may break the bond job because it changes the MAC from what dnsmasq expects. We could get around that with static IPs, but could we avoid the problem by reusing the MAC for the first interface and only generating a new one for the second?
> 
> Yeah, I also found the issue with DHCP. I'm looking at changing it per your suggestion /hold

Updated to use original mac (from metal3-dev-env) on first interface and new generated mac on 2nd.  DHCP appears to work now.

cybertron:
/lgtm

Yep, this is working for me now too. Thanks!

config_example.sh
```diff
@@ -650,6 +650,12 @@ set -x
 # More info here: https://github.com/openshift-metal3/dev-scripts/pull/1241#issuecomment-846067822
 # export BMO_WATCH_ALL_NAMESPACES="true"
 
+# For platforms that use nmstate configuration, the bond mode is set to the value.
+# The possible values are 'balance-rr', 'active-backup', 'broadcast', 'balance-xor',
+# 'balance-tlb', and 'balance-alb'.
+# The default is no bonding, so value is set to 'none'.
+# export BOND_CONFIG='none'
+
```
rwsu, line 658: It wasn't obvious to me that in order to use BOND_CONFIG, I had to also set BOND_PRIMARY_INTERFACE. Perhaps if BOND_CONFIG is not set to "none", then it should automatically set BOND_PRIMARY_INTERFACE to some value. It is also not clear what value I needed to set for BOND_PRIMARY_INTERFACE. I guessed "eth0", and that seemed to work.

rwsu (changes requested):

config_example.sh
bfournie: Good point, it is a bit clumsy but I wanted to keep the existing behavior for `BOND_PRIMARY_INTERFACE`.
I'll set it when` BOND_CONFIG!='none'`

bfournie: Updated

rwsu (commented):
/lgtm

cybertron:
/lgtm
/approve
/retest-required

This passed the bond job that at least partially exercises the new functionality so I think it should be good to go.

PR #1565, opened 2023-08-14
utils: allow mirroring single container image

mkowalski:


mkowalski:
Required-for: https://github.com/openshift/release/pull/42097

cybertron:
/lgtm

Tested this locally and I can see that the image was successfully mirrored into my local registry. Since it's not being used in any of the ci jobs yet I don't see how it could have caused the couple of remaining failures.

derekhiggins:
lgtm but before adding it just thought I'd point out the mirroring support to ASSETS_EXTRA_FOLDER that merged recently
https://github.com/openshift-metal3/dev-scripts/pull/1569/files 
does this help your usecase? is this still needed?

mkowalski:
> lgtm but before adding it just thought I'd point out the mirroring support to ASSETS_EXTRA_FOLDER that merged recently https://github.com/openshift-metal3/dev-scripts/pull/1569/files does this help your usecase? is this still needed?

Yes and no... If I read it correctly (correct if I'm wrong) this works in a way that I would need to craft a manifest deploying something, place this manifest in the `ASSETS_EXTRA_FOLDER` and then machinery takes care so that images used by the manifest are mirrored and manifest is updated to use the mirror.

But in our use case there is no any manifest anywhere, what we need is only a vanilla set of images present in our local registry. Later on we may or may not deploy something that uses them, but the main issue I see is that we do not have any manifest at install-time.

To put even more there, we may be deploying workload that needs an image from the local registry and then this thing that we deployed will deploy something else what also needs a local image (kind of a container deploying next containers)

Not sure if that explains my use case well but I'm happy to clarify further

derekhiggins:
/approve
> Not sure if that explains my use case well but I'm happy to clarify further

Yup, makes sense, thanks for the explanation,

PR #1566, opened 2023-08-15
Do not run 'adm release extract' with KUBECONFIG set

dtantsur:
Presumably after https://github.com/openshift/oc/pull/1521, the command
fails if KUBECONFIG does not exist. Since the command is run before
the cluster is created, extracting always fails. Unset KUBECONFIG.


dtantsur:
Will not work.

PR #1567, opened 2023-08-15
Remove agent workaround for kubeadmin-password

bfournie:
Now that the kubeadmin-password is generated this workaround can be removed.


rwsu (commented):
/lgtm

PR #1570, opened 2023-08-21
TEST: wondering if anything passes

derekhiggins:
/hold


romfreiman:
just verifying - is the host that runs the installer is fips enabled?

derekhiggins:
> just verifying - is the host that runs the installer is fips enabled?

I'm not aware of anything we've done to enable it so probably not

derekhiggins:
https://github.com/openshift/release/pull/42557

PR #1571, opened 2023-08-21
Remove v6v4 from if statement evaluation since it forces deployment to be disconnected 

jadhaj:


common.sh
```diff
@@ -469,7 +469,7 @@ if [[ ! -z ${AGENT_E2E_TEST_BOOT_MODE} ]]; then
 fi
 
 # For IPv6 (default case) mirror images are used since quay doesn't support IPv6
-if [[ -n "$MIRROR_IMAGES" || -z "${IP_STACK:-}" || "$IP_STACK" = "v6" || "$IP_STACK" = "v6v4" ]]; then
+if [[ -n "$MIRROR_IMAGES" || -z "${IP_STACK:-}" || "$IP_STACK" = "v6" ]]; then
```
andfasano, line 472: Reverting could broke https://github.com/openshift-metal3/dev-scripts/pull/1548, so it doesn't seem a good option. 
IMHO a better option would be to set `export MIRROR_IMAGES=true` when `IP_STACK` is set to `v6v4` in [network.sh](https://github.com/openshift-metal3/dev-scripts/blob/425b66a58cf266dafb7e3df8ce6aaf6b017d49d2/network.sh#L141) (and I guess jobs using it already explicitly set in the CI)

/cc @dtantsur @elfosardo

jadhaj: Yep, export MIRROR_IMAGES=true will solve it but will also make deployment a disconnected one 

The deployment is working dine with MIRROR_IMAGES unset and v6 as primary in dualstack

dtantsur:
/hold

The line is correct. With v6 as a primary stack, we can end up with v6-only during provisioning, which will make quay unavailable.

pamoedom:
> /hold
> 
> The line is correct. With v6 as a primary stack, we can end up with v6-only during provisioning, which will make quay unavailable.

Hi @dtantsur, just trying to clarify one thing here, quay.io is reachable via IPv6 since months ago (PROJQUAY-131), are you referring to registry.ci.openshift.org?

dtantsur:
No, I mean quay. If it's now reachable, we may reconsider the whole approach of forcing mirroring with IPv6. In that case, the change should also cover the normal V6 case.

pamoedom:
> No, I mean quay. If it's now reachable, we may reconsider the whole approach of forcing mirroring with IPv6. In that case, the change should also cover the normal V6 case.

Yes, it's publicly reachable via IPv6:
~~~
$ dig AAAA quay.io +short
2600:1f18:483:cf00:3599:c8ae:db8a:ed50
2600:1f18:483:cf02:cc5a:b6c5:df70:ab6d
2600:1f18:483:cf00:8030:5c1f:79f0:15df
2600:1f18:483:cf00:e11b:135b:b044:a79e
2600:1f18:483:cf01:57b9:2c4:9432:85c0
2600:1f18:483:cf01:1d94:746f:43c9:4239
2600:1f18:483:cf02:4fe3:1d26:b265:a5fa
2600:1f18:483:cf01:bd4f:a00e:1d1c:8607

---

$ nc -vz quay.io 443
Connection to quay.io (2600:1f18:483:cf01:57b9:2c4:9432:85c0) 443 port [tcp/https] succeeded!
~~~

PR #1572, opened 2023-08-22
appliance: remove unused appliance.raw disk image

danielerez:
After copying appliance.raw disk image for each node, the source file is no longer needed.
Hence, deleting the file to avoid storage overconsumption on CI machines.

### DIFF
--- a/agent/06_agent_create_cluster.sh
+++ b/agent/06_agent_create_cluster.sh
@@ -366,6 +366,10 @@ case "${AGENT_E2E_TEST_BOOT_MODE}" in
     # Attach the diskimage to nodes
     attach_appliance_diskimage master $NUM_MASTERS
     attach_appliance_diskimage worker $NUM_WORKERS
+
+    # Delete the unused appliance.raw file
+    # (to avoid storage overconsumption on the CI machine)
+    sudo rm -f "${OCP_DIR}/appliance.raw"
     ;;
 esac
 


danielerez:
/pj-rehearse pull-ci-openshift-appliance-master-e2e-compact-ipv4-static

agent/06_agent_create_cluster.sh
```diff
@@ -366,6 +366,10 @@ case "${AGENT_E2E_TEST_BOOT_MODE}" in
     # Attach the diskimage to nodes
     attach_appliance_diskimage master $NUM_MASTERS
     attach_appliance_diskimage worker $NUM_WORKERS
+
+    # Delete the unused appliance.raw file
+    # (to avoid storage overconsumption on the CI machine)
+    sudo rm -f "${OCP_DIR}/appliance.raw"
```
andfasano, line 372: This sounds like a task for the clean step?

danielerez: > This sounds like a task for the clean step?

We actually need it before the cleanup, intentionally to avoid storage exhaustion during cluster installation.
As the CI machines seems to have only about 200GiB, so a compact cluster with 3 hosts takes about 100% storage now.

andfasano: I see. Still, it seems a little bit fragile, as the underlying machine storage may vary. If you have a reasonably rough estimate of the minimum space required, it's possible to tweak the [MIN_SPACE_REQUIRED](https://github.com/openshift-metal3/dev-scripts/blob/425b66a58cf266dafb7e3df8ce6aaf6b017d49d2/sanitychecks.sh#L6) config var to ensure the environment has enough storage for the current configuration

danielerez: Yeah, it could still be fragile. But anyway, this file is redundant aftering creating the VMs, so sort of best effort...
Hmm, so how can we use MIN_SPACE_REQUIRED? Can we pass it in the e2e definitions (in release)?

danielerez: I see 'MIN_SPACE_REQUIRED' actually does nothing other than validation? https://github.com/openshift-metal3/dev-scripts//blob/91233d947557dbac45de8a820491de115fe7a9e8/sanitychecks.sh#L35

andfasano: Yes, it prevents moving further if the basic conditions are not met

andfasano: (MIN_SPACE_REQUIRED may be defined within a DEVSCRIPTS_CONFIG section in the CI step, ie)

andfasano: Ok, from a discussion with @danielerez looks like that anyhow this intermediate removal may be useful in providing some more breath (~25Gib), especially for those situations where to the storage is not excessive (and unfortunately this is not somethin easily configurable in the CI). The usage of a reasonable `MIN_SPACE_REQUIRED` value may help (in terms of preventing installation doomed to fail) especially for local setups.

/approve

PR #1574, opened 2023-08-25
Update OWNERS

derekhiggins:
elfosardo is a long term consitent contributor
Also remove flaper87 and hardys who are no
longer working on the project


derekhiggins:
@andfasano @dtantsur

andfasano:
Late to the party but definitely fine for me

PR #1576, opened 2023-08-25
AGENT-698: Agent support for OCI platform

bfournie:
Initial support for the external OCI platform. This sets the libvirt domain manufacturer and product to allow the assisted-service validations for this platform to pass.

Requires https://github.com/openshift/assisted-service/pull/5438 and https://github.com/openshift/installer/pull/7442

In addition, this requires changes to either:
a) use a minimalISO instead of a fullISO in the agent-based-installer 
b) not require a mimimalISO in assisted-service

### DIFF
--- a/agent/roles/manifests/templates/agent-cluster-install_yaml.j2
+++ b/agent/roles/manifests/templates/agent-cluster-install_yaml.j2
@@ -4,7 +4,7 @@ metadata:
   name: test-agent-cluster-install
   namespace: {{ cluster_namespace }}
 spec:
-{% if num_masters != "1" %}
+{% if (platform_type != "none") and (platform_type != "external") %}
   apiVIP: {{ api_vip }} 
   ingressVIP: {{ ingress_vip }} 
 {% endif %}
@@ -43,8 +43,12 @@ spec:
     - cidr: {{ external_subnet_v6 }}
 {% endif %}
     networkType: {{ network_type }}
-{% if platform_type == "none" %}
+{% if (platform_type == "none") or (platform_type == "external") %}
     userManagedNetworking: true
+{% endif %}
+{% if (platform_type == "external") %}
+  external:
+    platformName: oci
 {% endif %}
   provisionRequirements:
     controlPlaneAgents: {{ num_masters }}
--- a/agent/05_agent_configure.sh
+++ b/agent/05_agent_configure.sh
@@ -277,7 +277,7 @@ function enable_load_balancer() {
   local load_balancer_ip=${2}
   local HTTP_PORT=80
 
-  if [[ "${AGENT_PLATFORM_TYPE}" == "none" && "${NUM_MASTERS}" > "1" ]]; then
+  if [[ "${AGENT_PLATFORM_TYPE}" == "none" || "${AGENT_PLATFORM_TYPE}" == "external" ]] && [[ "${NUM_MASTERS}" > "1" ]]; then
 
     # setup haproxy as the load balancer
     if [[ "${IP_STACK}" == "v6" ]]; then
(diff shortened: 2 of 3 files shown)


bfournie:
Still a POC
/hold

pawanpinjarkar:
Below PRs will be useful to generate minimal ISO 
https://github.com/openshift/assisted-image-service/pull/137 
https://github.com/openshift/installer/pull/7450

pawanpinjarkar:
I successfully tested compact IPV4 cluster installation for baremetal  with minimal ISO generated for agent installer

bfournie:
I was able to get a successful install using the following env vars:
```
export INSTALLER_REPO_PR=7442
export AGENT_INSTALLER_API_SERVER_LOCAL_REPO=/home/stack/assisted-service
export AGENT_INSTALLER_API_SERVER_DOCKERFILE=Dockerfile.assisted-service.ocp
export AGENT_PLATFORM_TYPE=external
export OPENSHIFT_RELEASE_STREAM=4.14
export OPENSHIFT_RELEASE_TYPE=ci

```
In addition I had to manually change the installer code to include the define for `ExternalPlatformType PlatformType = External` from https://github.com/openshift/assisted-service/pull/5438 and the image-service changes https://github.com/openshift/assisted-image-service/pull/137

agent/06_agent_create_cluster.sh
```diff
@@ -99,6 +123,11 @@ function attach_agent_iso() {
         agent_iso="${OCP_DIR}/agent.iso"
     fi
 
+    if [[ "${AGENT_PLATFORM_TYPE}" == "external" ]]; then
```
andfasano, line 126: It seems to me that this block may be logically better placed in step 05 than 06

bfournie: Agree. Updated

agent/roles/manifests/templates/agent-cluster-install_yaml.j2
```diff
@@ -43,7 +43,7 @@ spec:
     - cidr: {{ external_subnet_v6 }}
 {% endif %}
     networkType: {{ network_type }}
-{% if platform_type == "none" %}
+{% if (platform_type == "none") or (platform_type == "external") %}
```
rwsu, line 46: In addition, now that the assisted-service PR to add external type merged, ExternalPlatformSpec should be added here. https://github.com/openshift/assisted-service/pull/5438/files#diff-8a5d8069ae1e2777920acfce8fd2d122f8e62c3e89b9c20b791b8cd35776dc9eR198

rwsu (commented):
/lgtm

agent/roles/manifests/templates/agent-cluster-install_yaml.j2
```diff
@@ -43,8 +43,12 @@ spec:
...
     userManagedNetworking: true
+{% endif %}
+{% if (platform_type == "external") %}
+  external:
+    platformName: external
```
pawanpinjarkar, line 51: `platformName: oci`

agent/roles/manifests/templates/agent-cluster-install_yaml.j2
```diff
@@ -4,7 +4,7 @@ metadata:
   name: test-agent-cluster-install
   namespace: {{ cluster_namespace }}
 spec:
-{% if num_masters != "1" %}
+{% if (platform_type != "none") and (platform_type != "external") %}
```
pawanpinjarkar, line 7: With `external` platform, other scenarios like compact, HA are also possible

agent/roles/manifests/templates/agent-cluster-install_yaml.j2
bfournie: Good catch, fixed.

agent/roles/manifests/templates/agent-cluster-install_yaml.j2
bfournie: Yes this fix is necessary to support that, otherwise the VIPs would be set which fails when UserManagedNetwork=true

agent/roles/manifests/templates/agent-cluster-install_yaml.j2
bfournie: Fixed

zaneb:
only the must-gather failed

PR #1579, opened 2023-09-06
Rename ipxeBaseURL to bootArtifactsBaseURL

pawanpinjarkar:
https://github.com/openshift/installer/pull/7450 renamed `ipxeBaseURL` to more generic `bootArtifactsBaseURL`. 
This PR fixes it in dev-scripts.


zaneb:
There's no way to do this without temporarily breaking the job, but at least it is optional.
/lgtm
/approve

PR #1581, opened 2023-09-11
Allow pushing custom image names when preparing a release

andfasano:
In some cases it may be useful to provide a custom image name - rather than using the folder name as a default - when preparing a local release.
This is particularly useful for the agent-based installer, since the same repo may be used to produce (more than one) different images shipped in the release payload (having their own name).

Also, enabling automatically local registry when using custom images also for the agent workflow

### DIFF
--- a/agent/04_agent_prepare_release.sh
+++ b/agent/04_agent_prepare_release.sh
@@ -20,9 +20,14 @@ source $SCRIPTDIR/oc_mirror.sh
 # To specify a custom Dockerfile set <ENTRYNAME>_DOCKERFILE, as a relative path of the Dockerfile
 # within the configured repo
 #
-# For example, to use a custom installer image:
+# To specify a custom image name, set <ENTRYNAME>_IMAGE with the required image name
+#
+# For example, to use a custom installer and assisted-service image:
 # export INSTALLER_LOCAL_REPO=~/go/src/github.com/openshift/installer
 # export INSTALLER_DOCKERFILE=images/installer/Dockerfile.ci
+# export ASSISTED_SERVICE_LOCAL_REPO=~/git/assisted-service
+# export ASSISTED_SERVICE_DOCKERFILE=Dockerfile.assisted-service.ocp
+# export ASSISTED_SERVICE_IMAGE=agent-installer-api-server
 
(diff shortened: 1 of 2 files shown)


agent/04_agent_prepare_release.sh
```diff
@@ -20,9 +20,14 @@ source $SCRIPTDIR/oc_mirror.sh
 # To specify a custom Dockerfile set <ENTRYNAME>_DOCKERFILE, as a relative path of the Dockerfile
 # within the configured repo
 #
-# For example, to use a custom installer image:
+# To specify a custom image name, set <ENTRYNAME>_IMAGE with the required image name
+#
+# For example, to use a custom installer and assisted-service image:
 # export INSTALLER_LOCAL_REPO=~/go/src/github.com/openshift/installer
 # export INSTALLER_DOCKERFILE=images/installer/Dockerfile.ci
+# export ASSISTED_SERVICE_REPO=~/git/assisted-service
```
bfournie, line 28: Should this be ?
export ASSISTED_SERVICE**_LOCAL**_REPO=~/git/assisted-service

andfasano: I don't think keeping a `*` in the name portion will work, `REPO_OVERRIDES` looks for names ending with just `_LOCAL_REPO`

bfournie: I think those ** got put in because I tried to make it Bold but I was really asking if 'Local' needs to be added  in 'ASSISTED_SERVICE_REPO'

andfasano: Good catch thanks those asteriks catpured completely my focus :)

andfasano:
> > /lgtm although it would be very helpful to add instructions to agent/README.md about how to use this and also include the comments in https://github.com/openshift-metal3/dev-scripts/blob/master/agent/04_agent_prepare_release.sh#L16-L21 in the README
> 
> Meanwhile I've slightly enhanced the 04 step readme, I'll plan to followup more detailed instructions for managing more than one repo changes in a subsequent PR

FYI https://github.com/openshift-metal3/dev-scripts/pull/1590

PR #1582, opened 2023-09-13
OCPBUGS-18956: Provide support for external platform for SNO

bfournie:
Currently setting the scenario to SNO uses platform 'none' but 'external' platform can also be used with SNO so adding that support. Also added 'external' platform in docs and comments.


agent/docs/none-platform.md
```diff
@@ -1,21 +1,25 @@
-# None Platform Support
+# None and External Platform Support
```
rwsu, line 1: I would suggest renaming the file to none-and-external-plaform.md.

rwsu (changes requested):

agent/docs/none-platform.md
bfournie: Done

common.sh
```diff
@@ -428,7 +428,9 @@ if [[ ! -z ${AGENT_E2E_TEST_SCENARIO} ]]; then
           export MASTER_MEMORY=32768
           export NUM_WORKERS=0
           export NETWORK_TYPE="OVNKubernetes"
-          export AGENT_PLATFORM_TYPE="none"
+          if [[ "${AGENT_PLATFORM_TYPE:-baremetal}" != "external" ]]; then
+            export AGENT_PLATFORM_TYPE="none"
+          fi
```
andfasano, line 433: If the two accepted values for `AGENT_PLATFORM_TYPE` are just `none` and `external`, then I'd say that a validation rule could be more clear, informing also the user of the invalid value (keeping a default value of `none` if not set, as before)

common.sh
```diff
@@ -428,7 +428,14 @@ if [[ ! -z ${AGENT_E2E_TEST_SCENARIO} ]]; then
           export MASTER_MEMORY=32768
           export NUM_WORKERS=0
           export NETWORK_TYPE="OVNKubernetes"
-          export AGENT_PLATFORM_TYPE="none"
+          if [[ "${AGENT_PLATFORM_TYPE:-baremetal}" != "baremetal" ]]; then
```
andfasano, line 431: What about just defaulting `AGENT_PLATFORM_TYPE` to `none` and applying the validation after, ie: 
```
export AGENT_PLATFORM_TYPE="${AGENT_PLATFORM_TYPE:-"none"}
if [[ "${AGENT_PLATFORM_TYPE}" != "external" ]]  && [[ "${AGENT_PLATFORM_TYPE}" != "none" ]]; then
...
```

bfournie: Updated

agent/docs/none-and-external-platform.md
```diff
@@ -1,21 +1,25 @@
-# None Platform Support
+# None and External Platform Support
 
 ## Configuration Variable
 
 To enable None platform for compact and HA clusters:
 ````
 export AGENT_PLATFORM_TYPE=none
 ````
+To enable External platform for compact and HA clusters:
```
pawanpinjarkar, line 9: For none platform, the default was SNO so saying `To enable None platform for compact and HA clusters:` is right for none but external always support all the 3 cluster topologies so please change `To enable External platform for compact and HA clusters:` to `To enable External platform for SNO, compact and HA clusters:`

pawanpinjarkar (commented):
minor suggestions.

PR #1584, opened 2023-09-18
Add extra network/nic to sample config

martinkennelly:
@cybertron


cybertron:
/ok-to-test
/lgtm

Thanks!

cybertron:
/approve

This is basically a doc change so I don't think we need to wait for a second approver. If someone objects to the wording or something they can propose a followup PR. :-)

PR #1589, opened 2023-09-27
AGENT-714: Agent support for using baremetal hosts from install-confi…

bfournie:
…g.yaml

Add a new env variable to control where baremetal hosts are set - either in install-config.yaml or agent-config.yaml. This can be used to test https://issues.redhat.com/browse/AGENT-552 for BM IPI compatibility.


pawanpinjarkar (commented):
/lgtm

agent/common.sh
```diff
@@ -10,6 +10,8 @@ export AGENT_USE_APPLIANCE_MODEL=${AGENT_USE_APPLIANCE_MODEL:-"false"}
 export AGENT_APPLIANCE_HOTPLUG=${AGENT_APPLIANCE_HOTPLUG:-"false"}
 export AGENT_PLATFORM_TYPE=${AGENT_PLATFORM_TYPE:-"baremetal"}
 
+export AGENT_INSTALL_CONFIG_HOSTS=${AGENT_INSTALL_CONFIG_HOSTS:-"false"}
```
andfasano, line 13: A couple of comments:
* A little bit of name bikesheeding, but I'd prefer something like `AGENT_FORCE_INSTALL_CONFIG` (I don't have a better alternative!) to indicate that we'll try to generate a config into install-config.yaml (a general name may also be valid for other fields?)
* If true, shouldn't make sense only when AGENT_PLATFORM_TYPE == baremetal?

bfournie: Sure, I'm fine with changing the name although we need to still indicate HOSTS in that name since this is really about adding additional host info to install-config, we can still use install-config.yaml without this  so its not like we're using it to select install-config.yaml.
How about AGENT_BM_HOSTS_IN_INSTALL_CONFIG? That would also identify that these are baremetal hosts which, as you rightly pointed out, we'll need an additional check for.

agent/roles/manifests/templates/install-config_yaml.j2
```diff
@@ -66,6 +69,34 @@ platform:
 {% for ingress_vip in i_vips %}
       - {{ ingress_vip }}
 {% endfor %}
+{% if agent_install_config_hosts == "true" %}
```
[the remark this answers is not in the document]
bfournie: Yeah this can be cleaned up. I think for now the best will be to add an additional file - install-config-bm-hosts.yaml.j2, and then choose in in the ansible tasks to use that file based on the env setting, similar to what we do for agent-config bonds https://github.com/openshift-metal3/dev-scripts/blob/master/agent/roles/manifests/tasks/install-config.yml#L21. 
This will simplify the base template a lot. We can eventually simplify it even more by splitting out `none` platform as you indicated.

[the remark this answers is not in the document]
bfournie: Updated to use the platform-type to choose install-config.yaml file to use.  We can eventually use more macros to split out the common sections, will save that for a future PR.

agent/roles/manifests/tasks/install-config.yml
```diff
@@ -7,6 +7,13 @@
   template:
     src: "templates/install-config_yaml.j2"
     dest: "{{ install_path }}/install-config.yaml"
+  when: platform_type != "baremetal"
+
+- name: write the install-config.yaml
+  template:
+    src: "templates/install-config_hosts_yaml.j2"
```
andfasano, line 14: I think the name of the template should contains the platform type, ie `install-config_baremetal_yaml.j2`

bfournie: Good point, renamed.

pawanpinjarkar (commented):
/lgtm

PR #1591, opened 2023-10-06
Skip host FIPS validation in installer by default

zaneb:
Since CI hosts do not run with FIPS mode enabled, disable checking for it in openshift-install by default. This setting can be overridden in the config by setting FIPS_VALIDATE=true.

### DIFF
--- a/common.sh
+++ b/common.sh
@@ -126,6 +126,12 @@ if [[ "$OPENSHIFT_RELEASE_TYPE" == "ga" ]]; then
     export OPENSHIFT_RELEASE_STREAM=${OPENSHIFT_VERSION%.*}
 fi
 
+if [ "${FIPS_MODE:-false}" = "true" ]; then
+    if ! [ "${FIPS_VALIDATE:-false}" = "true" ]; then
+        export OPENSHIFT_INSTALL_SKIP_HOSTCRYPT_VALIDATION="true"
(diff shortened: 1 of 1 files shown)


derekhiggins:
sounds good to me, looks like it needs a small update
> common.sh: line 129: FIPS_MODE: unbound variable

zaneb:
fixed, thanks

common.sh
```diff
@@ -126,6 +126,12 @@ if [[ "$OPENSHIFT_RELEASE_TYPE" == "ga" ]]; then
     export OPENSHIFT_RELEASE_STREAM=${OPENSHIFT_VERSION%.*}
 fi
 
+if [ "${FIPS_MODE:-false}" = "true" ]; then
+    if ! [ "${FIPS_VALIDATE:-false}" = "true" ]; then
+        export OPENSHIFT_INSTALL_SKIP_HOSTCRYPT_VALIDATION="true"
```
andfasano, line 131: Where's getting consumed this var? Wasn't able to find any reference in openshift install

derekhiggins: It doesn't exist yet, mention in https://issues.redhat.com/browse/OCPBUGS-15845

andfasano: Thanks, I've missed that.

/approve

PR #1592, opened 2023-10-09
Mirror manifest lists and use authfile explicitly

mkowalski:
We have observed that in specific environments (e.g. CI with IPv6-only)
there are missing default credentials for the local registry hosted at
`virthost.ostest.test.metalkube.org:5000` what causes errors like

```
Trying to pull [...]
Error: initializing source docker://virthost.ostest.test [...]
[...] /localimages/local-release-image: authentication required
```

In order to fix that, we are trying to explicitly pull the image with
`--authfile` so that the subsequent `podman run` has the image already
available.

---

We have also noticed that when release image constains manifest list but
the list contains only a single image, then `oc adm release mirror`
unpacks the list causing digests of the images to change. In order to
prevent that, we are explicitly requesting that manifest lists are kept
without unpacking.

This should solve the problem of inconsistency between results returned
by `oc adm release info` and `oc adm release mirror`.

### DIFF
--- a/04_setup_ironic.sh
+++ b/04_setup_ironic.sh
@@ -204,6 +204,7 @@ if [ ! -f "${CACHED_MACHINE_OS_BOOTSTRAP_IMAGE}" ]; then
 fi
 
 # cached images to the bootstrap VM
+sudo -E podman pull --authfile "${PULL_SECRET_FILE}" "${IRONIC_IMAGE}" || echo "WARNING: Could not pull latest $IRONIC_IMAGE; will try to use cached images instead"
 sudo podman run -d --net host --privileged --name httpd-${PROVISIONING_NETWORK_NAME} --pod ironic-pod \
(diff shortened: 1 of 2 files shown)


04_setup_ironic.sh
```diff
@@ -204,6 +204,7 @@ if [ ! -f "${CACHED_MACHINE_OS_BOOTSTRAP_IMAGE}" ]; then
 fi
 
 # cached images to the bootstrap VM
+sudo -E podman pull --authfile $PULL_SECRET_FILE ${IRONIC_IMAGE} || echo "WARNING: Could not pull latest $IRONIC_IMAGE; will try to use cached images instead"
```
dtantsur, line 207: nit: use consistent variable syntax (either $VAR or ${VAR}). Ideally also with "" around pull secret (may have spaces).

mkowalski: Of course, just fixed it

PR #1595, opened 2023-10-18
Try to work around libvirt instability

cjeanner:
On CS9 and RHEL9+ hypervisor, libvirt isn't using a main daemon with all the sockets - it's using a modular daemon approach, and systemd socket activation.
This leads to issues where clients can't connect to libvirt services, because the exposed systemd socket isn't good for any reason.

Ensuring virtproxyd.socket is running at that point should help a bit, though it may be better to try to switch back to the "old" libvirt way or, at least, drop the socket activation nightmare.

### DIFF
--- a/ocp_install_env.sh
+++ b/ocp_install_env.sh
@@ -288,7 +288,8 @@ function generate_ocp_install_config() {
         exit 1
       fi
     fi
-
+    # Try to fix libvirt stability before running bootstrap VM
+    sudo systemctl restart virtproxyd.socket
     cat > "${outdir}/install-config.yaml" << EOF
 apiVersion: v1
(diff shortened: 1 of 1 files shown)


derekhiggins:
/ok-to-test
looks like it could be a packaging problem, restarting libvirtd (which happens earlier in dev-scripts) results in the socket associated with virtproxyd.socket disappearing (at least in some circumstances), the dependency between the two mustn't be correct. could be worth opening a libvirtd bug in RHEL

cjeanner:
Closing this - there's a far better way to ensure things are stable, as exposed in #1603 
Note that I opened an equivalent patch against metal3-env-dev under https://github.com/metal3-io/metal3-dev-env/pull/1313 that should also help.

PR #1598, opened 2023-10-26
OCPBUGS-22417: Include extra manifests when building configImage

bfournie:
When building the configImage, the extra manifests were not being added to the directory where it was built.

### DIFF
--- a/agent/06_agent_create_cluster.sh
+++ b/agent/06_agent_create_cluster.sh
@@ -68,6 +68,11 @@ function create_factory_image() {
 
 function create_config_image() {
 
+    # Copy any extra manifests
+    if [ -d "${asset_dir}/openshift" ]; then
+        cp -r "${asset_dir}/openshift" "${config_image_dir}"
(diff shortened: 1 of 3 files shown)


agent/06_agent_create_cluster.sh
```diff
@@ -68,6 +68,11 @@ function create_factory_image() {
 
 function create_config_image() {
 
+    # Copy any extra manifests
+    if [ -d "${asset_dir}/openshift" ]; then
```
andfasano, line 72: It's not likely something subjected to change frequently, but just for the sake of clarity - and to avoid issues if amended in future, I'd suggest to export `$EXTRA_MANIFESTS_PATH` in `common.sh` and use it here.

bfournie: Updated to use common EXTRA_MANIFESTS_PATH

pawanpinjarkar (commented):
/lgtm

PR #1600, opened 2023-11-07
Avoid git fetch error when using PR number

elfosardo:
This patch fixes a git error when running step 4 multiple times and building the ironic-image with a specific PR number.

### DIFF
--- a/04_setup_ironic.sh
+++ b/04_setup_ironic.sh
@@ -68,8 +68,10 @@ for IMAGE_VAR in $(env | grep "_LOCAL_IMAGE=" | grep -o "^[^=]*") ; do
         IMAGE_PR_VAR=${IMAGE_VAR/_LOCAL_IMAGE}_PR
         IMAGE_PR=${!IMAGE_PR_VAR:-}
         if [[ -n ${IMAGE_PR:-} ]]; then
+	    if [[ $(git rev-parse --abbrev-ref HEAD) != pr${IMAGE_PR} ]]; then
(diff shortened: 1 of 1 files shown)


iurygregory (approved):

04_setup_ironic.sh
```diff
@@ -68,8 +68,10 @@ for IMAGE_VAR in $(env | grep "_LOCAL_IMAGE=" | grep -o "^[^=]*") ; do
         IMAGE_PR_VAR=${IMAGE_VAR/_LOCAL_IMAGE}_PR
         IMAGE_PR=${!IMAGE_PR_VAR:-}
         if [[ -n ${IMAGE_PR:-} ]]; then
+            if git rev-parse --abbrev-ref HEAD != pr${IMAGE_PR}; then
```
dtantsur, line 71: Is it some git-specific syntax? Otherwise you need [[ ]]

dtantsur: It does not seem to be a special syntax. You've added a condition that is always `false` because git returns an error:
```
$ git rev-parse --abbrev-ref HEAD != pr
master
!=
fatal: ambiguous argument '!=': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
```

elfosardo: thanks
I'll add the [[ ]]

PR #1601, opened 2023-11-08
AGENT-742: Add baremetal host BMC configuration to install-config

bfournie:
For the agent installer, add fields to install-config.yaml for the BMC and provisioning network. This allows baremetal hosts to be set up at install time.


rwsu:
It wasn't obvious to me from looking at this PR alone. To test this, you'll need to set `export AGENT_BM_HOSTS_IN_INSTALL_CONFIG=true `.

rwsu (commented):
/lgtm

agent/roles/manifests/templates/install-config_baremetal_yaml.j2
```diff
@@ -65,10 +65,22 @@ platform:
...
+    provisioningNetwork: {{ provisioning_network_profile }}
     hosts:
 {% for hostname in hostnames %}
     - name: {{ hostname }}
       bootMACAddress: {{ macs[loop.index0] }}
+      bmc:
+        address: {{ bmc_addresses[loop.index0] }}
+        username: {{ bmc_usernames[loop.index0] }}
+        password: {{ bmc_passwords[loop.index0] }}
+        disableCertificateVerification: false
```
ahussey-redhat, line 83: Please feel free to ignore this comment.
I noticed that `disableCertificateVerification` is hard coded here, but is configurable in the `assisted-service` https://github.com/openshift/assisted-service/blob/25d0ae75d6a4a268f7555339a74cc6b050ba23aa/internal/installcfg/installcfg.go#L21

Should it also be configurable here?

bfournie: Thanks @ahussey-redhat. So this setting doesn't affect the installation or the resulting baremetalhosts, and making it configurable would just require a new environment variable which wouldn't be used in any CI tests so I don't think we need it to be configurable.

ahussey-redhat:
That makes sense :)

pawanpinjarkar (commented):
/lgtm

agent/05_agent_configure.sh
```diff
@@ -245,6 +245,14 @@ function generate_cluster_manifests() {
...
   fi
 
+  # Set BMC info
+  nodes_bmc_usernames=$(printf '%s,' "${AGENT_NODES_BMC_USERNAMES[@]}")
+  export AGENT_NODES_BMC_USERNAMES_STR=${nodes_bmc_usernames::-1}
+  nodes_bmc_passwords=$(printf '%s,' "${AGENT_NODES_BMC_PASSWORDS[@]}")
+  export AGENT_NODES_BMC_PASSWORDS_STR=${nodes_bmc_passwords::-1}
+  nodes_bmc_addresses=$(printf '%s,' "${AGENT_NODES_BMC_ADDRESSES[@]}")
+  export AGENT_NODES_BMC_ADDRESSES_STR=${nodes_bmc_addresses::-1}
+
```
andfasano, line 255: I saw these values get logged. Not a big issue, but unless strictly required I'd wrap them in a `set +x` just to be on the safe side

bfournie: Changed to wrap with `set +x`

pawanpinjarkar (commented):
/lgtm

PR #1603, opened 2023-11-13
Properly manage libvirt services

cjeanner:
With newer CS9 and RHEL-9, libvirtd has moved to a modular, socket activated layout.

This new layout conflicts with the single, monolithic "libvirtd.service" that was restarted until now, creating issues with other projects that may wrap this metal3-dev-env.

There is a related PR against metal3-env-dev[1].

[1] https://github.com/metal3-io/metal3-dev-env/pull/1313

### DIFF
--- a/02_configure_host.sh
+++ b/02_configure_host.sh
@@ -10,6 +10,29 @@ source oc_mirror.sh
 
 early_deploy_validation
 
+#
+# Manage libvirtd services based on OS
+# Get a dedicated method, making it easier to duplicate
+# or move in order to expose it to some other parts if needed.
+# This is the same method as defined in metal3-dev-env via that
+# commit:
+# https://github.com/metal3-io/metal3-dev-env/pull/1313/commits/a6a79685986f9d7cb18c4eb680ee4d2a759e99dc
+#
+manage_libvirtd() {
+  case ${DISTRO} in
+      centos9|rhel9)
+          for i in qemu network nodedev nwfilter secret storage interface; do
+              sudo systemctl enable --now virt${i}d.socket
+              sudo systemctl enable --now virt${i}d-ro.socket
+              sudo systemctl enable --now virt${i}d-admin.socket
+          done
+          ;;
+      *)
+          sudo systemctl restart libvirtd.service
+        ;;
+esac
+}
+
 # Generate user ssh key
 if [ ! -f $HOME/.ssh/id_rsa.pub ]; then
     ssh-keygen -f ~/.ssh/id_rsa -P ""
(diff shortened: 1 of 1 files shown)


cjeanner:
Hmm I think there's an issue somewhere else in the tests?
`curl: (7) Failed to connect to ofcir-service.ofcir-system.svc.cluster.local port 443: No route to host`
not sure it has anything to do with my patch, does it?

The other tests also failed while trying to access cluster.local resources.

common.sh
```diff
@@ -509,3 +509,21 @@ if [[ -n "$MIRROR_IMAGES" || -z "${IP_STACK:-}" || "$IP_STACK" = "v6" || "$IP_ST
 fi
 
 export AGENT_TEST_CASES=${AGENT_TEST_CASES:-}
+
+#
+# Manage libvirtd services based on OS
+#
+manage_libvirtd() {
```
andfasano, line 516: Unless not needed elsewhere in future, it seems to me that the function defintion could be better placed in step 02, since it's the one that mostly takes care of the networking setup, rather than in `common.sh`, which is usually imported by all the other steps.

andfasano:
> Hmm I think there's an issue somewhere else in the tests? `curl: (7) Failed to connect to ofcir-service.ofcir-system.svc.cluster.local port 443: No route to host` not sure it has anything to do with my patch, does it?
> 
> The other tests also failed while trying to access cluster.local resources.

Yes, that's not directly related to the patch, but it's a CI related problem

common.sh
cjeanner: In fact, I'm wondering if we couldn't import metal3-dev-env `lib/common.sh` ? It seems most of the things are duplicated there, with some additions.
I hesitated putting this method in the 02, but since it's pretty generic, thought it would be best in common.sh to get it properly exposed in case of future iteration/additions.. I can of course move it.

cjeanner:
@andfasano not really sure about the cause of the failure for the e2e-agent-ha-dualstack :/. Prow isn't that nice in showing logs..

andfasano:
> Ah, seems to be unrelated to the change:
> 
> ```
> level=debug msg=OpenShift console route is admitted
> level=debug msg=Still waiting for the cluster to initialize: Cluster operators authentication, console, ingress, kube-    apiserver, monitoring are not available
> level=debug msg=Still waiting for the cluster to initialize: Cluster operators authentication, console, ingress, monitoring are not available
> level=debug msg=Still waiting for the cluster to in    itialize: Cluster operators authentication, console, monitoring are not available
> level=debug msg=Still waiting for the cluster to initialize: Cluster operators authentication, console are not available
> level=debug msg=Still waiting     for the cluster to initialize: Cluster operator authentication is not available
> level=debug msg=Still waiting for the cluster to initialize: Cluster operator etcd is degraded
> level=debug msg=Still waiting for the cluster to initializ    e: Cluster operator etcd is degraded
> bash: line 190: 17345 Killed                  timeout -s 9 105m make agent
> ```
> 
> apparently the cluster wasn't ready in time?

Right, the internal timeout interrupted the job, which took longer due the `Cluster operator etcd is degraded` problem

common.sh
cjeanner: 'k - I'll push an update tomorrow then. Defining a method in the 02 is still OK, so that it may be easier to move it around?

cjeanner:
/hole
will move the method definition accordingly

02_configure_host.sh
```diff
@@ -141,14 +164,14 @@ fi
 ZONE="\nZONE=libvirt"
 
 # Allow local non-root-user access to libvirt
-# Restart libvirtd service to get the new group membership loaded
 if ! id $USER | grep -q libvirt; then
   sudo usermod -a -G "libvirt" $USER
+  gpasswd -a "${USER}" libvirt
```
rabi, line 169: I think we need escalated privileges for this command as well?

PR #1604, opened 2023-11-17
AGENT-337: When using ZTP manifests, support both VIP and VIPs

bfournie:
In agent_cluster_install.yaml, include VIPS in addition to VIP to be compatible with the assisted-service change.
https://github.com/openshift/assisted-service/pull/5501

### DIFF
--- a/agent/roles/manifests/templates/agent-cluster-install_yaml.j2
+++ b/agent/roles/manifests/templates/agent-cluster-install_yaml.j2
@@ -6,7 +6,17 @@ metadata:
 spec:
 {% if (platform_type != "none") and (platform_type != "external") %}
   apiVIP: {{ api_vip }} 
+  apiVIPs:
+{% set a_vips = api_vips.split(',') %}
+{% for api_vip in a_vips %}
+    - {{ api_vip }}
+{% endfor %}
   ingressVIP: {{ ingress_vip }} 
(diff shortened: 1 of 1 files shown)


nmagnezi (approved):
/lgtm

filanov (approved):

pawanpinjarkar (approved):

filanov:
@zaneb can you please approve?

filanov:
in the latest version we removed api vip and ingress vip and replaced with api vips and ingress vips
the api was marked as deprecated for a year and now we removed it

bfournie:
I don't believe this fixed a blocking issue, Nir and I had verified that the singular work still worked with agent-cluster-install.yaml. It just is consistent now with the API. Also multiple VIPs are now supported in agent dev-scripts ZTP testing, this was an earlier oversight.

zaneb:
> in the latest version we removed api vip and ingress vip and replaced with api vips and ingress vips the api was marked as deprecated for a year and now we removed it

@filanov we don't remove fields from a kube API, you can only create a new version of the API without that field. But I believe that Nir and Bob have it under control.

filanov:
Yes we are adding a new api version, v1beta2

filanov:
@carbonin is taking over that work from Nir

PR #1605, opened 2023-11-21
Remove gpasswd command call

cjeanner:
gpasswd does the same as the usermod call we have just before. It was
mostly a mix with the newgrp command, but it doesn't help either.

Since there's no way to reload the groups system wide without a logout
and login, we can't really do anything...

### DIFF
--- a/02_configure_host.sh
+++ b/02_configure_host.sh
@@ -166,7 +166,6 @@ ZONE="\nZONE=libvirt"
 # Allow local non-root-user access to libvirt
 if ! id $USER | grep -q libvirt; then
   sudo usermod -a -G "libvirt" $USER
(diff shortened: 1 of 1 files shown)


dtantsur:
/approve
/hold

https://github.com/metal3-io/metal3-dev-env/pull/1317/ must merge first

cjeanner:
@dtantsur the related patch has been merged in the other project. The discussion I had there lead to "just remove the thing supposed to (not) reload groups and be off with that". Lemme know if this is also OK on this side.

TLDR; no way to reload groups, logout/login is needed, but due to CI and automation we may not be able to enforce this.

elfosardo:
this should be ok, and we were missing sudo anyway
let's see what the CI says

PR #1607, opened 2023-11-24
Enable libvirt proxy local socket

psathyan:
We have observed the below error in some of our deployments.

```
2023-11-23 15:20:02 level=error msg=Error: failed to dial libvirt: dial unix /var/run/libvirt/libvirt-sock: connect: no such file or directory
```

Using this patch, we are able to proceed when the above error is observed. Newer deployments with this patch have been successful.

### DIFF
--- a/02_configure_host.sh
+++ b/02_configure_host.sh
@@ -21,7 +21,7 @@ early_deploy_validation
 manage_libvirtd() {
   case ${DISTRO} in
       centos9|rhel9|rocky9)
-          for i in qemu network nodedev nwfilter secret storage interface; do
+          for i in qemu interface network nodedev nwfilter secret storage proxy; do
               sudo systemctl enable --now virt${i}d.socket
               sudo systemctl enable --now virt${i}d-ro.socket
               sudo systemctl enable --now virt${i}d-admin.socket


02_configure_host.sh
```diff
@@ -21,11 +21,13 @@ early_deploy_validation
 manage_libvirtd() {
   case ${DISTRO} in
       centos9|rhel9)
-          for i in qemu network nodedev nwfilter secret storage interface; do
+          for i in qemu interface network nodedev nwfilter secret storage; do
```
dtantsur, line 24: Just add proxy to this list?

02_configure_host.sh
```diff
@@ -21,11 +21,13 @@ early_deploy_validation
...
               sudo systemctl enable --now virt${i}d.socket
               sudo systemctl enable --now virt${i}d-ro.socket
               sudo systemctl enable --now virt${i}d-admin.socket
           done
+          sudo systemctl enable virtproxyd.service
```
dtantsur, line 29: I'm not familiar with this component, but normally you enable either the service or the activation socket, not both.

02_configure_host.sh
psathyan: Yes, we could do that. Will update in the next patch.

02_configure_host.sh
psathyan: True, it is either the socket or service. Also, we are enabling the daemon but not starting it.  I can remove this step.

bshephar (approved):
Confirming that this also works for me.

mkowalski (approved):
Tested and this fixes dev-scripts on a fresh Stream 9 installations

bshephar:
It seems I have no power here either.

PR #1609, opened 2023-11-29
Add build arg to custom release

andfasano:
This patch adds the possibility to specify an additional build arg when building a local repo. In addition, it ensures that the installer will be extracted from the mirror, so that the subsequent steps will use the updated one -  in case it was just rebuilt (this feature will avoid specifying `export KNI_INSTALL_FROM_GIT=true` when using a custom release)

### DIFF
--- a/agent/docs/custom-release.md
+++ b/agent/docs/custom-release.md
@@ -93,6 +93,18 @@ The previous example will build a new image using the `Dockerfile.assisted-servi
 `virthost.ostest.test.metalkube.org:5000/localimages/assisted-service:latest` and it will be used as `agent-installer-api-server` within the newly
 built release image.
 
+## Specify an image build arg
+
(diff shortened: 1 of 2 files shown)


agent/docs/custom-release.md
```diff
@@ -93,6 +93,18 @@ The previous example will build a new image using the `Dockerfile.assisted-servi
 `virthost.ostest.test.metalkube.org:5000/localimages/assisted-service:latest` and it will be used as `agent-installer-api-server` within the newly
 built release image.
 
+## Specify an image build arg
+
+It may also be useful to configure an additional build arg when building an image. The `<ENTRYNAME>_BUILD_ARG` could be 
+usef for that:
```
bfournie, line 99: Nit - typo "usef"

pawanpinjarkar (commented):
/lgtm

PR #1399, opened 2022-05-18
NO-ISSUE: Pin community.general collection to version 4.8.1

adriengentil:
The provisioning of equinix machines is broken because it fails with:
```
fatal: [localhost]: FAILED! => {"changed": false, "msg": "Unsupported parameters for (packet_device) module: tags Supported parameters include: always_pxe, auth_token, count,
 count_offset, device_ids, facility, features, hostnames, ipxe_script_url, locked, operating_system, plan, project_id, state, user_data, wait_for_public_IPv, wait_timeout"}
```

This issue seems to be related to the latest version (`5.0.0`) of the `community.general` ansible collection which drops the support for ansible 2.9 that we install (https://github.com/ansible-collections/community.general/blob/stable-5/CHANGELOG.rst#major-changes)

This change pins the `community.general` ansible collection to the previous version `4.8.1` which seems to work well.

PR #1619, opened 2023-12-15
MGMT-16396: Install ansible.posix collection

adriengentil:
Install posix collection in order to use synchronize module. It was a
built-in module before.

Remove the constraint on general collection as it was pinned because
rhel 8 used a old version of ansible, which is not the case anymore with
rhel 9 (https://github.com/openshift-metal3/dev-scripts/pull/1399).


adriengentil:
/retest
looks like flakes

Issue #1620, opened 2023-12-18
Wrong command in ./01_install_requirements.sh script


mnietoji:
+(./01_install_requirements.sh:102): yq -iy '.[3].dnf.nobest = "true"' /home/dev-scripts/metal3-dev-env/vm-setup/roles/packages_installation/tasks/centos_required_packages.yml
Error: unknown shorthand flag: 'y' in -y
Usage:
  yq eval [expression] [yaml_file1]... [flags]

This line is wrong
yq -iy '.[3].dnf.nobest = "true"' ${METAL3_DEV_ENV_PATH}/vm-setup/roles/packages_installation/tasks/centos_required_packages.yml

mnietoji:
yq --version
yq (https://github.com/mikefarah/yq/) version v4.40.5

mnietoji:
I have seen that yq -iy works with yq 3.2.3 but fails with yq 4.40.5, but i think v4 is being installed here:
dev-scripts/metallb/deploy_operator.sh:go install -mod='' [github.com/mikefarah/yq/v4@v4.13.3](http://github.com/mikefarah/yq/v4@v4.13.3)

bdlink:
yq v3 has not been supported since 2021, need to check all places yq is used and make sure they work for yq v4.18.1 and later versions.

bdlink:
Issue here seems to be that yq 3.4.3 (which is installed by the script) is not installed if another version is already present. removing the offending option -y does not work, because later uses of yq will fail (as incompatable with yq v4).

elfosardo:
this should help https://github.com/openshift-metal3/dev-scripts/pull/1668
when 4.x is available on pypi we can think about updating the command

PR #1621, opened 2023-12-19
Fix yq command to be compatible with yq v4

mnietoji:
Option -y is not available in yq version 4 that
is the one being installed
https://github.com/openshift-metal3/dev-scripts/issues/1620

### DIFF
--- a/01_install_requirements.sh
+++ b/01_install_requirements.sh
@@ -99,7 +99,7 @@ esac
 # during dnf upgrade
 sudo dnf -y install jq
 sudo python -m pip install yq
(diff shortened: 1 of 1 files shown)


eshulman2:
Hi @elfosardo @sadasu can you please review/comment on this patch.

elfosardo:
I beleive the latest available version in pypi is still 3.x 
I'd rather pin yq version installed so when the new version comes out we don't break dev-scripts

elfosardo:
/hold
this can be closed, yq version has been pinned

elfosardo:
/close
obsolete

Issue #1613, opened 2023-12-01
openshift-sdn support is about to be removed from the installer


danwinship:
openshift-sdn is now deprecated, and (most) users will no longer be able to install sdn-based clusters starting in 4.15. The installer will soon start rejecting install-configs with `networkType: OpenShiftSDN`. (https://github.com/openshift/installer/pull/7720). So dev-scripts needs to do one of:

1. Switch to using `NETWORK_TYPE=OVNKubernetes` regardless of `IP_STACK` (retroactively changing the meaning of some existing configs).
2. Switch to using `NETWORK_TYPE=OVNKubernetes` regardless of `IP_STACK` when installing 4.15 or later (causing some existing configs to silently change meaning when users move to 4.15).
3. Start requiring `NETWORK_TYPE` to always be explicitly specified when `IP_STACK=v4` (possibly just for 4.15+ installs) (breaking some existing configs but making it clear how to fix them, and not silently changing behavior on anyone).

I'm updating all of our `IP_STACK=v4` CI jobs (even pre-4.15 ones) to explicitly set `NETWORK_TYPE` (https://github.com/openshift/release/pull/46286, though that's incomplete), so once that's done, it would be possible for dev-scripts to change its defaults without affecting CI.

(For CI jobs that actually need to test openshift-sdn in 4.15+, we will be hacking around the installer change by [using manifests](https://github.com/openshift/installer/pull/7720#discussion_r1399735975), so you don't have to worry about that. dev-scripts itself should probably _not_ automatically hack around the installer change for users (and at any rate, the plan is for sdn to be removed from OCP entirely in 4.17, so you'll need to change something by then).

PR #1622, opened 2023-12-20
CORS-2969: Don't default to OpenShiftSDN after deprecation

zaneb:
OpenShiftSDN is being deprecated in 4.15, so switch the default to OVNKubernetes in the IPv4-only case (OVNk is already required for IPv6/dual-stack) to avoid installer failures.

This follows up on 8537fdd1e01fa07caa7f09ed03c6beafdacb9378, which sets OpenShiftSDN in the manifests (but not the install-config) iff it is *explicitly* requested.

### DIFF
--- a/network.sh
+++ b/network.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+source release_info.sh
+
 function nth_ip() {
   network=$1
   idx=$2
@@ -111,6 +113,11 @@ else
   exit 1
 fi
 
+function openshift_sdn_deprecated() {
+  # OpenShiftSDN is deprecated in 4.15 and later
+  printf '4.15\n%s\n' "$(openshift_version)" | sort -V -C
+}
+
 if [[ "$IP_STACK" = "v4" ]]
 then
   export CLUSTER_SUBNET_V4=${CLUSTER_SUBNET_V4:-"10.128.0.0/14"}
@@ -119,7 +126,11 @@ then
   export CLUSTER_HOST_PREFIX_V6=""
   export SERVICE_SUBNET_V4=${SERVICE_SUBNET_V4:-"172.30.0.0/16"}
   export SERVICE_SUBNET_V6=""
+if openshift_sdn_deprecated; then
+  export NETWORK_TYPE=${NETWORK_TYPE:-"OVNKubernetes"}
+else
   export NETWORK_TYPE=${NETWORK_TYPE:-"OpenShiftSDN"}
+fi
 elif [[ "$IP_STACK" = "v6" ]]; then
   export CLUSTER_SUBNET_V4=""
   export CLUSTER_SUBNET_V6=${CLUSTER_SUBNET_V6:-"fd01::/48"}
--- a/ocp_install_env.sh
+++ b/ocp_install_env.sh
@@ -1,3 +1,5 @@
+source release_info.sh
+
 eval "$(go env)"
 
 function get_arch() {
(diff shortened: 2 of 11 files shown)


danwinship:
"Fixes #1613"

But https://github.com/openshift/release/pull/46286 and https://github.com/openshift/release/pull/46338 should probably merge before this, to ensure that no CI jobs unexpectedly switch network types when the default changes.

ocp_install_env.sh
```diff
@@ -258,10 +258,15 @@ EOF
   fi
 }
 
+function openshift_sdn_deprecated() {
+  # OpenShiftSDN is deprecated in 4.15 and later
+  [[ "${NETWORK_TYPE}" = "OpenShiftSDN" ]] && printf '4.15\n%s\n' "$(openshift_version)" | sort -V -C
+}
```
danwinship, line 264: `openshift_sdn_deprecated` makes it sound like this just checks if openshift-sdn is deprecated, but it actually checks if sdn is deprecated _and_ it's the selected `NETWORK_TYPE`.

danwinship (commented):
This makes the control flow really weird... it still defaults to `NETWORK_TYPE=OpenShiftSDN` for IPv4 in `network.sh`, but then just overrides it later...

zaneb:
@danwinship Fixed the control flow, hopefully this actually works for the agent as well.

> But https://github.com/openshift/release/pull/46286 and https://github.com/openshift/release/pull/46338 should probably merge before this, to ensure that no CI jobs unexpectedly switch network types when the default changes.

Can't we merge this and abandon all of https://github.com/openshift/release/pull/46563, https://github.com/openshift/release/pull/46286, and https://github.com/openshift/release/pull/46338?

From my reading of the commit messages, there's a bunch of jobs that don't have -sdn- in the title where we don't actually care what network type they have. With this change they'll carry on as they always have on existing release branches, but switch to ovnk on new ones. It seems to me that that's what we want, and it avoids renaming everything in the world (multiple times).

network.sh
```diff
@@ -111,6 +111,11 @@ else
   exit 1
 fi
 
+function openshift_sdn_deprecated() {
+  # OpenShiftSDN is deprecated in 4.15 and later
+  printf '4.15\n%s\n' "$(openshift_version)" | sort -V -C
```
bfournie, line 116: Doesn't `function openshift_version()` need to be included in network.sh for this to work?

danwinship:
In _some_ cases it definitely makes sense to convert the SDN job to OVN-K (eg, the "virtualmedia" jobs, https://github.com/openshift/release/pull/46338).

But there are some places where there are two jobs that are completely identical except that one of them is sdn and one is ovn, so the right fix is to drop the sdn job. (I'm not sure if that case applies to any dev-scripts-based jobs though.)

The trickier case is where there are two jobs that are _nearly_ identical, like `e2e-metal-ipi-sdn` (-ipv4) vs `e2e-metal-ipi-ovn-ipv6`, where it's not clear if we still care about testing both IPv4 and IPv6 if we're not also testing both network plugins. And that seemed like a decision that was better to leave up to the individual repo owners to figure out.

> and it avoids renaming everything in the world (multiple times).

https://github.com/openshift/release/pull/46563 and https://github.com/openshift/release/pull/46286 don't rename any _jobs_; they just change job _configs_ to make it clearer when a job ends up using sdn.

(https://github.com/openshift/release/pull/46338 does rename jobs, but that's because it's actually flipping jobs from sdn to ovn.)

danwinship:
> And that seemed like a decision that was better to leave up to the individual repo owners to figure out.

(of course, you _are_ "the individual repo owners" for some of these jobs, so if you want to figure that now, go ahead...)

network.sh
bfournie: Maybe its related to the ordering of the sources in` agent/06_agent_create_cluster.sh`, I had to define `openshift_version()` in network.sh in order to successfully run `make agent`

andfasano:
/hold

Currently there are two separate issues (but somehow related) requiring to be addressed:

1. As [previously mentioned](https://github.com/openshift-metal3/dev-scripts/pull/1622#discussion_r1434140781) by @bfournie, the `openshift_version` function is not defined in `network.sh` and so it's not found
2. The current dev-scripts CI configuration was still pointing at [4.14](https://github.com/openshift/release/blob/273f642cef6ec841454c89a9f3ee37fd88fe1c65/ci-operator/config/openshift-metal3/dev-scripts/openshift-metal3-dev-scripts-master.yaml#L22), thus making pass the current builds. I'll open a new PR to fix this point

andfasano:
Requires https://github.com/openshift/release/pull/47487

network.sh
bfournie: Recommend moving `openshift_release_version()` to common.sh, that should take care of the ordering issue.

danwinship:
> But [openshift/release#46286](https://github.com/openshift/release/pull/46286) and [openshift/release#46338](https://github.com/openshift/release/pull/46338) should probably merge before this, to ensure that no CI jobs unexpectedly switch network types when the default changes.

done

> "Fixes #1613"

(as in, "you should add that to the initial comment so it closes that issue when this merges")

andfasano:
/hold cancel

Now the agent job is failing as expected

zaneb:
The dependencies here are much nastier than I realised, and I'm almost ready to give up on this change.
This whole pile of bash is completely unmaintainable :cry:

PR #1626, opened 2024-01-17
Fix delete PATH_CONF_DNSMASQ

rdiazcam:
When configuring dnsmasq sudo is needed to delete previous configuration files owned by root.


elfosardo:
/close
obsolete

PR #1627, opened 2024-01-19
Document METAL3_DEV_ENV variable

elfosardo:


elfosardo:
the issue with e2e-metal-ipi-serial-ipv4 is independent from this change

elfosardo:
/retest
giving it another chance

derekhiggins:
/hold
I already run install requirements multiple times, when I've made local changes to metal3-dev-env, I want to get this error message, it reminds me to set METAL3_DEV_ENV  (to avoid running sync_repo_and_patch)

If the error message doesn't happen my local updates would be silently zapped, without realising it I wouldn't be running the code I wanted and my changes would be lost.

elfosardo:
@derekhiggins @dtantsur please give it another look, thanks!

derekhiggins:
/unhold
/lgtm
thanks

PR #1628, opened 2024-01-23
network: fix generation of VIPs for single- and dual-stack setups

mkowalski:


cybertron:
/lgtm

Seems fine. I assume this is related to deploying single stack clusters on dual stack envs?

mkowalski:
/hold
Until manual testing proves that all works

Yes, the issue with current state is that we do not generate a valid install-config yaml for a single-stack cluster with dual-stack hosts. For such a scenario (allowed by syntax of config.sh) we put dual-stack VIPs in install-config.yaml but only single cluster network. This is illegal and does not pass the basic validations.

With this fix I want to be explicit as the VIP configuration goes. I am still not allowing to have single VIP setup with dual-stack cluster but extending feature set is not my goal, all I want first is to remove a scenario that prevents installation when in general it should work

cybertron:
/lgtm

Seems fine to me.

mkowalski:
/close

Old stuff, I don't even remember why I needed that

PR #1630, opened 2024-02-05
NO-ISSUE: In Centos9/RHEL9 update install location for network tools

paul-maidment:
The network-tools package is no longer available in RHEL9. This package supplies ifup/ifdown. ifup/ifdown are now found in the package NetworkManager-initscripts-updown.

This PR aims to address that by installing the correct package on RHEL9/Centos9

### DIFF
--- a/01_install_requirements.sh
+++ b/01_install_requirements.sh
@@ -80,6 +80,9 @@ case $DISTRO in
     sudo update-alternatives --install /usr/bin/pip3 pip3 /usr/bin/pip3.9 1
     ;;
   "centos9"|"rhel9")
+    # In RHEL9/Centos9 ifup/ifdown there is no package called network-scripts
+    # ifup/ifdown are found in NetworkManager-initscripts-updown instead.    
+    sudo dnf install NetworkManager-initscripts-updown
     sudo dnf -y install python3-pip
     if [[ $DISTRO == "centos9" ]]; then
       sudo dnf config-manager --set-enabled crb


01_install_requirements.sh
```diff
@@ -80,6 +80,9 @@ case $DISTRO in
     sudo update-alternatives --install /usr/bin/pip3 pip3 /usr/bin/pip3.9 1
     ;;
   "centos9"|"rhel9")
+    # In RHEL9/Centos9 ifup/ifdown there is no package called network-scripts
+    # ifup/ifdown are found in NetworkManager-initscripts-updown instead.    
+    sudo dnf install NetworkManager-initscripts-updown
```
elfosardo, line 85: is there any reason you want to be explicit with this here?
we're installing NetworkManager-initscripts-updown already from metal3-dev-env on centos/rhel 9
we install network-scripts in centos/rhel 8 only

paul-maidment: I recently ran the default installation of dev-scripts on a RHEL9 Beaker box and found this ifup/ifdown to be missing.

Actually it's the first time I have heard of https://github.com/metal3-io/metal3-dev-env so I don't typically use that for running dev-scripts and I imagine some users also don't use the dev-env (though I will be checking it out in future)

If you just run dev-scripts without the dev-env project then nothing is installed to replace the Centos8/RHEL8 dependency here. https://github.com/openshift-metal3/dev-scripts/pull/1630/files/50bf9ee158244b7fcd128194086b642c3735b43e#diff-9becdbb081652c615123932efab73b37017c3769239fcc74a194e9871cd8bc19R60

Soi I think this is a reasonable PR, especially since we already install something for RHEL8

PR #1642, opened 2024-03-11
Fix upgrade from CS8 compatibility

elfosardo:
systemctl requires sudo to run
also removing the || true to avoid covering hidden errors, the network service should always be present before the migration and if any ifcfg file is there


honza:
nit: more detail in the commit message would be nice

/lgtm

elfosardo:
> nit: more detail in the commit message would be nice
> 
> /lgtm

@honza thanks, you're right, I've updated the description

PR #1647, opened 2024-04-02
cleanup: use $OPENSHIFT_INSTALLER path

honza:
The ocp directory doesn't always have the binary present.  For example, the cleanup commands will fail when openshift-installer is built from git.  The var is populate in `common.sh` and the same pattern is used in `utils.sh` when creating the cluster.


elfosardo:
unrelated known issue

elfosardo:
unrelated issue

PR #1649, opened 2024-04-11
Add rocky9 support

derekhiggins:
CI job that ran on rocky9 with this branch
https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_release/50285/rehearse-50285-periodic-ci-openshift-release-master-nightly-4.15-e2e-metal-ipi-ovn-ipv6/1778423230030155776

### DIFF
--- a/02_configure_host.sh
+++ b/02_configure_host.sh
@@ -20,7 +20,7 @@ early_deploy_validation
 #
 manage_libvirtd() {
   case ${DISTRO} in
-      centos9|rhel9)
+      centos9|rhel9|rocky9)
           for i in qemu network nodedev nwfilter secret storage interface; do
               sudo systemctl enable --now virt${i}d.socket
               sudo systemctl enable --now virt${i}d-ro.socket
@@ -306,6 +306,8 @@ if [ "$MANAGE_BR_BRIDGE" == "y" ] ; then
     sudo virsh net-destroy ${BAREMETAL_NETWORK_NAME}
     sudo nmcli con del ${BAREMETAL_NETWORK_NAME}
     sudo virsh net-start ${BAREMETAL_NETWORK_NAME}
+    # Needed in IPv6 on some EL9 hosts for the bootstrap VM to get an IP
+    echo 0 | sudo dd of=/proc/sys/net/ipv6/conf/${BAREMETAL_NETWORK_NAME}/addr_gen_mode
     if [ "$INT_IF" ]; then #Need to bring UP the NIC after destroying the libvirt network
         sudo nmcli con up ${INT_IF}
     fi
--- a/01_install_requirements.sh
+++ b/01_install_requirements.sh
@@ -78,9 +78,9 @@ case $DISTRO in
     sudo alternatives --set python3 /usr/bin/python3.9
     sudo update-alternatives --install /usr/bin/pip3 pip3 /usr/bin/pip3.9 1
     ;;
-  "centos9"|"rhel9")
+  "centos9"|"rhel9"|"rocky9")
     sudo dnf -y install python3-pip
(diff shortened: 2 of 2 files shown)


02_configure_host.sh
```diff
@@ -306,6 +306,8 @@ if [ "$MANAGE_BR_BRIDGE" == "y" ] ; then
     sudo virsh net-destroy ${BAREMETAL_NETWORK_NAME}
     sudo nmcli con del ${BAREMETAL_NETWORK_NAME}
     sudo virsh net-start ${BAREMETAL_NETWORK_NAME}
+    # Needed in IPv6 on some EL9 hosts for the bootstrap VM to get an IP
+    echo 0 | sudo dd of=/proc/sys/net/ipv6/conf/${BAREMETAL_NETWORK_NAME}/addr_gen_mode
```
derekhiggins, line 310: @cybertron @mkowalski without this my bootstrap VM doesn't get any IP. Any idea why or if there is a better way I could apply it(I couldn't find a way to get it to work with nmcli)?

mkowalski: My only concern is that echo-ing into /proc/sys is very temporary and NetworkManager will revert that if at any point in time BAREMETAL_NETWORK_NAME goes down&up again. It may not be a real concern, but something to keep in mind.

Would this nmcli call work for you?
```
nmcli conn modify ${BAREMETAL_NETWORK_NAME} ipv6.addr-gen-mode 0
```

The ultimately persistent method is

```
/etc/NetworkManager/NetworkManager.conf
[...]
[connection]
ipv6.addr-gen-mode=0
```
to have it as a global default - that's what we do in real OCP clusters - https://github.com/openshift/machine-config-operator/blob/master/templates/common/openstack/files/ipv6-config.yaml

mkowalski: P.S. Be aware of a bug in older NetworkManager related to this feature - https://access.redhat.com/solutions/7016090. Again, it may be not relevant, but it's so subtle that at some point may be helpful to know about it

derekhiggins: I tried "nmcli conn modify ${BAREMETAL_NETWORK_NAME} ipv6.addr-gen-mode 0" but it didn't work 

I didn't try updating "/etc/NetworkManager/NetworkManager.conf" 

how would you feel about merging this as is and I'll circle back and try updating NetworkManager.conf once the CI nodes are all on EL9?

mkowalski: Right, indeed something is fishy with mode `0`...

```
[root@localhost ~]# nmcli conn show eno1 | grep addr-gen-mode
ipv6.addr-gen-mode:                     eui64
[root@localhost ~]# nmcli conn modify eno1 ipv6.addr-gen-mode 0
[root@localhost ~]# nmcli conn show eno1 | grep addr-gen-mode
ipv6.addr-gen-mode:                     eui64
[root@localhost ~]# nmcli conn modify eno1 ipv6.addr-gen-mode 1
[root@localhost ~]# nmcli conn show eno1 | grep addr-gen-mod
ipv6.addr-gen-mode:                     stable-privacy
[root@localhost ~]# nmcli conn modify eno1 ipv6.addr-gen-mode 2
[root@localhost ~]# nmcli conn show eno1 | grep addr-gen-mod
ipv6.addr-gen-mode:                     default-or-eui64
[root@localhost ~]# nmcli conn modify eno1 ipv6.addr-gen-mode eui64
[root@localhost ~]# nmcli conn show eno1 | grep addr-gen-mod
ipv6.addr-gen-mode:                     eui64
```

Go ahead with merging, your solution is perfectly valid one

PR #1651, opened 2024-04-16
CORS-3462: use static openshift-installer when possible

r4f4:
With https://github.com/openshift/installer/pull/8161, the baremetal installer is now part of the regular installer which is now statically linked.

Using the static binary has the advantage of working on both CS8 and CS9 VMs and is a necessary step for moving the Installer images to a RHEL9 base.

### DIFF
--- a/common.sh
+++ b/common.sh
@@ -174,6 +174,17 @@ fi
 
 export OPENSHIFT_RELEASE_TAG=$(echo $OPENSHIFT_RELEASE_IMAGE | sed -E 's/[[:alnum:]\/.-]*(release|okd).*://')
 
+if [[ -f ${OCP_DIR}/release_info.json ]]; then
+    release_version=$(jq -r ".metadata.version" ${OCP_DIR}/release_info.json | grep -oP "\d\.\d+")
+else
+    release_version=$(oc adm release info --registry-config "$PULL_SECRET_FILE" "$OPENSHIFT_RELEASE_IMAGE" -o json | jq -r ".metadata.version" | grep -oP "\d\.\d+")
+fi
+# On 4.16+ the baremetall installer was merged into the regular installer
+export DEFAULT_OPENSHIFT_INSTALL_CMD="openshift-install"
+if is_lower_version "${release_version}" 4.16; then
+    export DEFAULT_OPENSHIFT_INSTALL_CMD="openshift-baremetal-install"
+fi
+
 # Use "ipmi" for 4.3 as it didn't support redfish, for other versions
 # use "redfish", unless its CI where we use "mixed"
 if [[ "$OPENSHIFT_VERSION" == "4.3" ]]; then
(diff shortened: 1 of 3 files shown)


r4f4:
/retitle CORS-3024,CORS-2797: use static openshift-installer when possible

zaneb:
/retitle CORS-3462: use static openshift-installer when possible

common.sh
```diff
@@ -206,7 +206,12 @@ if [ "${UPSTREAM_IRONIC:-false}" != "false" ] ; then
 fi
 
 if [ -z "$KNI_INSTALL_FROM_GIT" ]; then
-    export OPENSHIFT_INSTALLER=${OPENSHIFT_INSTALLER:-${OCP_DIR}/openshift-baremetal-install}
+    # On 4.16+ the baremetall installer was merged into the regular installer
+    if is_lower_version $OPENSHIFT_VERSION 4.16; then
+        export OPENSHIFT_INSTALLER=${OPENSHIFT_INSTALLER:-${OCP_DIR}/openshift-baremetal-install}
```
zaneb, line 211: We also have `OPENSHIFT_INSTALLER_CMD` used in a couple of places, but not here (probably because of dependencies).

I think we need to define a `DEFAULT_OPENSHIFT_INSTALLER_CMD` and set it on the basis of the release version, then use that as the binary name here and as the default for `OPENSHIFT_INSTALLER_CMD` in the 2 places where that is defined.

dtantsur:
I wonder if this actually works. The CI fails with
```
rhcos.sh: line 26: ocp/cluster16/openshift-baremetal-install: No such file or directory
```

This file uses variable called OPENSHIFT_INSTALLER. I'm not sure why @zaneb suggested _CMD suffix, but we now need to update a lot more places.

andfasano:
/retest-required

@r4f4 in general the changes look fine to me, leaving the final approval to the metal folks since it will impact mostly metal jobs

common.sh
```diff
@@ -174,6 +174,12 @@ fi
 
 export OPENSHIFT_RELEASE_TAG=$(echo $OPENSHIFT_RELEASE_IMAGE | sed -E 's/[[:alnum:]\/.-]*(release|okd).*://')
 
+# On 4.16+ the baremetall installer was merged into the regular installer
+export DEFAULT_OPENSHIFT_INSTALLER_CMD="openshift-install"
+if is_lower_version ${OPENSHIFT_VERSION:-""} 4.16; then
```
dtantsur, line 179: OPENSHIFT_VERSION is empty in the CI run
```
 ++(common.sh:179): source(): is_lower_version '' 4.16
+++(common.sh:34): is_lower_version(): echo ' 4.16'
+++(common.sh:34): is_lower_version(): tr ' ' '\n'
+++(common.sh:34): is_lower_version(): sort -V
+++(common.sh:34): is_lower_version(): head -n1
++(common.sh:34): is_lower_version(): [[ '' != 4.16 ]]
++(common.sh:35): is_lower_version(): return 0
++(common.sh:180): source(): export DEFAULT_OPENSHIFT_INSTALLER_CMD=openshift-baremetal-install
++(common.sh:180): source(): DEFAULT_OPENSHIFT_INSTALLER_CMD=openshift-baremetal-install 
```

andfasano: Ops! It's the lowercase function `openshift_version` that returns the current payload version

zaneb: Then there's existing code that's wrong: https://github.com/openshift-metal3/dev-scripts/blob/6f6748827cb7d4f8f3d302ae99f17184cf363520/common.sh#L206

r4f4:
/hold
Depends on https://github.com/openshift-metal3/dev-scripts/pull/1652

zaneb:
/close
in favour of #1653

PR #1652, opened 2024-04-17
Assign OPENSHIFT_VERSION early if not manually defined

elfosardo:


01_install_requirements.sh
```diff
@@ -163,3 +163,10 @@ fi
 retry_with_timeout 5 60 "curl $OPENSHIFT_CLIENT_TOOLS_URL | sudo tar -U -C /usr/local/bin -xzf -"
 sudo chmod +x /usr/local/bin/oc
 oc version --client -o json
+
+if [[ -z "$OPENSHIFT_VERSION" ]]; then
+  source release_info.sh
+  save_release_info ${OPENSHIFT_RELEASE_IMAGE} ${OCP_DIR}
+  OPENSHIFT_VERSION=$(openshift_version)
+  export OPENSHIFT_VERSION
```
dtantsur, line 171: This file is not sourced, so it's probably not useful.

elfosardo: mmm right

zaneb (commented):
This is dragging us deeper and deeper into dependency hell.

I created `release_info.sh` so that we would _not_ have to include it everywhere that `utils.sh` is used.
This patch is now including both `release_info.sh` and `utils.sh` in `common.sh` so that basically every function is available everywhere and you have to have all of the data in the system in your head to know if any of it is going to work.

Bash is a horrible language to write something this big in. What we have now is borderline unmaintainable. This is going in exactly the wrong direction IMHO and I fear it will push us right over the edge. We need to make it more modular over time and not throw everything in one big pot.

common.sh
```diff
@@ -165,9 +166,22 @@ export OPENSHIFT_INSTALL_PATH="${OPENSHIFT_INSTALL_PATH:-$GOPATH/src/github.com/
...
+  sudo chmod +x /usr/local/bin/oc
+  oc version --client -o json
+  source release_info.sh
+  mkdir -p ${OCP_DIR}
+  save_release_info ${OPENSHIFT_RELEASE_IMAGE} ${OCP_DIR}
+  OPENSHIFT_VERSION=$(openshift_version)
+  export OPENSHIFT_VERSION
+fi
 if [[ -z "$OPENSHIFT_CI" ]]; then
   export OPENSHIFT_VERSION=${OPENSHIFT_VERSION:-$(echo $OPENSHIFT_RELEASE_IMAGE | sed "s/.*:\([[:digit:]]\.[[:digit:]]*\).*/\1/")}
```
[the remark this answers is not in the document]
zaneb: Misread it, this is for *not* CI.

elfosardo:
/hold
I'm not happy either about this, I was looking for a way to just get OPENSHIFT_VERSION early during the run to be available for the entire process, but at some point we got in a dependency loop.
Besides the fact that bash is definitely not the right way to go, we should find an alternative to provide all the needed info, especially a basic one like the running openshift version, at the start of the script.
I won't continue this, but I'll leave it here for the time being just as a reminder of the current status of dev-scripts.

PR #1653, opened 2024-04-17
CORS-3462: extract static openshift-installer from 4.16

zaneb:
With https://github.com/openshift/installer/pull/8161, the baremetal installer is now part of the regular installer which is now statically linked.

Using the static binary has the advantage of working on both CS8 and CS9 VMs and is a necessary step for moving the Installer images to a RHEL9 base.

### DIFF
--- a/03_build_installer.sh
+++ b/03_build_installer.sh
@@ -20,7 +20,11 @@ function build_installer() {
   # Build installer
   pushd .
   cd $OPENSHIFT_INSTALL_PATH
-  TAGS="${OPENSHIFT_INSTALLER_BUILD_TAGS:-libvirt baremetal}" DEFAULT_ARCH=$(get_arch) hack/build.sh
+  local default_tags="libvirt baremetal"
+  if [[ "${FIPS_MODE:-false}" = "true" && "${FIPS_VALIDATE:-false}" = "true" ]]; then
+      default_tags="${default_tags} fipscapable"
+  fi
+  TAGS="${OPENSHIFT_INSTALLER_BUILD_TAGS:-${default_tags}}" DEFAULT_ARCH=$(get_arch) hack/build.sh
   popd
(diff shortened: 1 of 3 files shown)


zaneb:
This is a potential alternative to #1651

03_build_installer.sh
```diff
@@ -20,7 +20,11 @@ function build_installer() {
   # Build installer
   pushd .
   cd $OPENSHIFT_INSTALL_PATH
-  TAGS="${OPENSHIFT_INSTALLER_BUILD_TAGS:-libvirt baremetal}" DEFAULT_ARCH=$(get_arch) hack/build.sh
+  local default_tags="libvirt baremetal"
+  if [ "${FIPS_MODE:-false}" = "true" && "${FIPS_VALIDATE:-false}" = "true" ]; then
```
dtantsur, line 24: You're mixing shell syntaxes here, I believe. Either

 ` [ "${FIPS_MODE:-false}" = "true" ] && [ "${FIPS_VALIDATE:-false}" = "true" ]`

or

 ` [[ "${FIPS_MODE:-false}" == "true" && "${FIPS_VALIDATE:-false}" == "true" ]]`

utils.sh
```diff
@@ -4,6 +4,14 @@ set -o pipefail
 
 source release_info.sh
 
+function default_installer_cmd() {
+    if is_lower_version "$(openshift_version "${OCP_DIR}")" "4.16" || [ "${FIPS_MODE:-false}" = "true" && "${FIPS_VALIDATE:-false}" = "true" ]; then
```
dtantsur, line 8: same here

03_build_installer.sh
zaneb: You're right, I misread the bit in the man page where it says not to use `-a`

PR #1661, opened 2024-05-29
OCPBUGS-34581: Fix bad dns agent test for ocp 4.16

andfasano:
This patch fixes the failing `AGENT_TEST_CASES` `bad_dns` script. Since Openshift 4.16, the nmtui main panel shows an additional "Radio" entry, thus requiring an additional step for a proper selection.


andfasano:
cc @rwsu

rwsu (commented):
/lgtm

PR #1662, opened 2024-05-30
Upgrade golang to 1.22.3

honza:


dtantsur:

The Go version does not affect the actually provisioned cluster. I don't want to hold an import Go update forever on unstable jobs.

PR #1668, opened 2024-06-21
Avoid installing yq twice

elfosardo:
Also install it early and pin its version


elfosardo:
/retest
not sure what happened here but it doesn't look related

elfosardo:
/retest
failure not related to this patch

elfosardo:
/retest
we're past the timeout issue

PR #1669, opened 2024-06-25
IP_STACK: v6 - allow disable MIRROR_IMAGES

hjensas:
IPv6 forces MIRROR_IMAGES and sets disableAllDefaultSources for OperatorHub, and the pull-secret contains only an entry for the mirror. This probably makes sense for disconnected installs, but not when there is actual connectivity.

Add another option, MIRROR_IMAGES_IPV6_DISABLE when unset the behaviour is as before. But when this is set - MIRROR_IMAGES is no longer implicitly enabled when "$IP_STACK" == "v6" and the release image override is not set.


network.sh
```diff
@@ -139,7 +139,9 @@ elif [[ "$IP_STACK" = "v6" ]]; then
   export SERVICE_SUBNET_V4=""
   export SERVICE_SUBNET_V6=${SERVICE_SUBNET_V6:-"fd02::/112"}
   export NETWORK_TYPE=${NETWORK_TYPE:-"OVNKubernetes"}
-  export MIRROR_IMAGES=true
```
derekhiggins, line 142: Instead of introducing a new variable could we just change this so that it only happens if unset? e.g. 
> export MIRROR_IMAGES=${MIRROR_IMAGES:-true}

hjensas: I changed it to do this. Also update all the conditions on `MIRROR_IMAGES` to treat value `false` the same way it treats `unset`.

hjensas: I changed it to this:
```
 # MIRROR_IMAGES -
-# Mirror latest ci images to local registry. This is always true for IPv6, but
-# can be turned off for an IPv4 install.
+# Mirror latest ci images to local registry. This is true by default for IPv6, but
+# can be turned on for an IPv4 install. To explicitly disable set it to false.
 # Default: true
 #
 #export MIRROR_IMAGES=true
```

NOTE the change `can be turned off for an IPv4` to `can be turned on for an IPv4`, I think that is more correct. It is by default left `unset` when `IP_STACK=v4`, so in that case one would have to explicitly enable it? The "Default: true" bit is still correct because the default for `IP_STACK` is `v6`.

bfournie:
For the agent installer I think the checks in the yaml files also need to change to check for 'true', e.g. `mirror_images` is set here:
https://github.com/openshift-metal3/dev-scripts/blob/master/agent/roles/manifests/vars/main.yml#L36

And its currently being checked in multiple places like this:
https://github.com/openshift-metal3/dev-scripts/blob/master/agent/roles/manifests/templates/install-config_baremetal_yaml.j2#L109

That should be changed to 
`{% if mirror_images == 'true' %}`

hjensas:
> For the agent installer I think the checks in the yaml files also need to change to check for 'true', e.g. `mirror_images` is set here: https://github.com/openshift-metal3/dev-scripts/blob/master/agent/roles/manifests/vars/main.yml#L36
> 
> And its currently being checked in multiple places like this: https://github.com/openshift-metal3/dev-scripts/blob/master/agent/roles/manifests/templates/install-config_baremetal_yaml.j2#L109
> 
> That should be changed to `{% if mirror_images == 'true' %}`

I changed it like this:
```
-mirror_images: "{{ lookup('env', 'MIRROR_IMAGES') != '' }}"
+mirror_images: "{{ lookup('env', 'MIRROR_IMAGES') | bool }}"
````

This means values `true`, `True`, `TRUE`, `1`, `yes` etc should make `{% if mirror_images %}` evaluate to `true` - empty string, `false`, `False`, `FALSE`, `no`, `0` etc it evaluates to `false`. So no need to update the conditions? :crossed_fingers:

bfournie:
> > For the agent installer I think the checks in the yaml files also need to change to check for 'true', e.g. `mirror_images` is set here: https://github.com/openshift-metal3/dev-scripts/blob/master/agent/roles/manifests/vars/main.yml#L36
> > And its currently being checked in multiple places like this: https://github.com/openshift-metal3/dev-scripts/blob/master/agent/roles/manifests/templates/install-config_baremetal_yaml.j2#L109
> > That should be changed to `{% if mirror_images == 'true' %}`
> 
> I changed it like this:
> 
> ```
> -mirror_images: "{{ lookup('env', 'MIRROR_IMAGES') != '' }}"
> +mirror_images: "{{ lookup('env', 'MIRROR_IMAGES') | bool }}"
> ```
> 
> This means values `true`, `True`, `TRUE`, `1`, `yes` etc should make `{% if mirror_images %}` evaluate to `true` - empty string, `false`, `False`, `FALSE`, `no`, `0` etc it evaluates to `false`. So no need to update the conditions? 🤞

Sounds good, thanks Harald

elfosardo:
the ipv4 job is currently foobar
we're waiting for https://github.com/openshift/release/pull/53433

hjensas:

https://github.com/openshift/release/pull/53433 is merged, so giving it a go.

hjensas:
I added another commit to use the `,,` to do case modification on MIRROR_IMAGES before comparing to `false` - this allows the user to use `False`, `FALSE` etc. Nice to have when ansible is used to write the configuration.

elfosardo:
/retest
/lgtm
it looks ok, just verifying failing jobs

elfosardo:
/override ci/prow/e2e-metal-ipi-bm-bond ci/prow/e2e-metal-ipi-serial-ipv4
failures are not related

elfosardo:
unrelated, known issue

elfosardo:
unrelated known issue

PR #1670, opened 2024-06-27
Fix yq -iy unsupported flag

maximsava12:
Deployment failed with:
  "Error: unknown shorthand flag: 'y' in -y"
This patch should fix the issue


elfosardo:
/hold
this should avoid the problem https://github.com/openshift-metal3/dev-scripts/pull/1668
it's not needed to update the command unless we explicitly install a version higher than 3
in any case currently the highest version available in pipy is 3.4.3 https://pypi.org/project/yq/

elfosardo:
/close
it's been a while

PR #1671, opened 2024-06-28
AGENT-925: Block port 22624 when adding day 2 worker node

rwsu:
22624 is the insecure port for the machine config server. Day 2
nodes should be communicating through the secure port 22623. By
blocking the insecure port for extra worker nodes, we force them
to use the secure port.

The filter isn't applied to master or worker nodes because they
fetch their ignition through the secure port during initial
installation.

### DIFF
--- a/agent/05_agent_configure.sh
+++ b/agent/05_agent_configure.sh
@@ -476,6 +476,36 @@ function get_nodes_bmc_info() {
 
 }
 
+function block_insecure_machine_config_server_port() {
+    existing_filter=$(sudo virsh nwfilter-list)
+    if [[ "$existing_filter" == *"block-insecure-machine-config-server"* ]]; then
+        sudo virsh nwfilter-undefine block-insecure-machine-config-server
+    fi
+    tmpfilter=$(mktemp --tmpdir "block-insecure-mcs--XXXXXXXXXX")
+    _tmpfiles="$_tmpfiles $tmpfilter"
+    echo "<filter name='block-insecure-machine-config-server' chain='root'>
+  <uuid>aaaaaaaa-aaaa-aaaa-aaaa-000000000001</uuid>
+  <rule action='drop' direction='in' priority='500'>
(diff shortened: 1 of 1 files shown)


agent/05_agent_configure.sh
```diff
@@ -476,6 +476,36 @@ function get_nodes_bmc_info() {
...
+        tmpdomain=$(mktemp --tmpdir "${name}--XXXXXXXXXX")
+        _tmpfiles="$_tmpfiles $tmpdomain"
+        sudo virsh dumpxml ${name} > ${tmpdomain}
+
+        sed -i '/interface type=\([^>]*\)>/a\
+      <filterref filter="block-insecure-machine-config-server"/>\' ${tmpdomain}
+
+        sudo virsh define ${tmpdomain}
+    done
+}
```
andfasano, line 506: Maybe I'm oversimplifying it a little bit, but wouldn't be enough to execute a simple ansible playbook like the following (similar to the ones used in metal3-dev-env):

```
tasks:
    - name: Block port 22624 using firewalld
      firewalld:
        port: 22624/tcp
        state: disabled
        immediate: yes
        permanent: yes
```

on each control plane (master) node? Given also that this is a specific requirement for the add nodes workflow, it could be probably a task executed only in the new step

agent/05_agent_configure.sh
```diff
@@ -519,6 +549,8 @@ if [[ "${AGENT_PLATFORM_TYPE}" == "external" ]] || [[ "${AGENT_PLATFORM_TYPE}" =
   set_device_mfg worker $NUM_WORKERS ${AGENT_PLATFORM_TYPE} ${AGENT_PLATFORM_NAME}
 fi
 
+block_insecure_machine_config_server_port extraworker $NUM_EXTRA_WORKERS
```
andfasano, line 552: As mentioned before, I was expecting the port to be blocked on the control plane nodes, more than on the hosts to be added

PR #1672, opened 2024-07-05
Move to Ansible 8

elfosardo:
Update metal3-dev-env hash and set default Ansible version to 8.0.0


elfosardo:
/hold
unbelievable....
going to change the commit message before merging

PR #1674, opened 2024-07-10
METAL-1089: Enable CRB repo in RHEL9

elfosardo:
Starting from RHEL9 some packages (e.g. libvirt-devel) have been moved to the CRB repo, which is not enabled by default. Add the proper command to enable it on RHEL9, noting that a valid subscription is needed for that.


elfosardo:
not related

PR #1675, opened 2024-07-11
Add support for Alma 9

cybertron:
This seems to have been dropped in the move from 8 to 9. Just adding it back in so we have equivalent functionality for 8 and 9-based distros (and because I'm still using it).


cybertron:
/test e2e-metal-ipi-ovn-dualstack-v6

All three nodes ended up in different states. I want to see if that behavior is consistent.

cybertron:
/retest-required

Unrelated failure pulling an image

PR #1676, opened 2024-07-15
Install python-devel package based on OS version

elfosardo:
In el9 python is already version 3.9 and the devel package is simply named python3-devel, the python39-devel package does not exist.


elfosardo:
/retest
giving it one more chance, but I'm going to override it afterwards as the failure does not depend from this change

elfosardo:
not related

PR #1682, opened 2024-07-23
Remove unwanted $

sdatko:
The existing code works a bit accidentally, because the command output is ignored anyway. If there would be no output redirection to /dev/null, then the command would always fail with something like ~"Command not found: This system is not yet registered".

Details: https://www.shellcheck.net/wiki/SC2091

### DIFF
--- a/01_install_requirements.sh
+++ b/01_install_requirements.sh
@@ -90,7 +90,7 @@ case $DISTRO in
     elif [[ $DISTRO == "rhel9" ]]; then
       # NOTE(raukadah): If a system is subscribed to RHEL subscription then
       # sudo subscription-manager identity will return exit 0 else 1.
-      if $(sudo subscription-manager identity > /dev/null 2>&1); then
+      if sudo subscription-manager identity > /dev/null 2>&1; then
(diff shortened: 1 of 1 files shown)


cjeanner (approved):

01_install_requirements.sh
```diff
@@ -90,7 +90,7 @@ case $DISTRO in
     elif [[ $DISTRO == "rhel9" ]]; then
       # NOTE(raukadah): If a system is subscribed to RHEL subscription then
       # sudo subscription-manager identity will return exit 0 else 1.
-      if $(sudo subscription-manager identity > /dev/null 2>&1); then
+      if (sudo subscription-manager identity > /dev/null 2>&1); then
```
dtantsur, line 93: Good catch, I always miss this :( Could you remove the parenthesis as well?

sdatko: Sure, done – I originally kept the parenthesis for readability only

raukadah (approved):

elfosardo:
unrelated

PR #1683, opened 2024-07-26
Remove Ironic inspector from clouds.yaml

honza:


elfosardo:
this works for OCP 4.17, but what about older versions?

honza:
Thanks, I'll update the patch.

honza:
The existing code implies that "noauth" is supported in 4.3-4.5.  Is that correct?

elfosardo:
> The existing code implies that "noauth" is supported in 4.3-4.5. Is that correct?

@honza I believe that's plausible 
and now I wonder if we should just remove all the logic since those releases are not supported since a looooong time
maybe in a different patch though

elfosardo:
/retest
gotta love CI

metal3-templater.go
```diff
@@ -90,6 +93,7 @@ func main() {
 		}
 
 		templateData.AuthType = "none"
+		templateData.OCPVersionUsesInspector = true
```
MahnoorAsghar, line 96: Why is this set to true? (Shouldn't it be false by default?)

MahnoorAsghar: 'noauth' is for the older releases, got it.

utils.sh
```diff
@@ -343,15 +344,23 @@ function generate_auth_template {
...
         IRONIC_CREDS="$IRONIC_USER:$IRONIC_PASSWORD"
-        INSPECTOR_USER=$((oc -n openshift-machine-api  get secret/metal3-ironic-inspector-password -o template --template '{{.data.username}}' || echo "") | base64 -d)
-        INSPECTOR_PASSWORD=$((oc -n openshift-machine-api  get secret/metal3-ironic-inspector-password -o template --template '{{.data.password}}' || echo "") | base64 -d)
-        INSPECTOR_CREDS="$INSPECTOR_USER:$INSPECTOR_PASSWORD"
+        if [[ "$OCP_VERSIONS_INSPECTOR" == *"$VERSION"* ]]; then
+            INSPECTOR_USER=$((oc -n openshift-machine-api  get secret/metal3-ironic-inspector-password -o template --template '{{.data.username}}' || echo "") | base64 -d)
+            INSPECTOR_PASSWORD=$((oc -n openshift-machine-api  get secret/metal3-ironic-inspector-password -o template --template '{{.data.password}}' || echo "") | base64 -d)
+            INSPECTOR_CREDS="$INSPECTOR_USER:$INSPECTOR_PASSWORD"
+        else
+            INSPECTOR_CREDS=""
```
elfosardo, line 352: we can probably remove this else statement, INSPECTOR_CREDS is not needed if we don't use inspector

hroyrh: Something is causing the jobs to fail ( [example](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift-metal3_dev-scripts/1683/pull-ci-openshift-metal3-dev-scripts-master-e2e-metal-ipi-bm/1845831262900588544#1:build-log.txt%3A19931-19976) ), can this be related.

elfosardo: yep, just tested this locally and I think we should remove -inspector-basic-auth in L362

elfosardo: if we can't change the arguments conditionally in metal3-templater.go the value should be -ironic-basic-auth=":" in case of inspector absence

honza: Updated.

I roved the else branch.  I made the `-inspector-basic-auth` flag optional in the templater and then I removed the flag at the call site.

PR #1684, opened 2024-07-29
Don't force use of a mirrored image in v6-primary

cybertron:
It isn't necessary to use a mirror with v6-primary since it's still dual stack and still has access to v4 public registries. By forcing the use of the mirrored image path, we actually break the no mirror use case because we point the deployment at an image that doesn't exist in the local registry.


mkowalski (approved):

elfosardo:
/override ci/prow/e2e-agent-ha-dualstack ci/prow/e2e-metal-ipi-virtualmedia
not really needed for this change

elfosardo:
failure looks related to an issue with a storage operator

PR #1694, opened 2024-08-20
NO-ISSUE: revert unwanted images mirroring

andfasano:
PR #1669 introduced a [check](https://github.com/openshift-metal3/dev-scripts/pull/1669/files#diff-a045819fa24dfbeb74522ddaf6eac77b5a4d39762752db45251e6101cd5c987dR77) that appears to be always enabled for the agent jobs, since the `IP_STACK` env var is exported only [after](https://github.com/openshift-metal3/dev-scripts/blob/bdd7bc71fe7949116e12973032d665c60a545c20/common.sh#L410) that check, thus consuming more resources/time also for those scenarios where is not strictly required.
`MIRROR_IMAGES` seems anyhow already enabled for v6 in [network.sh](https://github.com/openshift-metal3/dev-scripts/blob/bdd7bc71fe7949116e12973032d665c60a545c20/network.sh#L14), so this block appears to be redundant (except for the dualstack case which I'm not sure/remember if it was really required)


common.sh
```diff
@@ -73,10 +73,6 @@ export SSH_PUB_KEY="${SSH_PUB_KEY:-$(cat $HOME/.ssh/id_rsa.pub)}"
 
 # mirror images for installation in restricted network
 export MIRROR_IMAGES=${MIRROR_IMAGES:-}
-# For IPv6 (default case) mirror images are used since quay doesn't support IPv6
```
hjensas, line 76: I added this because I change the condition on [L488](https://github.com/openshift-metal3/dev-scripts/pull/1694/files#diff-a045819fa24dfbeb74522ddaf6eac77b5a4d39762752db45251e6101cd5c987dL488).

I wonder if the IPv6 jobs failing is related to this? If it is, we may end up with better result if we remove the first condition `-z "${IP_STACK:-}"` only here?

andfasano: Given that in `common.sh` there was already a point where `MIRROR_IMAGES` was set on [L223](https://github.com/openshift-metal3/dev-scripts/pull/1694/files#diff-a045819fa24dfbeb74522ddaf6eac77b5a4d39762752db45251e6101cd5c987dL223-L225) - and that's usually a dev-only feature when making a [custom release](https://github.com/openshift-metal3/dev-scripts/blob/master/agent/docs/custom-release.md) not used in CI -, my personal preference would be to limit the number of places where a config var is set to avoid increasing the complexity. 
IIRC the block on L488 it's not actively used in CI (cc @bfournie for confirmation), so I don't think it's an issue right now (and I feel it could be moved somewhere else eventually)

andfasano: > I wonder if the IPv6 jobs failing is related to this?

I suspect it's something else  - at least for the agent jobs, since locally it's working (maybe @elfosardo or @derekhiggins could comment on the metal jobs)

andfasano: It was not, in my local tests I wasn't using the `AGENT_USE_ZTP_MANIFESTS` config var (as it happens in the agent SNO CI job). So, I restored the initial checks to allow defining `OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE`, and now it should be ok (unfortunately the CI is broken at the moment...)

andfasano:
Hopefully the CI issue was resolved now, retesting

/retest

common.sh
bfournie: @andfasano correct, the `oc-mirror` case is not currently used in CI, when we looked into using it did not support nightly builds (https://issues.redhat.com/browse/AGENT-368). If that has changed we should make the default for mirroring since its a more common use case.

elfosardo:
/override ci/prow/e2e-metal-ipi-bm  ci/prow/e2e-metal-ipi-virtualmedia
unrelated

hjensas:
So, turns out this PR broke the RHOSO CI downstream.

This [condition](https://github.com/openshift-metal3/dev-scripts/pull/1694/files#diff-a045819fa24dfbeb74522ddaf6eac77b5a4d39762752db45251e6101cd5c987dR484):
`if [[ (-n "$MIRROR_IMAGES" && "${MIRROR_IMAGES,,}" != "false") || -z "${IP_STACK:-}" || "$IP_STACK" = "v6" ]]; then`

`MIRROR_IMAGE == false` + `IP_STACK = v6` leads to `TRUE` i.e ignoring the config of `MIRROR_IMAGES=false` which means we end up trying to use the local registry in L505 - i.e it breaks what I tried to implement in https://github.com/openshift-metal3/dev-scripts/pull/1669.

andfasano:
> So, turns out this PR broke the RHOSO CI downstream.
> 
> This [condition](https://github.com/openshift-metal3/dev-scripts/pull/1694/files#diff-a045819fa24dfbeb74522ddaf6eac77b5a4d39762752db45251e6101cd5c987dR484): `if [[ (-n "$MIRROR_IMAGES" && "${MIRROR_IMAGES,,}" != "false") || -z "${IP_STACK:-}" || "$IP_STACK" = "v6" ]]; then`
> 
> `MIRROR_IMAGE == false` + `IP_STACK = v6` leads to `TRUE` ? i.e ignoring the config of `MIRROR_IMAGES=false` which means we end up trying to use the local registry in L505 - i.e it breaks what I tried to implement in #1669.

Sorry but I don't know the RHOSO CI jobs, could you please share a link to the failing jobs? Assuming that they are not agent jobs, I guess the problem could be originated by [L509](https://github.com/openshift-metal3/dev-scripts/blob/d7ce3f44e0c0c65549784e1ebf9ef02533ce2339/common.sh#L509)

hjensas:
> > So, turns out this PR broke the RHOSO CI downstream.
> > This [condition](https://github.com/openshift-metal3/dev-scripts/pull/1694/files#diff-a045819fa24dfbeb74522ddaf6eac77b5a4d39762752db45251e6101cd5c987dR484): `if [[ (-n "$MIRROR_IMAGES" && "${MIRROR_IMAGES,,}" != "false") || -z "${IP_STACK:-}" || "$IP_STACK" = "v6" ]]; then`
> > `MIRROR_IMAGE == false` + `IP_STACK = v6` leads to `TRUE` ? i.e ignoring the config of `MIRROR_IMAGES=false` which means we end up trying to use the local registry in L505 - i.e it breaks what I tried to implement in #1669.
> 
> Sorry but I don't know the RHOSO CI jobs, could you please share a link to the failing jobs? Assuming that they are not agent jobs, I guess the problem could be originated by [L509](https://github.com/openshift-metal3/dev-scripts/blob/d7ce3f44e0c0c65549784e1ebf9ef02533ce2339/common.sh#L509)

The jobs are using the IPI to deploy OCP. And yes, the issue is that  [L509](https://github.com/openshift-metal3/dev-scripts/blob/d7ce3f44e0c0c65549784e1ebf9ef02533ce2339/common.sh#L509) overrides to use the local mirror.

Example logs of failing job here: https://sf.hosted.upshift.rdu2.redhat.com/logs/09/609/472164b16879772ff7f8cc8f1aa4ec21a64a807a/check-gitlab-cee/uni04delta-rhel9-rhoso18.0-ipv6-sealusa40-sdatko-upstream/93a65f0/logs/hypervisor/ci-framework-data/logs/devscripts/

You can see in the log that despite `MIRROR_IMAGES=False` - it still set `OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE` to the local mirror.
```
2024-08-29 11:06:52 ++(common.sh:484): [[ -n False ]]
2024-08-29 11:06:52 ++(common.sh:484): [[ false != \f\a\l\s\e ]]
2024-08-29 11:06:52 ++(common.sh:484): [[ -z v6 ]]
2024-08-29 11:06:52 ++(common.sh:484): [[ v6 = \v\6 ]]
2024-08-29 11:06:52 ++(common.sh:486): [[ oc-adm == \o\c\-\m\i\r\r\o\r ]]
2024-08-29 11:06:52 ++(common.sh:498): [[ ! -z '' ]]
2024-08-29 11:06:52 ++(common.sh:509): export =virthost.ocp.openstack.lab:5000/localimages/local-release-image:latest
2024-08-29 11:06:52 ++(common.sh:509): OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE=virthost.ocp.openstack.lab:5000/localimages/local-release-image:latest
```

I have proposed a potential fix: https://github.com/openshift-metal3/dev-scripts/pull/1700

PR #1697, opened 2024-08-27
NO-ISSUE: configure install-config.yaml proxy when required

andfasano:
This patch fixes the agent scenarios when a proxy is configured via the `INSTALLER_PROXY` config var. 
Without this patch, all the (connected) scenario - such as the following - are broken:

```
export INSTALLER_PROXY=true
```


andfasano:
Note: useful to test https://github.com/openshift/oc/pull/1857

pawanpinjarkar (commented):
/lgtm

PR #1699, opened 2024-08-29
Fix mirror config for v6-primary

cybertron:
It turns out we do need to mirror images for v6-primary clusters because during provisioning they only get an IP from the primary family, which means they can't pull images from the public ipv4 image registries. This fixes the logic so it is all consistent where it needs to be so v6-primary deployments will work as intended by default.


cybertron:
/test e2e-metal-ipi-ovn-dualstack-v6

I flipped this back to draft because I'm not necessarily expecting it to work yet, but I want to see how it behaves in CI since it has been different from my local results up until now.

cybertron:
/test e2e-metal-ipi-ovn-dualstack-v6

Cluster deployed this time but the job failed on image stream import. Seems unrelated, but let's see if it reproduces.

network.sh
```diff
@@ -148,6 +148,11 @@ elif [[ "$IP_STACK" = "v4v6" || "$IP_STACK" = "v6v4" ]]; then
   export SERVICE_SUBNET_V4=${SERVICE_SUBNET_V4:-"172.30.0.0/16"}
   export SERVICE_SUBNET_V6=${SERVICE_SUBNET_V6:-"fd02::/112"}
   export NETWORK_TYPE=${NETWORK_TYPE:-"OVNKubernetes"}
+  # Although v6-primary clusters have access to both v4 and v6, during provisioning
+  # the nodes only get v6 addresses so we need to mirror images or they can't pull.
+  if [[ "$IP_STACK" = "v6v4" ]]; then
+    export MIRROR_IMAGES=true
```
hjensas, line 154: Can we use `${MIRROR_IMAGES:-true}` here? like L142? So that we only set `true` if the user did not already set `MIRROR_IMAGES:-false` in config.

cybertron: Yes, done.

cybertron:
/test e2e-metal-ipi-ovn-dualstack-v6

Let's see if it fails in yet another different way.

cybertron:
/test e2e-metal-ipi-ovn-dualstack-v6

Trying with a single BMC driver.

PR #1701, opened 2024-09-12
Add additional macros for agent installer nmstate testing

bfournie:
Adds a couple macros that were useful in testing specific nmstate configurations.


agent/roles/manifests/templates/net_macros.yaml
```diff
@@ -5,6 +5,26 @@
...
 
+{%- macro interfacesmacid(name, mac) -%}
+      - name: {{ name }}
+        type: ethernet
+        state: up
+        identifier: mac-address
+        mac-address: {{ mac }}
+{%- endmacro %}
+
+{%- macro interfacesdisabled(name) -%}
```
rwsu, line 16: I had some trouble figuring out how or in what context this macro could be used. Could you add some comments?

bfournie: Sure, added comments for this new macros, let me know if they aren't clear. To add some additional info
- adding disabled interfaces to the NM Config section is a common thing in customer scenarios in order to support h/w with multiple nics that should be disabled to avoid them running DHCP and introducing network problems. These disabled interfaces often use the actual NIC names and therefore don't need entries in the MAC interface table
- the ` identifier: mac-address` field is supported by nmstate but not in common usage

I'm not sure either of these need a dedicated CI test but it may be useful to duplicate a customer scenario.

PR #1706, opened 2024-10-08
AGENT-966: use oc from current release image

andfasano:
This patch is required to test https://github.com/openshift/installer/pull/9066 

As a rule of thumb, node-joiner patches may require to update as well the related wrapping oc commands (which could be present in CI registries but not yet released to the usual mirror)


pawanpinjarkar (approved):
/lgtm

PR #1707, opened 2024-10-09
OCPBUGS-36506: Fix config for disconnected cluster

derekhiggins:
Ensure all but BMC traffic goes via the proxy


derekhiggins:
/hold
still testing

PR #1710, opened 2024-10-21
NO-JIRA: increase log level for day2 add nodes commands

andfasano:
This patch increases the `oc adm node-image ...` commands loglevel to help troubleshooting CI jobs failures


rwsu (commented):
/lgtm

PR #1711, opened 2024-10-23
AGENT-975: Add support for agent-based installer minimal ISO

bfournie:
Configure the installer to create a minimal ISO. Add support for a minimal ISO installation in a disconnected environment using the bootArtifactsBaseURL field. Made the PXE server into a more generic boot server that can be used to serve the rootfs for minimal ISO.


elfosardo:
/retest
bit worried about the virtualmedia job, the failure does not depend on this change of course, let's give it another try

bfournie:
All three of the BM jobs (metal-ipi-bm, ipi-bm-bond, and ipi-virtualmedia) are failing more often than not in the history and the failures for this PR are unrelated to the change and consistent with the other failures.

agent/roles/manifests/vars/main.yml
```diff
@@ -5,6 +5,7 @@ agent_extra_workers_ips: "{{ lookup('env', 'AGENT_EXTRA_WORKERS_IPS_STR') }}"
 agent_extra_workers_ipsv6: "{{ lookup('env', 'AGENT_EXTRA_WORKERS_IPSV6_STR') }}"
 agent_extra_workers_hostnames: "{{ lookup('env', 'AGENT_EXTRA_WORKERS_HOSTNAMES_STR') }}"
 agent_install_config_bm_hosts: "{{ lookup('env', 'AGENT_BM_HOSTS_IN_INSTALL_CONFIG') }}"
+agent_minimal_iso: "{{ lookup('env', 'AGENT_MINIMAL_ISO') }}"
```
rwsu, line 8: I think AGENT_MINIMAL_ISO should be added to config_example.sh.

agent/roles/manifests/templates/agent-config_yaml.j2
```diff
@@ -14,8 +14,11 @@ rendezvousIP: {{ ips[0] }}
 {% else %}
 rendezvousIP: {{ ipsv6[0] }}
 {% endif %}
-{% if boot_mode == "PXE" %}
-bootArtifactsBaseURL: {{ pxe_server_url }}
+{% if (boot_mode == "PXE") or (agent_minimal_iso == "true" and mirror_images) %}
```
rwsu, line 17: Could you explain the relationship between agent_minimal_iso and mirror_images? Does having AGENT_MINIMAL_ISO set to true require a user to also set MIRROR_IMAGES to true?

pawanpinjarkar:
Agree with @rwsu as above. Overall other changes look good hence approving. 

/approve

agent/roles/manifests/templates/agent-config_yaml.j2
bfournie: AGENT_MINIMAL_ISO doesn't imply that MIRROR_IMAGES is needed as the rootfs will be downloaded from mirror.openshift.com in a connected setup. However, if MIRROR_IMAGES is set, then the `bootArtifactsBaseURL` must also be set since the rootfs in that case must be retrieved from the local boot server.

I will add a comment describing this.

rwsu (commented):
/lgtm

agent/cleanup.sh
```diff
@@ -15,17 +15,16 @@ early_cleanup_validation
...
-  "DISKIMAGE" )
+if [[ "${AGENT_E2E_TEST_BOOT_MODE}" == "DISKIMAGE" ]]; then
     sudo rm -rf "${OCP_DIR}/cache"
     sudo rm -rf "${OCP_DIR}/temp"
     sudo podman rmi -f ${APPLIANCE_IMAGE} || true
-    ;;
-esac
+fi
+
+if [[ -d ${BOOT_SERVER_DIR} ]]; then
```
andfasano, line 24: Isn't equivalent to the previous case (which I found more readable)? Ie, the pxe boot server gets currently created only when AGENT_E2E_TEST_BOOT_MODE is PXE, isn't it?

agent/common.sh
```diff
@@ -25,12 +27,14 @@ export OPENSHIFT_INSTALLER_CMD="openshift-install"
...
-  export PXE_SERVER_DIR=${WORKING_DIR}/boot-artifacts
-  export PXE_SERVER_URL=http://$(wrap_if_ipv6 ${PROVISIONING_HOST_EXTERNAL_IP}):${AGENT_PXE_SERVER_PORT}
-  export PXE_BOOT_FILE=agent.x86_64.ipxe
-fi
+# Set required config vars for the http boot server
+# The necessary files will be copied to boot-artifacts by the installer for either:
+# 1. PXE, when the 'openshift-install agent create pxe-files' command is run
+# 2. Minimal ISO, when the 'openshift-install agent create image' command is run and bootArtifacts is set
+#    in install-config.yaml
+export BOOT_SERVER_DIR=${WORKING_DIR}/boot-artifacts
```
andfasano, line 35: As before, what's the requirement to define those config var if not PXE?

agent/cleanup.sh
bfournie: No we also need to create the boot server for minimal ISO in the disconnected case. Its similar to PXE in that a http server is providing the files written to `boot-artifacts`, but its not the same set of files for minimal ISO

andfasano:
/approve

(probably it would be useful to have a specific pxe job also here)

PR #1713, opened 2024-10-30
AGENT-991: Scenarios for 4 and 5 control plane replicas

rwsu:
Added 5CONTROL and 4CONTROL as additional AGENT_E2E_TEST_SCENARIOs.

Added post install validation to count number of master and worker nodes matches NUM_MASTERS and NUM_WORKERS.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@ all: default
 assisted: assisted_deployment bell
 
 # Deploy cluster with agent installer flow
-agent: agent_requirements requirements configure agent_build_installer agent_prepare_release agent_configure agent_create_cluster
+agent: agent_requirements requirements configure agent_build_installer agent_prepare_release agent_configure agent_create_cluster agent_post_install_validation
 
 # Deploy cluster with agent installer flow and adds nodes after initial install
 # Requires at least 1 extra worker node to be configured with disk size at least 100Gß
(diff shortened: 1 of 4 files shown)


common.sh
```diff
@@ -413,6 +413,20 @@ if [[ ! -z ${AGENT_E2E_TEST_SCENARIO} ]]; then
   fi
 
   case "$SCENARIO" in
+      "5CONTROL" )
+          export NUM_MASTERS=5
+          export MASTER_VCPU=4
+          export MASTER_DISK=100
+          export MASTER_MEMORY=32768
```
andfasano, line 420: I'm wondering if it could be better to have a 16G for the 4/5 cases, considering also the possibility to add some workers later - to not consume too much mem for the VMs

rwsu: I used 16G to test the 5 case and it worked fine. Let's try that.

bfournie:
It would be useful to update config_example.sh with the new scenarios...
https://github.com/openshift-metal3/dev-scripts/blob/master/config_example.sh#L720

Also the error message here https://github.com/openshift-metal3/dev-scripts/blob/master/common.sh#L382.

Otherwise looks good.

Makefile
```diff
@@ -7,7 +7,7 @@ all: default
 assisted: assisted_deployment bell
 
 # Deploy cluster with agent installer flow
-agent: agent_requirements requirements configure agent_build_installer agent_prepare_release agent_configure agent_create_cluster
+agent: agent_requirements requirements configure agent_build_installer agent_prepare_release agent_configure agent_create_cluster agent_post_install_validation
```
pawanpinjarkar, line 10: Nice addition!

pawanpinjarkar (commented):
/lgtm

PR #1719, opened 2024-11-14
Fix undefined variable in cleanup

cybertron:
This variable is not always defined and it's currently causing make clean to fail in non-agent use cases.


cybertron:
Oops, dupe of https://github.com/openshift-metal3/dev-scripts/pull/1718

PR #1720, opened 2024-11-15
METAL-991: Add script to set up CAPI e2e tests

honza:


06_create_cluster.sh
```diff
@@ -123,6 +123,20 @@ if [[ ! -z "${ENABLE_VIRTUAL_MEDIA_VIA_EXTERNAL_NETWORK}" ]]; then
     oc patch provisioning provisioning-configuration --type merge -p "{\"spec\":{\"virtualMediaViaExternalNetwork\":true}}"
 fi
 
+if [[ -n "${ENABLE_CAPI_E2E}" ]]; then
```
dtantsur, line 126: Could it be a small separate script that is invoked in the new job?

honza: Done

honza: ... now I'm wondering how we will call it from the job: https://github.com/openshift/release/pull/58839

honza: I added the env var back.

PR #1722, opened 2024-11-20
AGENT-963: make target to remove extraworker nodes from cluster

rwsu:
Adds a new target to remove all extraworker nodes:

make agent_remove_node

The nodes are deleted from the cluster. The VM is then shutdown and its disk reinitialized to be empty. Any agent ISO attached as /dev/sdc is also removed.

The extraworker nodes can be added back to the cluster by running:

make agent_add_node

They can be repeated added and removed using these two make targets.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,7 @@ agent: agent_requirements requirements configure agent_build_installer agent_pre
 
 # Deploy cluster with agent installer flow and adds nodes after initial install
 # Requires at least 1 extra worker node to be configured with disk size at least 100Gß
-agent_plus_add_node: agent agent_add_node
+agent_plus_add_node: agent agent_add_extraworker_nodes
 
 agent_requirements:
 	./agent/01_agent_requirements.sh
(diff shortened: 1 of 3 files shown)


agent/agent_remove_node.sh
andfasano, line 1: This file sounds more like a `agent_remove_all_extraworkers_nodes.sh`

Makefile
```diff
@@ -40,6 +40,9 @@ agent_tests:
 agent_add_node:
 	./agent/07_agent_add_node.sh
 
+agent_remove_node:
```
andfasano, line 43: Ona second check, it seems more suitable naming the targets as:

* `agent_add_extraworker_nodes`
* `agent_remove_extraworker_nodes`

rwsu: Good suggestion. I will make that change.

agent/07_agent_add_node.sh
```diff
@@ -54,7 +75,7 @@ case "${AGENT_E2E_TEST_BOOT_MODE}" in
     # Copy the generated PXE artifacts in the tftp server location
     # The local http server should be running and was started by
     # day 1 installtion.
-    cp $OCP_DIR/add-node//boot-artifacts/* ${PXE_SERVER_DIR}
+    cp $OCP_DIR/add-node//boot-artifacts/* ${BOOT_SERVER_DIR}
```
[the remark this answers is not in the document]
rwsu: Yes, we missed this change. The PXE and Bob's minimal ISO PR were being merged around the same time.

pawanpinjarkar (commented):
/lgtm

PR #1723, opened 2024-11-27
NO-JIRA: use default logging for oc adm node-image commands

andfasano:
Remove debug logging since too verbose


pawanpinjarkar (commented):
/lgtm

PR #1726, opened 2024-12-11
Add almalinux9 support

eb4x:
It's one of the other popular RHELatives.


eb4x:
Closing in favor of #1675 which does the same.

PR #1728, opened 2025-01-16
NO-JIRA: skip the add nodes step if no additional workers are defined

andfasano:
This patch skips the 07 step in case no extra workers are present.

### DIFF
--- a/agent/05_agent_configure.sh
+++ b/agent/05_agent_configure.sh
@@ -331,13 +331,18 @@ function generate_cluster_manifests() {
 }
 
 function write_extra_workers_ips() {
+  # not required in case no additional workers were defined
+  if [ "${NUM_EXTRAWORKERS:-0}" -eq 0 ]; then
+    return
+  fi
+
   # write extra worker IP addresses out to file, so that it can be used by 07_agent_add_node.sh
   if [[ "$IP_STACK" = "v6" ]]; then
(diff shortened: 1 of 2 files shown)


agent/05_agent_configure.sh
```diff
@@ -331,13 +331,18 @@ function generate_cluster_manifests() {
...
+    return
+  fi
+
   # write extra worker IP addresses out to file, so that it can be used by 07_agent_add_node.sh
   if [[ "$IP_STACK" = "v6" ]]; then
     extra_workers_ips_space_delimited=$(printf '%s ' "${AGENT_EXTRA_WORKERS_IPSV6[@]}")
   else
     extra_workers_ips_space_delimited=$(printf '%s ' "${AGENT_EXTRA_WORKERS_IPS[@]}")
   fi
-  mkdir -p "${SCRIPTDIR}/${OCP_DIR}/add-node/"
```
elfosardo, line 340: is this dir created somewhere else?

andfasano: Yes, there was an Ansible task [here](https://github.com/openshift-metal3/dev-scripts/blob/7f450ec80ccc6ca43bf8d43012870bc261d1eca4/agent/roles/manifests/tasks/add-nodes.yml#L1), created via the `manifests` role used in the  ansible playbook in the `generate_cluster_manifests`, invoked here in the same step file

pawanpinjarkar (commented):
/lgtm

PR #1731, opened 2025-01-27
OCPEDGE-1192: feat: initial arbiter node addition

eggfoobar:
added support for creating install for clusters with arbiter nodes

In 4.19 of OCP, we will be supporting a deployment with Two Node Openshift with Arbiter on the baremetal platform, this PR adds the arbiter configuration to the metal scripts to help deploy and test arbiter topologies for baremetal.

For more details see EP: https://github.com/openshift/enhancements/pull/1674


zaneb:
No idea about the inspection thing.

Creating a new node type and duplicating all of the config for it seems like a... high-maintenance way to do this. Couldn't we treat the arbiter node as a control-plane node everywhere except when generating the install config?

02_configure_host.sh
```diff
@@ -136,6 +136,7 @@ ansible-playbook \
     -e "baremetal_network_name=${BAREMETAL_NETWORK_NAME}" \
     -e "working_dir=$WORKING_DIR" \
     -e "num_masters=$NUM_MASTERS" \
+    -e "num_arbiters=$NUM_ARBITERS" \
```
dtantsur, line 139: The playbook we're using here comes from https://github.com/metal3-io/metal3-dev-env/ which is an upstream project and thus does not (and can not) know about OCP-specific node types.

eggfoobar: I see, okay. I'm trying to make the changes Zane suggested, that hopefully should get rid of these additions.

eggfoobar: So I had to keep the additional flavor due to needing to alter the size of the arbiter node, but things are a bit more straight forward now.

eggfoobar:
Hey @zaneb and @dtantsur, thanks for the recommendations, I trimmed back all the additional files to only the required parts for deploying the arbiter node. Currently, we don't need much of the other features of the scripts here, so the changes only revolve around altering the install config and adding a new arbiter flavor that will only get triggered when the arbiter is enabled. The `flavor` addition of the `arbiter` is there to help create smaller arbiter nodes in CI since we wish to test the heterogenous deployment.

eggfoobar:
/unhold

The issue I had mentioned earlier has been resolved, no reason to have a hold on this PR anymore.

dtantsur:
/approve

@eggfoobar could you add the new variables to  https://github.com/openshift-metal3/dev-scripts/blob/master/config_example.sh with an explanation how to use them? (e.g. how to set NUM_MASTERS)

eggfoobar:
> /approve
> 
> @eggfoobar could you add the new variables to https://github.com/openshift-metal3/dev-scripts/blob/master/config_example.sh with an explanation how to use them? (e.g. how to set NUM_MASTERS)

Yeah absolutely, updated!

eggfoobar:
@dtantsur This should be all set now, thanks for the input!

ocp_install_env.sh
```diff
@@ -181,6 +181,20 @@ EOF
     fi
 }
 
+function arbiterStanza() {
+    if [[ ! -z "${ENABLE_ARBITER:-}" ]]; then
```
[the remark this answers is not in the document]
eggfoobar: Good point, I want to keep it as is in the code since `ENABLE_ARBITER` in my mind would mean to turn on, updating the doc. Thank you!

PR #1733, opened 2025-03-03
AGENT-1127: Add ability to test operator configuration with ABI

bfournie:
Add a new export - AGENT_OPERATORS - that can take a comma-separated list of operators. The operator list will be stored in a file and pushed to the rendezvous node when it is up so that it can be configured in assisted-service by the agent installer client.

### DIFF
--- a/common.sh
+++ b/common.sh
@@ -409,6 +409,9 @@ export AGENT_BOOT_SERVER_PORT=${AGENT_BOOT_SERVER_PORT:-8089}
 # Enable MCE deployment
 export AGENT_DEPLOY_MCE=${AGENT_DEPLOY_MCE:-}
 
+# Test of agent operators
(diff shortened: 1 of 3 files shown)


common.sh
```diff
@@ -430,26 +433,38 @@ if [[ ! -z ${AGENT_E2E_TEST_SCENARIO} ]]; then
           export NUM_MASTERS=5
           export MASTER_VCPU=4
           export MASTER_DISK=100
+          if [[ "$AGENT_OPERATORS" =~ "mtv" ]]; then
+             export MASTER_VCPU=9
```
pawanpinjarkar, line 437: As `MASTER_VCPU` is always set to `9` irrespective of the cluster topology, it could be moved out of the switch case and overrriden only once at one place.

bfournie: Good point. Fixed.

PR #1734, opened 2025-03-05
OCPEDGE-1645: fix: change node role arbiter logic

eggfoobar:
changed the node role to be applied on index count and not on name, in ci the host name can be anything so we need to trigger based off of the order.

### DIFF
--- a/01_install_requirements.sh
+++ b/01_install_requirements.sh
@@ -85,7 +85,7 @@ case $DISTRO in
     ;;
   "centos9"|"rhel9"|"almalinux9"|"rocky9")
     sudo dnf -y install python3-pip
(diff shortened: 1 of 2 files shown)


eggfoobar:
/hold

Wait for https://github.com/openshift/release/pull/59584 to be merged in to validate the fix.

eggfoobar:
It seems like moving to ocp-4.19 in https://github.com/openshift/release/pull/62465 might require some other updates on this script due to the metal jobs failing. I'm okay with merging this before adding the presubmits since it's a simple enough change.

/unhold

eggfoobar:
Validated the fix locally

elfosardo:
/lgtm
failure not related to this change

PR #1738, opened 2025-03-18
agent: use release 'url' in ApplianceConfig

danielerez:
The appliance now supports using a release image URL: https://github.com/openshift/appliance/pull/352
Hence, we can now use a nightly/ci release url in the ApplianceConfig template.
Also no need for 'imageRegistry' as the appliance now builds the registry internally: https://github.com/openshift/appliance/pull/349

### DIFF
--- a/agent/roles/manifests/templates/appliance-config_yaml.j2
+++ b/agent/roles/manifests/templates/appliance-config_yaml.j2
@@ -2,11 +2,8 @@ apiVersion: v1beta1
 kind: ApplianceConfig
 ocpRelease:
   version: {{ version }}
-  channel: candidate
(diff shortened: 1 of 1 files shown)


agent/roles/manifests/templates/appliance-config_yaml.j2
```diff
@@ -1,12 +1,8 @@
 apiVersion: v1beta1
 kind: ApplianceConfig
 ocpRelease:
-  version: {{ version }}
-  channel: candidate
-  cpuArchitecture: {{ ansible_architecture }}
+  url: {{ image }}
```
pawanpinjarkar, line 4: @danielerez isn't `version` still needed?

PR #1742, opened 2025-04-07
AGENT-378: support for isolated network

andfasano:
This PR adds the support to enable a fully disconnected (baremetal) network


bfournie:
I tested this and it seems to work fine.
Could you add this new field to config_example.sh also?

PR #1743, opened 2025-04-07
AGENT-1108: Gather console screenshots as part of agent-gather

rwsu:
The screenshots can aid in diagnosing issues in agent-tui. Currently they are saved in $OCP_DIR but aren't being archived by agent-gather.

Adds a screenshot after the bad_dns fix is applied to verify the dns configuration was changed correctly.


pawanpinjarkar (commented):
/lgtm

PR #1744, opened 2025-04-08
NO-ISSUE: Code restructure

pawanpinjarkar:
- Moved utility functions ( to be reused for iso-builder) into agent/e2e/agent-tui/utils.sh
- No need to pass CLUSTER_NAME to test-fix-wrong-dns.sh as its a global var

PR #1745, opened 2025-04-10
AGENT-1134: Dev-scripts support to build OVE ISO

pawanpinjarkar:
Dev-scripts support to build OVE ISO 

- Reuse **`make agent`** target (instead of introducing a new one)
- Build agent OVE ISO
- Boot up the required VMs
- Set custom rendezvous node (Use **`AGENT_RENDEZVOUS_NODE_HOSTNAME`**) and auto-configure rendezvousIP via TUI
- Take TUI screenshots during install flow
- Verify that the assisted install web UI is running and accessible
- Increase VCPU allocation for certain operators requiring more resources
- Install required dependencies
- Set isolated network to true for truly disconnected OVE env
- Temporarily skip common steps not relevant for OVE ISO installs
- Temporarily skip generating unused manifests and config files
- Reclaim disk space by cleaning up cache and unnecessary files 
(Use `**AGENT_CLEANUP_ISO_BUILDER_CACHE_LOCAL_DEV**` for **local/dev testing only to save the diskspace**) **Not recommended for CI**
----
To test, set as below

   ```
    export AGENT_E2E_TEST_BOOT_MODE=ISO_NO_REGISTRY
    export AGENT_RENDEZVOUS_NODE_HOSTNAME="${CLUSTER_NAME:-ostest}_master_1"
    export AGENT_CLEANUP_ISO_BUILDER_CACHE_LOCAL_DEV=true

    make agent
```

 ISO will be created at
    `ocp/ostest/iso_builder/<OCP_VERSION>/ove/output/agent-ove.x86_64.iso`

### DIFF
--- a/agent/01_agent_requirements.sh
+++ b/agent/01_agent_requirements.sh
@@ -4,12 +4,22 @@ set -o pipefail
 
 SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )/.." && pwd )"
 
+LOGDIR=${SCRIPTDIR}/logs
+source $SCRIPTDIR/logging.sh
 source $SCRIPTDIR/common.sh
 source $SCRIPTDIR/agent/common.sh
 source $SCRIPTDIR/utils.sh
+source $SCRIPTDIR/ocp_install_env.sh
 source $SCRIPTDIR/validation.sh
 
 early_deploy_validation
+
+function clone_agent_installer_utils() {
+  # Clone repo, if not already present
+  if [[ ! -d $OPENSHIFT_AGENT_INSTALER_UTILS_PATH ]]; then
+    sync_repo_and_patch go/src/github.com/openshift/agent-installer-utils https://github.com/openshift/agent-installer-utils.git
+  fi
+}
 
 if [[ -z ${AGENT_E2E_TEST_SCENARIO} ]]; then
     printf "\nAGENT_E2E_TEST_SCENARIO is missing or empty. Did you forget to set the AGENT_E2E_TEST_SCENARIO env var in the config_<USER>.sh file?"
@@ -49,4 +59,17 @@ fi
 if [[ "${AGENT_E2E_TEST_BOOT_MODE}" == "ISCSI" ]]; then
     # Install shell to administer local storage
     sudo dnf -y install targetcli
+fi
+
+if [[ "${AGENT_E2E_TEST_BOOT_MODE}" == "ISO_NO_REGISTRY" ]]; then
+   sudo dnf -y install xorriso coreos-installer syslinux skopeo
+    
+   early_deploy_validation
+
+   write_pull_secret
+
+   # Extract an updated client tools from the release image
+   extract_oc "${OPENSHIFT_RELEASE_IMAGE}"
+
+   clone_agent_installer_utils
 fi
--- a/agent/06_agent_create_cluster.sh
+++ b/agent/06_agent_create_cluster.sh
@@ -1,5 +1,6 @@
 #!/usr/bin/env bash
 set -euxo pipefail
+shopt -s nocasematch
 
 SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )/.." && pwd )"
 
@@ -12,6 +13,7 @@ source $SCRIPTDIR/validation.sh
 source $SCRIPTDIR/release_info.sh
 source $SCRIPTDIR/agent/common.sh
 source $SCRIPTDIR/agent/iscsi_utils.sh
+source $SCRIPTDIR/agent/e2e/agent-tui/utils.sh
 
 early_deploy_validation
 
@@ -78,6 +80,38 @@ function create_config_image() {
 
     # Copy the auth files to OCP_DIR so wait-for command can access it
     cp -r ${config_image_dir}/auth ${asset_dir}
+}
+
+function create_agent_iso_no_registry() {
+  # Create agent ISO without registry a.k.a. OVE ISO
+  local asset_dir=${1}
+  pushd .
+  cd $OPENSHIFT_AGENT_INSTALER_UTILS_PATH/tools/iso_builder
+  ./hack/build-ove-image.sh --pull-secret-file "${PULL_SECRET_FILE}" --release-image-url "${OPENSHIFT_RELEASE_IMAGE}" --ssh-key-file "${SSH_KEY_FILE}" --dir "${asset_dir}"
+  popd
+}
+
+# Deletes all files and directories under asset_dir
+# example, ocp/ostest/iso_builder/4.19.* 
+# except the final generated ISO file (agent-ove.x86_64.iso), 
+# to free up disk space while preserving the built artifact.
+# Note: This optional cleanup is relevant only when the
+# AGENT_CLEANUP_ISO_BUILDER_CACHE_LOCAL_DEV is set as as true, 
+function cleanup_diskspace_agent_iso_noregistry() {
+ local asset_dir=${1%/}  # Remove trailing slash if present
+
+  # Iterate over all versioned directories matching 4.19.*
+  for dir in "$asset_dir"/4.19.*; do
+    [ -d "$dir" ] || continue
+
+    echo "Cleaning up directory: $dir"
+
+    # Delete all files and symlinks except the agent-ove.x86_64.iso
+    sudo find "$dir" \( -type f -o -type l \) ! -name 'agent-ove.x86_64.iso' -print -delete
+
+    # Remove any empty directories left behind
+    sudo find "$dir" -type d -empty -print -delete
+  done
 }
 
 function set_device_config_image() {
@@ -150,6 +184,52 @@ function attach_appliance_diskimage() {
         # Boot machine from the appliance disk image
         sudo virt-xml ${name} --edit target=sda --disk="boot_order=1" --start
     done
+}
+
+function attach_agent_iso_no_registry() {
+    set_file_acl
+
+    local base_dir=$SCRIPTDIR/$OCP_DIR
+    local iso_name="agent-ove.$(uname -p).iso"
+    local agent_iso_no_registry=$(find "$base_dir" -type f -name "$iso_name" 2>/dev/null | head -n 1)
+
+    for (( n=0; n<${2}; n++ ))
+    do
+        name=${CLUSTER_NAME}_${1}_${n}
+        sudo virt-xml ${name} --add-device --disk "${agent_iso_no_registry}",device=cdrom,target.dev=sdc
+        sudo virt-xml ${name} --edit target=sda --disk="boot_order=1"
+        sudo virt-xml ${name} --edit target=sdc --disk="boot_order=2" --start
+    done
+}
+
+function automate_rendezvousIP_selection(){
+  for (( n=0; n<${2}; n++ ))
+    do
+        name=${CLUSTER_NAME}_${1}_${n}
+        # Take screenshots of console before running the automation that configures the rendezvousIP. 
+        # The screenshot may help us see if agent-tui has reached the expected success state.
+        sudo virsh screenshot $name "${OCP_DIR}/${name}_console_screenshot_before_automation_configures_rendezvousIP.ppm"
+
+        ./agent/e2e/agent-tui/automate-no-registry-agent-tui.sh $name
+
+         # Take screenshot of the console after running the automation that configures the rendezvousIP.
+         sudo virsh screenshot $name "${OCP_DIR}/${name}_console_screenshot_after_automation_configures_rendezvousIP.ppm"
+        echo "Finished configuring the rendezvousIP via agent-tui for $name"
+    done
+}
+
+function check_assisted_install_UI(){
+  local rendezvousIP=$(getRendezvousIP)
+  local url="http://$rendezvousIP:3001"
+  while true; do
+    if curl -s -o /dev/null -w "%{http_code}" "$url" | grep -q "^200$"; then
+      echo "Assisted install UI is up: $url"
+      break
+    else
+      echo "Assisted install UI not ready, retrying in 30 seconds..."
+      sleep 30
+    fi
+  done
 }
 
 function get_node0_ip() {
@@ -435,9 +515,11 @@ function put_operator_file() {
   ssh "${ssh_opts[@]}" "sudo cp /home/core/operators.yaml /etc/assisted/manifests/."
 }
 
-asset_dir="${1:-${OCP_DIR}}"
-config_image_dir="${1:-${OCP_DIR}/configimage}"
-openshift_install="$(realpath "${OCP_DIR}/openshift-install")"
+if [[ "${AGENT_E2E_TEST_BOOT_MODE}" != "ISO_NO_REGISTRY" ]]; then
+  asset_dir="${1:-${OCP_DIR}}"
+  config_image_dir="${1:-${OCP_DIR}/configimage}"
+  openshift_install="$(realpath "${OCP_DIR}/openshift-install")"
+fi
 
 case "${AGENT_E2E_TEST_BOOT_MODE}" in
   "ISO" )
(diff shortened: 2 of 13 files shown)

### CODE REVIEW

File: agent/e2e/agent-tui/utils.sh [hunk 1]
```diff
@@ -0,0 +1,114 @@
+#!/bin/bash
+
+# source common.sh
+set +x
+
+### utils function for sending various keys and/or text to the console
```
pawanpinjarkar, line 6: For better tracking, this PR should be merged after https://github.com/openshift-metal3/dev-scripts/pull/1744 which introduces code refactoring.

agent/08_agent_clone_agent_installer_utils.sh
andfasano, line 1: Step numbering doesn't match the workflow, as it is used before than step 06. The name is also confusing, it's just about preparing the "ISO builder" (the fact that it's store in the agent-installer-utils it's a technical detail).
Also, I'm not sure such few tasks are worth having an entire new separate step, they could easily fit in the `agent_create_cluster` one (within the `create_agent_iso_no_registry` function)

Makefile
```diff
@@ -13,6 +13,12 @@ agent: agent_requirements requirements configure agent_build_installer agent_pre
 # Requires at least 1 extra worker node to be configured with disk size at least 100Gß
 agent_plus_add_node: agent agent_add_extraworker_nodes
 
+# Create an agent (OVE) ISO that includes agent-tui, assisted-install-ui, and release images,
+# enabling cluster installation in disconnected environments without requiring a registry.
+# TODO: Enhance agent_create_cluster to programmatically interact with the UI to initiate cluster installation.
+# Currently, agent_create_cluster boots the agent OVE ISO, configures the rendezvousIP and verifies if UI is running.
+agent_iso_no_registry: agent_requirements requirements configure agent_clone_agent_installer_utils agent_create_cluster
```
andfasano, line 20: Any specific reason for skipping the `agent_configure` step? Apart for the extraworker, it's also required for enabling the isolated network

agent/06_agent_create_cluster.sh
```diff
@@ -152,6 +162,40 @@ function attach_appliance_diskimage() {
...
+    local agent_iso_no_registry=$(find "$base_dir" -type f -name "$iso_name" 2>/dev/null | head -n 1)
+
+    for (( n=0; n<${2}; n++ ))
+    do
+        name=${CLUSTER_NAME}_${1}_${n}
+        sudo virt-xml ${name} --add-device --disk "${agent_iso_no_registry}",device=cdrom,target.dev=sdc
+        sudo virt-xml ${name} --edit target=sda --disk="boot_order=1"
+        sudo virt-xml ${name} --edit target=sdc --disk="boot_order=2" --start
+        echo "Waiting for 3 mins to arrive at agent-tui screen"
+        sleep 180
```
andfasano, line 179: I don't think it's necessary to wait 3 minutes for each node, it should be quicker to start all of them and _then_ wait for no more than 3 minutes in total

agent/06_agent_create_cluster.sh
```diff
@@ -499,6 +545,23 @@ case "${AGENT_E2E_TEST_BOOT_MODE}" in
...
+    mkdir -p ${asset_dir}
+    create_agent_iso_no_registry ${asset_dir}
+
+    attach_agent_iso_no_registry master $NUM_MASTERS
+    attach_agent_iso_no_registry worker $NUM_WORKERS
+
+    # In case of SNO, wait for additional 3 minutes for the node
+    # to finish booting up and make sure UI is up
+    if [[ "${NUM_MASTERS}" == "1" ]]; then
+      sleep 180
```
andfasano, line 561: Why this additional requirement?

agent/06_agent_create_cluster.sh
```diff
@@ -528,17 +591,24 @@ if [[ ! -z $AGENT_OPERATORS ]]; then
     put_operator_file
 fi
 
-wait_for_cluster_ready
+# Current goal is to only verify if the nodes are booted fine,
+# TUI sets the rendezvous IP correctly and UI is accessible.
+# The next goal is to simulate adding the cluster details via UI
+# and complete the cluster installation.
+if [[ "${AGENT_E2E_TEST_BOOT_MODE}" != "ISO_NO_REGISTRY" ]]; then
+    wait_for_cluster_ready
```
andfasano, line 599: maybe an exit 0 would be sufficient to temporary skip those parts not yet applicable for the `ISO_NO_REGISTRY`

agent/e2e/agent-tui/test-no-registry-agent-tui.sh
andfasano, line 1: This is not a test, it's the automation of configuring the rendezvous IP

agent/e2e/agent-tui/test-no-registry-agent-tui.sh
```diff
@@ -0,0 +1,30 @@
...
+source $SCRIPTDIR/agent/e2e/agent-tui/utils.sh
+
+set +x
+shopt -s nocasematch
+
+node_name=$1
+
+# The following commands sends key presses through "virsh send-key" to interact
+# with agent-tui
+if [[ "$node_name" == "${CLUSTER_NAME}_master_0" ]]; then
```
andfasano, line 14: This works, but I hoped for a more robust way to identify the rendezvous node (given that is an arbitrary decision, it could be stored in a RENDEZVOUS_NODE_HOSTNAME for example)

agent/e2e/agent-tui/utils.sh
```diff
@@ -73,4 +81,33 @@ function pressKeys(){
     local keyCode="KEY_"$c
     _pressKey $keyCode
   done
-}
\ No newline at end of file
+}
+
+function typeIPAddress() {
```
pawanpinjarkar, line 86: Rethinking, maybe this is not needed as the existing `pressKeys()` could be used

agent/01_agent_requirements.sh
```diff
@@ -50,3 +50,7 @@ if [[ "${AGENT_E2E_TEST_BOOT_MODE}" == "ISCSI" ]]; then
     # Install shell to administer local storage
     sudo dnf -y install targetcli
 fi
+
+if [[ "${AGENT_E2E_TEST_BOOT_MODE}" == "ISO_NO_REGISTRY" ]]; then
+    sudo dnf -y install xorriso coreos-installer
```
pawanpinjarkar, line 55: Also needs isohybrid

config_example.sh
```diff
@@ -794,7 +794,9 @@ set -x
 # This config variable is used only by the agent based installer and is optional.
 # The default value for AGENT_E2E_TEST_BOOT_MODE is 'ISO'.
 # For the openshift-appliance flow, set the boot mode to 'DISKIMAGE'.
-# AGENT_E2E_TEST_BOOT_MODE=PXE
+# To generate PXE assets, set the boot mode to 'PXE'.
+# To generate an ISO without registry a.k.a. OVE ISO, set the boot mode to 'ISO_NO_REGISTRY'.
```
rwsu, line 798:
```suggestion
# To generate an ISO that can be used in a disconnected environment without using a registry a.k.a. OVE ISO, set the boot mode to 'ISO_NO_REGISTRY'.
```

agent/06_agent_create_cluster.sh
```diff
@@ -12,6 +12,7 @@ source $SCRIPTDIR/validation.sh
 source $SCRIPTDIR/release_info.sh
 source $SCRIPTDIR/agent/common.sh
 source $SCRIPTDIR/agent/iscsi_utils.sh
+source $SCRIPTDIR/agent/e2e/agent-tui/utils.sh
```
rwsu, line 15: In run_agent_test_cases(), bad_dns test case should not run if AGENT_E2E_TEST_BOOT_MODE=ISO_NO_REGISTRY.

rwsu (changes requested):

agent/e2e/agent-tui/utils.sh
rwsu: I don't think it is needed.

agent/08_agent_clone_agent_installer_utils.sh
```diff
@@ -0,0 +1,28 @@
...
+SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )/.." && pwd )"
+
+LOGDIR=${SCRIPTDIR}/logs
+source $SCRIPTDIR/logging.sh
+source common.sh
+source utils.sh
+source ocp_install_env.sh
+source validation.sh
+
+function clone_agent_installer_utils() {
```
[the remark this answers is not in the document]
pawanpinjarkar: yes, file 08 is not needed.

Makefile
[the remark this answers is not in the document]
pawanpinjarkar: May be keeping a new target helps as the `make agent` target has some steps which are not relevant for OVE such as `agent_build_installer`

agent/06_agent_create_cluster.sh
pawanpinjarkar: If we don't wait for the agent TUI to show up and finish the automated steps, each VM just boots up and all of them waits at the first screen of TUI to enter rendezvousIP and then code execution just proceeds to the point where it expects the assisted install UI to be up but in reality, code execution never gets a chance to set rendezvousIP.

agent/06_agent_create_cluster.sh
pawanpinjarkar: Right, its not needed. `check_assisted_install_UI` wait for the UI to be available anyway.

agent/06_agent_create_cluster.sh
pawanpinjarkar: yes, we will do `exit 0` as soon as assiseted install UI is up and none of the other actions/tests needs to be done for OVE case for now

agent/06_agent_create_cluster.sh
andfasano: There's no need to serialize them, and to wait NUM_MASTERS * NUM_WORKERS * 3 minutes (in the compact case 9 additional minutes). So:

* Start all the instances simultaneously
* Wait 3 minutes
* Loop over all the instances and set the agent TUI accordingly on the role

Makefile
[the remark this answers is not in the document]
andfasano: I'm not sure they could become relevant in future. I'm fine for now to proceed with a new target, and eventually review it later (to be merged with the `agent` one), if it will not fit properly

andfasano:
@pawanpinjarkar did you also maybe check the `clean` target with this new boot mode?

pawanpinjarkar:
> @pawanpinjarkar did you also maybe check the `clean` target with this new boot mode?

@andfasano yes, I am cleaning the directory here https://github.com/openshift-metal3/dev-scripts/pull/1745/files#diff-414fad1e37343146b4f215663b307230240277cd0043033465cd5e3360d1eec7R32-R35

Makefile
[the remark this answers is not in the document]
pawanpinjarkar: right

agent/06_agent_create_cluster.sh
```diff
@@ -80,6 +81,15 @@ function create_config_image() {
     cp -r ${config_image_dir}/auth ${asset_dir}
 }
 
+function create_agent_iso_no_registry() {
+  # Create agent ISO without registry a.k.a. OVE ISO
+  local asset_dir=${1}
+  pushd .
+  cd $OPENSHIFT_AGENT_INSTALER_UTILS_PATH/tools/iso_builder
+  ./hack/build-ove-image.sh --pull-secret-file "${PULL_SECRET_FILE}" --release-image-url "${OPENSHIFT_RELEASE_IMAGE}" --dir "${asset_dir}"
```
rwsu, line 89: Could we add this to enable ssh'ing into the nodes?
```suggestion
  ./hack/build-ove-image.sh --pull-secret-file "${PULL_SECRET_FILE}" --release-image-url "${OPENSHIFT_RELEASE_IMAGE}" --dir "${asset_dir}" --ssh-key-file ~/.ssh/id_rsa.pub
```

agent/06_agent_create_cluster.sh
```diff
@@ -152,6 +162,45 @@ function attach_appliance_diskimage() {
...
+        sudo virt-xml ${name} --edit target=sda --disk="boot_order=1"
+        sudo virt-xml ${name} --edit target=sdc --disk="boot_order=2" --start
+    done
+}
+
+function automate_rendezvousIP_selection(){
+  for (( n=0; n<${2}; n++ ))
+    do
+        name=${CLUSTER_NAME}_${1}_${n}
+        ./agent/e2e/agent-tui/automate-no-registry-agent-tui.sh $name
```
rwsu, line 185: We should take console screenshots before and after the script.
Example:
 [     sudo virsh screenshot $name "${OCP_DIR}/${name}_console_screenshot_before_dns_fix.ppm" ](https://github.com/openshift-metal3/dev-scripts/blob/39a6bd2cad1d811b090c1e78a08b35d8cd03962e/agent/06_agent_create_cluster.sh#L296C4-L296C91)

rwsu (changes requested):

agent/01_agent_requirements.sh
```diff
@@ -50,3 +60,16 @@ if [[ "${AGENT_E2E_TEST_BOOT_MODE}" == "ISCSI" ]]; then
     # Install shell to administer local storage
     sudo dnf -y install targetcli
 fi
+
+if [[ "${AGENT_E2E_TEST_BOOT_MODE}" == "ISO_NO_REGISTRY" ]]; then
+   sudo dnf -y install xorriso coreos-installer syslinux
```
bfournie, line 65: Since iso_builder requires `skopeo` this should also be included, otherwise get:
`/home/stack/go/src/github.com/openshift/agent-installer-utils/tools/iso_builder/hack/helper.sh: line 90: skopeo: command not found`

agent/06_agent_create_cluster.sh
```diff
@@ -528,14 +604,14 @@ if [[ ! -z $AGENT_OPERATORS ]]; then
     put_operator_file
 fi
 
+
 wait_for_cluster_ready
```
bfournie, line 608: Seems that a changed is needed [here](https://github.com/openshift-metal3/dev-scripts/blob/master/agent/06_agent_create_cluster.sh#L219) since the `openshift-install` path isn't created. I'm getting
```
Assisted install UI is up: http://192.168.111.20:3001
./agent/06_agent_create_cluster.sh: line 273: /home/stack/dev-scripts-bfournie-ove/ocp/ostest/openshift-install: No such file or directory
```

pawanpinjarkar: The code execution should skip after the UI is up and the error you saw was due to this [typo](https://github.com/openshift-metal3/dev-scripts/pull/1745/files#diff-51babdd488dfb3c638b47aaeb87403b34eb32bed9422adf3632ac9372460ebc7R572) This will be fixed in upcoming rebase

agent/cleanup.sh
```diff
@@ -29,6 +29,10 @@ if [[ "${AGENT_E2E_TEST_BOOT_MODE}" == "DISKIMAGE" ]]; then
     sudo podman rmi -f ${APPLIANCE_IMAGE} || true
 fi
 
+if [[ "${AGENT_E2E_TEST_BOOT_MODE}" == "ISO_NO_REGISTRY" ]]; then
+    sudo rm -rf "${OCP_DIR}/iso_builder"
```
bfournie, line 33: I found that `make clean` needs to be run twice because of this directory. The first time its run ocp/ostest isn't fully removed [here](https://github.com/openshift-metal3/dev-scripts/blob/master/ocp_cleanup.sh#L18) because its not run as sudo, so when subsequently running `make agent` we get a failure [here]( https://github.com/openshift-metal3/dev-scripts/blob/master/sanitychecks.sh#L10).

Probably should also remove ${OCP_DIR} here since it will still exist.

pawanpinjarkar: nice. Thanks @bfournie

pawanpinjarkar: You won't not get this issue if you set `AGENT_CLEANUP_ISO_BUILDER_CACHE_LOCAL_DEV` to `true`as it then removes the appliance cache and all other files except the generated OVE ISO to save the disk space.

ocp_cleanup.sh
```diff
@@ -15,7 +15,11 @@ fi
 if [ -d ${OCP_DIR} ]; then
     ${OCP_DIR}/openshift-install --dir ${OCP_DIR} --log-level=debug destroy bootstrap
     ${OCP_DIR}/openshift-install --dir ${OCP_DIR} --log-level=debug destroy cluster
-    rm -rf ${OCP_DIR}
+    if [[ "${AGENT_E2E_TEST_BOOT_MODE}" == "ISO_NO_REGISTRY" ]]; then
+      sudo rm -rf "${OCP_DIR}"
+    else
+      rm -rf ${OCP_DIR}
```
rwsu, line 21: Wouldn't it be easier to always do `sudo rm -rf "${OCP_DIR}"`?

pawanpinjarkar: Good point — since this cleanup block runs in all contexts and the goal is simply to ensure ${OCP_DIR} is removed reliably, using sudo unconditionally here actually makes the logic simpler and more robust.

PR #1747, opened 2025-04-14
Pin collections to latest ansible 2.15 compatible versions

lmiccini:
ansible.utils just released a new version that is incompatible with ansible-2.15 (ansible-collections/ansible.utils@7059201) and this results in:

```
    TASK [common : Show networks data for debugging (common role)] *****************
    task path: /home/dev-scripts/metal3-dev-env/vm-setup/roles/common/tasks/main.yml:42
    [WARNING]: Collection ansible.netcommon does not support Ansible version
    2.15.12
    [WARNING]: Collection ansible.utils does not support Ansible version 2.15.12
    redirecting (type: filter) ansible.builtin.nthhost to ansible.netcommon.nthhost
    redirecting (type: filter) ansible.builtin.nthhost to ansible.netcommon.nthhost
```

this is a proposal to pin these collections until we can upgrade to 2.16.

### DIFF
--- a/01_install_requirements.sh
+++ b/01_install_requirements.sh
@@ -135,7 +135,9 @@ sudo python -m pip install ansible=="${ANSIBLE_VERSION}"
 
 pushd ${METAL3_DEV_ENV_PATH}
 ansible-galaxy install -r vm-setup/requirements.yml
-ansible-galaxy collection install --upgrade ansible.netcommon ansible.posix ansible.utils community.general
+# Let's temporarily pin these collections to the latest compatible with ansible-2.15
+#ansible-galaxy collection install --upgrade ansible.netcommon ansible.posix ansible.utils community.general
+ansible-galaxy collection install 'ansible.netcommon<=7.2.0' ansible.posix 'ansible.utils<=5.1.2' community.general
 ANSIBLE_FORCE_COLOR=true ansible-playbook \
   -e "working_dir=$WORKING_DIR" \
   -e "virthost=$HOSTNAME" \


elfosardo:
upstream workaround https://github.com/metal3-io/metal3-dev-env/pull/1517

01_install_requirements.sh
```diff
@@ -135,7 +135,9 @@ sudo python -m pip install ansible=="${ANSIBLE_VERSION}"
 
 pushd ${METAL3_DEV_ENV_PATH}
 ansible-galaxy install -r vm-setup/requirements.yml
-ansible-galaxy collection install --upgrade ansible.netcommon ansible.posix ansible.utils community.general
+# Let's temporarily pin these collections to the latest compatible with ansible-2.15
+#ansible-galaxy collection install --upgrade ansible.netcommon ansible.posix ansible.utils community.general
+ansible-galaxy collection install 'ansible.netcommon<=7.2.0' ansible.posix 'ansible.utils<=5.1.2' community.general
```
dtantsur, line 140: Did they actually broke the compatibility in 5.1.3? If not, I think we need to allow for patch releases and do `<7.3.0` and `<5.2.0` instead. WDYT?

lmiccini: We could do ansible.utils<6.0.0  and ansible.netcommon<8.0.0 as those are the ones bringing the breaking changes. I just went with the latest released (and tested) versions to be on the safe side.

dtantsur: My worry is about missing future bug fixes. Would you consider a follow-up once we merge this and unblock the CI?

PR #1749, opened 2025-04-14
Relax pinned versions to use breaking ones as upper limits

lmiccini:
Let's use ansible.utils<6.0.0 and ansible.netcommon<8.0.0 as ceiling so we can benefit from minor releases / bug fixes.


elfosardo:
/retest
all tests look good, dualstack failed because of failed mirroring

PR #1750, opened 2025-04-15
fix for the incorrectly formatted

zhaozhanqi:
base on https://github.com/openshift-metal3/dev-scripts/pull/1747,  Found the following error 

```
ansible-galaxy collection install 'ansible.netcommon<=7.2.0' ansible.posix 'ansible.utils<=5.1.2' community.general

ERROR! Neither the collection requirement entry key 'name', nor 'source' point to a concrete resolvable collection artifact. Also 'name' is not an FQCN. A valid collection name must be in the format <namespace>.<collection>. Please make sure that the namespace and the collection name contain characters from [a-zA-Z0-9_] only.
```


zhaozhanqi:
> I'm not sure where you're seeing this as it correctly passes our CI (and has been tested locally too) please also this looks obsolete, rebase your repo

this issue happens on rhel8,  sounds like rhel9 works well.

elfosardo:
yeah, we don't really use rhel8/centos8 anymore

PR #1752, opened 2025-04-25
OCPBUGS-55386: Update oc-mirror to use -v2

bfournie:
Use version v2 of oc-mirror. This version has many improvements over v1 but the primary reason for using v2 is that it allows nightly or CI builds to be tested via the release field in the ImageSetConfiguration.

Note that oc-mirror is not currently used in any CI tests and has strictly been used for local testing with the agent-based installer. As a result its not necessary to keep backwards compatibility with v1.

It's expected that oc-mirror will eventually be used for setting up a mirror for disconnected testing instead of 'oc adm release mirror' since its the most common way to set up mirrors by our customers; 'oc adm release mirror' does not provide for installing additional operators.


oc_mirror.sh
```diff
@@ -78,33 +78,21 @@ function create_file_imageset() {
...
   platform:
-    architectures:
-      - "amd64"
-    channels:
-      - name: candidate-${OPENSHIFT_RELEASE_STREAM}
-        minVersion: $latest_release
-        maxVersion: $latest_release
-        type: ocp
+    graph: true
+    release: $OPENSHIFT_RELEASE_IMAGE
```
rwsu, line 87: I think a similar change need to be made for create_registry_imageset.

bfournie: As discussed, that's for mirror-to-mirror publish which is only valid for v1. I will remove it.

PR #1755, opened 2025-04-30
AGENT-1194: Fix premature oc extraction

pawanpinjarkar:
The `01_agent_requirements.sh` tries to extract `oc` binary much before its installed via `01_install_requirements.sh` causing the CI job to fail with `oc: command not found` error. This worked locally because we have `oc` installed.

Also, moved git clone agent-installer-utils into a more logical place `agent/06_agent_create_cluster.sh`

### DIFF
--- a/agent/01_agent_requirements.sh
+++ b/agent/01_agent_requirements.sh
@@ -14,13 +14,6 @@ source $SCRIPTDIR/validation.sh
 
 early_deploy_validation
 
-function clone_agent_installer_utils() {
-  # Clone repo, if not already present
-  if [[ ! -d $OPENSHIFT_AGENT_INSTALER_UTILS_PATH ]]; then
-    sync_repo_and_patch go/src/github.com/openshift/agent-installer-utils https://github.com/openshift/agent-installer-utils.git
-  fi
-}
-
 if [[ -z ${AGENT_E2E_TEST_SCENARIO} ]]; then
     printf "\nAGENT_E2E_TEST_SCENARIO is missing or empty. Did you forget to set the AGENT_E2E_TEST_SCENARIO env var in the config_<USER>.sh file?"
     invalidAgentValue
@@ -63,6 +56,4 @@ fi
 
 if [[ "${AGENT_E2E_TEST_BOOT_MODE}" == "ISO_NO_REGISTRY" ]]; then
    sudo dnf -y install xorriso coreos-installer syslinux skopeo
-
-   clone_agent_installer_utils
 fi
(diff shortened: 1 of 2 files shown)


agent/01_agent_requirements.sh
```diff
@@ -66,8 +66,5 @@ if [[ "${AGENT_E2E_TEST_BOOT_MODE}" == "ISO_NO_REGISTRY" ]]; then
     
    early_deploy_validation
 
-   # Extract an updated client tools from the release image
-   extract_oc "${OPENSHIFT_RELEASE_IMAGE}"
```
andfasano, line 70: It does not seem correct to me, since an initial version of `oc` is downloaded in the very first step https://github.com/openshift-metal3/dev-scripts/blob/29bb68793c0825229d9757c135fa090ab4572436/01_install_requirements.sh#L172

agent/01_agent_requirements.sh
```diff
@@ -63,8 +63,6 @@ fi
 
 if [[ "${AGENT_E2E_TEST_BOOT_MODE}" == "ISO_NO_REGISTRY" ]]; then
    sudo dnf -y install xorriso coreos-installer syslinux skopeo
-    
-   early_deploy_validation
```
pawanpinjarkar, line 67: Already called from https://github.com/openshift-metal3/dev-scripts/pull/1755/commits/f11cb0d2cfc437a6b21822da280664f6ebe5d943#diff-054af3cea588fc3c166a66dbc56c926ab31695db4f6d47d2af638f759303106dR15

bfournie: Seems reasonable to remove this duplicate call.

agent/01_agent_requirements.sh
andfasano: Ah, I see. I think the 01_agent_requirements should be used only for basic installations of dependency strictly required by ABI (and probably also the iso builder checkout could be moved later).

Btw, could you please open a proper JIRA task for this issue (AGENT is ok)? Otherwise it could become diffcult to track all the changes, ie NO-ISSUE IMHO should be used only in very rare cases, thanks.

pawanpinjarkar: Whats the next best target for checking out the iso builder code? I don't want to creare a new target which only clones the code. Can't be embeded in 03_build_installer.sh either.

andfasano: IMHO it's in `06_agent_create_cluster.sh` step. This is the step where we usually create the ISO to be attached to the VMs. In this case we're simply following a different approach, but still the goal it's to generate an ISO

pawanpinjarkar: Done

PR #1757, opened 2025-05-05
OCPEDGE-1891: Add redfish_verify_ca support

slintes:
This is a follow up to https://github.com/metal3-io/metal3-dev-env/pull/1528

For TNF we need to have a SSL secured redfish endpoint. The PR mentioned above adds support for creation of private key and certificate, configuration of sushy, and configuration of ironic's redfish driver.

This PR adds setting certificate verification for the BMH, based on the redfish_verify_ca option in ironic nodes json.


bfournie:
Note that this line https://github.com/openshift-metal3/dev-scripts/pull/1757/files#diff-c5e795042845913905126f98aa1813bf9b03fab8f94b38a39ea52b91be930000R84 has broken the agent-based installer when using BMC configuration, i.e. when 
`AGENT_BM_HOSTS_IN_INSTALL_CONFIG=true`

It causes the generated install-config.yaml to have improper formatting
 ```
    bmc:
        address: redfish+https://192.168.111.1:8000/redfish/v1/Systems/b155087e-ec87-4e31-8417-6e5e2ae79e51
        username: admin
        password: password
        disableCertificateVerification: true      networkConfig:
```

PR #1758, opened 2025-05-06
AGENT-1196: OVE create dummy files

pawanpinjarkar:
Creates dummy kubeconfig and kubeadmin password file in case of ISO_NO_REGISTRY to make CI happy.


rwsu (commented):
/lgtm

PR #1761, opened 2025-05-07
appliance: copy disk images to WORKING_DIR

danielerez:
Copy appliance.raw files to /opt/dev-scripts in order to avoid storage exhaustion.

### DIFF
--- a/agent/06_agent_create_cluster.sh
+++ b/agent/06_agent_create_cluster.sh
@@ -170,7 +170,7 @@ function attach_appliance_diskimage() {
     set_file_acl
 
     local config_image_drive="sdd"
-    local appliance_disk_image="${OCP_DIR}/appliance.raw"
+    local appliance_disk_image="${WORKING_DIR}/appliance.raw"
 
     for (( n=0; n<${2}; n++ ))
     do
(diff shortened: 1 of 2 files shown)


agent/06_agent_create_cluster.sh
```diff
@@ -170,7 +170,7 @@ function attach_appliance_diskimage() {
     set_file_acl
 
     local config_image_drive="sdd"
-    local appliance_disk_image="${OCP_DIR}/appliance.raw"
+    local appliance_disk_image="${WORKING_DIR}/appliance.raw"
```
andfasano, line 173: Usually for transient artifacts $OCP_DIR it's a better choice, and it also gets cleanead up. $WORKING_DIR may be used to store stuff across different runs.

danielerez: Yeah, but any other alternative? The 200GB in OCP_DIR is not enough for 3 disk images :(

andfasano: Maybe I'm wrong, but isn't that more an hw related issue (ie how the disks are mounted)? How would you be sure that $WORKING_DIR will always have the sufficient storage? Not a direct solution, but a preliminary sanity check could fail fast the build (we did something similar in the past https://github.com/openshift-metal3/dev-scripts/blob/8261816c228722eac8d69adf0bb0fcbbdb104c0a/sanitychecks.sh#L32)

danielerez: Good idea, added.
Also pushed a PR for release (so we could rehearse the test): https://github.com/openshift/release/pull/64651

PR #1762, opened 2025-05-08
Don't restart NetworkManager unnecessarily

cybertron:
Restarting NM breaks VPN connections, which are required to build images for some OpenShift projects, notably machine-config-operator. Forcing a NetworkManager restart every time we update the node, even when no changes to the NetworkManager package were made, causes problems when working on patches to MCO.

This change checks the NM version before and after the dnf update and only restarts NM if the version changed.


dtantsur:
/lgtm

I no longer remember why we even restart it...

PR #1766, opened 2025-05-21
AGENT-1211: Validate and Enforce Expected Size Limit for OVE ISO

pawanpinjarkar:
With 4.19 tech preview, the expected ISO size is approximately 36GB. Fail, if the generated ISO size is larger.


agent/06_agent_create_cluster.sh
```diff
@@ -95,6 +95,24 @@ function create_agent_iso_no_registry() {
...
 }
 
+function assert_agent_no_registry_iso_size(){
+  base_dir=$SCRIPTDIR/$OCP_DIR
+  iso_name="agent-ove.$(uname -p).iso"
+  agent_iso_no_registry=$(find "$base_dir" -type f -name "$iso_name" 2>/dev/null | head -n 1)
+  iso_size=$(stat -c%s "$agent_iso_no_registry")
+  
+  # With 4.19 tech preview, the expected ISO size is approximately 36GB
+  iso_threshold=36
```
andfasano, line 105: Since this is just a basic sanity check, it won't hurt to support some minor adjustments, so I'd set the threshold size to a value slightly bigger, ie 40Gb

agent/06_agent_create_cluster.sh
```diff
@@ -95,6 +95,24 @@ function create_agent_iso_no_registry() {
   popd
 }
 
+function assert_agent_no_registry_iso_size(){
+  base_dir=$SCRIPTDIR/$OCP_DIR
+  iso_name="agent-ove.$(uname -p).iso"
```
andfasano: given than this info is already required in `attach_agent_iso_no_registry`, I'd suggest to calculate it directly in `attach_agent_iso_no_registry` (in case of success) and save it into an env var, so that it could be consumed later by the functions requiring it

agent/06_agent_create_cluster.sh
```diff
@@ -95,6 +95,24 @@ function create_agent_iso_no_registry() {
...
+  agent_iso_no_registry=$(find "$base_dir" -type f -name "$iso_name" 2>/dev/null | head -n 1)
+  iso_size=$(stat -c%s "$agent_iso_no_registry")
+  
+  # With 4.19 tech preview, the expected ISO size is approximately 36GB
+  iso_threshold=40
```
andfasano: I'd expect we'll bump this value from time to time, so maybe it could be more convenient to store it in a config var with a default value (easier to maintain)

agent/06_agent_create_cluster.sh
```diff
@@ -95,6 +95,24 @@ function create_agent_iso_no_registry() {
...
+  iso_size_limit=$((iso_threshold * 1024 * 1024 * 1024))
+
+  if (( iso_size > iso_size_limit )); then
+    echo "Error: OVE ISO size of $agent_iso_no_registry exceeds ${iso_threshold}GB"
+    exit 1
+  else
+    echo "OVE ISO size is within the acceptable limit."
```
andfasano: I'd remove it this log trace in case of success

common.sh
```diff
@@ -405,6 +405,7 @@ export NETWORKING_MODE=${NETWORKING_MODE:-}
 export AGENT_E2E_TEST_BOOT_MODE=${AGENT_E2E_TEST_BOOT_MODE:-"ISO"}
 export AGENT_CLEANUP_ISO_BUILDER_CACHE_LOCAL_DEV=${AGENT_CLEANUP_ISO_BUILDER_CACHE_LOCAL_DEV:-"false"}
 export AGENT_RENDEZVOUS_NODE_HOSTNAME=${AGENT_RENDEZVOUS_NODE_HOSTNAME:-${CLUSTER_NAME}_master_0}
+export AGENT_OVE_ISO_SIZE==${AGENT_OVE_ISO_SIZE:-40}
```
bfournie:
```suggestion
export AGENT_OVE_ISO_SIZE=${AGENT_OVE_ISO_SIZE:-40}
```

agent/06_agent_create_cluster.sh
```diff
@@ -95,6 +95,19 @@ function create_agent_iso_no_registry() {
...
 
+function assert_agent_no_registry_iso_size(){
+  agent_iso_no_registry=$(get_agent_iso_no_registry)
+  iso_size=$(stat -c%s "$agent_iso_no_registry")
+  
+  # With 4.19 tech preview, the expected ISO size is approximately 36GB
+  iso_size_limit=$(($AGENT_OVE_ISO_SIZE * 1024 * 1024 * 1024))
+
+  if (( iso_size > iso_size_limit )); then
+    echo "Error: OVE ISO size of $agent_iso_no_registry exceeds ${AGENT_OVE_ISO_SIZE}GB"
```
rwsu: Let's have the message also print out the iso_size so that we can see by how much it is over.

rwsu (changes requested):

rwsu (commented):
/lgtm

PR #1768, opened 2025-05-27
NO-ISSUE: Take screenshot after each console key press

rwsu:
Add additional screenshots to help with debugging issues with console interaction in CI tests.

### DIFF
--- a/agent/e2e/agent-tui/test-fix-wrong-dns.sh
+++ b/agent/e2e/agent-tui/test-fix-wrong-dns.sh
@@ -65,9 +65,8 @@ pressDown "Select Quit" ${numStepsToQuit}
 pressEnter "Select 'Quit' menu item"
 pressEsc "Esc from network tree view" 2
 sleep 3
-pressTab "Goto <Quit> button to exit agent-tui" 1
 
 # wait for check to update, to visually see the release image
 # check change to passing.
 sleep 10
-pressEnter "Exit agent-tui"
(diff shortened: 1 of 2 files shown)


rwsu:
ansible galaxy issue
/retest-required

rwsu:
With https://github.com/openshift/agent-installer-utils/pull/101 merged, the blank screen still shows after exiting nmtui.

The retest passed so it appears to be an intermittent issue. I was able to figure out what may be causing the previous failure. After pressing ESC, a TAB key press afterwards sometimes does not register. So the cursor stays on <Configure Network>, instead of intended target <Quit>. The next key press is ENTER, so the agent-tui goes back into nmtui and that is what we see in the last screenshot. Disabling the ESC key in the checks page fixes this problem.

But the core issue is the blank screen after exiting nmtui. What is causing that is unclear.

rwsu:
https://github.com/openshift/agent-installer-utils/pull/102 merged. 
/test e2e-agent-bad-dns

rwsu:
The tab key didn't register in the last job. The CI release image should have contained https://github.com/openshift/agent-installer-utils/pull/102. Retesting.
/test e2e-agent-bad-dns

rwsu:
bad-dns job passed. Testing a few more times.
/test e2e-agent-bad-dns

rwsu:
bad-dns job passed again. Let's test one more time.
/test e2e-agent-bad-dns

PR #1771, opened 2025-06-11
NO-ISSUE: Renable xtrace for agent/06_agent_create_cluster.sh

rwsu:
xtrace was inadvertently disabled in
https://github.com/openshift-metal3/dev-scripts/pull/1745

This patch moves the getRendezvousIP function out of e2e/agent-tui/utils.sh and into common.sh.

e2e/agent-tui/utils.sh contains "set +x" which disables xtrace.

06_agent_create_cluster.sh no longer sources e2e/agent-tui/util.sh so that it doesn't inherit the "set +x" call.


andfasano:
/lgtm

btw, is there any other case where it should be re-enabled?

PR #1773, opened 2025-06-13
OCPEDGE-1932: feat: support abi with arbiter node

eggfoobar:
Add support for agent based arbiter install in the dev scripts. This also adds the `TWONODE` scenario which should capture both two node arbiter and two node fencing deployments.


common.sh
```diff
@@ -456,6 +456,13 @@ if [[ ! -z ${AGENT_E2E_TEST_SCENARIO} ]]; then
           export MASTER_MEMORY=32768
           export NUM_WORKERS=0
           ;;
+      "TWONODE" )
```
bfournie: Please add the new defines to config_example.sh

agent/05_agent_configure.sh
```diff
@@ -63,6 +63,10 @@ function configure_node() {
     local hostname="$(printf "${WORKER_HOSTNAME_FORMAT}" "${node_num}")"
     local ip=$((base_ip + ${NUM_MASTERS} + node_num))
   fi
+  if [[ "$node_type" == "arbiter" ]]; then
+    local hostname="$(printf "arbiter-%d" "${node_num}")"
```
rwsu: To be consistent, I think a hostname format should be created here.
```suggestion
    local hostname="$(printf "${ARBITER_HOSTNAME_FORMAT}" "${node_num}")"
```

common.sh
rwsu: Would it be better to call this TNA? There will be other two node configurations coming in the future for TNF.

I would also add a notion of NUM_ARBITERS and the following here:
```
export ENABLE_ARBITER="true"
export NUM_ARBITERS=1
export ARBITER_VCPU=2
export ARBITER_MEMORY=16384
export ARBITER_DISK=100
```

agent/06_agent_create_cluster.sh
```diff
@@ -543,6 +543,9 @@ case "${AGENT_E2E_TEST_BOOT_MODE}" in
 
     attach_agent_iso master $NUM_MASTERS
     attach_agent_iso worker $NUM_WORKERS
+    if [[ ! -z "${ENABLE_ARBITER:-}" ]]; then
```
rwsu: Instead have
attach_agent_iso arbiter $NUM_ARBITERS

common.sh
```diff
@@ -395,7 +395,7 @@ export AGENT_WAIT_FOR_INSTALL_COMPLETE=${AGENT_WAIT_FOR_INSTALL_COMPLETE:-true}
 # Agent specific configuration 
 
 function invalidAgentValue() {
-  printf "Found invalid value \"$AGENT_E2E_TEST_SCENARIO\" for AGENT_E2E_TEST_SCENARIO. Supported values: 'COMPACT_IPXX', 'HA_IPXX', 'SNO_IPXX', '4CONTROL_IPXX', or '5CONTROL_IPXX', where XX is 'V4', 'V6', or 'V4V6'"
+  printf "Found invalid value \"$AGENT_E2E_TEST_SCENARIO\" for AGENT_E2E_TEST_SCENARIO. Supported values: 'COMPACT_IPXX', 'HA_IPXX', 'SNO_IPXX', 'TWONODE', '4CONTROL_IPXX', or '5CONTROL_IPXX', where XX is 'V4', 'V6', or 'V4V6'"
```
rwsu:
```suggestion
  printf "Found invalid value \"$AGENT_E2E_TEST_SCENARIO\" for AGENT_E2E_TEST_SCENARIO. Supported values: 'COMPACT_IPXX', 'HA_IPXX', 'SNO_IPXX', 'TNA', '4CONTROL_IPXX', or '5CONTROL_IPXX', where XX is 'V4', 'V6', or 'V4V6'"
```

agent/05_agent_configure.sh
eggfoobar: Sure thing I can add that in there

common.sh
eggfoobar: Originally we had avoided adding `NUM_ARBITERS` to keep things simple for the arbiter additions. If it's okay with you, I would like to avoid adding that in, in this PR since that would make the change a bit much, I can alter it in a different PR so it'll be easier to follow. WDYT?

eggfoobar: > Would it be better to call this TNA?

Yeah that might be a better way to do it, I can change it to TNA to keep in line with things like SNO.

common.sh
```diff
@@ -456,6 +456,13 @@ if [[ ! -z ${AGENT_E2E_TEST_SCENARIO} ]]; then
           export MASTER_MEMORY=32768
           export NUM_WORKERS=0
           ;;
+      "TWONODE" )
+          export NUM_MASTERS=2
+          export MASTER_VCPU=4
+          export MASTER_DISK=100
+          export MASTER_MEMORY=32768
+          export NUM_WORKERS=0
```
bfournie, line 479: Might want to add the values for ARBITER_VCPU, MEMORY, and DISK here instead of defining that in a separate file.
```
export ARBITER_VCPU=2
export ARBITER_MEMORY=16384
export ARBITER_DISK=100
```

bfournie (commented):
I've tested this with https://github.com/openshift/installer/pull/9780 and was able to install the cluster although there was an issue etcd being degraded which may require an additonal runtimecfg change.

common.sh
eggfoobar: Sure thing, updated

common.sh
eggfoobar: Sounds good, changed to TNA and added to config example

common.sh
eggfoobar: Good suggestion, added them in there now, thanks!

agent/06_agent_create_cluster.sh
eggfoobar: Thinking the same thing as above, about holding this off for a separate PR since we would need to make more changes to the current flow for arbiter installs, wdyt @rwsu ?

agent/05_agent_configure.sh
```diff
@@ -63,6 +63,10 @@ function configure_node() {
     local hostname="$(printf "${WORKER_HOSTNAME_FORMAT}" "${node_num}")"
     local ip=$((base_ip + ${NUM_MASTERS} + node_num))
   fi
+  if [[ "$node_type" == "arbiter" ]]; then
+    local hostname="$(printf "${ARBITER_HOSTNAME_FORMAT}" "${node_num}")"
```
andfasano, line 63: I'm not clear about the specific requirement for having a different hostname for the arbiter. Can you please clarify it? Is it just for troubleshooting or it's maybe an hard dependency for something else?

common.sh
andfasano: I'd also agree that having an explicit NUM_ARBITERS it would be more clear, preferably in this PR, and it allows for explicit sanity check (ie is it supported any value > 1)?

common.sh
```diff
@@ -456,6 +457,17 @@ if [[ ! -z ${AGENT_E2E_TEST_SCENARIO} ]]; then
           export MASTER_MEMORY=32768
           export NUM_WORKERS=0
           ;;
+      "TNA" )
+          export ENABLE_ARBITER="true"
+          export NUM_MASTERS=2
+          export MASTER_VCPU=4
+          export MASTER_DISK=100
```
andfasano, line 473: Added a comment here: https://github.com/openshift-metal3/dev-scripts/pull/1773#discussion_r2149449950.

Not sure if the NUM_ARBITER will always be equal to 1, but it seems to a good occasion to make it more clear (and add any required validation)

common.sh
eggfoobar: Sure thing, I can make the change for that

eggfoobar:
> @eggfoobar this PR adds the arbiter support in dev-scripts. Do you have any plan also to enhanced the OpenShift CI with presubmit jobs and, moreover, release gating ones? Thanks

Yes that's right, the plan is release gating in the future, this specific change will help us create some agent specific CI, we are currently running regular CI with IPI install using these scripts.

agent/05_agent_configure.sh
eggfoobar: Good point, this is mainly to help make things easier to troubleshoot, there is no hard requirement on hostname

agent/05_agent_configure.sh
```diff
@@ -63,6 +63,10 @@ function configure_node() {
...
+    local ip=$((base_ip + ${NUM_MASTERS} + node_num))
```
andfasano, line 64: This doesn't look fully correct to me. The subsequent call

```
configure_node "arbiter" "0"
```

could generate an ip conflict in case a worker (or an extra-worker) will be added. That's because the `configure_node` assign ips following the convention:

```
base_ip | NUM_MASTERS | NUM_WORKERS | NUM_EXTRAWORKERS
```

In order to ensure the new role type will be correctly handled, it should be something like 

```
base_ip | NUM_MASTERS | NUM_ARBITERS | NUM_WORKERS | NUM_EXTRAWORKERS
```

agent/05_agent_configure.sh
andfasano: Ok thanks

agent/05_agent_configure.sh
eggfoobar: Yup, you're right, I ran into this, will be updating the code with the NUM_ARBITERS

agent/06_agent_create_cluster.sh
eggfoobar: Swapped out to using NUM_ARBITERS, thanks for the feedback

common.sh
eggfoobar: Yeah good point, currently we don't have any limitation in the code but we want this script to just be 1 arbiter and 2 masters. I added a validation step to enforce that.

agent/05_agent_configure.sh
```diff
@@ -63,6 +63,10 @@ function configure_node() {
     local hostname="$(printf "${WORKER_HOSTNAME_FORMAT}" "${node_num}")"
     local ip=$((base_ip + ${NUM_MASTERS} + node_num))
   fi
+  if [[ "$node_type" == "arbiter" ]]; then
+    local hostname="$(printf "${ARBITER_HOSTNAME_FORMAT}" "${node_num}")"
+    local ip=$((base_ip + ${NUM_MASTERS} + ${NUM_ARBITERS} + ${NUM_WORKERS} + ${NUM_EXTRA_WORKERS} + node_num))
```
andfasano: I suppose also the other node types should take into account the num arbiters, so that:

```
master_ip       = base + N
arbiter_ip      = base + #masters + N
worker_ip       = base + #masters + #arbiters + N
extra_worker_ip = base + #masters + #arbiters + #workers + N
```

eggfoobar: You're right, makes more sense that way, I updated with the flow you've described, thanks!

eggfoobar:
Hey @andfasano I'm not sure how important of a signal the serial run is, currently it seems to be timing out often due to its length, should we skip it for this since IPV6 passed?

andfasano:
> Hey @andfasano I'm not sure how important of a signal the serial run is, currently it seems to be timing out often due to its length, should we skip it for this since IPV6 passed?

Technically speaking this PR touched some common files also used by the metal job, so it'd be better to investigate the current failure reasons of the job. Also, does it fail only for this PR or not?

eggfoobar:
Unfortunately in recent runs only my PR has executed that job, I'll check in another PR. However, the one failure that did happen outside of the timeout passed during the timeout, so I'm not sure how related that is.

andfasano:
> Unfortunately in recent runs only my PR has executed that job, I'll check in another PR. However, the one failure that did happen outside of the timeout passed during the timeout, so I'm not sure how related that is.

Maybe someone from the metal could help on that? cc @elfosardo

eggfoobar:
@andfasano I seems the sharding has helped with the serial lane, the latest failure that was skipped is unrelated to this repo and is being resolved elsewhere. Otherwise, I think this PR is good to go

config_example.sh
```diff
@@ -496,6 +486,26 @@ set -x
...
+#export ARBITER_HOSTNAME_FORMAT=arbiter-%d
+
+# NUM_ARBITERS -
+# Indicate number of arbiter nodes to deploy.
+# Arbiter nodes take the place of a master node but run only critical containers to maintain HA for the cluster,
+# it is a TechPreview for OCP 4.19, so `export FEATURE_SET="TechPreviewNoUpgrade"` must be set.
+# Default: 0
+#
+# Furthermore, since an Arbiter node takes the place of a master node the NUM_MASTERS count
+# must be set to `export NUM_MASTERS=2` and not the default `3` to avoid even number etcd members.
```
bfournie, line 503: Seems like this is done in the script and so NUM_MASTERS doesn't need to be set

eggfoobar: Ah, yup for agent based that's true, but for IPI install, you would need to set this independently.

config_example.sh
```diff
@@ -496,6 +486,26 @@ set -x
...
+# Set a custom hostname format for arbiters. This is a format string that should
+# include one %d field, which will be replaced with the number of the node.
+# Default: "arbiter-%d"
+#
+#export ARBITER_HOSTNAME_FORMAT=arbiter-%d
+
+# NUM_ARBITERS -
+# Indicate number of arbiter nodes to deploy.
+# Arbiter nodes take the place of a master node but run only critical containers to maintain HA for the cluster,
+# it is a TechPreview for OCP 4.19, so `export FEATURE_SET="TechPreviewNoUpgrade"` must be set.
```
andfasano, line 499: @eggfoobar can you please add the proper doc for the FEATURE_SET new config var as well (ok also in a subsequent PR) ? As per the current implementation it could be set indipendently (and probably in future also another var for the feature gates could be useful)

eggfoobar: Sure thing, it was a pre-existing flag that I saw in the code and didn't notice there was no doc for it. I'll open a PR to add that and some feature gate addons would be nice as well. Thanks!

PR #1775, opened 2025-07-02
AGENT-1193: Custom openshift-install can be used by appliance

rwsu:
If the openshift-install binary was built from source, this change creates a symlink to it in the appliance's asset directory. Placing it in the asset directory will signal to the appliance to use it instead of downloading it from the release.

This allows us to test changes in the installer repo without having to wait for it to merge and appear in an offcial release image.

Hard coding to version 4.19 in the directory structures used by the *agent_iso_no_registry functions has also be removed.

Works in conjunction with: https://github.com/openshift/agent-installer-utils/pull/122

### DIFF
--- a/agent/03_agent_build_installer.sh
+++ b/agent/03_agent_build_installer.sh
@@ -8,11 +8,6 @@ source $SCRIPTDIR/logging.sh
 source $SCRIPTDIR/common.sh
 source $SCRIPTDIR/agent/common.sh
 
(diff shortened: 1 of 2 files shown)


eggfoobar:

Testing out if lane passes

rwsu:
/hold 
May need to be revised to account for https://github.com/openshift/agent-installer-utils/pull/82

PR #1783, opened 2025-08-08
Allow limiting virtual media emulator to one IP family

dtantsur:
It's a common case by customers. Requires:
- https://review.opendev.org/c/openstack/sushy-tools/+/956904

I also need to fix up the BMC addresses, otherwise the logic does not kick in.


iurygregory:
the sushy-tools patch is merged

dtantsur:
> the sushy-tools patch is merged

We need a release though

iurygregory:
I've pushed https://review.opendev.org/c/openstack/releases/+/957055

dtantsur:
/retest

Huh, it has been a while

PR #1785, opened 2025-08-19
OCPBUGS-57386: Add retries for external dependencies

jupierce:

PR #1786, opened 2025-08-19
OCPBUGS-57386: Add retries for external dependencies

jupierce:
Replaces more dnf invocations than https://github.com/openshift-metal3/dev-scripts/pull/1785 with function.


andfasano:
/hold

It looks like https://github.com/openshift-metal3/dev-scripts/pull/1785 should land before the current one?

jupierce:
@andfasano this pull request is a superset of #1785 and covers significantly more surface area. I didn't know which the team would be more comfortable with. #1785 is the minimum to address issues discussed in OCPBUGS-57386.  This PR covers the same class of issues, but across more of the code base.

PR #1791, opened 2025-08-25
NO-ISSUE: Bump AGENT_OVE_ISO_SIZE to 50Gb

andfasano:
Currently the OVE ISO size is slightly larger than the previous 40Gb, so bumping it to a greater value


andfasano:

The patch does not affect any metal job, and currently they are broken

PR #1792, opened 2025-08-28
OCPBUGS-60977: fix: fencing credentials for tnf require hostname

eggfoobar:
updated fencing credentials to add hostname instead of node name for ipv6 and dualstack since they force fqdn hostnames for the node names.

### DIFF
--- a/utils.sh
+++ b/utils.sh
@@ -318,13 +318,18 @@ function node_map_to_install_config_fencing_credentials() {
     credentials:
 EOF
     for ((idx=0;idx<$(($NUM_MASTERS));idx++)); do
-      name="$(printf $MASTER_HOSTNAME_FORMAT ${idx})"
+      hostname="$(printf $MASTER_HOSTNAME_FORMAT ${idx})"
+      # IP V6 and DualStack will force FQDN hostname for the VMs, we need to update
+      # this here to correctly set the hostname for the fencing credentials.
+      if [[ $IP_STACK != 'v4' ]]; then
+        hostname="${hostname}.${CLUSTER_DOMAIN}"
+      fi
       username=$(node_val ${idx} "driver_info.username")
       password=$(node_val ${idx} "driver_info.password")
       address=$(node_val ${idx} "driver_info.address")
 
       cat <<EOF
-    - hostname: ${name}
+    - hostname: ${hostname}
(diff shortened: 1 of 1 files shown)


jaypoulz:
/lgtm

I think we'll need to rework this when we harden this on the cluster side to incorporate some kind of "static" identification, but this will do for now.

utils.sh
```diff
@@ -318,13 +318,18 @@ function node_map_to_install_config_fencing_credentials() {
     credentials:
 EOF
     for ((idx=0;idx<$(($NUM_MASTERS));idx++)); do
-      name="$(printf $MASTER_HOSTNAME_FORMAT ${idx})"
+      hostname="$(printf $MASTER_HOSTNAME_FORMAT ${idx})"
+      # IP V6 and DualStack will force FQDN hostname for the VMs, we need to update
+      # this here to correctly set the hostname for the fencing credentials.
+      if [[ $IP_STACK != 'v4' ]]; then
```
MahnoorAsghar, line 324: I'm sorry if I'm plain wrong here, but it looks like hostname remains the same for dualstack and ipv6, but changes format for the v4 case which is contrary to the comment and PR description?

MahnoorAsghar:
/test e2e-metal-ipi-serial-ipv4-1of2
Required job failing with:

>  expected:
>       warning: Cannot refresh available updates:
>         Reason: NoChannel
>         Message: The update channel has not been configured.
>       Upstream update service: http://172.30.77.46:8000/graph
>       Channel: test-channel
>       No updates available. You may still upgrade to a specific release image with --to-image or wait for new updates to be available.

utils.sh
eggfoobar: Oh this should be correct, so in this if statement, we're just saying if you're not in `v4` than we use FQDN hostname, which is just the `$(printf $MASTER_HOSTNAME_FORMAT ${idx})` + `$CLUSTER_DOMAIN`. 

So if you're `v4` you get the normal `printf $MASTER_HOSTNAME_FORMAT ${idx}` but if you're not `v4` and are `v6 or v4v6`, then you get the FQDN version of that.

MahnoorAsghar: My bad, thats absolutely correct

PR #1794, opened 2025-09-02
AGENT-1292: Create agent OVE UI ISO using Dockerfile

bfournie:
Create the agent OVE UI ISO with the Dockerfile added for Konflux integration in order to exercise this build method. Add a flag to switch between using this Dockerfile and the build-ove-image.sh script.

### DIFF
--- a/agent/06_agent_create_cluster.sh
+++ b/agent/06_agent_create_cluster.sh
@@ -3,6 +3,7 @@ set -euxo pipefail
 shopt -s nocasematch
 
 SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )/.." && pwd )"
+ARCH=$(uname -m)
 
 LOGDIR=${SCRIPTDIR}/logs
 source $SCRIPTDIR/logging.sh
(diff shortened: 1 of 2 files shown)


bfournie:
Confirmed that installation works the same with the new flag when using 
`export AGENT_E2E_TEST_BOOT_MODE=ISO_NO_REGISTRY`

agent/06_agent_create_cluster.sh
```diff
@@ -90,7 +90,17 @@ function create_agent_iso_no_registry() {
...
   cd $OPENSHIFT_AGENT_INSTALER_UTILS_PATH/tools/iso_builder
-  ./hack/build-ove-image.sh --pull-secret-file "${PULL_SECRET_FILE}" --release-image-url "${OPENSHIFT_RELEASE_IMAGE}" --ssh-key-file "${SSH_KEY_FILE}" --dir "${asset_dir}"
+  if [ "${AGENT_OVE_ISO_FROM_CONTAINER}" == false ]; then
+    ./hack/build-ove-image.sh --pull-secret-file "${PULL_SECRET_FILE}" --release-image-url "${OPENSHIFT_RELEASE_IMAGE}" --ssh-key-file "${SSH_KEY_FILE}" --dir "${asset_dir}"
+  else
+    # Build the ISO in the container image
+    make build-ove-iso-container PULL_SECRET_FILE="${PULL_SECRET_FILE}" RELEASE_IMAGE_URL="${OPENSHIFT_RELEASE_IMAGE}"
+    # Retrieve ISO from container
+    ./hack/iso-from-container.sh
+    local iso_name="agent-ove.$(uname -p).iso"
```
andfasano: Patching here the filename for the desired arch doesn't seem fully correct to me. I don't want to block this PR, but it should be fixed directly in the agent-installer-utils repo

bfournie: I'm not quite sure I see the issue, as this is just being used to get the filename of the ISO that the container will have generated, similar to what we do in:
here - https://github.com/bfournie/dev-scripts/blob/master/agent/06_agent_create_cluster.sh#L62
here - https://github.com/bfournie/dev-scripts/blob/master/agent/06_agent_create_cluster.sh#L150
and here - https://github.com/bfournie/dev-scripts/blob/master/agent/06_agent_create_cluster.sh#L159

Can you let me know what should fixed in agent-installer-utils?

andfasano: I think that the root issue is that the `ARCH` parameter is not defined here in the build script (and by default is set to [x86_64](https://github.com/openshift/agent-installer-utils/blob/0dc0681302c17227913913b43439ddc00243489e/tools/iso_builder/hack/helper.sh#L58)). So there's the implicit assumption that the _current (uname -a)_ is the right one, which potentially couldn't match. 
A slightly better fix just here could be to define a local `ARCH` var as `uname -a`, and use it for building the ISO

config_example.sh
```diff
@@ -979,3 +979,11 @@ set -x
...
 #
 # export AGENT_ISOLATED_NETWORK=""
+
+# AGENT_OVE_ISO_FROM_CONTAINER
+# Default: false
+#
+# When enabled, uses buildah to create a container image
+# for the OVE UI ISO
+#
+# export AGENT_OVE_ISO_FROM_CONTAINER=""
```
andfasano: On a second thought, do we really need this additional configuration variable (the lesser the better)? Now that we're moving to Konflux as the default build system, I think we should stick to the Dockerfile as the official/default way of building the ISO. The script could remain but only for local debugging in agent-installer-utils.
(As a separate discussion thread, we should also avoid having different ways of using it outside and inside the Dockerfile)

bfournie: Yes I am fine with removing it, I had just added it to allow a cutover but since we can build the ISO successfully with Dockerfile we should use that method.

agent/06_agent_create_cluster.sh
bfournie: OK, added use of ARCH i all places and pass it into the makefile. An additional change in https://github.com/openshift/agent-installer-utils/pull/162 will use the ARCH

andfasano:
/lgtm

/hold
Maybe is it better to wait for a green e2e-agent-io-registry-compact-ipv4 run?

bfournie:
removing hold, e2e-agent-io-registry-compact-ipv4 is now passing
/unhold

PR #1797, opened 2025-09-19
OPNET-708: Add asset files for new CI job to test br-ex.

emy:
As per [OPNET-708](https://issues.redhat.com/browse/OPNET-708) we found out that we do not have a reliable way to test br-ex in NMState. This PR is a preparation for the work in the release repository.


emy:
CC: @cybertron

PR #1798, opened 2025-09-22
feat: add support for TNF agent based install

eggfoobar:


agent/roles/manifests/templates/install-config_baremetal_yaml.j2
```diff
@@ -21,6 +21,20 @@ controlPlane:
   hyperthreading: Enabled
   name: master
   replicas: {{ num_masters }} 
+{% if num_arbiters == "0" and num_masters == "2" %}
```
andfasano: what about moving this logic in the code, trying to keep the jinja template as simple as possible? It could become something like `{% if enable_fencing %}...`

agent/roles/manifests/templates/install-config_baremetal_yaml.j2
```diff
@@ -21,6 +21,20 @@ controlPlane:
...
+{% set hostnames = agent_nodes_hostnames.split(',') %}
```
andfasano: Also below the same code for extracting hostnames and bmc details is used, as a minor point it could be the occasion to refactor it and remove the duplication (for the jinja vars definition)

eggfoobar: That's a great point, wasn't too familiar with jinja var scoping, but I went ahead and moved the duplicated instantiated vars at the top of the file, is that what you had in mind for this?

agent/roles/manifests/templates/install-config_baremetal_yaml.j2
eggfoobar: Sure thing, I went ahead and added the variable to handle the logic on the code side.

common.sh
```diff
@@ -397,6 +397,13 @@ fi
 
 export ENABLE_LOCAL_REGISTRY=${ENABLE_LOCAL_REGISTRY:-}
 
+# Helper variable, not meant to be configurable by user, will be set if two node fencing is detected
+# and used as reference for TNF configurations.
+export ENABLE_TWO_NODE_FENCING=""
+if [[ ${NUM_ARBITERS} -eq 0 ]] && [[ ${NUM_MASTERS} -eq 2 ]]; then
+  export ENABLE_TWO_NODE_FENCING="true"
```
andfasano, line 407: This is a little bit confusing for me. Below there is already a logic to enable the 2n fendince when the AGENT_E2E_TEST_SCENARIO is TNF, which makes sense to me.

Here I was expecting something more like
```
export ENABLE_TWO_NODE_FENCING=${ENABLE_TWO_NODE_FENCING:"false"}
```

Either to avoid an unset var either to allow the user to override it _if_ required (usually when not using AGENT_E2E_TEST_SCENARIO).

eggfoobar: Ah sorry, I had forgotten another place I wanted to use this, I replaced the `if` logic in the ipi flow as well, so this logic just sets the correct var for that flow. As well we explicitly set the var in the agent flow. 

I updated the code to allow the user to change it if in the future we want to force fencing credentials being present, but otherwise this will just be on auto pilot

andfasano: Ok, it wasn't clear to me that the same config var was used in both the bm IPI workflow and the agent one. A better way to highlight that could be to add a check, ie something like `if [[ -z ${AGENT_E2E_TEST_SCENARIO} ]]; then` (it may also prevent weird mistakes in case both vars were set _and_ also AGENT_E2E_TEST_SCENARIO)

eggfoobar: Oh good catch, updated the code to also check for that so it only gets checked when AGENT_E2E_TEST_SCENARIO is **not** set, and allows the agent flow to dictate that themselves, thanks @andfasano !

agent/roles/manifests/templates/install-config_baremetal_yaml.j2
```diff
@@ -21,6 +26,16 @@ controlPlane:
   hyperthreading: Enabled
   name: master
   replicas: {{ num_masters }} 
+{% if enable_two_node_fencing %}
+  fencing:
+    credentials:
+{% for hostname in hostnames %}
+    - hostname: {{hostname}}
```
jaypoulz, line 33: pretty sure we also allow disabling certificate verification.

agent/roles/manifests/templates/install-config_baremetal_yaml.j2
```diff
@@ -1,4 +1,9 @@
 {% import 'net_macros.yaml' as net %}
+{% set hostnames = agent_nodes_hostnames.split(',') %}
+{% set bmc_addresses = agent_nodes_bmc_addresses.split(',') %}
```
jaypoulz, line 3: One thing we don't have a 1/1 mapping for is ipmi support for BMC.
This is supported for provisioning via the baremetal operator via ironic, but is *NOT* supported for TNF.

I know it looks like duplicate information, but I would avoid reusing the same variables as bmc used for the baremetalhost definitions because we don't have feature parity and may need to support a divergent feature set from the provisioning side.

agent/roles/manifests/templates/install-config_baremetal_yaml.j2
eggfoobar: Updated and added statement to supply the certificate verification.

agent/roles/manifests/templates/install-config_baremetal_yaml.j2
eggfoobar: This was discussed earlier and is not relevant for this PR anymore, added BMC validation for redfish since that was discussed as being more important here.

PR #1799, opened 2025-09-25
Generate FIPS-compatible ssh keys

zaneb:
The default SSH key format in CentOS appears to have changed from RSA to ED25519, however this is not compatible with FIPS. Explicitly request an RSA key to avoid blocking FIPS installs.

### DIFF
--- a/02_configure_host.sh
+++ b/02_configure_host.sh
@@ -35,7 +35,7 @@ esac
 
 # Generate user ssh key
 if [ ! -f $HOME/.ssh/id_rsa.pub ]; then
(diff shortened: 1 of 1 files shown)


zaneb:
The default was [changed in OpenSSH 9.5](https://www.openssh.com/txt/release-9.5). OpenSSH was updated from 8.7 to 9.9 [in RHEL9.8](https://issues.redhat.com/browse/RHEL-108912).

zaneb:
Merged this manually since (a) it's clearly fine, and (b) the world is broken.

PR #1803, opened 2025-10-22
NO-JIRA: fix: add golang download mirror url to avoid google bucket url

eggfoobar:


01_install_requirements.sh
```diff
@@ -170,6 +171,7 @@ ANSIBLE_FORCE_COLOR=true ansible-playbook \
   -e "working_dir=$WORKING_DIR" \
   -e "virthost=$HOSTNAME" \
   -e "go_version=$GO_VERSION" \
+  -e "go_custom_mirror=$GO_CUSTOM_MIRROR"
```
pawanpinjarkar: need a `\`

eggfoobar: Phew, good catch, fixed!

PR #1807, opened 2025-11-07
Add extra worker to cluster after dev-scripts install is complete

MahnoorAsghar:
Assisted by: Claude

### DIFF
--- a/.add_worker_implementation_summary.md
+++ b/.add_worker_implementation_summary.md
@@ -1,0 +1,234 @@
+# Implementation Summary: Post-Installation Worker Node Addition
+
+## What Was Created
+
+### Core Scripts
+
+1. **`add_worker_node.sh`** - Main script to add worker nodes
+   - Creates libvirt VM with customizable resources
+   - Configures virtual BMC (vbmc or sushy-tools)
+   - Generates BareMetalHost and Secret manifests
+   - Provides step-by-step instructions for completion
+   - Automatically finds available BMC ports and generates unique MAC addresses
+
+2. **`remove_worker_node.sh`** - Cleanup script to remove worker nodes
+   - Drains and deletes node from cluster
+   - Removes BareMetalHost and Machine resources
+   - Destroys VM and cleans up disk/NVRAM
+   - Removes BMC configuration
+
+3. **`auto_approve_csrs.sh`** - Helper script for CSR approval
+   - Auto-approves pending CSRs for specified duration
+   - Useful for development/testing scenarios
+   - Default 30-minute duration
+
+### Documentation
+
+1. **`docs/add-worker-post-install.md`** - Complete documentation
+   - Detailed usage instructions
+   - Configuration options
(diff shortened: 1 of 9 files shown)


.add_worker_implementation_summary.md
dtantsur, line 1: You should not commit this file

add_worker_node.sh
```diff
@@ -0,0 +1,306 @@
...
+  username: $(echo -n "admin" | base64 -w0)
+  password: $(echo -n "password" | base64 -w0)
+---
+apiVersion: metal3.io/v1alpha1
+kind: BareMetalHost
+metadata:
+  name: ${CLUSTER_NAME}-${WORKER_NAME}
+  namespace: openshift-machine-api
+  labels:
+    infraenvs.agent-install.openshift.io: ${CLUSTER_NAME}
```
dtantsur, line 257: This should not be here: it requires assisted components

add_worker_node.sh
```diff
@@ -0,0 +1,306 @@
...
+spec:
+  online: true
+  bootMACAddress: ${WORKER_MAC}
+  bmc:
+    address: ${BMC_FULL_ADDRESS}
+    credentialsName: ${CLUSTER_NAME}-${WORKER_NAME}-bmc-secret
+    disableCertificateVerification: true
+  automatedCleaningMode: disabled
```
dtantsur, line 265: Hmm, why?

remove_worker_node.sh
```diff
@@ -0,0 +1,125 @@
...
+else
+    CLUSTER_ACCESSIBLE=true
+fi
+
+# Delete BareMetalHost
+if [ "$CLUSTER_ACCESSIBLE" = true ]; then
+    BMH_NAME="${CLUSTER_NAME}-${WORKER_NAME}"
+    if oc get baremetalhost -n openshift-machine-api "$BMH_NAME" &>/dev/null; then
+        echo "Deleting BareMetalHost ${BMH_NAME}..."
+        oc delete baremetalhost -n openshift-machine-api "$BMH_NAME" || true
```
dtantsur, line 51: I wonder why errors are ignored in all these calls

remove_worker_node.sh
```diff
@@ -0,0 +1,125 @@
...
+    else
+        echo "BareMetalHost ${BMH_NAME} not found in cluster"
+    fi
+    
+    # Check for and delete corresponding Machine
+    echo "Checking for Machine resources..."
+    MACHINES=$(oc get machine -n openshift-machine-api -o name | grep -i "$WORKER_NAME" || true)
+    if [ -n "$MACHINES" ]; then
+        echo "Found machines: $MACHINES"
+        echo "$MACHINES" | xargs -r oc delete -n openshift-machine-api || true
```
dtantsur, line 68: Are you sure it's the right process? I think you need to scale down MachineSet, otherwise it will create another machine

PR #1811, opened 2025-11-18
Ignore signatures when using oc-mirror

bfournie:
Nightly builds don't include signatures so ignore when using oc-mirror.

### DIFF
--- a/oc_mirror.sh
+++ b/oc_mirror.sh
@@ -72,7 +72,7 @@ function mirror_to_file() {
    config=${1}
 
    pushd ${WORKING_DIR}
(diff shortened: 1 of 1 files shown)


cybertron:
/lgtm

CI failures look related to ansible galaxy, not the patch.

andfasano:
/test e2e-agent-sno-ipv6

since it's using a mirror, I wasn't expecting it to fail

bfournie:
> /test e2e-agent-sno-ipv6
> 
> since it's using a mirror, I wasn't expecting it to fail

Note that this isn't using` oc-mirror`, only `oc adm release mirror` so doesn't test this new parameter

PR #1814, opened 2025-11-21
Fix IPv6 VIPs DNS lookup for baremetal deployments

mnietoji:
Issue when using Dual stack networking as EXTERNAL_SUBNET_V4 is defined. Instead, use IP_STACK for setting API_VIPS and INGRESS_VIPS for either IPV4, IPV6 or Dual stack


cybertron:
/test e2e-metal-ipi-bm

Failed on an infra issue before it started testing the patch, seems like we should make sure it passes this job though.

cybertron:
/lgtm

Looks reasonable to me and passed the bm jobs.

mnietoji:
hi, what else is  needed to merge this PR?

PR #1819, opened 2025-12-01
OCPBUGS-62790: Restore agent MASTER_MEMORY default to 16GiB

zaneb:
The default MASTER_MEMORY was changed in some agent scenarios in order
to work around a problem in OKD where FCOS would use only 20% of RAM for
the ephemeral storage instead of 50%. Other scenarios were later changed
to match when tests started failing in 4.19 due to OCPBUGS-62790.

Return the defaults to the minimum values validated by assisted-service
in ABI, so that any future regressions like OCPBUGS-62790 are caught
immediately.

This depends on https://github.com/openshift/installer/pull/10133, including backports to 4.20 and 4.19.

### DIFF
--- a/common.sh
+++ b/common.sh
@@ -472,28 +472,28 @@ if [[ ! -z ${AGENT_E2E_TEST_SCENARIO} ]]; then
           export NUM_MASTERS=5
           export MASTER_VCPU=${MASTER_VCPU:-4}
           export MASTER_DISK=${MASTER_DISK:-100}
-          export MASTER_MEMORY=${MASTER_MEMORY:-24576}
+          export MASTER_MEMORY=${MASTER_MEMORY:-16384}
           export NUM_WORKERS=0
           ;;
       "4CONTROL" )
           export NUM_MASTERS=4
           export MASTER_VCPU=${MASTER_VCPU:-4}
           export MASTER_DISK=${MASTER_DISK:-100}
-          export MASTER_MEMORY=${MASTER_MEMORY:-24576}
+          export MASTER_MEMORY=${MASTER_MEMORY:-16384}
           export NUM_WORKERS=0
           ;;
       "COMPACT" )
           export NUM_MASTERS=3
           export MASTER_VCPU=${MASTER_VCPU:-4}
           export MASTER_DISK=${MASTER_DISK:-100}
-          export MASTER_MEMORY=${MASTER_MEMORY:-32768}
+          export MASTER_MEMORY=${MASTER_MEMORY:-16384}
           export NUM_WORKERS=0
           ;;
       "TNA" )
           export NUM_MASTERS=2
           export MASTER_VCPU=${MASTER_VCPU:-8}
           export MASTER_DISK=${MASTER_DISK:-100}
-          export MASTER_MEMORY=${MASTER_MEMORY:-32768}
+          export MASTER_MEMORY=${MASTER_MEMORY:-16384}
           export NUM_ARBITERS=1
           export ARBITER_VCPU=${ARBITER_VCPU:-2}
           export ARBITER_MEMORY=${ARBITER_MEMORY:-8192}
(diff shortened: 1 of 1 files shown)


zaneb:
https://github.com/openshift/installer/pull/10133 (4.21), https://github.com/openshift/installer/pull/10140 (4.20), and https://github.com/openshift/installer/pull/10153 (4.19) have all merged.

zaneb:
Tests appear to be still failing with "no space left on disk" when pulling images, even though the fix has worked as intended and created a separate tmpfs for the ostree (and even though there is still 815MiB free):

```
Filesystem      Size  Used Avail Use% Mounted on
tmpfs           3.2G  122M  3.1G   4% /run
tmpfs           7.9G  6.8G  1.1G  87% /run/ephemeral_base
/dev/loop0      7.8G  7.0G  815M  90% /run/ephemeral
tmpfs           4.0G  2.4G  1.7G  60% /usr
tmpfs           7.9G   84M  7.8G   2% /tmp
```

This implies that even without the coreos layering change, we would be exceeding the limit for 16GiB servers now. In my local testing 16GiB servers worked OK, so I am not sure what is different here.

zaneb:
/retest
This works fine locally with 4.21 nightly. Unclear why it is not working here (where we're also using 4.21 nightly).

zaneb:
Although it's still not clear to me why this always worked for me locally, it does appear that [OCPBUGS-76450](https://issues.redhat.com/browse/OCPBUGS-76450) is a likely reason for it failing here.

bfournie:
Do we need more master memory when running iso-no-registry?  It looks like its getting stuck at the hosts discovery page and log shows ` host status: <pending-for-input>"`

zaneb:
> Do we need more master memory when running iso-no-registry? It looks like its getting stuck at the hosts discovery page and log shows ` host status: <pending-for-input>"`

Yes, log shows:

```
{ID:has-memory-for-role Status:failure Message:Require at least 17.15 GiB RAM for role master, found only 16.00 GiB}
```

I've upped it to 17.5GiB in that scenario. 16GiB seems to be fine for workers.

bfournie:
In the 5-node test it looks like its running out of space.
df on master-0 is                                                                                                                                                
```
  tmpfs           7.9G  7.0G  917M  89% /run/ephemeral_base
  /dev/loop0      7.8G  7.1G  703M  92% /run/ephemeral    
```

We are seeing a lot of no space left on device errors
`Jun 09 00:53:54 master-0 kubelet.sh[6763]: E0609 04:53:54.711053    6763 pod_workers.go:1324] "Error syncing pod, skipping" err="[failed to \"StartContainer\" for \"kube-controller-manager\" with ImagePullBackOff: \"Back-off pulling image \\\"quay-proxy.ci.openshift.org/openshift/ci@sha256:3a1a1b0d28a9ccf3975a8130a47faa8d6b6ff839bbbebf96ca772e9ef042d3b1\\\": ErrImagePull: unable to pull image or OCI artifact: pull image err: copying system image from manifest list: writing blob: adding layer with blob \\\"sha256:9804b85b1335932fd36975e50d59918c0c9cb228bd73f136cba45f9074bb2c1c\\\"/\\\"\\\"/\\\"sha256:a7a9d44b2b76e3b949c80aa6f660497815aeaf59f76bbc79ef39b766a2c9afa6\\\": unpacking failed (error: exit status 1; output: write /usr/bin/kube-apiserver: no space left on device); artifact err: image reference: reference is a container image, not an OCI artifact\", failed to \"StartContainer\" for \"cluster-policy-controller\" with ImagePullBackOff: \"Back-off pulling image \\\"quay-proxy.ci.openshift.org/openshift/ci@sha256:5df0c7356a54b7763d88b864dfa139ee09ba5dfdc7b9341332ef3cc8e06e793a\\\": ErrImagePull: unable to pull image: copying system image from manifest list: writing blob: adding layer with blob \\\"sha256:5ab75c71069351042dcf2c7a6a33f5de09ff630a2fadeecc5d86a38bf4b46838\\\"/\\\"\\\"/\\\"sha256:0eca9fb113becd0621a57b8ca92ae94a5f951ba4255950b7eae379c5a22f0c46\\\": unpacking failed (error: exit status 1; output: write /usr/share/misc/magic.mgc: no space left on device)\\nexhausting input failed (error: write /var/lib/containers/storage/overlay-layers/tmp/temp-dir-1708365073/0-addition: no space left on device)\"]" pod="kube-system/bootstrap-kube-controller-manager-master-0" podUID="e8713f2ac2dd91c32a23c4122361cbb0"`


I think for the 5-node, 4-node, and ha scenarios we need to use 24GB

zaneb:
> I think for the 5-node, 4-node, and ha scenarios we need to use 24GB

It's an essential requirement that HA work in 16.
It's possible to imagine changing the validation limits for 4/5-node, but this should be a last resort.

PR #1820, opened 2025-12-02
AGENT-1383: Use agent-iso-builder image

pawanpinjarkar:
Switch from cloning agent-installer-utils repo to using the `agent-iso-builder` container image when creating OVE ISOs for disconnected environments.

###   What changed

  - Extract ISO builder tooling from a container image instead of cloning the git repo
  - Auto-detect the right image version based on the OpenShift release (e.g., `registry.ci.openshift.org/ocp/4.21:agent-iso-builder`)
  - Clean up the extracted files and container image when done

###   Why

  This is faster, more consistent with CI, and automatically matches the tooling version to your OpenShift version.

 ---Description drafted with assistance from Claude

### DIFF
--- a/agent/06_agent_create_cluster.sh
+++ b/agent/06_agent_create_cluster.sh
@@ -83,14 +83,13 @@ function create_config_image() {
 }
 
 function create_agent_iso_no_registry() {
-  # Clone agent-installer-utils
-  if [[ ! -d $OPENSHIFT_AGENT_INSTALER_UTILS_PATH ]]; then
-    sync_repo_and_patch go/src/github.com/openshift/agent-installer-utils https://github.com/openshift/agent-installer-utils.git
-  fi
-  # Create agent ISO without registry a.k.a. OVE ISO
   local asset_dir=${1}
+
+  id=$(podman create --pull always --authfile "${PULL_SECRET_FILE}" "${AGENT_ISO_BUILDER_IMAGE}") &&  podman cp "${id}":/src "${asset_dir}" &&  podman rm "${id}"
+
+  # Create agent ISO without registry a.k.a. OVE ISO
   pushd .
-  cd $OPENSHIFT_AGENT_INSTALER_UTILS_PATH/tools/iso_builder
+  cd "${asset_dir}"/src
   # Build the ISO in the container image
   make build-ove-iso-container PULL_SECRET_FILE="${PULL_SECRET_FILE}" RELEASE_IMAGE_URL="${OPENSHIFT_RELEASE_IMAGE}" ARCH=${ARCH}
   # Retrieve ISO from container
@@ -98,6 +97,7 @@ function create_agent_iso_no_registry() {
   local iso_name="agent-ove.${ARCH}.iso"
   echo "Moving ${iso_name} to ${asset_dir}"
   mv ./output-iso/${iso_name} "${asset_dir}"
+  rm -rf "${asset_dir}"/src
   popd
 }
 
--- a/agent/cleanup.sh
+++ b/agent/cleanup.sh
@@ -23,10 +23,15 @@ function agent_remove_iscsi_disks() {
       done
 }
 
-if [[ "${AGENT_E2E_TEST_BOOT_MODE}" == "DISKIMAGE" ]]; then
+if [[ -d "${OCP_DIR}/cache" && -d "${OCP_DIR}/temp" ]]; then
     sudo rm -rf "${OCP_DIR}/cache"
     sudo rm -rf "${OCP_DIR}/temp"
     sudo podman rmi -f ${APPLIANCE_IMAGE} || true
+fi
+
+if [[ -d "${OCP_DIR}/iso_builder" ]]; then
+    sudo rm -rf "${OCP_DIR}/iso_builder"
+    sudo podman rmi -f ${AGENT_ISO_BUILDER_IMAGE} || true
 fi
 
 if [[ -d ${BOOT_SERVER_DIR} ]]; then
--- a/agent/common.sh
+++ b/agent/common.sh
@@ -26,6 +26,14 @@ export ISCSI_DEVICE_NAME=${ISCSI_DEVICE_NAME:-"/dev/sdb"}
 # See: https://github.com/openshift/appliance
 export APPLIANCE_IMAGE=${APPLIANCE_IMAGE:-"quay.io/edge-infrastructure/openshift-appliance:latest"}
 
+if [ "${AGENT_E2E_TEST_BOOT_MODE}" == "ISO_NO_REGISTRY" ] ; then
+    full_ocp_version=$(skopeo inspect --authfile $PULL_SECRET_FILE docker://$OPENSHIFT_RELEASE_IMAGE | jq -r '.Labels["io.openshift.release"]')
+    major_minor_patch_version=$(echo "\"$full_ocp_version\"" | jq -r 'split("-")[0]')
+    major_minor_version=$(echo $major_minor_patch_version | cut -d'.' -f1,2 )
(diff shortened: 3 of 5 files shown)


agent/common.sh
```diff
@@ -26,6 +26,8 @@ export ISCSI_DEVICE_NAME=${ISCSI_DEVICE_NAME:-"/dev/sdb"}
 # See: https://github.com/openshift/appliance
 export APPLIANCE_IMAGE=${APPLIANCE_IMAGE:-"quay.io/edge-infrastructure/openshift-appliance:latest"}
 
+export AGENT_OVE_ISO_BUILDER_IMAGE=${AGENT_OVE_ISO_BUILDER_IMAGE:-"quay.io/edge-infrastructure/agent-ove-iso-builder:latest"}
```
pawanpinjarkar: Value needs to be changed

pawanpinjarkar:
From [ci/prow/e2e-agent-compact-ipv4-iso-no-registry](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift-metal3_dev-scripts/1820/pull-ci-openshift-metal3-dev-scripts-master-e2e-agent-compact-ipv4-iso-no-registry/1996290008822059008) logs 👇 

```
++(/root/dev-scripts/agent/common.sh:34): echo 'Using AGENT_ISO_BUILDER_IMAGE: registry.ci.openshift.org/ocp/4.21:agent-iso-builder'
Using AGENT_ISO_BUILDER_IMAGE: registry.ci.openshift.org/ocp/4.21:agent-iso-builder
............
............
+(./agent/06_agent_create_cluster.sh:89): create_agent_iso_no_registry(): mkdir -p /root/dev-scripts/ocp/ostest/iso_builder/src
++(./agent/06_agent_create_cluster.sh:91): create_agent_iso_no_registry(): podman create --authfile /opt/dev-scripts/pull_secret.json registry.ci.openshift.org/ocp/4.21:agent-iso-builder
Trying to pull registry.ci.openshift.org/ocp/4.21:agent-iso-builder...
Getting image source signatures
Copying blob sha256:cb7a4ba5e2e4ab7f8919986f6a894bac1ce72610998fc7e6a92686347ce309f2
Copying config sha256:be0491b532837c9edb7c241e5a7df82407ea57b1a8f6f2d169f4c2a0258f729c
Writing manifest to image destination
+(./agent/06_agent_create_cluster.sh:91): create_agent_iso_no_registry(): id=81cb2db60b0efee6cfd9586146fdd395734dba65a01af5cd62ceb6ebc8675ccb
+(./agent/06_agent_create_cluster.sh:91): create_agent_iso_no_registry(): podman cp 81cb2db60b0efee6cfd9586146fdd395734dba65a01af5cd62ceb6ebc8675ccb:/src /root/dev-scripts/ocp/ostest/iso_builder/src
+(./agent/06_agent_create_cluster.sh:91): create_agent_iso_no_registry(): podman rm 81cb2db60b0efee6cfd9586146fdd395734dba65a01af5cd62ceb6ebc8675ccb
81cb2db60b0efee6cfd9586146fdd395734dba65a01af5cd62ceb6ebc8675ccb
+(./agent/06_agent_create_cluster.sh:94): create_agent_iso_no_registry(): pushd .
~/dev-scripts ~/dev-scripts
+(./agent/06_agent_create_cluster.sh:95): create_agent_iso_no_registry(): cd /root/dev-scripts/ocp/ostest/iso_builder/src/src
+(./agent/06_agent_create_cluster.sh:97): create_agent_iso_no_registry(): make build-ove-iso-container PULL_SECRET_FILE=/opt/dev-scripts/pull_secret.json RELEASE_IMAGE_URL=registry.build09.ci.openshift.org/ci-op-p0lir155/release@sha256:f70333f1e5b5630323326ee4b9404ef6708a1b600aea272259db497961f7e474 ARCH=x86_64
make[1]: Entering directory '/root/dev-scripts/ocp/ostest/iso_builder/src/src'
# Build the container with specific capabilities to support podman used by openshift-appliance
[1/2] STEP 1/14: FROM registry.ci.openshift.org/ocp/4.21:agent-preinstall-image-builder AS builder
```

agent/cleanup.sh
```diff
@@ -29,6 +29,11 @@ if [[ "${AGENT_E2E_TEST_BOOT_MODE}" == "DISKIMAGE" ]]; then
     sudo podman rmi -f ${APPLIANCE_IMAGE} || true
 fi
 
+if [[ "${AGENT_E2E_TEST_BOOT_MODE}" == "ISO_NO_REGISTRY" ]]; then
```
bfournie: Could you check if "${OCP_DIR}/iso_builder" exists and remove it instead of checking BOOT_MODE?

pawanpinjarkar: The main purpose of this block is cleaning up the container image - the directory removal is secondary since `make clean` already purges OCP_DIR. I kept it purposefully same as the DISKIMAGE code block above.

bfournie: Sure, the main issue I see with checks in the clean script is that the user can change the settings and then do a clean & make and these won't get cleaned up, or the image removed. So it would be good to be able to do this without relying on a env variable, which may have changed since the make.

rwsu: I agree with Bob, things should not be left uncleaned if AGENT_E2E_TEST_BOOT_MODE is changed. AGENT_ISO_BUILDER_IMAGE would go uncleaned if the MODE is changed to something else.

agent/06_agent_create_cluster.sh
```diff
@@ -83,21 +83,21 @@ function create_config_image() {
...
 
 function create_agent_iso_no_registry() {
-  # Clone agent-installer-utils
-  if [[ ! -d $OPENSHIFT_AGENT_INSTALER_UTILS_PATH ]]; then
-    sync_repo_and_patch go/src/github.com/openshift/agent-installer-utils https://github.com/openshift/agent-installer-utils.git
-  fi
-  # Create agent ISO without registry a.k.a. OVE ISO
   local asset_dir=${1}
+
+  id=$(podman create --authfile "${PULL_SECRET_FILE}" "${AGENT_ISO_BUILDER_IMAGE}") &&  podman cp "${id}":/src "${asset_dir}" &&  podman rm "${id}"
```
rwsu:
```suggestion
  id=$(podman create --pull always --authfile "${PULL_SECRET_FILE}" "${AGENT_ISO_BUILDER_IMAGE}") &&  podman cp "${id}":/src "${asset_dir}" &&  podman rm "${id}"
```

rwsu (changes requested):

config_example.sh
```diff
@@ -854,6 +854,7 @@ set -x
 # To generate an ISO that can be used in a disconnected environment 
 # without using a registry a.k.a. OVE ISO, set the boot mode to 'ISO_NO_REGISTRY'.
 # export AGENT_E2E_TEST_BOOT_MODE=ISO
+# export AGENT_ISO_BUILDER_IMAGE="registry.ci.openshift.org/ocp/4.21:agent-iso-builder"
```
andfasano, line 857: As this is expected to be provided by the CI, I think I would prefer to leave it empty and fail fast in case of misconfiguration - rather than providing a potentially valid but not consistent value

pawanpinjarkar: Isn't this just can example? In actual, the value will be set according to major minor version https://github.com/openshift-metal3/dev-scripts/pull/1820/files#diff-2873b86d2f543deb97334c2e888ee7f4f0549cad18d8eb4ec0d3c9e86170031cR33
Perhaps, adding a comment explaining this here should be enough?

andfasano: Ah right, sorry I misread it for a real config.

PR #1823, opened 2025-12-03
OCPBUGS-66339: Partially revert "Enable TLS validation for Redfish emulator"

dtantsur:
In the original change I assumed that any 4.21 version already contains
the BMC CA.  This is not the case for upgrade jobs, which now fail.

On top of that, dev-scripts is used to run jobs on older builds that
also haven't caught up yet.

I'm keeping the metal3-dev-env bump in place since it does not affect
anything. I also keep settings the BMC CA in the install-config so
that we don't accidentally get into the same situation later.

This reverts commit 1276f86ef38cd08933025e77da3fb6ce561196ec.

### DIFF
--- a/ocp_install_env.sh
+++ b/ocp_install_env.sh
@@ -361,8 +361,6 @@ function generate_ocp_install_config() {
       NETWORK_TYPE=OVNKubernetes
     fi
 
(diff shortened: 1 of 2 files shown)


dtantsur:
/retest

Unrelated, hopefully transient

PR #1825, opened 2025-12-04
OPNET-748: Support running with external loadbalancer

cybertron:
Add an EXTERNAL_LOADBALANCER option to the config settings that configures the external loadbalancer and tells the cluster to use it instead of the internal one.

### DIFF
--- a/06_create_cluster.sh
+++ b/06_create_cluster.sh
@@ -29,6 +29,10 @@ if [[ ! -z "$INSTALLER_PROXY" ]]; then
   fi
 fi
 
+if [ -n "$EXTERNAL_LOADBALANCER" ]; then
+  ./external_loadbalancer.sh &
+fi
+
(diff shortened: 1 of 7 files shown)


06_create_cluster.sh
```diff
@@ -29,6 +29,10 @@ if [[ ! -z "$INSTALLER_PROXY" ]]; then
   fi
 fi
 
+if [ -n "$EXTERNAL_LOADBALANCER" ]; then
+  ./external_loadbalancer.sh &
+fi
```
mkowalski, line 34: One detail I would like to see is a failure of the script if you set `EXTERNAL_LOADBALANCER` but do not set `ENABLE_BOOTSTRAP_STATIC_IP=true`. Otherwise we will proceed with openshift-install which will fail and you will end up in a debugging nightmare

cybertron: Good call. I've added a validation for this.

PR #1830, opened 2025-12-19
 AGENT-1193: Add mirror-path and registry-cert support for OVE ISO builds

rwsu:
Add support for using pre-mirrored images and custom registry certificates when building OVE ISOs, enabling disconnected deployments with custom registries.
    
Changes for mirror-path:
- Pass --mirror-path to build-ove-image.sh when MIRROR_IMAGES is enabled
- Appliance skips oc-mirror and uses pre-mirrored images when --mirror-path is provided

Changes for registry-cert:
- Add --registry-cert parameter for custom registry TLS certificates
- Support both script build method (faster for development/debugging) and container build method
- Use unified mechanism across both build methods
- Convert to REGISTRY_CERT make variable for containerized builds

Assisted-by: Claude Sonnet 4.5 <noreply@anthropic.com>

### DIFF
--- a/agent/iso_no_registry.sh
+++ b/agent/iso_no_registry.sh
@@ -1,0 +1,102 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# OVE (OpenShift Virtualization Edition) ISO building utilities
+# Functions for creating agent ISOs without embedded registry
+
+# Check if using a custom registry (not upstream quay.io or CI registry)
+function is_custom_registry() {
+  [[ ! "${OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE}" =~ quay\.io ]] && \
+  [[ ! "${OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE}" =~ registry\.ci\.openshift\.org ]]
+}
+
+# Determine release image URL based on mirror configuration
+function get_release_image_url() {
+  if [[ "${MIRROR_IMAGES}" == "true" ]] && [[ -n "$(get_repo_overrides)" ]]; then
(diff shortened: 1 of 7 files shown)


agent/iso_no_registry.sh
```diff
@@ -0,0 +1,106 @@
...
+
+  pushd .
+  cd "${asset_dir}"/src
+
+  # Determine release image URL
+  local release_image_url=$(get_release_image_url)
+  if [[ "${MIRROR_IMAGES}" == "true" ]]; then
+    echo "Using mirrored release image: ${release_image_url}"
+  else
+    echo "Using upstream release image: ${release_image_url}"
```
andfasano: guess it's a typo, both urls are the same

rwsu: The get_release_image_url function determined the value of ${release_image_url}. I don't think it is that meaningful to have two different log messages and it is causing some confusion. Will consolidate into a single message.

agent/iso_no_registry.sh
```diff
@@ -0,0 +1,104 @@
...
+
+# Create agent ISO without registry (OVE ISO)
+function create_agent_iso_no_registry() {
+  local asset_dir=${1}
+
+  # Update release_info.json as its needed by CI tests
+  save_release_info ${OPENSHIFT_RELEASE_IMAGE} ${OCP_DIR}
+
+  # Temporarily use custom agent-iso-builder image for testing
+  AGENT_ISO_BUILDER_IMAGE="quay.io/rwsu1/agent-iso-builder:dev-scripts"
```
bfournie: Need to remove temp change.

bfournie:
Looks good, just fix the temporary image and will approve.

PR #1831, opened 2025-12-22
METAL-1705: Re-apply "Enable TLS validation for Redfish emulator"

dtantsur:
Reverts the partial revert that disabled the validation in the CI.
Now that we've crossed the release boundary, it should be safe to do.

Starting with version 4.22 instead of 4.21 to be extra safe.

This reverts commit df885c49eb5ef7adf8c9338248060e5016e9e295.


dtantsur:
/retest

Mirrors

PR #1834, opened 2026-01-02
AGENT-1445: Add operator and manifest validation for ISO_NO_REGISTRY mode

bfournie:
Adds a new validation to verify that expected operators are installed after cluster deployment in ISO_NO_REGISTRY mode.

Adds a new validation to verify all manifests are applied properly to the cluster.

### DIFF
--- a/agent/08_agent_post_install_validation.sh
+++ b/agent/08_agent_post_install_validation.sh
@@ -6,6 +6,174 @@ SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )/.." && pwd )"
 LOGDIR=${SCRIPTDIR}/logs
 source $SCRIPTDIR/logging.sh
 source $SCRIPTDIR/common.sh
+source $SCRIPTDIR/release_info.sh
+
+function validate_installed_manifests() {
+    echo "Validating installed operator manifests..."
+
+    # Determine OCP version - manifest validation only supported in 4.21+
+    ocp_version=$(openshift_version ${OCP_DIR})
+    echo "Detected OpenShift version: ${ocp_version}"
+
+    major_version=$(echo "$ocp_version" | cut -d. -f1)
+    minor_version=$(echo "$ocp_version" | cut -d. -f2)
+
+    # Skip validation for versions older than 4.21
+    if [ "$major_version" -lt 4 ] || [ "$major_version" -eq 4 -a "$minor_version" -le 20 ]; then
+        echo "Manifest validation not supported for version ${ocp_version}, skipping."
+        return 0
+    fi
+
+    # Suppress trace output for data processing
+    set +x
+
+    # Get the ConfigMap as JSON
+    if ! configmap_json=$(oc get configmap/olm-operator-manifests -n assisted-installer -o json 2>&1); then
+        set -x
+        echo "ERROR: Could not find ConfigMap olm-operator-manifests in assisted-installer namespace"
+        return 1
+    fi
+
+    # Get all keys that end with .metadata.yaml
+    metadata_keys=$(echo "$configmap_json" | jq -r '.data | keys[] | select(endswith(".metadata.yaml"))')
+
+    if [ -z "$metadata_keys" ]; then
+        set -x
+        echo "ERROR: No metadata files found in ConfigMap"
+        return 1
+    fi
+
+    missing_resources=()
+
+    for metadata_key in $metadata_keys; do
+        # Get the metadata content (plain YAML)
+        metadata_yaml=$(echo "$configmap_json" | jq -r ".data[\"$metadata_key\"]")
+
+        # Extract manifest filenames (lines starting with "- " in YAML list)
+        manifest_files=$(echo "$metadata_yaml" | grep "^- " | sed 's/^- //')
+
+        for manifest_file in $manifest_files; do
+            # Get the base64-encoded manifest content and decode it
+            manifest_yaml=$(echo "$configmap_json" | jq -r ".data[\"$manifest_file\"]" | base64 -d)
+
+            # Extract kind and name from the manifest YAML
+            kind=$(echo "$manifest_yaml" | sed -n 's/^kind: *//p' | head -1 | tr -d '\r')
+            name=$(echo "$manifest_yaml" | sed -n 's/^  name: *//p' | head -1 | tr -d '\r')
+
+            if [ -z "$kind" ] || [ -z "$name" ]; then
+                set -x
+                echo "    WARNING: Could not extract kind or name from $manifest_file"
+                set +x
+                continue
+            fi
+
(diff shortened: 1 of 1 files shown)


bfournie:
/retitle AGENT-1445: Add version-specific operator validation for ISO_NO_REGISTRY mode

agent/08_agent_post_install_validation.sh
```diff
@@ -29,6 +200,12 @@ if [[ "${AGENT_E2E_TEST_BOOT_MODE}" == "ISO_NO_REGISTRY" ]]; then
     echo "SUCCESS: API server connection established and ClusterVersion is available."
     # Run subsequent commands after successful cluster setup
     oc get packagemanifests -n openshift-marketplace
+
```
andfasano, line 203: I just noticed that the current wait condition at [line 15](https://github.com/openshift-metal3/dev-scripts/pull/1834/changes#diff-ccfc36732325a5612c0c5a84d5c6f58f517441bb0f2933308d4f9146026e7874L15) seems not complete: it'd be better to check that also `Progressing` is `False`

agent/08_agent_post_install_validation.sh
```diff
@@ -29,6 +200,12 @@ if [[ "${AGENT_E2E_TEST_BOOT_MODE}" == "ISO_NO_REGISTRY" ]]; then
...
+    # Validate expected operators are installed
+    validate_installed_operators || exit 1
```
andfasano, line 205: Q: Is there any risk to have a false positive due an incomplete operator index?

agent/08_agent_post_install_validation.sh
bfournie: Added check for `Progressing`

agent/08_agent_post_install_validation.sh
bfournie: Tightened up the checks to ensure that if the ConfigMap or metadata files not found it will report an error. 

For the operators not being in the configured catalog, that will cause an earlier failure during the installation, e.g.
`FATAL 2026/02/13 12:39:02  [ERROR]  : invalid default channel configuration for package "numaresources-operator": the default channel "4.21" was filtered out, a new default channel must be configured for this package`

If operators are included in the installation, so are included in the manifests, but are not included in `appliance-config.yaml `the manifests won't be applied so this check will detect it and report an error.

agent/agent_post_install_validation.sh
```diff
@@ -4,6 +4,176 @@ set -euxo pipefail
...
+    else
+        echo "SUCCESS: All expected operator manifests are applied."
+    fi
+}
+
+function validate_installed_operators() {
+    echo "Validating installed operators..."
+
+    # Define expected operators per version
+    expected_operators_4_20=(
```
rwsu: Instead of creating lists of operators here to verify, what do you think about reading the [operators listed in the appliance-config.yamls](https://github.com/openshift/agent-installer-utils/blob/af4cc9a823969a8dac38f6fca3baaef4eb3755c4/tools/iso_builder/config/4.20/appliance-config.yaml#L12) and generating a list from it?

bfournie: Good point @rwsu. I've updated it to use the operators in `appliance-config.yaml` based on version.

agent/08_agent_post_install_validation.sh
```diff
@@ -4,6 +4,165 @@ set -euxo pipefail
...
+    fi
+
+    missing_resources=()
+
+    for metadata_key in $metadata_keys; do
+        echo "Processing metadata: $metadata_key"
+
+        # Get the metadata content (plain YAML)
+        set +x
+        metadata_yaml=$(echo "$configmap_json" | jq -r ".data[\"$metadata_key\"]")
```
andfasano, line 50: I found this trace (and the next one) too verbose and confusing. I think it could be safely skipped (ie drop the set +x/-x - btw not sure if it was intentional but they also look swapped) to reduce the high level of noise currently present in the log. The only relevant bit that could be logged is an eventual not found resource

andfasano: (also, a slightly but not zero risk of leaking some sensisble data)

agent/agent_post_install_validation.sh
```diff
@@ -4,6 +4,165 @@ set -euxo pipefail
...
+        echo "SUCCESS: All expected operator manifests are applied."
+    fi
+}
+
+function validate_installed_operators() {
+    echo "Validating installed operators..."
+
+    # Determine OCP version
+    ocp_version=$(openshift_version ${OCP_DIR})
+    echo "Detected OpenShift version: ${ocp_version}"
```
andfasano: Not sure about this, it's a bit that should be already known in the logs I guess

agent/08_agent_post_install_validation.sh
```diff
@@ -4,6 +4,165 @@ set -euxo pipefail
...
+        for missing_res in "${missing_resources[@]}"; do
+            echo "  - $missing_res"
+        done
+        return 1
+    else
+        echo "SUCCESS: All expected operator manifests are applied."
+    fi
+}
+
+function validate_installed_operators() {
```
andfasano, line 106: Also this method has a lot of noise, mostly derived by the previous set +x/-x - so removing them should also help here

agent/08_agent_post_install_validation.sh
```diff
@@ -4,6 +4,165 @@ set -euxo pipefail
...
+    for op in "${expected_operators[@]}"; do
+        echo "  - $op"
+    done
+
+    # Get list of installed operators (just the names, first column)
+    installed_operators=$(oc get operators -o custom-columns=NAME:.metadata.name --no-headers)
+
+    missing_operators=()
+    for expected_op in "${expected_operators[@]}"; do
+        if ! echo "$installed_operators" | grep -q "^${expected_op}\."; then
```
andfasano, line 155: Maybe I'm wrong but I think also this line produces an unnecessary amount of traces

andfasano (commented):
As a general comment the code looks fine, but the output log is very confusing with a lot of noise. Since this part will be read in case of failure, it must be concise and short for an immediate understanding.
An additional problem noted, the post validation step does not produce a log file, so it'd be nice to fix also that.

agent/08_agent_post_install_validation.sh
bfournie: Removed a lot of verbosity. Only output should be warnings and final error messages of resources are missing.

agent/08_agent_post_install_validation.sh
bfournie: Yep, removed verbosity.

agent/08_agent_post_install_validation.sh
bfournie: Removed.

PR #1835, opened 2026-01-12
METAL-1230: Add support for aarch64 workers on x86

honza:
Similar to regular workers, we can now say:

```sh
export NUM_ARM_WORKERS=${NUM_ARM_WORKERS:-"0"}
```

And if you use a multi-arch payload, you can inspect aarch64 nodes on x86 using qemu emulation. Qemu emulation of aarch64 workers was recently merged in metal3-dev-env upstream so we bump the version here.

Assisted by Claude, tested by hand on my machine.


dtantsur:
Needs rebase (and seem to have a few unrelated changes - maybe split out bumping metal3-dev-env?)

honza:
> maybe split out bumping metal3-dev-env

Removed, no longer necessary

elfosardo:
failures are not related to this change
we should consider overriding the failing jobs

elfosardo:
/retest

giving them another chance before overriding

PR #1836, opened 2026-01-13
AGENT-1309: Modify NoRegistryClusterInstall storage requirements

pawanpinjarkar:
Modify NoRegistryClusterInstall storage requirements

### DIFF
--- a/common.sh
+++ b/common.sh
@@ -546,6 +546,12 @@ if [[ ! -z ${AGENT_E2E_TEST_SCENARIO} ]]; then
     if [ "${SCENARIO}" == "HA" ]; then
        export WORKER_VCPU=5
     fi
+    # Increase disk storage requirements for NoRegistryClusterInstall aka agent OVE ISO
+    case "$SCENARIO" in
+      "SNO"|"COMPACT"|"HA" )
+        export MASTER_DISK=220
+        ;;
+    esac
   fi
 
   if [[ "$AGENT_OPERATORS" =~ "mtv" ]]; then


common.sh
```diff
@@ -546,6 +546,10 @@ if [[ ! -z ${AGENT_E2E_TEST_SCENARIO} ]]; then
     if [ "${SCENARIO}" == "HA" ]; then
        export WORKER_VCPU=5
     fi
+    # Increase disk storage requirements for NoRegistryClusterInstall aka agent OVE ISO
+    export MASTER_DISK=220
+    export WORKER_DISK=220
+    export ARBITER_DISK=220
```
andfasano: Until the TNF will be explicitly supported (and currently it is not), I'd prefer to not have any explicit configuration setting for it

common.sh
```diff
@@ -546,6 +546,13 @@ if [[ ! -z ${AGENT_E2E_TEST_SCENARIO} ]]; then
     if [ "${SCENARIO}" == "HA" ]; then
        export WORKER_VCPU=5
     fi
+    # Increase disk storage requirements for NoRegistryClusterInstall aka agent OVE ISO
+    case "$SCENARIO" in
+      "SNO"|"COMPACT"|"HA" )
+        export MASTER_DISK=220
+        export WORKER_DISK=220
```
andfasano:
```suggestion
```
The requirement is only for master nodes, not for the workers

andfasano: @pawanpinjarkar I just noticed that https://github.com/openshift/installer/pull/10176 force 220Gb also for the workers, so please fix it as well. To avoid keep blocking the CI, let's land this PR as it is and then please fix both installer/dev-scripts repos for the worker

andfasano:

This patch does not seem related

common.sh
pawanpinjarkar: https://github.com/openshift/installer/pull/10254 fixes the installer

PR #1838, opened 2026-01-20
Remove el8 support (glibc too old for recent oc CLI)

elfosardo:
Recent OpenShift releases require glibc 2.34+ which is only available in el9. The oc CLI from OCP 4.14+ won't run on el8.

- Remove el8 case from install requirements
- Update validation to require version >= 9
- Remove oc.rhel8 extraction fallback
- Simplify CI Dockerfile to RHEL9 only
- Update README prerequisites

### DIFF
--- a/01_install_requirements.sh
+++ b/01_install_requirements.sh
@@ -86,31 +86,6 @@ fi
 # Install ansible, other packages are installed via
 # vm-setup/install-package-playbook.yml
 case $DISTRO in
-  "centos8"|"rhel8"|"almalinux8"|"rocky8")
-    # install network-scripts package to be able to use legacy network commands
-    sudo dnf install -y network-scripts
-    if [[ $DISTRO == "centos8" ]] && [[ "$NAME" != *"Stream"* ]]; then
-        echo "CentOS is not supported, please switch to CentOS Stream / RHEL / Rocky / Alma"
-        exit 1
-    fi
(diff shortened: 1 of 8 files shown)


01_install_requirements.sh
```diff
@@ -193,8 +168,7 @@ if [[ "${NODES_PLATFORM}" == "baremetal" ]] ; then
 fi
 
 # needed if we are using locally built images
-# We stop any systemd service so we can run in a container, since
-# there's no RPM/systemd version available for RHEL8
+# We stop any systemd service so we can run in a container
```
dtantsur: I wonder if it's needed at all now

common.sh
```diff
@@ -392,14 +392,9 @@ if [ -n "$EXTERNAL_LOADBALANCER" -a -z "$ENABLE_BOOTSTRAP_STATIC_IP" ]; then
   exit 1
 fi
 
-# TODO(bnemec): Once https://github.com/ansible/ansible/pull/75537 merges this
-# can be removed.
 ALMA_PYTHON_OVERRIDE=
 source /etc/os-release
 export DISTRO="${ID}${VERSION_ID%.*}"
```
dtantsur, line 396: Can we remove the remaining 3 lines too? Or at least ALMA_PYTHON_OVERRIDE?

01_install_requirements.sh
elfosardo: they're not, I'll remove them

common.sh
elfosardo: we can remove L395 but DISTRO is still used

PR #1840, opened 2026-01-24
capi: remove metadata when copying secret

honza:
Validated here https://github.com/openshift/release/pull/73807


damdo (commented):
/lgtm

damdo:
@honza do we still need this?

PR #1841, opened 2026-01-26
Update metal3-dev-env hash

elfosardo:
Bump Ansible version to align with metal3-dev-env requirements.


cybertron:
/lgtm

The job failures don't appear related and the other agent jobs passed, so it looks to me like this should be fine.

PR #1842, opened 2026-01-26
OCPBUGS-74430: Fix screen progression issues with agent OVE UI

bfournie:
Fixes multiple issues with the behaviour of the UI-driven installation for Agent OVE, i.e. ISO_NO_REGISTRY test.

- Added a delay before starting the installation. Without this delay the ignition_config_overrides do not set since the IGNITION_CONFIG_OVERRIDE feature had not yet been read from InfraEnv
- Handle progression from the Cluster screen to allow cluster creation to complete and Operators page to load
- Fix Host Discovery page loading to proceed when Next button is available

In addition debug has been added to save the cluster and host status retrieved from the API along with additional screenshots.

### DIFF
--- a/agent/isobuilder/ui_driven_cluster_installation/main.go
+++ b/agent/isobuilder/ui_driven_cluster_installation/main.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"encoding/json"
 	"fmt"
 	"log"
 	"net/http"
@@ -36,6 +37,7 @@ var (
 
 func main() {
 	logrus.Info("Launching headless browser...")
+	time.Sleep(1 * time.Minute)
 	chromiumPath, _ := launcher.LookPath()
   	url := launcher.New().Bin(chromiumPath).NoSandbox(true).Headless(true).MustLaunch()
 	browser := rod.New().ControlURL(url).MustConnect()
@@ -55,6 +57,11 @@ func main() {
 			screenshotPath = filepath.Join(cwd, ocpDir)
 	}
 
+	// Wait 1 minute before entering cluster details to allow backend initialization
+	// This mimics manual usage where a user takes time to read and fill in the form
+	logrus.Info("Waiting 1 minute to allow backend initialization...")
+	time.Sleep(1 * time.Minute)
+
 	logrus.Info("Enter cluster details")
 	err = clusterDetails(page, filepath.Join(screenshotPath, "01-cluster-details.png"))
 	if err != nil {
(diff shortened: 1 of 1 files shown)


agent/isobuilder/ui_driven_cluster_installation/main.go
```diff
@@ -55,6 +57,11 @@ func main() {
 			screenshotPath = filepath.Join(cwd, ocpDir)
 	}
 
+	// Wait 1 minute before entering cluster details to allow backend initialization
+	// This mimics manual usage where a user takes time to read and fill in the form
+	logrus.Info("Waiting 1 minute to allow backend initialization...")
```
andfasano, line 62: Froma pure logging point of view, until a new LOGS_DIR env var will be set (inherited from the dev-scripts calling environment), and configured for the current logger, I don't think the logs will be visible anywhere
Note: the ds LOGS_DIR already points to a folder gathered at the end of the job and published as CI artifacts, so in that way it will be automatically taken (aka don't log any sensible info)

agent/isobuilder/ui_driven_cluster_installation/main.go
```diff
@@ -55,6 +57,11 @@ func main() {
 			screenshotPath = filepath.Join(cwd, ocpDir)
 	}
 
+	// Wait 1 minute before entering cluster details to allow backend initialization
```
andfasano, line 60: What is the exact problem here? What is the exact backed initialization we're waiting for? 
Is it go-rod too fast? Btw note that the `SlowMotion` could be also used to slow it down, ie:

```
browser := rod.New().
		ControlURL(url).
		SlowMotion(500 * time.Millisecond).
		MustConnect()
```

agent/isobuilder/ui_driven_cluster_installation/main.go
andfasano:
```
logOutputDir     = os.Getenv("LOGDIR")
...
func initLogger() {
	logFileName := "ui_driven_cluster_installation.log"
	logBaseDir := "."
	if logOutputDir != "" {
		logBaseDir = logOutputDir
	}

	file, err := os.OpenFile(filepath.Join(logBaseDir, logFileName), os.O_CREATE|os.O_WRONLY, 0644)
	if err != nil {
		panic(err)
	}
	logrus.SetOutput(io.MultiWriter(os.Stdout, file))
}
```

agent/isobuilder/ui_driven_cluster_installation/main.go
```diff
@@ -36,6 +37,7 @@ var (
 
 func main() {
 	logrus.Info("Launching headless browser...")
+	time.Sleep(1 * time.Minute)
```
andfasano, line 40: Why 1 minute? As soon as the UI is available the user should be able to jump on it and start filling the details

agent/isobuilder/ui_driven_cluster_installation/main.go
```diff
@@ -63,30 +70,53 @@ func main() {
...
+			} else {
+				logrus.Error("Cluster creation timeout - cluster not found after multiple retries")
+				saveFullPageScreenshot(page, filepath.Join(screenshotPath, "02-operators-error.png"))
+				log.Fatal("Failed to load cluster - cluster details not found")
+			}
+		} else {
+			// No error page, cluster loaded successfully
+			logrus.Infof("Operators page loaded successfully (attempt %d)", attempt)
+			break
+		}
```
andfasano, line 98: @bfournie I think at this point this project may deserve some relevant refactoring to make it maintenable and simpler (also to change in case in future enh of the UI). I like the original @pawanpinjarkar ideas of having 1 method per page, and the main should really looks like something extremely simple (meta-code):

```
initLogger()
page := initBrowser()

fillAndSubmitClusterDetails(...)
fillAndSubmitVirtualizationBundle(...)
fillAndSubmitHostDiscovery(...)
fillAndSubmitVerifyStorage(...)
(and so on)
```

Every page submitting method should have its own logic to understand if an error happened or not, and then fill the required fields.
And the logic should be based only on the UI elements (as the user would do)

agent/isobuilder/ui_driven_cluster_installation/main.go
```diff
@@ -63,30 +70,53 @@ func main() {
...
 	}
 
 	logrus.Info("Select virtualization bundle")
+
+	// Save API state for debugging
+	saveClusterAPIState(screenshotPath, "02-operators")
```
andfasano, line 104: I don't think this is required, and potentially overlapping with the agent-gather. We must have a single way of debugging issues, I think we already agreed on adding the SSH key to the ISO for letting the agent-gather to do its work

agent/isobuilder/ui_driven_cluster_installation/main.go
```diff
@@ -63,30 +70,53 @@ func main() {
...
+
 	err = virtualizationBundle(page, filepath.Join(screenshotPath, "02-operators.png"))
 	if err != nil {
 		log.Fatalf("failed to select virtualization bundle: %v", err)
 	}
 
+	// Wait for operators page to finish loading before clicking Next
+	logrus.Info("Waiting for operators page to finish loading...")
+	time.Sleep(5 * time.Second)
```
andfasano, line 113: Also this wait here seems not strictly required

agent/isobuilder/ui_driven_cluster_installation/main.go
```diff
@@ -172,8 +205,126 @@ func virtualizationBundle(page *rod.Page, path string) error {
...
+
+	logrus.Info("Found Next button, waiting for it to be enabled...")
+	logrus.Info("This may take several minutes while hosts complete discovery and validations...")
+
+	// Poll host status while waiting for Next button to be enabled
+	startTime := time.Now()
+	ticker := time.NewTicker(30 * time.Second)
+	defer ticker.Stop()
+
+	go func() {
```
andfasano, line 227: I don't think we should do at all this kind of actions here (polling the AS to infer the status). As the user would do, the expected approach is to wait until:
1) An error is shown in a popup
2) The next button becomes enabled

bfournie:
/hold
Marking as hold for now

bfournie:
Some of the screen progression issues seem to be resolved in 4.22 so closing this for now
/close

PR #1846, opened 2026-01-28
MGMT-22968: Add infrastructure-only mode and Landing Zone VM support

eliorerz:
This commit adds a new `infra_only` Makefile target and support for Landing Zone VMs to dev-scripts. This enables users to create baremetal-like test infrastructure (VMs, networks, BMC emulation) without deploying an OpenShift cluster.

Changes:

1. New `infra_only` Makefile target
   - Runs 'requirements' and 'configure' steps only
   - Stops before cluster deployment (build_installer, ironic, etc.)
   - Useful for testing external deployment tools (Enclave, custom installers)

2. Landing Zone VM support
   - Added NUM_LANDINGZONE environment variable (default: 0)
   - Added landingzone flavor with configurable specs:
     * LANDINGZONE_MEMORY (default: 8192 MB)
     * LANDINGZONE_DISK (default: 60 GB) * LANDINGZONE_VCPU (default: 4)
   - Landing Zone VM acts as deployment host with dual network access

3. Network topology fix for Landing Zone use case
   - Worker VMs: cluster network only (BMC network removed)
   - Landing Zone VM: both BMC and cluster networks
   - Ensures proper network isolation for testing scenarios
   - Only applies when NUM_LANDINGZONE > 0

4. Documentation updates
   - Added "Infrastructure-Only Mode" section to README
   - Added "Deploying with Landing Zone VMs" section to README
   - Documented network topology and use cases

Use case: Testing external deployment tools that need infrastructure without cluster deployment, such as Enclave for bare metal OpenShift installations.

Backward compatibility: All existing targets (default, agent, assisted) work unchanged. New variables default to 0 (no Landing Zone VMs).


eliorerz:
FYI @carbonin

mkowalski:
/approve

AFAIK what you call "landing zone vm" is called "bastion host" and "landing zone" itself is "everything in my cloud subscription what can possibly interact with each other.

But it's just a naming so whatever works for you, works

javipolo (approved):

elfosardo:
/retest
failures do not seem related to the patch

README.md
```diff
@@ -452,6 +475,33 @@ in `config_$USER.sh` (check `config_example.sh` for recommended defaults):
 #export WORKER_VCPU=4
 ```
 
+### Deploying with Landing Zone VMs
+
+Landing Zone VMs are special-purpose VMs that have access to both the BMC (provisioning)
+and cluster (baremetal) networks. They are useful as deployment hosts for external
+tooling that needs to manage worker nodes via BMC while also having access to the
+cluster network.
+
```
dtantsur, line 490: I assume that a full deployment won't work in this topology, and one must use `make infra-only` and something like GORi. Let's clarify that this is not just yet another VM type but rather a whole different architecture.

eliorerz: Sure, I've :

  1. Added a prominent Note at the top clarifying that Landing Zone VMs represent a different deployment architecture, that make (full deployment) is not supported, and that make infra_only + external tooling (e.g., Enclave) must be used instead.
  2. Added an explicit `make infra_only` command block showing what to run after configuration.
  3. Renamed "Worker VMs" to "Cluster VMs (masters/workers)" in the network topology list for clarity.
  4. Improved the closing paragraph to describe the Landing Zone VM as a bridge between the BMC management plane and the cluster network.

PR #1853, opened 2026-02-13
Add config templates and use-template.sh for common deployment scenarios

elfosardo:
Add pre-built configuration templates under config-templates/ covering IPI (ipv4, ipv6, dualstack, compact) and agent-based (SNO, HA, compact) deployments. Templates only set values that override defaults, so they stay minimal and don't drift as defaults change.

The use-template.sh helper copies a template to config_$USER.sh and preserves the CI_TOKEN across template switches.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,17 @@ Considering that this is a new install on a clean OS, the next tasks should be p
 
 5. Create a config file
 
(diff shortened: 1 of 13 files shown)


hroyrh:
Looks good.
/lgtm

PR #1866, opened 2026-03-13
Use set -euxo pipefail in 01_install_requirements

elfosardo:
Replace set -ex with set -euxo pipefail for consistency with
02_configure_host.sh and 04_setup_ironic.sh.

Guard optional environment variables with :- defaults to avoid
nounset errors and fix a $ARCH typo that should be $GOARCH.

Also fix common.sh to handle the case where Go is not yet installed
when sourced by 01_install_requirements.sh — skip 'go env' if the
go binary is absent and default GOPATH to empty.

### DIFF
--- a/01_install_requirements.sh
+++ b/01_install_requirements.sh
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-set -ex
+set -euxo pipefail
 
 
 source logging.sh
@@ -10,7 +10,7 @@ source validation.sh
 
 early_deploy_validation true
 
-if [ -z "${METAL3_DEV_ENV}" ]; then
+if [ -z "${METAL3_DEV_ENV:-}" ]; then
   export REPO_PATH=${WORKING_DIR}
   sync_repo_and_patch metal3-dev-env https://github.com/metal3-io/metal3-dev-env.git
   pushd ${METAL3_DEV_ENV_PATH}
@@ -153,7 +153,7 @@ GO_CHECKSUM="$(
 )"
 
 if [ -z "$GO_CHECKSUM" ]; then
-  echo "Error: Could not find checksum for $VERSION ($OS/$ARCH)" >&2
+  echo "Error: Could not find checksum for $VERSION ($OS/$GOARCH)" >&2
 else
   echo "Checksum: $GO_CHECKSUM"
 fi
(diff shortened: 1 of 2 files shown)


01_install_requirements.sh
```diff
@@ -41,7 +41,7 @@ sudo sh -c "echo 'max_parallel_downloads=8' >> /etc/dnf/dnf.conf"
 # in the upgrade command,but this is more explicit and complete
 sudo dnf -y clean all
 
-old_version=$(sudo dnf info NetworkManager | grep Version | cut -d ':' -f 2)
+old_version=$(sudo dnf info NetworkManager | { grep Version || true; } | cut -d ':' -f 2)
```
dtantsur: How can `grep` fail if `dnf info` does not fail? In fact, this will work even for packages that are not installed.

elfosardo: right, dnf will spit out the Version as long as it knows about the package
I'll remove this

common.sh
```diff
@@ -6,9 +6,10 @@ export PATH="/usr/local/go/bin:$HOME/.local/bin:$PATH"
 # Set a PS4 value which logs the script name and line #.
 export PS4='+(${BASH_SOURCE}:${LINENO}): ${FUNCNAME[0]:+${FUNCNAME[0]}(): }'
 
-eval "$(go env)"
-
-export PATH="${GOPATH}/bin:$PATH"
+if command -v go &>/dev/null; then
+    eval "$(go env)"
+fi
+export PATH="${GOPATH:-}/bin:$PATH"
```
dtantsur: nit: this could go under `if` block

common.sh
```diff
@@ -169,7 +170,7 @@ if [ -z "${OPENSHIFT_RELEASE_IMAGE:-}" ]; then
   fi
 fi
 export OPENSHIFT_RELEASE_IMAGE="${OPENSHIFT_RELEASE_IMAGE:-$LATEST_CI_IMAGE}"
-export OPENSHIFT_INSTALL_PATH="${OPENSHIFT_INSTALL_PATH:-$GOPATH/src/github.com/openshift/installer}"
+export OPENSHIFT_INSTALL_PATH="${OPENSHIFT_INSTALL_PATH:-${GOPATH:-}/src/github.com/openshift/installer}"
```
dtantsur: If GOPATH is not set, the result will be meaningless, so we can just as well not set it at all

01_install_requirements.sh
```diff
@@ -153,7 +153,7 @@ GO_CHECKSUM="$(
 )"
 
 if [ -z "$GO_CHECKSUM" ]; then
-  echo "Error: Could not find checksum for $VERSION ($OS/$ARCH)" >&2
+  echo "Error: Could not find checksum for $VERSION ($OS/$GOARCH)" >&2
```
zaneb, line 156: I believe this line is currently unreachable with `set -u`

zaneb (commented):
/approve

01_install_requirements.sh
elfosardo: mmm why?
set -u triggers on unset variables only, not empty ones
GO_CHECKSUM is always set on L141, which could be empty

PR #1867, opened 2026-03-18
NO-JIRA: Set agent DISABLE_IMAGE_POLICY for local installs

bfournie:
In CI, for agent-based installations the environment variable OPENSHIFT_INSTALL_EXPERIMENTAL_DISABLE_IMAGE_POLICY is set and this is passed down to assisted-service. When doing local testing we also need to set this environment variable.

### DIFF
--- a/common.sh
+++ b/common.sh
@@ -181,6 +181,12 @@ if [[ -z "$OPENSHIFT_CI" ]]; then
   export OPENSHIFT_VERSION=${OPENSHIFT_VERSION:-$(echo $OPENSHIFT_RELEASE_IMAGE | sed "s/.*:\([[:digit:]]\.[[:digit:]]*\).*/\1/")}
 fi
 
+# Disable image policy verification for local development
+# In CI, this is set via the test configuration (see https://github.com/openshift/installer/pull/10379)
+if [ "${OPENSHIFT_CI}" != "true" ] && [ "${OPENSHIFT_RELEASE_TYPE}" = "nightly" ]; then
+    export OPENSHIFT_INSTALL_EXPERIMENTAL_DISABLE_IMAGE_POLICY="true"
+fi
+
 export OPENSHIFT_RELEASE_TAG=$(echo $OPENSHIFT_RELEASE_IMAGE | sed -E 's/[[:alnum:]\/.-]*(release|okd).*://')
 
 # Use "ipmi" for 4.3 as it didn't support redfish, for other versions


common.sh
```diff
@@ -181,6 +181,12 @@ if [[ -z "$OPENSHIFT_CI" ]]; then
   export OPENSHIFT_VERSION=${OPENSHIFT_VERSION:-$(echo $OPENSHIFT_RELEASE_IMAGE | sed "s/.*:\([[:digit:]]\.[[:digit:]]*\).*/\1/")}
 fi
 
+# Disable image policy verification for local development
+# In CI, this is set via the test configuration (see https://github.com/openshift/installer/pull/10379)
+if [ "${OPENSHIFT_CI}" != "true" ]; then
+    export OPENSHIFT_INSTALL_EXPERIMENTAL_DISABLE_IMAGE_POLICY="true"
```
andfasano, line 187: This does not seem correct for all the situation. In particular for the release jobs (or nightly) we want to test a signed payload, isn't it? (I mean locally)

bfournie: For nightly builds we'll get a failure if we don't set this. This is required to be set because nightly images are not signed. It is set in CI. 

For CI builds this setting is not needed. I can add a clause in to only use for nightly builds.

andfasano: Shouldn't be required also for the [stable](https://amd64.ocp.releases.ci.openshift.org/#4-stable) stream? Or viceversa, it could be applied only for the [CI](https://amd64.ocp.releases.ci.openshift.org/#4.22.0-0.ci) stream

bfournie: Do you mean stable e.g. quay.io/openshift-release-dev/ocp-release:4.21.7-x86_64 ?
No we don't need to disable the signature checking on that, only on nightly builds when we are testing locally

bfournie:
I've tested this with:
1. 
```
export OPENSHIFT_RELEASE_STREAM=4.22
export OPENSHIFT_RELEASE_TYPE=ci
```
OPENSHIFT_INSTALL_EXPERIMENTAL_DISABLE_IMAGE_POLICY is not set and not needed for install.

2. 
`export OPENSHIFT_RELEASE_IMAGE=quay.io/openshift-release-dev/ocp-release:4.22.0-ec.3-x86_64`
OPENSHIFT_INSTALL_EXPERIMENTAL_DISABLE_IMAGE_POLICY is not set and not needed for install.

3. 
```
export OPENSHIFT_RELEASE_STREAM=4.21
export OPENSHIFT_RELEASE_TYPE=nightly
```
OPENSHIFT_INSTALL_EXPERIMENTAL_DISABLE_IMAGE_POLICY not needed for install.

4. 
```
export OPENSHIFT_RELEASE_STREAM=4.22
export OPENSHIFT_RELEASE_TYPE=nightly
```
OPENSHIFT_INSTALL_EXPERIMENTAL_DISABLE_IMAGE_POLICY set and needed for install, otherwise installation will fail.

common.sh
andfasano: ok thanks

zaneb:
I think this is handled now by #1865

bfournie:
> I think this is handled now by #1865

ABI won't get that setting though as its in create_cluster()

PR #1869, opened 2026-03-19
NO-JIRA: Set customizable TNA NUM_WORKERS in common.sh

fracappa:
Changing NUM_WORKERS env var in TNA to be customizable.
In this way TNA clusters with day-0 workers could be installed using dev-scripts


common.sh
dtantsur: Do you want to change this for consistency?

eggfoobar: That sounds like a great idea to future proof it, we might want to validate workers for TNF in the future, @fracappa feel free to add that as well

PR #1880, opened 2026-04-02
CORENET-6097: Open BGP port 179 for no-overlay mode

zhaozhanqi:
Add firewall rule to open port 179/tcp in the libvirt zone during host configuration. This port is required for BGP routing when no-overlay mode is enabled for the default network.

### DIFF
--- a/02_configure_host.sh
+++ b/02_configure_host.sh
@@ -122,6 +122,9 @@ sudo systemctl enable --now firewalld
 # access.
 configure_chronyd
 
+# Open BGP port for no-overlay mode enabled for default network
+sudo firewall-cmd --zone=libvirt --add-port=179/tcp
+
 export VNC_CONSOLE=true
 if [[ $(uname -m) == "aarch64" ]]; then
   VNC_CONSOLE=false


zhaozhanqi:
@mkowalski Hi, could you help take a look this? we need this patch to setup no-overlay cluster in our CI. thanks.

zhaozhanqi:
@bfournie @sadasu  Would you like to help review this?  thanks

zhaozhanqi:
@dtantsur Thanks for helping add the test label
Could you help check this can be approved? I don't think the failed job is not related the changes

02_configure_host.sh
```diff
@@ -122,6 +122,9 @@ sudo systemctl enable --now firewalld
 # access.
 configure_chronyd
 
+# Open BGP port for no-overlay mode enabled for default network
+sudo firewall-cmd --zone=libvirt --add-port=179/tcp
```
mkowalski, line 126: I don't like the fact it does not persist the reboot but I checked and seems like almost everyone is doing it non-persistent way here (https://github.com/search?q=repo%3Aopenshift-metal3%2Fdev-scripts%20firewall-cmd&type=code)

02_configure_host.sh
```diff
@@ -122,6 +122,9 @@ sudo systemctl enable --now firewalld
 # access.
 configure_chronyd
 
+# Open BGP port for no-overlay mode enabled for default network
```
tssurya, line 125: why is this only a requirement for default network/no-overlay?
don't we need to open this port in case of north-south traffic scenario for RouteAdvertisements with CUDN?

tssurya (approved):
/lgtm

tssurya:
@dtantsur or @sadasu PTAL

PR #1883, opened 2026-04-22
AGENT-1553: Add operator images to OVE mirror imageset

rwsu:
When AGENT_E2E_TEST_BOOT_MODE=ISO_NO_REGISTRY, append the operators section from the agent-iso-builder appliance-config.yaml to the oc-mirror imageset so operator images are included in the mirror and available to the appliance builder via --mirror-path.

Uses podman cp to extract the config from the iso-builder container image (which has no shell or python3) and parses it with python3 on the host.

Assisted-by: Claude Sonnet 4.6 <noreply@anthropic.com>

### DIFF
--- a/oc_mirror.sh
+++ b/oc_mirror.sh
@@ -81,6 +81,58 @@ mirror:
   - name: registry.redhat.io/rhel9/support-tools:latest
 EOF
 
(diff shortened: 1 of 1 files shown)


bfournie:
/retitle AGENT-1553: Add operator images to OVE mirror imageset

PR #1885, opened 2026-04-23
NO-JIRA: limit fencing credentials to master nodes only.

fracappa:
The agent install config-template iterated over all BMH nodes when generating fencing credentials. With NUM_WORKERS [now](https://github.com/openshift-metal3/dev-scripts/pull/1869) customizable in TNA/TNF scenario, this produces credentials for worker nodes too, causing install validation failures.


agent/roles/manifests/templates/install-config_baremetal_yaml.j2
```diff
@@ -38,7 +38,7 @@ controlPlane:
 {% if enable_two_node_fencing %}
   fencing:
     credentials:
-{% for hostname in hostnames %}
+{% for hostname in hostnames[:num_masters|int] %}
```
andfasano: There's a long chain but IINW this derives from `AGENT_NODES_HOSTNAMES`. As a quick fix it could work, but I won't give it for granted the ordering (first masters, then workers), so I'd evaluate also a more robust solution

fracappa: Hi @andfasano, thanks for the feedback!
I agree with you that the previous solution worked as a quick fix but it requires granted ordering in the node evaluation. Just pushed a commit that introduces dedicated master-only BMC arrays.

PR #1889, opened 2026-04-28
NO-JIRA: fix: set WORKER_DISK default to 100GB for TNA/TNF scenarios

fracappa:
Set `WORKER_DISK` default to 100GB for TNA and TNF scenarios. 
 
 Previously these scenarios didn't set `WORKER_DISK` because workers were hardcoded to 0.
 
After #1869 made `NUM_WORKERS` customizable, workers now inherit the global default of 60GB, which may be insufficient for OpenShift node.


dtantsur:
/ok-to-test
/lgtm

I wonder if we need to change the default globally.

fracappa:
@dtantsur I did have the same concern.

iurygregory:
@elfosardo @tdomnesc can you take a look? Thanks

PR #1891, opened 2026-05-12
OCPBUGS-85432: Increase resources for lvm in agent iso-no-registry

bfournie:
By default the lvm-operator is now included as a dependency for CNV in the iso-no-registry test. The MASTER_VCPU must be increased to 10 in order for the master roles to be correctly assigned. In addition an additional disk must be created, as LVM requires at least one non-installation HDD/SSD disk on the host.


bfournie:
Tested iso-no-registry with https://github.com/openshift/assisted-service/pull/10285 and confirmed successful installation.

PR #1892, opened 2026-05-12
Add WORKER_POOL_OS_IMAGE_STREAM

honza:
Allow deploying worker nodes with a different CoreOS stream than the default.
When WORKER_POOL_OS_IMAGE_STREAM is set (e.g. "rhel-10"), the worker BMH
manifests and MachineSet hostSelector are configured to use the specified
stream label, and a MachineConfigPool manifest is generated to set
osImageStream for the worker pool.

This enables testing heterogeneous clusters where workers boot with a
different IPA and on-disk OS stream than control-plane nodes.

This depends on https://github.com/openshift/installer/pull/10721

### DIFF
--- a/config_example.sh
+++ b/config_example.sh
@@ -619,6 +619,16 @@ set -x
 #export WORKER_DISK=60
 #export WORKER_VCPU=4
 
+# WORKER_COREOS_STREAM -
+# Override the CoreOS stream label on worker BareMetalHost manifests and the
+# worker MachineSet hostSelector. This controls which IPA (kernel/initrd/rootfs)
+# is used during PXE boot for worker nodes. For example, set to "rhel-10" to
+# boot workers with the RHEL 10 IPA while masters remain on the default stream.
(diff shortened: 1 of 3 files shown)


config_example.sh
```diff
@@ -619,6 +619,16 @@ set -x
 #export WORKER_DISK=60
 #export WORKER_VCPU=4
 
+# WORKER_COREOS_STREAM -
+# Override the CoreOS stream label on worker BareMetalHost manifests and the
+# worker MachineSet hostSelector. This controls which IPA (kernel/initrd/rootfs)
+# is used during PXE boot for worker nodes. For example, set to "rhel-10" to
+# boot workers with the RHEL 10 IPA while masters remain on the default stream.
+# Requires matching stream-prefixed images in the machine-os-images container.
+# Default: unset (use the installer's default, typically "rhel-9").
```
zaneb: typically `rhel-10` now

honza:
/hold

I'm adding some helpers to the installer to make this easier.

PR #1894, opened 2026-05-15
AGENT-1429: Add support for dual stack ISO_NO_REGISTRY

pawanpinjarkar:
Add support for dual stack in UI automation for ISO_NO_REGISTRY

### DIFF
--- a/agent/isobuilder/ui_driven_cluster_installation/main.go
+++ b/agent/isobuilder/ui_driven_cluster_installation/main.go
@@ -26,6 +26,28 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+const (
+	defaultControlPlaneCount = 3
+	downloadAttempts         = 3
+)
+
+var (
+	downloadLogsFlag = flag.Bool("download-logs", false, "Download installation logs from UI and exit")
+	clusterName      = os.Getenv("CLUSTER_NAME")
+	baseDomain       = os.Getenv("BASE_DOMAIN")
+	rendezvousIP     = os.Getenv("RENDEZVOUS_IP")
+	ocpDir           = os.Getenv("OCP_DIR")
+	apiVips          = os.Getenv("API_VIPS")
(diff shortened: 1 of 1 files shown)


agent/isobuilder/ui_driven_cluster_installation/main.go
```diff
@@ -395,6 +404,9 @@ func logConfiguration() {
 	logrus.Infof("FEATURE_SET=%s", featureSet)
 	logrus.Infof("FEATURE_GATES=%s", featureGates)
 	logrus.Infof("CONTROL_PLANE_COUNT=%d", controlPlaneCount)
+	if getScenarioType() == "HA" {
+		logrus.Infof("COMPUTE_PLANE_COUNT=%d", workers)
```
bfournie: This should handle a string from os.Getenv("NUM_WORKERS")
```suggestion
		logrus.Infof("COMPUTE_PLANE_COUNT=%s", workers)
```

agent/isobuilder/ui_driven_cluster_installation/main.go
```diff
@@ -507,8 +519,37 @@ func networkingDetails(page *rod.Page, path string) error {
...
-	page.MustElement("#form-input-ingressVips-0-ip-field").MustInput(ingressVips)
+	if ipStack == "v4v6" {
+		dualStackRadio := page.MustElement("#form-radio-stackType-dualStack-field")
+		dualStackRadio.MustClick()
+		err = saveFullPageScreenshot(page, timestampedPath(path, "after-dualstack"))
+		if err != nil {
+			return err
+		}
+		apiVips = strings.Split(apiVips,",")[0]
+		ingressVips = strings.Split(ingressVips,",")[0]
```
bfournie: This should local variables instead of shadowing the global variables, e.g.[](url)
```suggestion
        apiVipList := strings.Split(apiVips, ",")[0]
		ingressVipList := strings.Split(ingressVips, ",")[0]
```
And use those values below.
Also need to add a space as above.

PR #1896, opened 2026-05-19
AGENT-1520: Use local copy of agent-installer-utils to build OVE ISO

bfournie:
Use a local copy of AIU when creating the ISO to allow overrides of appliance-config etc.


bfournie:
ci/prow/e2e-agent-compact-ipv4 is failing because of https://redhat.atlassian.net/browse/OCPBUGS-88701. Overriding

PR #1898, opened 2026-05-29
METAL-1845: shellcheck fixes part 1

tdomnesc:


01_install_requirements.sh
```diff
@@ -118,7 +118,7 @@ esac
 # for CentOS/RHEL so we have to install from pip. We do not want to
 # overwrite an existing installation of the golang version, though,
 # so check if we have a yq before installing.
-if ! which yq 2>&1 >/dev/null; then
+if ! which yq >/dev/null 2>&1; then
```
zaneb, line 121: nit: if you do
```suggestion
if ! which yq &>/dev/null; then
```
then nobody ever has to think about whether this is the right order or not (it is now) again.

sanitychecks.sh
```diff
@@ -13,26 +13,26 @@ function verifyClean {
...
     exit 1 
   fi
 }
 
 function verifyFreeSpace {
-  AVAIL=$(df -h --output=avail -B 1G $WORKING_DIR | tail -n 1)
+  AVAIL=$(df -h --output=avail -B 1G "$WORKING_DIR" | tail -n 1)
 
-  if (( $AVAIL < $MIN_SPACE_REQUIRED )); then 
+  if (( "$AVAIL" < "$MIN_SPACE_REQUIRED" )); then
```
zaneb:
```suggestion
  if (( AVAIL < MIN_SPACE_REQUIRED )); then
```

utils.sh
```diff
@@ -122,58 +123,58 @@ function create_cluster() {
...
                   # Get digest of the pushed image
-                  DIGEST=$(podman inspect --format "{{.Digest}}" $IMAGE_MIRRORED)
+                  DIGEST=$(podman inspect --format "{{.Digest}}" "$IMAGE_MIRRORED")
                   IMAGE_MIRRORED="${IMAGE_MIRRORED}@${DIGEST}"
                 fi
-                sed -i -e "s%${IMAGE}%${IMAGE_MIRRORED}%g" $ASSET_NEW
+                sed -i -e "s%${IMAGE}%${IMAGE_MIRRORED}%g" "$ASSET_NEW"
             done
-        done
+        done < <(find "${ASSETS_EXTRA_FOLDER}" \( -name \*.yml -or -name \*.yaml \) -print0)
```
zaneb, line 177: yowza

utils.sh
```diff
@@ -239,20 +240,21 @@ function node_map_to_install_config_hosts() {
     start_idx="$2"
     role="$3"
 
-    for ((idx=$start_idx;idx<$(($1 + $start_idx));idx++)); do
+    for ((idx=start_idx;idx<$((num_hosts + start_idx));idx++)); do
```
zaneb:
```suggestion
    for ((idx=start_idx; idx < num_hosts + start_idx; idx++)); do
```

utils.sh
```diff
@@ -334,8 +336,9 @@ function node_map_to_install_config_fencing_credentials() {
   fencing:
     credentials:
 EOF
-    for ((idx=0;idx<$(($NUM_MASTERS));idx++)); do
-      hostname="$(printf $MASTER_HOSTNAME_FORMAT ${idx})"
+    for ((idx=0;idx<$((NUM_MASTERS));idx++)); do
```
zaneb:
```suggestion
    for ((idx=0; idx < NUM_MASTERS; idx++)); do
```

zaneb (commented):
"part 1 of 754" 😆
/approve

utils.sh
tdomnesc: What do you mean here?

01_install_requirements.sh
tdomnesc: First form is more visible and and is everywhere in the repo, let's leave it like this.

utils.sh
elfosardo: it's less intuitive than the FOR loop, but definitely "more" correct

PR #1903, opened 2026-06-03
METAL-1845: shellcheck fixes part 3

tdomnesc:


agent/e2e/agent-tui/utils.sh
```diff
@@ -69,23 +69,25 @@ function pressKeys(){
   if [[ -n "$node" ]]; then
     echo "$msg $text on node $node"
   else
-   echo $msg
+   echo "$msg"
   fi
 
+  # shellcheck disable=SC2034
```
MahnoorAsghar, line 75: Arent these variables unused?

tdomnesc: They are but I'm not trying to fix bugs here, just get shellcheck compliance.
Especially in the agent part where I don't know the author's intention.

PR #1908, opened 2026-06-09
OCPEDGE-2737: Enable aarch64 native KVM support for agent-based installation

fonta-rh:
## Summary

Enable IPI and agent-based installation (ABI) on native aarch64 KVM hosts (e.g. AWS Graviton bare metal). Five independent fixes that together unblock the full deployment path on aarch64:

1. **CPU model fix** — Narrow the `{% if is_aarch64 %}` conditional in the VM template so native KVM falls through to `host-passthrough` instead of the emulation-only `cortex-a57`
2. **Hardcoded x86_64 references** — Replace `x86_64` strings with `$(uname -m)` / `${ARCH}` across agent scripts and RHCOS boot image resolution
3. **CDROM bus fix** — Set `target.bus` explicitly on all `virt-xml` CDROM attachment calls (`scsi` on aarch64, `sata` on x86_64) to prevent the `virt` machine type from defaulting to USB (which doesn't exist)
4. **Go tarball arch fix** — Patch `metal3-dev-env` Ansible defaults after checkout to use `{{ GOARCH }}` instead of hardcoded `amd64` in the Go download URL, so the correct arm64 tarball is fetched on aarch64 hosts
5. **CDROM eject on aarch64** — After VMs boot from the agent ISO, eject the CDROM media on aarch64 hosts. AAVMF firmware on the `virt` machine type does not reliably honor `boot_order` for SCSI CDROM vs virtio disk, causing VMs to reboot into the ISO instead of from disk after the agent writes the OS

## Changes

| File | Fix |
|------|-----|
| `02_configure_host.sh` | Narrow CPU model conditional for native aarch64 KVM |
| `agent/06_agent_create_cluster.sh` | Add `CDROM_BUS` variable; set `target.bus` on all 5 `virt-xml` CDROM lines; add `eject_agent_iso()` function for aarch64 |
| `agent/common.sh` | PXE boot filename: `x86_64` → `$(uname -m)` |
| `agent/iscsi_utils.sh` | iSCSI boot filenames (DHCP bootp + iPXE) |
| `agent/07_agent_add_extraworker_nodes.sh` | Extra worker node ISO filename |
| `agent/iso_no_registry.sh` | OVE ISO cleanup exclusion pattern |
| `agent/01_agent_requirements.sh` | oc-mirror download URL |
| `rhcos.sh` | RHCOS format key lookup (`.architectures.x86_64` → host arch) |
| `01_install_requirements.sh` | Sed patch for metal3-dev-env Go tarball arch |

## Test plan

- [x] Full OCP 4.22 fencing-IPI deployment on AWS c7g.metal (Graviton3, native aarch64 KVM) — 2m+1a arbiter, 50m41s, 35/35 COs Available
- [x] Both nodes + arbiter Ready, etcd stable at revision 8
- [x] Verified x86_64 deployments unaffected (all substitutions evaluate to `x86_64` / `sata` on x86 hosts; Go sed patch is no-op when arch matches; CDROM eject is gated on aarch64)
- [x] `02_configure_host.sh`: `<os>`, `<features>`, VNC conditionals still fire correctly (only CPU block narrowed)
- [x] Go 1.24.10 arm64 installed via patched metal3-dev-env, checksum verified
- [x] ABI end-to-end on aarch64 with CDROM eject fix

Supersedes #1910.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/01_install_requirements.sh
+++ b/01_install_requirements.sh
@@ -26,8 +26,12 @@ if [ -z "${METAL3_DEV_ENV:-}" ]; then
   sed -i '/ANSIBLE_VERSION/{ s/10\.7\.0/8.7.0/; }' lib/common.sh
 
   # Go tarball defaults hardcode linux-amd64; use GOARCH passed as extra var.
-  sed -i 's/go_tarball: "go{{ go_version }}.linux-amd64.tar.gz"/go_tarball: "go{{ go_version }}.linux-{{ GOARCH | default('\''amd64'\'') }}.tar.gz"/' \
-    vm-setup/roles/packages_installation/defaults/main.yml
+  # Upstream fix: https://github.com/metal3-io/metal3-dev-env/pull/1694
+  GO_DEFAULTS="vm-setup/roles/packages_installation/defaults/main.yml"
+  if grep -q 'linux-amd64' "${GO_DEFAULTS}"; then
+    sed -i 's/go_tarball: "go{{ go_version }}.linux-amd64.tar.gz"/go_tarball: "go{{ go_version }}.linux-{{ GOARCH | default('\''amd64'\'') }}.tar.gz"/' \
+      "${GO_DEFAULTS}"
+  fi
 
   popd
 fi
--- a/01_install_requirements.sh
+++ b/01_install_requirements.sh
@@ -25,6 +25,10 @@ if [ -z "${METAL3_DEV_ENV:-}" ]; then
   # Patch metal3-dev-env to use Ansible 8.x on centos9/rhel9.
   sed -i '/ANSIBLE_VERSION/{ s/10\.7\.0/8.7.0/; }' lib/common.sh
 
+  # Go tarball defaults hardcode linux-amd64; use GOARCH passed as extra var.
+  sed -i 's/go_tarball: "go{{ go_version }}.linux-amd64.tar.gz"/go_tarball: "go{{ go_version }}.linux-{{ GOARCH | default('\''amd64'\'') }}.tar.gz"/' \
+    vm-setup/roles/packages_installation/defaults/main.yml
+
   popd
 fi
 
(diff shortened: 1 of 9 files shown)


01_install_requirements.sh
```diff
@@ -25,6 +25,10 @@ if [ -z "${METAL3_DEV_ENV:-}" ]; then
   # Patch metal3-dev-env to use Ansible 8.x on centos9/rhel9.
   sed -i '/ANSIBLE_VERSION/{ s/10\.7\.0/8.7.0/; }' lib/common.sh
 
+  # Go tarball defaults hardcode linux-amd64; use GOARCH passed as extra var.
+  sed -i 's/go_tarball: "go{{ go_version }}.linux-amd64.tar.gz"/go_tarball: "go{{ go_version }}.linux-{{ GOARCH | default('\''amd64'\'') }}.tar.gz"/' \
+    vm-setup/roles/packages_installation/defaults/main.yml
```
dtantsur: Could you submit this change upstream and adjust this workaround to account for the future fix?

02_configure_host.sh
```diff
@@ -129,6 +129,15 @@ export VNC_CONSOLE=true
 if [[ $(uname -m) == "aarch64" ]]; then
   VNC_CONSOLE=false
   echo "libvirt_cdrombus: scsi" >> vm_setup_vars.yml
+  # On native aarch64 KVM (e.g. AWS Graviton), the upstream metal3-dev-env
+  # template hardcodes cortex-a57 for all aarch64 VMs. That model only works
+  # under qemu emulation; native KVM requires host-passthrough. Narrow the
+  # CPU conditional so native aarch64 falls through to host-passthrough,
+  # without affecting the <os> or <features> sections that must still fire.
+  TEMPLATE="${VM_SETUP_PATH}/roles/libvirt/templates/baremetalvm.xml.j2"
```
dtantsur, line 138: Same

PR #1909, opened 2026-06-09
AGENT-1534: Change memory limits for agent

bfournie:
Change the limits for HA to fix oversubscription issues.

### DIFF
--- a/common.sh
+++ b/common.sh
@@ -514,11 +514,11 @@ if [[ ! -z ${AGENT_E2E_TEST_SCENARIO} ]]; then
           export NUM_MASTERS=3
           export MASTER_VCPU=${MASTER_VCPU:-4}
           export MASTER_DISK=${MASTER_DISK:-100}
-          export MASTER_MEMORY=${MASTER_MEMORY:-32768}
+          export MASTER_MEMORY=${MASTER_MEMORY:-24576}
           export NUM_WORKERS=2
           export WORKER_VCPU=${WORKER_VCPU:-4}
           export WORKER_DISK=${WORKER_DISK:-100}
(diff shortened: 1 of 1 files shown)


common.sh
```diff
@@ -514,11 +514,11 @@ if [[ ! -z ${AGENT_E2E_TEST_SCENARIO} ]]; then
           export NUM_MASTERS=3
           export MASTER_VCPU=${MASTER_VCPU:-4}
           export MASTER_DISK=${MASTER_DISK:-100}
-          export MASTER_MEMORY=${MASTER_MEMORY:-32768}
+          export MASTER_MEMORY=${MASTER_MEMORY:-24576}
           export NUM_WORKERS=2
           export WORKER_VCPU=${WORKER_VCPU:-4}
           export WORKER_DISK=${WORKER_DISK:-100}
-          export WORKER_MEMORY=${WORKER_MEMORY:-9000}
+          export WORKER_MEMORY=${WORKER_MEMORY:-16384}
```
andfasano, line 521: No need of line 572 anymore

bfournie: Yep, good catch.

zaneb, line 521: I don't understand how increasing this will help. If the theory is that all of the kernels are allocating all of the pages available to them as cache and causing memory contention on the hypervisor that is causing the bootstrap to stall, then the solution is to allocate less memory. This change does that in total, but increasing the memory for workers doesn't contribute to that, it just makes it harder to detect regressions when we require more RAM to install than what we say we do.

bfournie: Agreed, this really didn't contribute to the fix, just came about when we sized the existing machines we realized we had additional overhead for workers. We can back this out, either as separate commit or as part of the other sizing change.

PR #1911, opened 2026-06-09
METAL-1863: Convert install-config generation from bash heredoc to Ansible

elfosardo:
Replace the generate_ocp_install_config() bash heredoc in ocp_install_env.sh with an Ansible role and Jinja2 template.

05_create_install_config.sh now sources the existing config chain for environment variable resolution, generates YAML extra-vars via a config bridge script, and calls ansible-playbook to render install-config.yaml.

This is the first step toward migrating dev-scripts from imperative bash to declarative Ansible roles. The extra worker manifest generation remains in bash for now.

Assisted-By: Claude Opus 4.6


05_create_install_config.sh
```diff
@@ -14,22 +14,79 @@ early_deploy_validation
...
 
-    # Create a nodes.json file
-    mkdir -p "${OCP_DIR}"
-    jq '{nodes: .}' "${NODES_FILE}" | tee "${BAREMETALHOSTS_FILE}"
+    # Generate base config vars from the resolved bash environment
+    SCRIPTDIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+    "${SCRIPTDIR}/ansible/vars_bridge.sh"
 
-    # Create install config for openshift-installer
-    generate_ocp_install_config "${OCP_DIR}"
```
zaneb, line 29: Can we delete this function now?

05_create_install_config.sh
```diff
@@ -14,22 +14,79 @@ early_deploy_validation
...
+    fi
+
+    # Proxy variables (when INSTALLER_PROXY is set)
+    if [[ -n "${INSTALLER_PROXY:-}" ]]; then
+        cat >> "${EXTRA_VARS_FILE}" << PROXY_EOF
+devscripts_http_proxy: "${HTTP_PROXY}"
+devscripts_https_proxy: "${HTTPS_PROXY}"
+devscripts_no_proxy: "${NO_PROXY}"
+PROXY_EOF
+    fi
```
zaneb, line 79: It, uh, feels a bit like like we replaced the install-config heredoc with an equally long ansible config heredoc + 500 lines of YAML 😬

PR #1912, opened 2026-06-12
AGENT-1522: Add internalreleaseimage manifest with v1 version

pawanpinjarkar:
Temporarily adds internalreleaseimage extra manifest with `machineconfiguration.openshift.io/v1` version


andfasano:
@pawanpinjarkar unfortunately this is not sufficient, because the extra manifest must be placed inside the ISO

PR #1913, opened 2026-06-15
AGENT-1522: override IRI manifest

andfasano:
Temporarily adds internalreleaseimage extra manifest with machineconfiguration.openshift.io/v1 version

(replaces https://github.com/openshift-metal3/dev-scripts/pull/1912)


andfasano:
This approach wasn't properly working with the multi-pr command

PR #1916, opened 2026-06-19
OCPEDGE-2934: feat: add macAddress as alternative identifier for fencing credentials

fracappa:
Introduce FENCING_CREDENTIALS_IDENTIFIER env var to control whether fencing credentials identify nodes by hostname (default) or macAddress.


MahnoorAsghar:
Looks alright to me :)

agent/05_agent_configure.sh
```diff
@@ -130,6 +130,9 @@ function configure_node() {
       AGENT_NODES_IPSV6+=("$ipv6")
     fi
     AGENT_NODES_MACS+=("$node_mac")
+    if [[ "$node_type" == "master" ]]; then
+      AGENT_MASTER_MACS+=("$node_mac")
```
lucaconsalvi, line 134: **Bug:** `AGENT_MASTER_MACS` is populated here in `configure_node()` (called by `get_static_ips_and_macs()`), but the alternative code path — `get_baremetal_ips_and_macs()` — never declares or populates this array.

Since `generate_cluster_manifests()` unconditionally serializes it (`"${AGENT_MASTER_MACS[@]}"`), and the script runs with `set -euxo pipefail`, this will crash with `unbound variable` on every `NODES_PLATFORM=baremetal` deployment — even with the default `hostname` mode.

Suggested fix — mirror the `AGENT_MASTER_HOSTNAMES` pattern in `get_baremetal_ips_and_macs()`:
- Add `AGENT_MASTER_MACS=()` to the initialization block
- Add `AGENT_MASTER_MACS+=("$mac")` inside the `if (( i < NUM_MASTERS ))` block

Also worth adding `AGENT_MASTER_MACS=()` to the init block in `get_static_ips_and_macs()` to match the established pattern.

common.sh
```diff
@@ -412,6 +412,9 @@ if [[ -z ${AGENT_E2E_TEST_SCENARIO:-} ]] && [[ ${NUM_ARBITERS} -eq 0 ]] && [[ ${
   export ENABLE_TWO_NODE_FENCING="true"
 fi
 
+# Controls whether fencing credentials use "hostname" or "macAddress" to identify nodes
+export FENCING_CREDENTIAL_IDENTIFIER=${FENCING_CREDENTIAL_IDENTIFIER:-hostname}
```
lucaconsalvi, line 416: **Suggestion:** Any value other than the exact string `macAddress` silently falls through to hostname mode in both code paths. A typo like `macaddress` or `mac` would be silently ignored.

The `BMC_DRIVER` validation just below establishes a precedent for early validation — worth adding a similar guard:

utils.sh
```diff
@@ -468,19 +468,24 @@ function node_map_to_install_config_fencing_credentials() {
...
-      if [[ $IP_STACK != 'v4' ]]; then
-        hostname="${hostname}.${CLUSTER_DOMAIN}"
-      fi
       username=$(node_val ${idx} "driver_info.username")
       password=$(node_val ${idx} "driver_info.password")
       address=$(node_val ${idx} "driver_info.address")
 
+      if [[ "${FENCING_CREDENTIAL_IDENTIFIER}" == "macAddress" ]]; then
+        identifier_key="macAddress"
+        identifier_value=$(node_val ${idx} "ports[0].address")
```
lucaconsalvi, line 477: **Suggestion:** `node_val` uses `jq -r`, which returns the literal string `"null"` with exit code 0 when a JSON path doesn't exist. If the node JSON lacks `ports[0].address`, the generated YAML would contain `- macAddress: null` — which YAML parsers interpret as a null value.

Consider adding a guard:


Also — the original code had a helpful comment explaining why the FQDN append is needed for IPv6/DualStack. That "why" was lost in the refactor. Worth restoring above the `if [[ $IP_STACK != 'v4' ]]` check.

agent/roles/manifests/templates/install-config_baremetal_yaml.j2
```diff
@@ -37,18 +37,23 @@ controlPlane:
...
+{% for i in range(master_hostnames | length) %}
+{% if fencing_credential_identifier == "macAddress" %}
+    - macAddress: {{ master_macs[i] }}
+{% else %}
+    - hostname: {{ master_hostnames[i] }}
+{% endif %}
+      address: {{ master_bmc_addresses[i] }}
+      username: {{ master_bmc_usernames[i] }}
+      password: {{ master_bmc_passwords[i] }}
+      certificateVerification: {{ 'Disabled' if bmc_verify_cas[i] == "False" else 'Enabled' }}
```
lucaconsalvi: **Nit (pre-existing):** This uses `bmc_verify_cas[i]` (the all-nodes array from line 6) instead of `master_bmc_verify_cas[i]` (the master-specific array set on line 44). Invisible in 2-master/0-worker TNF but technically wrong. Since the PR touches this line, good opportunity to fix.

lucaconsalvi (commented):
Thanks for this PR, Francesco! Clean approach to supporting macAddress-based fencing credentials — the `identifier_key`/`identifier_value` pattern in `utils.sh` is nicely done.

A few findings below — the first one (missing `AGENT_MASTER_MACS` in the baremetal code path) is a deployment blocker that affects all baremetal deployments, not just macAddress mode.

**Minor:** The PR description references `FENCING_CREDENTIALS_IDENTIFIER` (plural) but the code consistently uses `FENCING_CREDENTIAL_IDENTIFIER` (singular). Worth fixing for searchability.

dhensel-rh (commented):
comment removed. Luca caught already.

PR #1917, opened 2026-06-22
OCPEDGE-2771: fix: place BiP ISO under WORKING_DIR and set SELinux label for QEMU access

fracappa:


utils.sh
```diff
@@ -270,12 +270,15 @@ NMEOF
     fi
 
     local bip_iso="${iso_dir}/rhcos-bip.iso"
+    rm -f "${bip_iso}"
```
MahnoorAsghar: ...why?

fracappa: When using the `WORKING_DIR` fix, `create_bip_cluster()` also needs to remove any stale `rhcos-bip.iso` before calling coreos-installer iso ignition embed, since the command refuses to overwrite existing output files.

Without this, a re-run after make clean fails with:
```
Error: persisting output file to /iso/rhcos-bip.iso
Caused by:
    File exists (os error 17)
```
This happens because make clean removes ocp/ostest/ (under home dir) but not the ISO directory under WORKING_DIR/${CLUSTER_NAME}/iso/.

PR #1921, opened 2026-06-29
Update metal3-dev-env pin to dba726c2

fonta-rh:
## Summary

- Bump `metal3-dev-env` pinned commit from `f5c0b85` → `dba726c` to pick up [metal3-io/metal3-dev-env#1694](https://github.com/metal3-io/metal3-dev-env/pull/1694) (Go tarball arch detection + CPU model template fix for aarch64 KVM)
- Remove the Go tarball `sed` workaround in `01_install_requirements.sh` — upstream now uses `go_arch` (auto-detected from `ansible_architecture`)
- Remove the CPU model `sed` workaround in `02_configure_host.sh` — upstream template now gates `cortex-a57` on `effective_domain_type == 'qemu'`, so native KVM correctly falls through to `host-passthrough`
- Ansible 8.x downgrade `sed` remains — CentOS Stream 9 ships Python 3.9 and Ansible 9+ requires Python 3.10+
- **Mount libvirt socket into vbmc and sushy-tools containers** — required by upstream commit `97ecdb7` which switched the BMC libvirt connection from `qemu+ssh://` to `qemu:///system` (local socket) to fix CI flakiness where SSH-based connections caused vbmc to ack power-on without libvirt executing it

### Why the libvirt socket mount is needed

The new metal3-dev-env pin includes [97ecdb7](https://github.com/metal3-io/metal3-dev-env/commit/97ecdb787e019ab3975a17c89929c572a1434657) ("vbmc: use local libvirt socket instead of SSH") which made two coordinated changes:

1. **Ansible role** (`virtbmc/tasks/setup_tasks.yml`): changed `vbmc_libvirt_uri` from `qemu+ssh://root@.../system` to `qemu:///system`
2. **Container startup** (`02_configure_host.sh`): added `-v /var/run/libvirt:/var/run/libvirt` to vbmc and sushy-tools containers

dev-scripts consumes the Ansible roles via `setup-playbook.yml` (so conf.py now contains `qemu:///system`), but has its own container startup code in `04_setup_ironic.sh` and `agent/05_agent_configure.sh`. Without adding the socket mount here too, sushy-tools tries to connect to the local libvirt socket which isn't available inside the container, causing nodes to hang at `registering` indefinitely.

## Test plan

- [x] aarch64 Graviton fencing-IPI deployment — 2-node cluster, all COs Available, Redfish BMC endpoints responding
- [ ] CI passes on x86_64 (standard IPI/agent flows)

### DIFF
--- a/01_install_requirements.sh
+++ b/01_install_requirements.sh
@@ -19,19 +19,11 @@ if [ -z "${METAL3_DEV_ENV:-}" ]; then
   # TODO -- come up with a plan for continuously updating this
   # Note we only do this in the case where METAL3_DEV_ENV is
   # unset, to enable developer testing of local checkouts
-  git reset f5c0b859717e71c35701905161caa6e65221b3fb --hard
+  git reset dba726c297571cc24a5205e873ad9e66ad1c719c --hard
 
   # Ansible 9+ requires Python 3.10+, but CentOS Stream 9 ships Python 3.9.
   # Patch metal3-dev-env to use Ansible 8.x on centos9/rhel9.
   sed -i '/ANSIBLE_VERSION/{ s/10\.7\.0/8.7.0/; }' lib/common.sh
-
-  # Go tarball defaults hardcode linux-amd64; use GOARCH passed as extra var.
-  # Upstream fix: https://github.com/metal3-io/metal3-dev-env/pull/1694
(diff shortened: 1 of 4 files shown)


elfosardo:
/lgtm

if CI is happy, we're happy

PR #1925, opened 2026-07-07
Exclude openssl from upgrade

tdomnesc:
https://github.com/python/cpython/issues/151504
  breaks SSL connections with openssl* > 3.5.7
  Remove once Python backport in CentOS Stream 9

PR #1927, opened 2026-07-08
Use --exclude instead of version-pinning for openssl workaround

fultonj:
The previous approach installed openssl-3.5.5-3.el9 and used versionlock, which fails on systems where that specific version is not available (e.g. RHEL 9.4 with openssl-3.0.7).

Replace with --exclude='openssl*' on the dnf upgrade command, which works on any RHEL/CentOS 9 system regardless of the installed openssl version.

### DIFF
--- a/01_install_requirements.sh
+++ b/01_install_requirements.sh
@@ -51,14 +51,11 @@ MAX_RETRIES=5
 # Delay between attempts (in seconds)
 _YUM_RETRY_BACKOFF=15
 
-# TODO: remove once Python fixed OpenSSL regression
-sudo dnf install -y --allowerasing openssl-3.5.5-3.el9 openssl-devel-3.5.5-3.el9 openssl-libs-3.5.5-3.el9
-sudo dnf install -y python3-dnf-plugin-versionlock
-sudo dnf versionlock add openssl openssl-libs openssl-devel
-
 attempt=1
 while (( attempt <= MAX_RETRIES )); do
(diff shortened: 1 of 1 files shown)


fultonj:
@bfournie  I work on a team with CI that uses dev-scripts on RHEL9.4 that broke from https://github.com/openshift-metal3/dev-scripts/pull/1925 and I'm hoping you'd consider this alternative instead.

bfournie:
> @bfournie I work on a team with CI that uses dev-scripts on RHEL9.4 that broke from #1925 and I'm hoping you'd consider this alternative instead.

John - I think @tdomnesc is the best one to ask about
https://github.com/openshift-metal3/dev-scripts/pull/1925

tdomnesc:
@fultonj - the `--exclude='openssl*'` option does not work if the base CentOS Stream 9 image already comes with the new openssl version.
The real fix is to have https://redhat.atlassian.net/browse/RHEL-191730 solved.
Meanwhile maybe I can come up with an option that checks openssl available packages and installs something less than 3.5.7 instead of hardcoding a version.

elfosardo:
/hold
this does not work

tdomnesc:
https://github.com/openshift-metal3/dev-scripts/pull/1928
This should also work for you

fultonj:
Thank you @tdomnesc 

Closing this PR in favor of https://github.com/openshift-metal3/dev-scripts/pull/1928

PR #1928, opened 2026-07-08
install last available openssl version <  3.5.7

tdomnesc:
Instead of hardcoding a version


fultonj:
Thanks! Worked on RHEL9.4.

danpawlik:
Thank you for that patch

bfournie:
I think we can safely override the `ci/prow/e2e-metal-ipi-ovn-ipv4` test, its past the requirements issue.

PR #1932, opened 2026-07-21
Add optional top-of-rack LLDP switch emulation

mkowalski:
## What

`ENABLE_LLDP_TOR` runs lldpd on the hypervisor bound to the libvirt tap devices (`vnet*`). Frames transmitted on a tap are delivered straight into the VM NIC, so cluster nodes receive LLDPDUs inbound exactly like from a production top-of-rack switch, without any bridge `group_fwd_mask` change.

## Why

The libvirt bridges do not forward the link-local scoped LLDP group address (`01:80:C2:00:00:0E`) between ports, and NetworkManager >= 1.59.1 filters locally-sourced LLDP frames (NM commit `9a79eba502`), so cluster nodes deployed by dev-scripts never see any LLDP neighbor.

Primary consumer is the kubernetes-nmstate LLDP e2e test, which since nmstate/kubernetes-nmstate#1549 expects an `lldp-switch` neighbor on the primary NIC. Upstream solves this for kubevirtci with `cluster/lldpd-switch.sh`; this adds the dev-scripts equivalent so the OpenShift `e2e-handler-ovn-ipv4` lane can pass again (see openshift/kubernetes-nmstate#765).

## How

- `lldp/configure_lldp_tor.sh`: installs lldpd (EPEL is already enabled by `01_install_requirements.sh`), writes `/etc/lldpd.d/lldp-tor.conf` (tx-interval 5, configurable system name defaulting to `lldp-switch`, interface pattern `vnet*` so only libvirt taps are touched, never physical host NICs), enables and restarts the service. lldpd tracks interface add/remove via netlink, so taps created when the VMs boot later (or recreated on node reboot) are picked up automatically.
- `lldp/cleanup_lldp_tor.sh`: best-effort teardown, wired unconditionally into `host_cleanup.sh`.
- Opt-in hook at the end of `02_configure_host.sh`, defaults in `common.sh`, documentation in `config_example.sh`.

Mirrors the `ENABLE_BGP_TOR` pattern.

---
This PR description was generated using AI. Please verify before acting on it.

### DIFF
--- a/02_configure_host.sh
+++ b/02_configure_host.sh
@@ -512,3 +512,8 @@ if [ "${PERSISTENT_IMAGEREG}" == true ] ; then
     sudo systemctl start nfs-server
     sudo exportfs -a
 fi
+
+# Optionally emulate a top-of-rack LLDP switch towards the cluster VMs
+if [[ -n "${ENABLE_LLDP_TOR:-}" ]]; then
+    lldp/configure_lldp_tor.sh
+fi
--- a/common.sh
+++ b/common.sh
@@ -708,3 +708,8 @@ if [ "${OPENSHIFT_CI}" == true ] ; then
 fi
 
 export ENABLE_CAPI_E2E=${ENABLE_CAPI_E2E:-}
+
+# Optional LLDP top-of-rack switch emulation on the libvirt tap devices
+# (see config_example.sh)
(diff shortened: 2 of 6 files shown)


mkowalski:
e2e-agent-compact-ipv4-iso-no-registry passed and we don't touch it

PR #1936, opened 2026-07-28
Fix image builds that don't use --build-arg

cybertron:
As part of the shellcheck fixes, this variable was quoted. However, when the variable is empty this results in a bogus parameter being passed to podman build which caused it to fail.

This appears to have been designed so it could be concatenated, but the variable is reset each time through the loop so I don't think we actually need that. This splits the argument name from its value so we pass valid arguments even when the variable is empty. Passing --build-arg '' seems to work fine with the image build I tested.


cybertron:
/approve

Dual stack failed on something unrelated, and it's not clear to me that CI would even exercise this code path.

PR #1922, opened 2026-06-30
Add baremetal ABI support for NODES_PLATFORM=baremetal

fonta-rh:
## Summary

- Add `NODES_PLATFORM=baremetal` support to the agent-based installation (`make agent`) pipeline for deploying on real baremetal nodes via a standalone provisioning host (no libvirt)
- Bypass three libvirt-dependent code paths: host configuration (early return after platform-agnostic setup), MAC/IP discovery (`get_baremetal_ips_and_macs()` reads from NODES_FILE + config vars), and ISO delivery (`mount_agent_iso_baremetal()` via Redfish VirtualMedia)
- Add dnsmasq DNS configuration from `BAREMETAL_API_VIP`/`BAREMETAL_INGRESS_VIP` config vars, and document all new `BAREMETAL_*` variables in `config_example.sh`
- Add RHEL 10 / CentOS Stream 10 support to `01_install_requirements.sh` (package availability fixes, metal3-dev-env compatibility patches)
- Add `BAREMETAL_MACS` config var for multi-NIC servers where PXE boot MAC ≠ data network MAC

## Context

This is part of [OCPEDGE-2775](https://issues.redhat.com/browse/OCPEDGE-2775) — enabling Two-Node with Fencing (TNF) deployment on real baremetal hardware. A companion tool (`make baremetal-adopt` from [OCPEDGE-2774](https://issues.redhat.com/browse/OCPEDGE-2774)) generates the `config_baremetal_fencing.sh` that exports `NODES_PLATFORM=baremetal`, `NODES_FILE`, bridge overrides, and network vars.

Target topology: **standalone provisioning host without libvirt** (e.g., RDU2 HPE e920t nodes with iLO BMC via Redfish).

Validated with three end-to-end deploys on RDU2 baremetal (OCP 4.22.0-rc.5 and 4.22.4), both nodes Ready, all operators Available, full TNF pipeline (auth → setup → fencing → after-setup → update-setup) completed.

## Changes

| File | Change |
|------|--------|
| `01_install_requirements.sh` | RHEL 10/CS10 support: distro case branch, CodeReady Builder repo, Go tarball arch fix for metal3-dev-env, remove unavailable packages (genisoimage → xorriso, python3-bcrypt) |
| `02_configure_host.sh` | Early return for baremetal — runs firewall, NTP, DNS, then `exit 0` (skips all libvirt VM/network/pool creation) |
[...]

PR #1937, opened 2026-07-29
OCPBUGS-100315: Fix baremetal registry and pull secret after #1922

tusharjadhav3302:
## Summary

Fixes two regressions introduced by #1922 where the early `exit 0` in
`02_configure_host.sh` for `NODES_PLATFORM=baremetal` skipped both local
registry setup and `REGISTRY_CREDS` file creation.

### 1. Baremetal without mirroring (`MIRROR_IMAGES` unset)
`write_pull_secret()` in `utils.sh` unconditionally merged `REGISTRY_CREDS`
(`private-mirror-ocp.json`), causing:
```
jq: error: Could not open file /home/zuul/private-mirror-ocp.json: No such file or directory
```
Affects ShiftStack/RHOSO CI jobs on real baremetal hardware.

### 2. Baremetal with mirroring (`MIRROR_IMAGES=true`)
The local registry at port 5000 never starts, causing `04_setup_ironic.sh`
to fail with "connection refused" when mirroring release images.
Affects virtual-baremetal CI using sushy-emulator (e.g. RHOSO `uni04delta-ipv6` jobs).

### `02_configure_host.sh` — registry setup in baremetal early-exit path
- `use_registry "podman"`: starts registry, creates `REGISTRY_CREDS` via podman login
- `use_registry ""` (quay): starts registry
- No registry: creates empty `{}` authfile for `write_pull_secret()` compatibility
- Mirrors the existing pattern in the libvirt path (lines 471–487)

### `utils.sh` — conditional REGISTRY_CREDS merge in write_pull_secret()
- Guards `REGISTRY_CREDS` merge behind `use_registry ""` check
- When mirroring is disabled: merges personal pull secret + CI registry token only
- Applies to both `OPENSHIFT_CI=true` and the standard CI_TOKEN code paths

## Root cause

#1922 assumed baremetal deployments never need a local registry. This is
true for "real" baremetal (standalone provisioning hosts pulling from
external registries), but not for:
- Virtual-baremetal setups using `MIRROR_IMAGES=true`
- Any baremetal path where `write_pull_secret()` is called without mirroring

## Test plan

- [ ] `NODES_PLATFORM=baremetal`, `MIRROR_IMAGES` unset: empty authfile created,
      `write_pull_secret()` succeeds without registry
- [ ] `NODES_PLATFORM=baremetal`, `MIRROR_IMAGES=true`: registry starts,
      `oc adm release mirror` succeeds
- [ ] `NODES_PLATFORM=libvirt`: no behavior change (early-exit block not reached)
- [ ] Verified against ShiftStack RHOSO CI job failure on serval70 (2026-07-28)

## Related

- Regression from: #1922
- Supersedes: #1938
- Downstream workaround: openstack-k8s-operators/ci-framework#4078

### DIFF
--- a/02_configure_host.sh
+++ b/02_configure_host.sh
@@ -97,6 +97,29 @@ if [[ "${NODES_PLATFORM}" == "baremetal" ]]; then
 
     switch_to_internal_dns
 
+    # Add a /etc/hosts entry for $LOCAL_REGISTRY_DNS_NAME
+    sudo sed -i "/${LOCAL_REGISTRY_DNS_NAME}/d" /etc/hosts
+    echo "${PROVISIONING_HOST_EXTERNAL_IP} ${LOCAL_REGISTRY_DNS_NAME}" | sudo tee -a /etc/hosts
+
+    # When MIRROR_IMAGES is configured, the local registry must be running
+    # before 04_setup_ironic.sh attempts to mirror release images into it.
+    # This covers virtual-baremetal (sushy-emulator) CI environments that
+    # set NODES_PLATFORM=baremetal together with MIRROR_IMAGES=true.
+    if use_registry "podman"; then
+        setup_local_registry
+        rm -f "${REGISTRY_CREDS}"
+        sudo podman login --authfile "${REGISTRY_CREDS}" \
+            -u "${REGISTRY_USER}" -p "${REGISTRY_PASS}" \
+            "${LOCAL_REGISTRY_DNS_NAME}":"${LOCAL_REGISTRY_PORT}"
+    elif use_registry ""; then
+        setup_local_registry
+    else
+        echo '{}' | sudo dd of="${REGISTRY_CREDS}"
+    fi
+    if [ -f "${REGISTRY_CREDS}" ]; then
+        sudo chown "$USER":"$USER" "${REGISTRY_CREDS}"
+    fi
+
     exit 0
 fi
 
--- a/utils.sh
+++ b/utils.sh
@@ -1009,8 +1009,12 @@ function write_pull_secret() {
     if [ "${OPENSHIFT_CI}" == true ]; then
         # We don't need to fetch a personal pull secret with the
         # token, but we still need to merge what we're given with the
-        # credentials for the local reigstry.
-        jq -s '.[0] * .[1]' "${REGISTRY_CREDS}" "${PERSONAL_PULL_SECRET}" > "${PULL_SECRET_FILE}"
+        # credentials for the local registry when mirroring is enabled.
+        if use_registry ""; then
+            jq -s '.[0] * .[1]' "${REGISTRY_CREDS}" "${PERSONAL_PULL_SECRET}" > "${PULL_SECRET_FILE}"
+        else
+            cp "${PERSONAL_PULL_SECRET}" "${PULL_SECRET_FILE}"
+        fi
         return
     fi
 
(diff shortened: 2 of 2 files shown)


tusharjadhav3302:
@pfontani this is a follow-up to #1922 — would appreciate your review

tusharjadhav3302:
ci-framework has a workaround merged (openstack-k8s-operators/ci-framework#4078) that pre-creates the empty file. This PR fixes the root cause in dev-scripts itself so other consumers are not affected.

02_configure_host.sh
```diff
@@ -97,6 +97,13 @@ if [[ "${NODES_PLATFORM}" == "baremetal" ]]; then
 
     switch_to_internal_dns
 
+    # Other scripts may read REGISTRY_CREDS even when local registry mirroring
+    # is disabled; ensure a valid empty authfile exists before the early return.
+    if ! use_registry "quay"; then
```
fonta-rh: Are you using "quay" here because that's what is on the libvirt path? I think this works but it might be a little clearer if we also use "" like in utils.sh (there's nothing special about quay here, it's just not podman, I think). This also catches "podman" registry, which might not matter but it might also be unintended

fonta-rh (approved):
Thanks for such a quick fix. I'd update the "quay" to "" if possible, to make it clearer that we're looking for a non-registry condition. It does work, though, I think, so it's not a blocker. Approving all the same (just informative, I don't have permissions)

02_configure_host.sh
tusharjadhav3302: @fonta-rh 
Updated to use_registry "" to match the guard in utils.sh.
Thanks for the review.

mkowalski:
/retitle OCPBUGS-100315: Fix baremetal registry and pull secret after #1922

mkowalski:
One possible gap in the `MIRROR_IMAGES=true` baremetal case: registry DNS resolution.

In the libvirt path, an `/etc/hosts` entry for `LOCAL_REGISTRY_DNS_NAME` (default `virthost.<cluster>.<domain>`) is added before the `podman login` (02_configure_host.sh lines ~478–480), but that code sits *after* the baremetal `exit 0`, and the baremetal `configure_dnsmasq()` only adds api/ingress VIP records. So in the new baremetal block, `podman login ${LOCAL_REGISTRY_DNS_NAME}:${LOCAL_REGISTRY_PORT}` and the subsequent release-image mirroring in `04_setup_ironic.sh` may fail on name resolution — unless the affected CI jobs already export a resolvable `LOCAL_REGISTRY_DNS_NAME`.

Suggest adding the `/etc/hosts` entry to the baremetal block too, e.g. before the `use_registry` checks:

```bash
sudo sed -i "/${LOCAL_REGISTRY_DNS_NAME}/d" /etc/hosts
echo "${PROVISIONING_HOST_EXTERNAL_IP} ${LOCAL_REGISTRY_DNS_NAME}" | sudo tee -a /etc/hosts
```

(and possibly a dnsmasq host record so cluster nodes can resolve the mirror as well).

---
This message was generated using AI. Please verify before acting on it.
Assisted-By: Claude Fable 5

tusharjadhav3302:
> One possible gap in the `MIRROR_IMAGES=true` baremetal case: registry DNS resolution.
> 
> In the libvirt path, an `/etc/hosts` entry for `LOCAL_REGISTRY_DNS_NAME` (default `virthost.<cluster>.<domain>`) is added before the `podman login` (02_configure_host.sh lines ~478–480), but that code sits _after_ the baremetal `exit 0`, and the baremetal `configure_dnsmasq()` only adds api/ingress VIP records. So in the new baremetal block, `podman login ${LOCAL_REGISTRY_DNS_NAME}:${LOCAL_REGISTRY_PORT}` and the subsequent release-image mirroring in `04_setup_ironic.sh` may fail on name resolution — unless the affected CI jobs already export a resolvable `LOCAL_REGISTRY_DNS_NAME`.
> 
> Suggest adding the `/etc/hosts` entry to the baremetal block too, e.g. before the `use_registry` checks:
> 
> ```shell
> sudo sed -i "/${LOCAL_REGISTRY_DNS_NAME}/d" /etc/hosts
> echo "${PROVISIONING_HOST_EXTERNAL_IP} ${LOCAL_REGISTRY_DNS_NAME}" | sudo tee -a /etc/hosts
> ```
> 
> (and possibly a dnsmasq host record so cluster nodes can resolve the mirror as well).
> 
> This message was generated using AI. Please verify before acting on it. Assisted-By: Claude Fable 5

Added /etc/hosts entry to the baremetal block

mkowalski:
~If I understand correctly, failure of `e2e-metal-ipi-ovn-bgp-virt-ipv4` is now a signal for this PR. Can we run this job with this change and see if it brings the green result?~

Actually, seems like this statement was incorrect. Looking deeper, this job does not run `nodes_platform=baremetal` path. The comment https://redhat.atlassian.net/browse/OCPBUGS-100315?focusedCommentId=17808943 is probably incorrect or the chain of what broke what was more complex

PR #1940, opened 2026-07-31
AGENT-1579: Agent-based installer - add static networking support for ISO_NO_REGISTRY mode

bfournie:
Add TUI automation to configure static IPs and hostnames via nmtui for agent-based ISO_NO_REGISTRY installations. Creates a NEW NetworkManager connection (rather than editing the existing DHCP one) to ensure `--copy-network` is triggered by coreos-installer, preserving the static config across reboots.

See https://github.com/openshift/assisted-installer-agent/pull/1374

### DIFF
--- a/agent/06_agent_create_cluster.sh
+++ b/agent/06_agent_create_cluster.sh
@@ -387,8 +387,24 @@ function verify_static_networking(){
         expected_hostname="$(printf "$hostname_format" "$n")"
 
         echo "Verifying ${node_type}-${n} is reachable at static IP ${node_ip} with hostname ${expected_hostname}"
-        local actual_hostname
-        if ! actual_hostname=$(timeout 60 ${SSH} "core@${node_ip}" hostname 2>/dev/null); then
+
+        # The UI reports the console URL as soon as it is available, but the
+        # nodes are still rebooting and finalizing (applying the static IP via
+        # coreos-installer --copy-network) at that point, so the static IP is not
+        # immediately reachable. Retry until the node comes up before failing.
+        local actual_hostname=""
+        local retries=40   # up to ~20 min at 30s interval
+        local i
+        for (( i=1; i<=retries; i++ )); do
+            if actual_hostname=$(${SSH} "core@${node_ip}" hostname 2>/dev/null) && [[ -n "${actual_hostname}" ]]; then
+                break
+            fi
+            actual_hostname=""
+            echo "Waiting for ${node_type}-${n} to be reachable at ${node_ip} (attempt ${i}/${retries})"
+            sleep 30
+        done
+
+        if [[ -z "${actual_hostname}" ]]; then
             echo "ERROR: static_ip test case failed - could not ssh to ${node_type}-${n} at static IP ${node_ip}"
             return 1
         fi
--- a/agent/06_agent_create_cluster.sh
+++ b/agent/06_agent_create_cluster.sh
@@ -335,6 +335,44 @@ function automate_rendezvousIP_selection(){
     done
 }
 
+function automate_static_networking(){
+  local node_type=$1
+  local node_count=$2
+  local base_ip=80
+
+  # Compute IP offset: masters start at 80, workers continue after masters,
+  # arbiters continue after workers
+  local offset=0
+  case "$node_type" in
+    master) offset=0 ;;
+    worker) offset=$NUM_MASTERS ;;
+    arbiter) offset=$(( NUM_MASTERS + NUM_WORKERS )) ;;
+  esac
+
+  local hostname_format
+  case "$node_type" in
+    master) hostname_format="$MASTER_HOSTNAME_FORMAT" ;;
+    worker) hostname_format="$WORKER_HOSTNAME_FORMAT" ;;
+    arbiter) hostname_format="$ARBITER_HOSTNAME_FORMAT" ;;
+  esac
+
+  for (( n=0; n<node_count; n++ ))
+    do
+        name=${CLUSTER_NAME}_${node_type}_${n}
+        node_ip="192.168.111.$(( base_ip + offset + n ))"
+        # shellcheck disable=SC2059
+        node_hostname="$(printf "$hostname_format" "$n")"
+
+        sudo virsh screenshot "$name" "${OCP_DIR}/${name}_console_screenshot_before_static_networking.ppm"
+
+        echo "Configuring static IP ${node_ip} and hostname ${node_hostname} on ${name}"
+        ./agent/e2e/agent-tui/automate-no-registry-agent-tui-static.sh "$name" "$node_ip" "$node_hostname"
+
+        sudo virsh screenshot "$name" "${OCP_DIR}/${name}_console_screenshot_after_static_networking.ppm"
+        echo "Finished configuring static networking for $name"
+    done
+}
+
 function check_assisted_install_UI(){
   local rendezvousIP
   rendezvousIP=$(getRendezvousIP)
@@ -838,9 +876,15 @@ case "${AGENT_E2E_TEST_BOOT_MODE}" in
         echo "aarch64: CDROM media ejected from all VMs"
     fi
 
-    automate_rendezvousIP_selection master "$NUM_MASTERS"
-    automate_rendezvousIP_selection worker "$NUM_WORKERS"
-    automate_rendezvousIP_selection arbiter "$NUM_ARBITERS"
+    if [[ "${AGENT_ISO_NO_REGISTRY_STATIC_NETWORKING:-false}" == "true" ]]; then
+        automate_static_networking master "$NUM_MASTERS"
+        automate_static_networking worker "$NUM_WORKERS"
+        automate_static_networking arbiter "$NUM_ARBITERS"
+    else
+        automate_rendezvousIP_selection master "$NUM_MASTERS"
+        automate_rendezvousIP_selection worker "$NUM_WORKERS"
+        automate_rendezvousIP_selection arbiter "$NUM_ARBITERS"
+    fi
 
     check_assisted_install_UI
 
--- a/agent/06_agent_create_cluster.sh
+++ b/agent/06_agent_create_cluster.sh
@@ -335,41 +335,69 @@ function automate_rendezvousIP_selection(){
     done
 }
 
+function hostname_format_for_node_type(){
+  case "$1" in
+    master) echo "$MASTER_HOSTNAME_FORMAT" ;;
+    worker) echo "$WORKER_HOSTNAME_FORMAT" ;;
+    arbiter) echo "$ARBITER_HOSTNAME_FORMAT" ;;
+  esac
+}
+
+# Configuration step of the 'static_ip' test case: drive the agent TUI to
+# assign a static IP and hostname to each node of the given type. The static
+# IPs are derived from staticIPForNode (see agent/common.sh).
 function automate_static_networking(){
   local node_type=$1
   local node_count=$2
-  local base_ip=80
-
-  # Compute IP offset: masters start at 80, workers continue after masters,
-  # arbiters continue after workers
-  local offset=0
-  case "$node_type" in
-    master) offset=0 ;;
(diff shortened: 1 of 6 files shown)


bfournie:
This appears to work well, removing the WIP

The install was successfully
```
time="2026-08-03T17:18:58+02:00" level=info msg="Waiting for 5m0s ..."
time="2026-08-03T17:24:03+02:00" level=info msg="Console URL is available."
time="2026-08-03T17:24:03+02:00" level=info msg="Screenshot saved to/home/stack/dev-scripts-bfournie-latest/ocp/ostest/09-installation-progress-10.png, with type of image/png"

$ oc get nodes
NAME       STATUS   ROLES                         AGE   VERSION
master-0   Ready    control-plane,master,worker   37m   v1.36.2
master-1   Ready    control-plane,master,worker   71m   v1.36.2
master-2   Ready    control-plane,master,worker   72m   v1.36.2

```
And can log in via the static IP to all of the nodes 

```
$ ssh core@192.168.111.80
Red Hat Enterprise Linux CoreOS 10.2.20260730-1
  Part of OpenShift 5.0, RHCOS is a Kubernetes-native operating system
  managed by the Machine Config Operator (`clusteroperator/machine-config`).

WARNING: Direct SSH access to machines is not recommended; instead,
make configuration changes via `machineconfig` objects:
  https://docs.openshift.com/container-platform/5.0/architecture/architecture-rhcos.html

---
Last login: Mon Aug  3 16:07:53 2026 from 192.168.111.1
[systemd]
Failed Units: 2
  NetworkManager-wait-online.service
  systemd-network-generator.service
[core@master-0 ~]$ exit
logout
Connection to 192.168.111.80 closed.
[stack@r640-u02 dev-scripts-bfournie-latest]$ ssh core@192.168.111.81
The authenticity of host '192.168.111.81 (192.168.111.81)' can't be established.
ED25519 key fingerprint is SHA256:806YjC6vkr4R/2xuwpeP3GuGAKSFQ2fVMkv6t0THmAU.
This key is not known by any other names.
Are you sure you want to continue connecting (yes/no/[fingerprint])? yes
Warning: Permanently added '192.168.111.81' (ED25519) to the list of known hosts.
Red Hat Enterprise Linux CoreOS 10.2.20260730-1
  Part of OpenShift 5.0, RHCOS is a Kubernetes-native operating system
  managed by the Machine Config Operator (`clusteroperator/machine-config`).

WARNING: Direct SSH access to machines is not recommended; instead,
make configuration changes via `machineconfig` objects:
```

agent/e2e/agent-tui/test-static-ip.sh
andfasano, line 1: a little bit of bikeshedding: the name here doesn't convey properly IMHO the nature of the test. This e2e script applies the first step (by automating the agent TUI, but that's a technical detail) of a test to verify that a network config applied via agent TUI is effectively applied at the end of the installation

agent/e2e/agent-tui/test-static-ip.sh
```diff
@@ -0,0 +1,165 @@
+#!/bin/bash
```
andfasano, line 1: Not strictly blocking, but I'd find really useful to have a brief description here about what the script is meant to perform: it will avoid reading all the commands one by one

config_example.sh
```diff
@@ -1009,6 +994,14 @@ set -x
...
 # export AGENT_OVE_ISO_SIZE=40
 
+# AGENT_ISO_NO_REGISTRY_STATIC_NETWORKING controls whether the ISO_NO_REGISTRY
+# test configures static networking on each node via the TUI (nmtui) instead of
+# using DHCP. When set to 'true', the TUI automation script will navigate nmtui
+# to configure a static IP on each node before proceeding with rendezvous IP
+# selection. Static IPs are assigned starting at 192.168.111.80 (rendezvous node),
+# incrementing by 1 for each additional node.
+# Requires: AGENT_E2E_TEST_BOOT_MODE=ISO_NO_REGISTRY
+# export AGENT_ISO_NO_REGISTRY_STATIC_NETWORKING=true
```
andfasano: I am little bit confused here. In the past we implemented the the `AGENT_TEST_CASES` approach, with the idea of running additional e2e test cases directly within dev-scripts. The `bad_dns` one also focused on the agent TUI. Rather than defining a new config var, I think it would be better to keep re-using the `AGENT_TEST_CASES` mechanism.

As mentioned also previously, from a strict point of view of the test, what we need is:

1) Modify the network configuration of just one of the nodes, assigning a static IP different from the usual DHCP ones, using the agent TUI
2) At the end of installation, _verify_ that the node effectively has been assigned the static IP.

A `[[ $AGENT_TEST_CASES =~ "static_ip" ]]` may help activating the test when required, and also adding the code where required.

agent/06_agent_create_cluster.sh
```diff
@@ -335,6 +335,44 @@ function automate_rendezvousIP_selection(){
...
+
+        sudo virsh screenshot "$name" "${OCP_DIR}/${name}_console_screenshot_before_static_networking.ppm"
+
+        echo "Configuring static IP ${node_ip} and hostname ${node_hostname} on ${name}"
+        ./agent/e2e/agent-tui/automate-no-registry-agent-tui-static.sh "$name" "$node_ip" "$node_hostname"
+
+        sudo virsh screenshot "$name" "${OCP_DIR}/${name}_console_screenshot_after_static_networking.ppm"
+        echo "Finished configuring static networking for $name"
+    done
+}
```
andfasano, line 371: For what regards the e2e tests, it would be sufficient to just modify _one_ of the nodes (maybe using an hard-coded ip). That could also help in reducing the complexity of the test code

bfournie: I'd like to continue updating all nodes to be a realistic test of an environment without DHCP which is a common case we're heard from customers.

config_example.sh
bfournie: Updated to add verification and use `$AGENT_TEST_CASES`

bfournie:
Confirmed on compact installation
```
$ grep Verified logs/06_agent_create_cluster-2026-08-14-221801.log
2026-08-14 23:51:53 +(./agent/06_agent_create_cluster.sh:416): verify_static_networking(): echo 'Verified master-0: reachable at 192.168.111.80 with hostname master-0'
2026-08-14 23:51:53 Verified master-0: reachable at 192.168.111.80 with hostname master-0
2026-08-14 23:51:54 +(./agent/06_agent_create_cluster.sh:416): verify_static_networking(): echo 'Verified master-1: reachable at 192.168.111.81 with hostname master-1'
2026-08-14 23:51:54 Verified master-1: reachable at 192.168.111.81 with hostname master-1
2026-08-14 23:51:56 +(./agent/06_agent_create_cluster.sh:416): verify_static_networking(): echo 'Verified master-2: reachable at 192.168.111.82 with hostname master-2'
2026-08-14 23:51:56 Verified master-2: reachable at 192.168.111.82 with hostname master-2
```

bfournie:
Remaining tests are permafailing early with
```
[WARNING]: Skipping Galaxy server https://galaxy.ansible.com/api/. Got an
unexpected error when getting available versions of collection equinix.cloud:
Unknown error when attempting to call Galaxy at
'https://galaxy.ansible.com/api/v3/collections/equinix/cloud/versions/0.12.0/':
<urlopen error [Errno -2] Name or service not known>. <urlopen error [Errno -2]
Name or service not known>
ERROR! Unknown error when attempting to call Galaxy at 'https://galaxy.ansible.com/api/v3/collections/equinix/cloud/versions/0.12.0/': <urlopen error [Errno -2] Name or service not known>. <urlopen error [Errno -2] Name or service not known>
Starting galaxy collection install process
Process install dependency map
```

agent/e2e/agent-tui/test-static-ip.sh
```diff
@@ -0,0 +1,180 @@
...
+#
+# test-static-ip.sh - first step of the 'static_ip' e2e test case.
+#
+# This script performs the *configuration* step of the static IP test: for a
```
andfasano: This comment is just to improve the readability and future maintenance.

It remains very difficult to understand exactly what the test is adding, without forcing the (human) reader to go line by line and decode it. Rather than adding comments in each section - IMHO they do not really add any real value to the understanding, and still forcing the user to read that point, I think it would be more friendly to have just an initial comment describing a real case of the fields that the code is going to add (how this is done it's just a technical detail). Asking to do that to Claude produced something like:

```
# This script configures static networking for a node via TUI/nmtui.                                                                                   
# It creates a new static IPv4 connection that persists after installation. 
#
# Example configuration applied:                                                                                                                       
# ┌─────────────────┬──────────────────┬──────────────────────┐                                                                                        
# │ Parameter       │ Example Value    │ Purpose              │                                                                                        
# ├─────────────────┼──────────────────┼──────────────────────┤                                                                                        
# │ Interface       │ enp2s0           │ Network interface    │                                                                                        
# │ IP Address      │ 192.168.111.30   │ Static IPv4 (/24)    │                                                                                        
# │ Gateway         │ 192.168.111.1    │ Default route        │                                                                                        
# │ DNS Server      │ 192.168.111.1    │ Name resolution      │                                                                                        
# │ Hostname        │ node-0           │ System hostname      │                                                                                        
# │ Rendezvous IP   │ 192.168.111.29   │ Cluster discovery    │                                                                                        
# └─────────────────┴──────────────────┴──────────────────────┘       
```

config_example.sh
```diff
@@ -1026,6 +1010,19 @@ set -x
...
+# using DHCP, then verifies at the end of the installation that each node came up
+# with the configured static IP and hostname. During the agent TUI automation a
+# new NetworkManager connection is created on each node with a static IPv4
+# address and hostname; static IPs are assigned starting at 192.168.111.80
+# (rendezvous node), incrementing by 1 for each additional node. After the
+# installation completes, the test ssh's to each node at its assigned static IP
+# and confirms the hostname matches, failing the deployment if any node is not
+# reachable at its static IP.
+# This test case requires AGENT_E2E_TEST_BOOT_MODE=ISO_NO_REGISTRY and is only
+# supported when IP_STACK=v4.
```
andfasano, line 1024: Is there a check to prevent that? Ie when IP_STACK != v4?

agent/06_agent_create_cluster.sh
```diff
@@ -335,6 +335,88 @@ function automate_rendezvousIP_selection(){
...
+    master) echo "$MASTER_HOSTNAME_FORMAT" ;;
+    worker) echo "$WORKER_HOSTNAME_FORMAT" ;;
+    arbiter) echo "$ARBITER_HOSTNAME_FORMAT" ;;
+  esac
+}
+
+# Configuration step of the 'static_ip' test case: drive the agent TUI to
+# assign a static IP and hostname to each node of the given type. The static
+# IPs are derived from staticIPForNode (see agent/common.sh).
+function automate_static_networking(){
```
andfasano: Some name bikeshedding (but for e2e test IMHO is as well as important).  The name of the test `static_ip` does not match with the name of the functions used for implementing and verifying it: `automate_static_networking` and `verify_static_networking`. As an additional point, these names are also pretty generic.
I think it would be useful to establish a little bit of shared naming convention for the e2e tests, and I'd suggest the following one:

* `test_case_do_<test case name>` for the function containing the implementation of the test (the activation or triggering code)
* `test_case_verify_<test case name>` for the verification code - something similar to the current code

So in for this patch we could have:
* `test_case_do_static_ip`
* `test_case_verify_static_ip`

(the same renaming could be applied for the bad_dns, ofc)

agent/e2e/agent-tui/test-static-ip.sh
bfournie: Updated comments to add a similar table and reduced the in-line comments to improve readability.

agent/06_agent_create_cluster.sh
bfournie: Updated both tests to use common naming mechanism.

config_example.sh
bfournie: Added check for IPv4 only

PR #1929, opened 2026-07-09
OPNET-787: Add optional top-of-rack BGP speaker for the baremetal network

mkowalski:
## What

`ENABLE_BGP_TOR=true` deploys an FRR container on the hypervisor host network that acts as a top-of-rack BGP speaker for the baremetal (`ostestbm`) network:

- accepts **dynamic BGP sessions** from cluster nodes anywhere on the external subnet (`bgp listen range`, so DHCP node addresses need no per-peer config)
- installs routes learned from the cluster into the host kernel via zebra, so prefixes advertised by the cluster (e.g. VIPs) are reachable from the hypervisor over the BGP paths
- `no bgp ebgp-requires-policy` (RFC 8212 relaxation), required for route exchange with FRR ≥ 8 defaults

Deployment happens at the end of `02_configure_host.sh` (the network exists by then and the speaker survives `ocp_cleanup`, matching its ToR role); teardown is wired into `host_cleanup.sh`. Follows the `metallb/` add-on layout. ASNs and the FRR image are configurable, defaults documented in `config_example.sh`:

```bash
export ENABLE_BGP_TOR=true
#export BGP_TOR_ASN=64513         # speaker ASN
#export BGP_CLUSTER_ASN=64512     # expected cluster ASN
#export BGP_TOR_IMAGE=quay.io/frrouting/frr:9.1.0
```

Listen ranges and address-family activation are derived from `EXTERNAL_SUBNET_V4`/`V6` (dual-stack aware; families without a subnet are omitted).

## Why

Primary consumer is BGP-based VIP management (openshift/enhancements#1982, OPNET-595): the cluster advertises API/Ingress VIPs via BGP and needs an external peer playing the ToR role. Also generally useful for any BGP-on-baremetal testing (e.g. MetalLB BGP mode without the e2e harness, OVN-K route advertisements).

## Validation

- Live-tested against a running BGP-VIP cluster: 3 control-plane FRR sessions re-establish within ~1 min of (re)deploying the speaker; advertised VIP /32s are ECMP-installed on the host; full cleanup→configure cycle verified (container, `$WORKING_DIR/bgp-tor`, firewall port).
- `bash -n` + shellcheck clean on the new scripts.
- Rendered v4-only `frr.conf` is byte-identical to the config used throughout the enhancement-1982 demo runs.

PR #1945, opened 2026-08-06
BGP VIP management: render an IPv6 ToR peer on dual-stack deployments

mkowalski:
Follow-up to #1939: with `IP_STACK=v4v6` the install-config carries one API/ingress VIP per address family, but `bgpVIPConfig` only rendered the IPv4 ToR peer — the secondary-family VIPs had no same-family BGP session to be advertised over. Render a second peer for the ToR's address on the external IPv6 subnet when one exists (override: `BGP_VIP_PEER_ADDRESS_V6`). The ToR speaker from #1929 already listens on both families.

Validated on a live dual-stack dev-scripts cluster: both sessions establish per node and both VIP families are advertised to the ToR (v4 /32s and v6 /128s).

Needed for the dual-stack conversion of the `e2e-metal-ipi-bgp-vip*` CI lanes (openshift/release#82912).

---
This PR description was generated using AI. Please verify before acting on it.

### DIFF
--- a/bgp/configure_bgp_tor.sh
+++ b/bgp/configure_bgp_tor.sh
@@ -62,6 +62,8 @@ router bgp ${BGP_TOR_ASN}
  no bgp ebgp-requires-policy
  neighbor CLUSTER peer-group
  neighbor CLUSTER remote-as ${BGP_CLUSTER_ASN}
+ neighbor CLUSTER password ${BGP_VIP_PASSWORD:-dev-scripts-bgp}
+ neighbor CLUSTER bfd
 ${LISTEN_RANGES}${ADDRESS_FAMILIES}!
 EOF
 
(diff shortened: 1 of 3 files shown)


mkowalski:
Pushed 67c4c6e: BGP_VIP_MANAGEMENT deployments now set every optional peer field (port, holdTime/keepaliveTime, password, bfdEnabled, ebgpMultiHop; overridable via `BGP_VIP_*` knobs), and the ToR is configured to match (peer-group password, BFD with bfdd enabled, BFD UDP ports opened). Rationale: the timer-format and dead port-field bugs found on openshift/installer#10718 were invisible to e2e because the defaults-only happy path never rendered those template branches. Pairs with the runtime-state assertion added to the verify step in openshift/release#82912.

This comment was generated using AI. Please verify before acting on it.

PR #1947, opened 2026-08-14
Reduce race condition in extra worker BMH application

elfosardo:
Wait for all worker Machines to have a BareMetalHost assigned before applying extra worker manifests.  This gives slow-registering workers (e.g. Redfish, ~17 min) time to claim their BMH slot, reducing the chance of machine-api grabbing the extra worker for the primary MachineSet.

The wait is non-fatal (default 120 retries * 10s = 20 min): if not all workers register within the timeout (e.g. permanently-unreachable worker BMH in some dualstack environments), the extra workers are applied anyway to preserve existing fallback behavior.  The retry count is configurable via EXTRA_WORKER_WAIT_RETRIES.

Assisted-By: Claude Opus 4.6

### DIFF
--- a/06_create_cluster.sh
+++ b/06_create_cluster.sh
@@ -92,25 +92,31 @@ fi
 
 if [[ -n "${APPLY_EXTRA_WORKERS:-}" ]]; then
     if [[ ${NUM_EXTRA_WORKERS} -ne 0 && -s "${OCP_DIR}/extra_host_manifests.yaml" ]]; then
-        # Wait for all worker Machines to have a BMH assigned before applying
-        # extra workers, to prevent the machine-api from claiming extra worker
-        # hosts for the primary MachineSet (race when regular workers are slow
-        # to register via Redfish).
+        # Give regular worker Machines time to claim their BMHs before
+        # introducing extra workers.  This reduces the chance of machine-api
+        # grabbing an extra worker host for the primary MachineSet when a
+        # regular worker is slow to register (e.g. Redfish).  If they don't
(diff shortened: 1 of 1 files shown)


elfosardo:
/retest

ah great, the same unrelated test failed here and in another job in a different patch

elfosardo:
/retest

more unrelated failures

06_create_cluster.sh
```diff
@@ -92,25 +92,31 @@ fi
...
+            ASSIGNED=$(oc get machines -n openshift-machine-api \
+                -l machine.openshift.io/cluster-api-machine-role=worker \
+                -o json 2>/dev/null | \
+                jq "[.items[].metadata.annotations // {} | .[\"metal3.io/BareMetalHost\"] // empty] | length")
+            if [[ "${ASSIGNED}" -ge "${NUM_WORKERS}" ]]; then
+                echo "All ${NUM_WORKERS} worker machines have a host assigned"
+                break
+            fi
+            echo "Waiting: ${ASSIGNED}/${NUM_WORKERS} worker machines have a host assigned"
+            sleep 10
```
dtantsur, line 115: This is 20 minutes of waiting by default, are you sure the environments with unreachable workers won't timeout anyway?

elfosardo: the timeout is a  non fatal wait, and 20 minutes should be ok in the overall time budget
we can evaluate after some runs if we want to reduce it to 15 or even 10 minute, but I would start with this

PR #1948, opened 2026-08-14
Revert argcomplete manual pinning. Was fixed in pypi

tdomnesc:
3.7.2 requires Python 3.10+ and 3.7.1 was yanked

This reverts commit 0a4d7c70ad7c423d00496315062b91f94417dd0b. This reverts commit f68bc574f57a75402a010184be7e0fcbbced6a3c.


tdomnesc:
I'm giving up on not required jobs, they fail for unrelated reasons, won't retry.

PR #1951, opened 2026-08-17
AGENT-1581: For ABI allow pre-built OVE ISO via AGENT_OVE_ISO_SOURCE

bfournie:
When AGENT_E2E_TEST_BOOT_MODE=ISO_NO_REGISTRY, add an optional AGENT_OVE_ISO_SOURCE setting that lets the user supply a pre-built agent-ove ISO instead of building one locally. When set, the local build (create_agent_iso_no_registry) and its build-only cleanup steps are skipped, and the ISO is obtained from the source into the expected asset directory.

AGENT_OVE_ISO_SOURCE accepts one of:
  - a quay.io/redhat-user-workloads/... container image ref (ISO extracted from the image via podman create/cp)
  - a local path/filename to an already-downloaded ISO

### DIFF
--- a/agent/06_agent_create_cluster.sh
+++ b/agent/06_agent_create_cluster.sh
@@ -90,19 +90,6 @@ function create_config_image() {
 }
 
 
-function assert_agent_no_registry_iso_size(){
-  agent_iso_no_registry=$(get_agent_iso_no_registry)
-  iso_size=$(stat -c%s "$agent_iso_no_registry")
-  
-  # With 4.19 tech preview, the expected ISO size is approximately 36GB
-  iso_size_limit=$((AGENT_OVE_ISO_SIZE * 1024 * 1024 * 1024))
-
-  if (( iso_size > iso_size_limit )); then
-    echo "Error: OVE ISO size of $agent_iso_no_registry is ${iso_size}, which exceeds the ${AGENT_OVE_ISO_SIZE}GB limit."
-    exit 1
-  fi
-}
-
 function set_device_config_image() {
 
     for (( n=0; n<${2}; n++ ))
@@ -808,20 +795,24 @@ case "${AGENT_E2E_TEST_BOOT_MODE}" in
     # Run a script from agent-installer-utils which internally uses openshift-appliance
     asset_dir=$SCRIPTDIR/$OCP_DIR/iso_builder
     mkdir -p "${asset_dir}"
-    create_agent_iso_no_registry "${asset_dir}"
 
-    assert_agent_no_registry_iso_size
+    if [[ -n "${AGENT_OVE_ISO_SOURCE}" ]]; then
+      # A pre-built OVE ISO was provided - fetch it instead of building one locally.
+      fetch_agent_iso_no_registry "${asset_dir}"
+    else
+      create_agent_iso_no_registry "${asset_dir}"
 
-    if [[ "$AGENT_CLEANUP_ISO_BUILDER_CACHE_LOCAL_DEV" == "true" ]]; then
-      # reclaim disk space by deleting unwanted cache, other files
-      cleanup_diskspace_agent_iso_noregistry "${asset_dir}"
-    fi
+      if [[ "$AGENT_CLEANUP_ISO_BUILDER_CACHE_LOCAL_DEV" == "true" ]]; then
+        # reclaim disk space by deleting unwanted cache, other files
+        cleanup_diskspace_agent_iso_noregistry "${asset_dir}"
+      fi
 
-    # Clean up registry data to save disk space after ISO is created
-    if [[ "${MIRROR_IMAGES}" == "true" ]]; then
-      echo "Cleaning up registry data at ${REGISTRY_DIR} to save disk space"
-      sudo rm -rf "${REGISTRY_DIR}/data"
-      echo "Registry data cleanup complete"
+      # Clean up registry data to save disk space after ISO is created
+      if [[ "${MIRROR_IMAGES}" == "true" ]]; then
+        echo "Cleaning up registry data at ${REGISTRY_DIR} to save disk space"
+        sudo rm -rf "${REGISTRY_DIR}/data"
+        echo "Registry data cleanup complete"
+      fi
     fi
 
     attach_agent_iso_no_registry master "$NUM_MASTERS"
--- a/agent/iso_no_registry.sh
+++ b/agent/iso_no_registry.sh
@@ -57,6 +57,104 @@ function build_ove_iso_container() {
   mv "./output-iso/${iso_name}" "${asset_dir}"
 }
 
+# Inject the local SSH public key into a pre-built OVE ISO's embedded live ignition.
+#
+# Locally-built ISOs (build-ove-image.sh --ssh-key-file) but a pre-built ISO
+# fetched via AGENT_OVE_ISO_SOURCE may not. This adds the key to the 'core' user.
+# Note that if the pre-built ISO already has the ssh key this will not create
+# a duplicate but it will overwrite the current key.
+function inject_ssh_key_into_ove_iso() {
+  local iso="${1}"
+  local dir base tmp ssh_pub coreos_installer
+
+  if [[ ! -f "${SSH_KEY_FILE}" ]]; then
+    echo "Warning: SSH_KEY_FILE (${SSH_KEY_FILE}) not found; skipping SSH key injection." >&2
+    return 0
+  fi
+
+  dir=$(dirname "${iso}")
+  base=$(basename "${iso}")
+  ssh_pub=$(cat "${SSH_KEY_FILE}")
+  tmp=$(mktemp -d)
+
+  echo "Injecting SSH key from ${SSH_KEY_FILE} into ${iso}"
+
+  coreos_installer=(sudo podman run --privileged --rm -v /run/udev:/run/udev
+    -v "${dir}:/data" -v "${tmp}:/cfg" -w /data quay.io/coreos/coreos-installer:release)
+
+  # Extract the ISO's existing embedded live ignition so we can preserve it.
+  if ! "${coreos_installer[@]}" iso ignition show "/data/${base}" > "${tmp}/iso.ign" 2>/dev/null \
+      || ! jq -e . "${tmp}/iso.ign" >/dev/null 2>&1; then
+    # No embedded ignition (or unreadable) - start from a minimal 3.2.0 config.
+    echo '{"ignition":{"version":"3.2.0"}}' > "${tmp}/iso.ign"
+  fi
+
+  # Append our key to the 'core' user (create it if absent), keeping any existing keys.
+  jq --arg key "${ssh_pub}" '
+    .passwd = (.passwd // {}) |
+    .passwd.users = (.passwd.users // []) |
+    if any(.passwd.users[]; .name == "core")
+    then .passwd.users |= map(
+           if .name == "core"
+           then .sshAuthorizedKeys = ((.sshAuthorizedKeys // []) + [$key] | unique)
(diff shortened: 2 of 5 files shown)


bfournie:
Successfully installed using Konflux storage
`export AGENT_OVE_ISO_SOURCE=quay.io/redhat-user-workloads/ocp-art-tenant/art-agent-installer-iso@sha256:8d8e1cb2f9eedc5be5a5f7910286a9f99a068a833d7d8d2919715a17d96505c3`

```
$ oc get nodes
NAME       STATUS   ROLES                         AGE   VERSION
master-0   Ready    control-plane,master,worker   18m   v1.35.6
master-1   Ready    control-plane,master,worker   64m   v1.35.6
master-2   Ready    control-plane,master,worker   64m   v1.35.6
$ oc get clusterversion
NAME      VERSION   AVAILABLE   PROGRESSING   SINCE   STATUS
version   4.22.9    True        False         9m55s   Cluster version is 4.22.9
```

agent/isobuilder/ui_driven_cluster_installation/main.go
```diff
@@ -612,7 +612,29 @@ func networkingDetails(page *rod.Page, path string) error {
 		}
 	}
 
-	page.MustElement("#form-input-sshPublicKey-field").MustInput(sshPublicKey)
+	// The "Host SSH Public Key" section offers a "Use the same host discovery SSH key"
+	// checkbox. When it is checked - e.g. the boot ISO already provided a host discovery
+	// SSH key (as with a pre-built OVE ISO) - the manual key text field is not rendered,
+	// so unconditionally waiting for it would hang forever. Only enter the key if the
+	// field is actually present and empty; otherwise the cluster already has an SSH key.
+	sshField, sshErr := page.Timeout(15 * time.Second).Element("#form-input-sshPublicKey-field")
```
pawanpinjarkar: Do we need a timeout here? If we do need, can we lower it to 2 seconds like https://github.com/openshift-metal3/dev-scripts/blob/master/agent/isobuilder/ui_driven_cluster_installation/main.go#L321 if there isn't any animation associated, otherwise, lets keep same smaller timeouts like other places.

agent/isobuilder/ui_driven_cluster_installation/main.go
```diff
@@ -612,7 +612,29 @@ func networkingDetails(page *rod.Page, path string) error {
...
+	if sshErr != nil || sshField == nil {
+		logrus.Info("SSH public key field not present (using host discovery SSH key); skipping input")
+	} else {
+		sshField.MustScrollIntoView()
```
pawanpinjarkar: I haven't checked the actual UI but `MustScrollIntoView` should only be needed when the element is not visible in the screenshot area and we want explicitly to get it visible in the scrollable page when taking a screenshot, e.g. a next button might not always be visible unless we scroll the element into the view like here https://github.com/openshift-metal3/dev-scripts/blob/master/agent/isobuilder/ui_driven_cluster_installation/main.go#L518

agent/isobuilder/ui_driven_cluster_installation/main.go
```diff
@@ -612,7 +612,29 @@ func networkingDetails(page *rod.Page, path string) error {
...
+		existing := strings.TrimSpace(sshField.MustProperty("value").String())
+		if existing == "" {
+			if inputErr := rod.Try(func() {
+				sshField.Timeout(30 * time.Second).MustWaitInteractable().MustInput(sshPublicKey)
```
pawanpinjarkar: Similarly, why to wait for 30 seconds?

agent/isobuilder/ui_driven_cluster_installation/main.go
bfournie: Good point, lowered it to 2 seconds consistent with other existence checks.

agent/isobuilder/ui_driven_cluster_installation/main.go
bfournie: Thanks, removed this.

agent/isobuilder/ui_driven_cluster_installation/main.go
bfournie: Reduced to 5 seconds. This should probably be longer than the existence check above as it waiting for the field to be interactable.

agent/isobuilder/ui_driven_cluster_installation/main.go
```diff
@@ -612,7 +612,28 @@ func networkingDetails(page *rod.Page, path string) error {
 		}
 	}
 
-	page.MustElement("#form-input-sshPublicKey-field").MustInput(sshPublicKey)
+	// The "Host SSH Public Key" section offers a "Use the same host discovery SSH key"
```
andfasano: @bfournie is the checkbox appearing under special conditions? Because in the regular 4.22 disconnected UI I do not see it.

agent/06_agent_create_cluster.sh
```diff
@@ -808,20 +808,28 @@ case "${AGENT_E2E_TEST_BOOT_MODE}" in
...
-    assert_agent_no_registry_iso_size
+    if [[ -n "${AGENT_OVE_ISO_SOURCE}" ]]; then
+      # A pre-built OVE ISO was provided - fetch it instead of building one locally.
+      fetch_agent_iso_no_registry "${asset_dir}"
 
-    if [[ "$AGENT_CLEANUP_ISO_BUILDER_CACHE_LOCAL_DEV" == "true" ]]; then
-      # reclaim disk space by deleting unwanted cache, other files
-      cleanup_diskspace_agent_iso_noregistry "${asset_dir}"
-    fi
+      assert_agent_no_registry_iso_size
```
andfasano: the assert looks duplicated?

agent/isobuilder/ui_driven_cluster_installation/main.go
bfournie: I only saw this when testing an ISO downloaded from console.redhat.com, i.e. using Above-the-Sea to add the ssh key.

agent/iso_no_registry.sh
```diff
@@ -57,6 +57,104 @@ function build_ove_iso_container() {
...
+    if [[ ! -f "${AGENT_OVE_ISO_SOURCE}" ]]; then
+      echo "Error: OVE ISO file not found: ${AGENT_OVE_ISO_SOURCE}" >&2
+      exit 1
+    fi
+    cp "${AGENT_OVE_ISO_SOURCE}" "${dest}"
+  fi
+
+  # A pre-built ISO has no SSH key baked in - add ours so we can ssh into the
+  # live/bootstrap OVE environment (and the installed nodes once it propagates).
+  inject_ssh_key_into_ove_iso "${dest}"
```
andfasano, line 153: This is somehow confusing for me. The user has always the possibility to specify the SSH key in the disconnected UI. So:
1) Such action looks unnecessary in the normal flow, where the SSH key it's already fille by the UI automation tool
2) If the SSH key is provided in the ISO (because it was downloaded from the SaaS UI), it should be pre-filled in the UI, with no special action required from the user (apart the possibility to override it)

andfasano (commented):
I think this PR is putting together two different things (despite the title):
1) fetching the ISO from another location
2) Injecting the SSH key in the ISO

The first part looks fine fine to me in general, but the second one does not seem well defined and may deserve more discussion. I would then recommend splitting the PR in two different ones, for a quicker landing

agent/iso_no_registry.sh
bfournie: This is **only** for debug before we get to the point of it being filled by the automation tool if the ISO did not already have a pre-filled SSH key. For example if it was downloaded from developer.redhat.com, not the the SaaS UI, or from the Konflux output of build pipeline. This is important ability to debug that we'd lose

agent/isobuilder/ui_driven_cluster_installation/main.go
andfasano: Just as a note, I don't think we ever agreed such behavior, because we don't have the concept of a "discovery" key. The expected behavior is that, if the user provides a SSH key from the SaaS UI, it will be used to automatically pre-fill the field (and, if the user wants, that value could be overriden)

bfournie:
> I think this PR is putting together two different things (despite the title):
> 
> 1. fetching the ISO from another location
> 2. Injecting the SSH key in the ISO
> 
> The first part looks fine fine to me in general, but the second one does not seem well defined and may deserve more discussion. I would then recommend splitting the PR in two different ones, for a quicker landing

I can split this in 2 PRs but as per the comment above, not injecting the ssh key greatly removes our ability to debug certain ISOs e.g. ones from the Konflux build pipeline output

agent/iso_no_registry.sh
andfasano: Ok, thanks for clarifying I got somehow confused with the previous ssh code. Please disregard this comment. The only improvement that I could see will be to avoid adding an SSH key if using an ISO that already contains it. Not sure if in this case the appending will risk to duplicate the same key twice

andfasano:
> > I think this PR is putting together two different things (despite the title):
> > 
> > 1. fetching the ISO from another location
> > 2. Injecting the SSH key in the ISO
> > 
> > The first part looks fine fine to me in general, but the second one does not seem well defined and may deserve more discussion. I would then recommend splitting the PR in two different ones, for a quicker landing
> 
> I can split this in 2 PRs but as per the comment above, not injecting the ssh key greatly removes our ability to debug certain ISOs e.g. ones from the Konflux build pipeline output

After your further comments, the injection part looks fine, while I think it'd be better to remove the part about the UI automation, since it refers to a behavior that wasn't properly expected (and I think it's bug, worth a separate discussion)

agent/isobuilder/ui_driven_cluster_installation/main.go
bfournie: Yes, let's treat this as a separate issue. I'll create a bug for it and remove the handling of it from the PR (and create a separate one if needed).

agent/06_agent_create_cluster.sh
bfournie: Yes, fixed

bfournie:
> After your further comments, the injection part looks fine, while I think it'd be better to remove the part about the UI automation, since it refers to a behavior that wasn't properly expected (and I think it's bug, worth a separate discussion)

Yes, I updated the PR to remove the UI-driven section for the SSH key. Will create a bug for that and handle it in a separate PR.

agent/06_agent_create_cluster.sh
```diff
@@ -808,22 +808,28 @@ case "${AGENT_E2E_TEST_BOOT_MODE}" in
...
-      echo "Registry data cleanup complete"
+      # Clean up registry data to save disk space after ISO is created
+      if [[ "${MIRROR_IMAGES}" == "true" ]]; then
+        echo "Cleaning up registry data at ${REGISTRY_DIR} to save disk space"
+        sudo rm -rf "${REGISTRY_DIR}/data"
+        echo "Registry data cleanup complete"
+      fi
     fi
 
+    assert_agent_no_registry_iso_size
```
andfasano: This location doesn't look good, because in some cases the ISO could be gone. On a second thought, I'm wondering if we really still need this check - at the beginning of the project we weren't sure about the exact size, but now we're reached quite a stable point for what regards the content of the ISO. 
So I would be now more inclined to remove this check at all (and simplify a little bit the code)

bfournie: Yeah that function references 4.19 and its size. Removing the function and the call.

bfournie: Updated to remove use of AGENT_OVE_ISO_SIZE

PR #1952, opened 2026-08-19
OPNET-772: Add NAT64/DNS64 support for IPv6-only clusters on IPv4-only hosts

emy:
Adds optional NAT64 (TAYGA) + DNS64 (unbound) support so an IPv6-only cluster
  (IP_STACK=v6) can be deployed on an IPv4-only host (HOST_IP_STACK=v4). Enable
  with ENABLE_NAT64=true; it is a no-op otherwise.

  The helpers live in nat64.sh (sourced by 02_configure_host.sh): bring IPv6 up on
  the baremetal bridge, configure the TAYGA NAT64 tunnel, run an unbound DNS64
  resolver (dns64-synthall, since the host has no native IPv6 egress) and point the
  host and every cluster-facing libvirt network at it, and rewrite node BMC
  addresses to the IPv6 baremetal address so the IPv6-only in-cluster Ironic can
  reach them after pivot. The second commit regenerates the sushy-tools BMC cert
  with the IPv6 SAN, required on OCP >= 4.22 where certificate verification is
  always on.

  01_install_requirements.sh installs tayga/unbound, common.sh validates the
  IP_STACK/HOST_IP_STACK requirement, and host_cleanup.sh/ocp_cleanup.sh tear the
  configuration down.

### DIFF
--- a/01_install_requirements.sh
+++ b/01_install_requirements.sh
@@ -110,8 +110,15 @@ case $DISTRO in
     ;;
   "rhel10"|"centos10")
     sudo dnf -y install python3-pip
-    if sudo subscription-manager identity > /dev/null 2>&1; then
-      sudo subscription-manager repos --enable "codeready-builder-for-rhel-10-$(arch)-rpms" || true
+    if [[ $DISTRO == "centos10" ]]; then
+      sudo dnf config-manager --set-enabled crb
+      sudo dnf -y install epel-release
+    elif [[ $DISTRO == "rhel10" ]]; then
+      if sudo subscription-manager identity > /dev/null 2>&1; then
+        sudo subscription-manager repos --enable "codeready-builder-for-rhel-10-$(arch)-rpms" || true
+      fi
+      # EPEL provides tayga (needed for NAT64); mirror the EL9 rhel handling.
+      sudo dnf -y install https://dl.fedoraproject.org/pub/epel/epel-release-latest-10.noarch.rpm
     fi
     sudo ln -s /usr/bin/python3 /usr/bin/python || true
     PYTHON_DEVEL="python3-devel"
--- a/01_install_requirements.sh
+++ b/01_install_requirements.sh
@@ -207,6 +207,19 @@ if [[ "${NODES_PLATFORM:-}" == "baremetal" ]] ; then
     sudo dnf -y install ipmitool
 fi
 
+# Install NAT64 dependencies if enabled
+if [[ "${ENABLE_NAT64:-false}" == "true" ]]; then
+    echo "Installing NAT64 dependencies (TAYGA, unbound)..."
+    # TAYGA provides the NAT64 translation; unbound provides DNS64 synthesis.
+    # unbound is used rather than CoreDNS because its built-in dns64 module
+    # reliably synthesizes AAAA records (the CoreDNS dns64 plugin build did not).
+    sudo dnf -y install tayga unbound
+    # We run our own DNS64 unbound instance on a dedicated port; make sure the
+    # stock unbound.service does not also grab port 53 and clash with the host
+    # NetworkManager resolver.
+    sudo systemctl disable --now unbound.service 2>/dev/null || true
+fi
+
 retry_with_timeout 5 60 "curl -L $OPENSHIFT_CLIENT_TOOLS_URL | sudo tar -U -C /usr/local/bin -xzf -"
 sudo chmod +x /usr/local/bin/oc
 oc version --client -o json
--- a/02_configure_host.sh
+++ b/02_configure_host.sh
@@ -484,8 +484,12 @@ fi
 
 # When NAT64 is enabled, ensure IPv6 forwarding rules are set for the bridge
 if [[ "${ENABLE_NAT64}" == "true" ]]; then
-  sudo ip6tables -A FORWARD --in-interface "${BAREMETAL_NETWORK_NAME}" -j ACCEPT 2>/dev/null || true
-  sudo ip6tables -A FORWARD --out-interface "${BAREMETAL_NETWORK_NAME}" -m state --state RELATED,ESTABLISHED -j ACCEPT 2>/dev/null || true
+  # Check-then-add so re-running does not accumulate duplicate rules; cleanup_nat64
+  # removes these with the matching -D commands.
+  sudo ip6tables -C FORWARD --in-interface "${BAREMETAL_NETWORK_NAME}" -j ACCEPT 2>/dev/null || \
+    sudo ip6tables -A FORWARD --in-interface "${BAREMETAL_NETWORK_NAME}" -j ACCEPT
+  sudo ip6tables -C FORWARD --out-interface "${BAREMETAL_NETWORK_NAME}" -m state --state RELATED,ESTABLISHED -j ACCEPT 2>/dev/null || \
+    sudo ip6tables -A FORWARD --out-interface "${BAREMETAL_NETWORK_NAME}" -m state --state RELATED,ESTABLISHED -j ACCEPT
 fi
 
 # Switch NetworkManager to internal DNS
--- a/02_configure_host.sh
+++ b/02_configure_host.sh
@@ -4,6 +4,7 @@ set -euxo pipefail
 source logging.sh
 source common.sh
 source network.sh
+source nat64.sh
 source utils.sh
 source validation.sh
 source oc_mirror.sh
@@ -293,6 +294,13 @@ if [ "${NUM_EXTRA_WORKERS}" -ne 0 ] || [ "${NUM_ARM_WORKERS}" -ne 0 ]; then
   fi
 fi
 
+# For NAT64 (IPv6-only cluster on an IPv4-only host) the BMC emulator must be
+# reached over IPv6 by the in-cluster Ironic pods; rewrite the node BMC addresses
+# to the host's IPv6 baremetal address.
+if [[ "${ENABLE_NAT64}" == "true" ]]; then
+    nat64_fixup_bmc_addresses
+fi
+
 # shellcheck disable=SC2034
 ZONE="\nZONE=libvirt"
 
@@ -474,6 +482,12 @@ if [ "$EXT_IF" ]; then
   sudo $IPTABLES -A FORWARD --in-interface "${BAREMETAL_NETWORK_NAME}" -j ACCEPT
 fi
 
+# When NAT64 is enabled, ensure IPv6 forwarding rules are set for the bridge
+if [[ "${ENABLE_NAT64}" == "true" ]]; then
+  sudo ip6tables -A FORWARD --in-interface "${BAREMETAL_NETWORK_NAME}" -j ACCEPT 2>/dev/null || true
+  sudo ip6tables -A FORWARD --out-interface "${BAREMETAL_NETWORK_NAME}" -m state --state RELATED,ESTABLISHED -j ACCEPT 2>/dev/null || true
+fi
+
 # Switch NetworkManager to internal DNS
 if [ "$MANAGE_BR_BRIDGE" == "y" ]; then
   switch_to_internal_dns
@@ -536,6 +550,12 @@ fi
 sudo virsh net-list | grep "${PROVISIONING_NETWORK_NAME}" || sudo virsh net-start "${PROVISIONING_NETWORK_NAME}"
 sudo virsh net-list | grep "${BAREMETAL_NETWORK_NAME}" || sudo virsh net-start "${BAREMETAL_NETWORK_NAME}"
 
+# Configure NAT64/DNS64 if enabled
+if [[ "${ENABLE_NAT64}" == "true" ]]; then
+    configure_nat64_bridge_ipv6
+    configure_tayga
+    configure_dns64
+fi
 
 # Setup a single nfs export for image registry
 if [ "${PERSISTENT_IMAGEREG}" == true ] ; then
--- a/02_configure_host.sh
+++ b/02_configure_host.sh
@@ -101,6 +101,11 @@ if [[ "${NODES_PLATFORM}" == "baremetal" ]]; then
     # Add a /etc/hosts entry for $LOCAL_REGISTRY_DNS_NAME
     sudo sed -i "/${LOCAL_REGISTRY_DNS_NAME}/d" /etc/hosts
     echo "${PROVISIONING_HOST_EXTERNAL_IP} ${LOCAL_REGISTRY_DNS_NAME}" | sudo tee -a /etc/hosts
+    # Under NAT64 the cluster is IPv6-only; also publish an AAAA record so nodes can
+    # reach host services (e.g. the image-registry NFS export) over the host's IPv6
+    # baremetal address. The sed cleanup above matches by name, so it removes both
+    # lines on re-run (idempotent).
+    [ -n "${PROVISIONING_HOST_EXTERNAL_IP_V6:-}" ] && echo "${PROVISIONING_HOST_EXTERNAL_IP_V6} ${LOCAL_REGISTRY_DNS_NAME}" | sudo tee -a /etc/hosts
 
     # When MIRROR_IMAGES is configured, the local registry must be running
     # before 04_setup_ironic.sh attempts to mirror release images into it.
@@ -500,6 +505,11 @@ fi
 # Add a /etc/hosts entry for $LOCAL_REGISTRY_DNS_NAME
 sudo sed -i "/${LOCAL_REGISTRY_DNS_NAME}/d" /etc/hosts
 echo "${PROVISIONING_HOST_EXTERNAL_IP} ${LOCAL_REGISTRY_DNS_NAME}" | sudo tee -a /etc/hosts
+# Under NAT64 the cluster is IPv6-only; also publish an AAAA record so nodes can
+# reach host services (e.g. the image-registry NFS export) over the host's IPv6
+# baremetal address. The sed cleanup above matches by name, so it removes both
+# lines on re-run (idempotent).
+[ -n "${PROVISIONING_HOST_EXTERNAL_IP_V6:-}" ] && echo "${PROVISIONING_HOST_EXTERNAL_IP_V6} ${LOCAL_REGISTRY_DNS_NAME}" | sudo tee -a /etc/hosts
 
 if use_registry "podman"; then
     # Remove any previous file, or podman login panics when reading the
--- a/02_configure_host.sh
+++ b/02_configure_host.sh
@@ -555,6 +555,10 @@ if [[ "${ENABLE_NAT64}" == "true" ]]; then
     configure_nat64_bridge_ipv6
     configure_tayga
     configure_dns64
+    # Make the sushy BMC cert valid for the IPv6 baremetal address so the
+    # IPv6-only in-cluster Ironic can reach the BMC. Must run before step 05
+    # embeds this cert into the install-config trust bundle.
+    nat64_fixup_sushy_cert
 fi
 
 # Setup a single nfs export for image registry
--- a/common.sh
+++ b/common.sh
@@ -433,18 +433,6 @@ if [[ "${BOOTSTRAP_IN_PLACE}" == "true" ]]; then
   export NETWORK_TYPE="OVNKubernetes"
 fi
 
-# Validate NAT64 configuration
-if [[ "${ENABLE_NAT64:-false}" == "true" ]]; then
-  if [[ "${IP_STACK:-v6}" != "v6" ]]; then
-    error "ENABLE_NAT64=true requires IP_STACK=v6 (got IP_STACK=${IP_STACK:-v6})"
-    exit 1
-  fi
-  if [[ "${HOST_IP_STACK:-${IP_STACK:-v6}}" != "v4" ]]; then
-    error "ENABLE_NAT64=true requires HOST_IP_STACK=v4 (got HOST_IP_STACK=${HOST_IP_STACK:-${IP_STACK:-v6}})"
-    exit 1
-  fi
-fi
-
 # Defaults the DISABLE_MULTICAST variable
 export DISABLE_MULTICAST=${DISABLE_MULTICAST:-false}
 
@@ -642,6 +630,33 @@ if [[ ! -z ${AGENT_E2E_TEST_SCENARIO} ]]; then
   if [[ "${AGENT_E2E_TEST_BOOT_MODE}" == "PXE" ]] || [[ "${AGENT_E2E_TEST_BOOT_MODE}" == "ISCSI" ]]; then
     LIBVIRT_FIRMWARE=bios
   fi
+fi
+
+# Validate NAT64 configuration. This must run after the agent scenario parsing
+# above, which derives IP_STACK from AGENT_E2E_TEST_SCENARIO; validating earlier
+# would only ever see the default IP_STACK and silently pass for v4/dual-stack
+# agent scenarios.
+if [[ "${ENABLE_NAT64:-false}" == "true" ]]; then
+  if [[ "${IP_STACK:-v6}" != "v6" ]]; then
+    error "ENABLE_NAT64=true requires IP_STACK=v6 (got IP_STACK=${IP_STACK:-v6})"
+    exit 1
+  fi
+  if [[ "${HOST_IP_STACK:-${IP_STACK:-v6}}" != "v4" ]]; then
+    error "ENABLE_NAT64=true requires HOST_IP_STACK=v4 (got HOST_IP_STACK=${HOST_IP_STACK:-${IP_STACK:-v6}})"
+    exit 1
+  fi
+  # The IPv6-only in-cluster Ironic can only reach the BMC emulator over IPv6,
+  # which only sushy/redfish listens on; vbmc (ipmi) binds IPv4 only. Reject the
+  # ipmi and mixed drivers rather than silently leaving those BMCs unreachable.
+  if [[ "${BMC_DRIVER}" != "redfish" && "${BMC_DRIVER}" != "redfish-virtualmedia" ]]; then
+    error "ENABLE_NAT64=true requires a redfish-based BMC_DRIVER (redfish or redfish-virtualmedia), got BMC_DRIVER=${BMC_DRIVER}"
+    exit 1
+  fi
+  # NAT64 provides external registry access, so image mirroring is skipped and no
+  # local registry is created. Default MIRROR_IMAGES to false here (before the
+  # registry-override decision later in this file) so the installer is not pointed
+  # at a registry that was never stood up.
+  export MIRROR_IMAGES=${MIRROR_IMAGES:-false}
 fi
 
 if [[ ! -z ${AGENT_E2E_TEST_BOOT_MODE} ]]; then
--- a/common.sh
+++ b/common.sh
@@ -433,6 +433,18 @@ if [[ "${BOOTSTRAP_IN_PLACE}" == "true" ]]; then
   export NETWORK_TYPE="OVNKubernetes"
 fi
 
+# Validate NAT64 configuration
+if [[ "${ENABLE_NAT64:-false}" == "true" ]]; then
+  if [[ "${IP_STACK:-v6}" != "v6" ]]; then
+    error "ENABLE_NAT64=true requires IP_STACK=v6 (got IP_STACK=${IP_STACK:-v6})"
+    exit 1
+  fi
+  if [[ "${HOST_IP_STACK:-${IP_STACK:-v6}}" != "v4" ]]; then
+    error "ENABLE_NAT64=true requires HOST_IP_STACK=v4 (got HOST_IP_STACK=${HOST_IP_STACK:-${IP_STACK:-v6}})"
+    exit 1
+  fi
+fi
+
 # Defaults the DISABLE_MULTICAST variable
 export DISABLE_MULTICAST=${DISABLE_MULTICAST:-false}
 
--- a/common.sh
+++ b/common.sh
@@ -645,12 +645,35 @@ if [[ "${ENABLE_NAT64:-false}" == "true" ]]; then
     error "ENABLE_NAT64=true requires HOST_IP_STACK=v4 (got HOST_IP_STACK=${HOST_IP_STACK:-${IP_STACK:-v6}})"
     exit 1
   fi
-  # The IPv6-only in-cluster Ironic can only reach the BMC emulator over IPv6,
-  # which only sushy/redfish listens on; vbmc (ipmi) binds IPv4 only. Reject the
-  # ipmi and mixed drivers rather than silently leaving those BMCs unreachable.
-  if [[ "${BMC_DRIVER}" != "redfish" && "${BMC_DRIVER}" != "redfish-virtualmedia" ]]; then
-    error "ENABLE_NAT64=true requires a redfish-based BMC_DRIVER (redfish or redfish-virtualmedia), got BMC_DRIVER=${BMC_DRIVER}"
-    exit 1
+  # The IPv6-only in-cluster Ironic reaches redfish (sushy) over the host's native
+  # IPv6 address, and reaches ipmi (vbmc, IPv4-only) via NAT64 translation of its
+  # IPv4 address (see nat64_fixup_bmc_addresses). Allow redfish, redfish-virtualmedia,
+  # ipmi and mixed; idrac/others are not emulated here, so still reject them.
+  case "${BMC_DRIVER}" in
+    redfish|redfish-virtualmedia|ipmi|mixed) ;;
+    *)
+      error "ENABLE_NAT64=true supports BMC_DRIVER redfish, redfish-virtualmedia, ipmi or mixed, got BMC_DRIVER=${BMC_DRIVER}"
+      exit 1
+      ;;
+  esac
+  # ipmi BMCs are reached through NAT64 by embedding their RFC1918 IPv4 address in
+  # NAT64_PREFIX. RFC 6052 forbids using the well-known prefix 64:ff9b::/96 for
+  # non-global (RFC1918) IPv4, so mixed/ipmi need a ULA prefix. Default one here
+  # (exported before network.sh applies its own 64:ff9b default), and reject an
+  # explicit well-known prefix rather than failing opaquely at BMC registration.
+  if [[ "${BMC_DRIVER}" == "mixed" || "${BMC_DRIVER}" == "ipmi" ]]; then
+    export NAT64_PREFIX=${NAT64_PREFIX:-"fd00:64::/96"}
+    if [[ "${NAT64_PREFIX}" == 64:ff9b::* ]]; then
+      error "ENABLE_NAT64=true with BMC_DRIVER=${BMC_DRIVER} cannot use the well-known NAT64 prefix ${NAT64_PREFIX} to reach the RFC1918 BMC network; set a ULA prefix, e.g. NAT64_PREFIX=fd00:64::/96"
+      exit 1
+    fi
+    # nat64_embed_v4 embeds the BMC's IPv4 in the last 32 bits, which is only the
+    # correct RFC 6052 layout for a /96 whose text ends in "::". Reject anything
+    # else here rather than silently producing an unreachable BMC address.
+    if [[ ! "${NAT64_PREFIX}" =~ ::/96$ ]]; then
+      error "ENABLE_NAT64=true with BMC_DRIVER=${BMC_DRIVER} requires a /96 NAT64_PREFIX ending in '::' (e.g. fd00:64::/96) to embed the ipmi BMC IPv4 address, got NAT64_PREFIX=${NAT64_PREFIX}"
+      exit 1
+    fi
   fi
   # NAT64 provides external registry access, so image mirroring is skipped and no
   # local registry is created. Default MIRROR_IMAGES to false here (before the
--- a/config_example.sh
+++ b/config_example.sh
@@ -446,9 +446,12 @@ set -x
 # IPv4-only hosts (HOST_IP_STACK=v4). Uses TAYGA for NAT64 translation and
 # unbound for DNS64 synthesis, enabling cluster VMs to reach external IPv4
 # resources (e.g. container registries) via synthesized IPv6 addresses.
-# Requires: IP_STACK=v6, HOST_IP_STACK=v4, and a redfish-based BMC_DRIVER
-# (redfish or redfish-virtualmedia) since the IPv6-only in-cluster Ironic can
-# only reach the BMC emulator over IPv6 (vbmc/ipmi listens on IPv4 only).
+# Requires: IP_STACK=v6 and HOST_IP_STACK=v4. Supported BMC_DRIVERs: redfish,
+# redfish-virtualmedia, ipmi and mixed. redfish (sushy) is reached over the host's
+# native IPv6 address; ipmi (vbmc, IPv4-only) is reached through NAT64 by embedding
+# its IPv4 address in NAT64_PREFIX. Because that embedding targets the RFC1918
+# baremetal network, BMC_DRIVER=mixed/ipmi requires a ULA NAT64_PREFIX (defaulted to
+# fd00:64::/96) -- the well-known 64:ff9b::/96 cannot carry RFC1918 (RFC 6052).
 # Default: false
 #
 #export ENABLE_NAT64=true
@@ -456,9 +459,11 @@ set -x
 # NAT64_PREFIX -
 # IPv6 prefix used for NAT64 address translation. Packets sent to addresses
 # in this prefix are translated to IPv4 by TAYGA.
-# Default: "64:ff9b::/96"
+# Default: "64:ff9b::/96" (redfish drivers), or "fd00:64::/96" when
+# BMC_DRIVER=mixed/ipmi (those reach the RFC1918 BMC network through NAT64).
 # Note: The well-known prefix 64:ff9b::/96 cannot reach RFC1918 addresses.
 # Use a ULA prefix (e.g. "fd00:64::/96") if you need to reach private IPv4.
+# Only the /96 length is supported for BMC address embedding.
 #
 #export NAT64_PREFIX="64:ff9b::/96"
 
--- a/config_example.sh
+++ b/config_example.sh
@@ -444,9 +444,11 @@ set -x
 # ENABLE_NAT64 -
 # Enable NAT64/DNS64 to allow IPv6-only clusters (IP_STACK=v6) to run on
 # IPv4-only hosts (HOST_IP_STACK=v4). Uses TAYGA for NAT64 translation and
-# CoreDNS for DNS64 synthesis, enabling cluster VMs to reach external IPv4
+# unbound for DNS64 synthesis, enabling cluster VMs to reach external IPv4
 # resources (e.g. container registries) via synthesized IPv6 addresses.
-# Requires: IP_STACK=v6 and HOST_IP_STACK=v4
+# Requires: IP_STACK=v6, HOST_IP_STACK=v4, and a redfish-based BMC_DRIVER
+# (redfish or redfish-virtualmedia) since the IPv6-only in-cluster Ironic can
+# only reach the BMC emulator over IPv6 (vbmc/ipmi listens on IPv4 only).
 # Default: false
 #
 #export ENABLE_NAT64=true
@@ -471,6 +473,24 @@ set -x
 # Default: "192.168.255.1"
 #
 #export NAT64_V4_ADDR="192.168.255.1"
+
+# NAT64_V6_ADDR -
+# TAYGA's own IPv6 address (source of the ICMPv6 errors it generates, e.g.
+# for PMTUD). Defaults to the 3rd address of EXTERNAL_SUBNET_V6, which is
+# on-link on the baremetal bridge. If you carve TAYGA its own subnet, set this
+# to an address outside any locally-attached subnet and route it to the NAT64
+# TUN device so error replies are delivered symmetrically.
+# Default: nth_ip(EXTERNAL_SUBNET_V6, 3)
+#
+#export NAT64_V6_ADDR="fd2e:6f44:5dd8:c956::3"
+
+# NAT64_DNS64_UPSTREAM -
+# Space-separated upstream resolver(s) the DNS64 unbound instance forwards to.
+# By default these are auto-discovered from the host's resolver config; set this
+# explicitly when the host has no usable upstream in resolv.conf (otherwise DNS64
+# falls back to 8.8.8.8, which is usually unreachable on a firewalled NAT64 host).
+#
+#export NAT64_DNS64_UPSTREAM="10.0.0.53"
 
 # ENABLE_BOOTSTRAP_STATIC_IP -
 # Configure a static IP for the bootstrap VM external NIC
--- a/config_example.sh
+++ b/config_example.sh
@@ -441,6 +441,37 @@ set -x
 #export EXTERNAL_SUBNET_V4="192.168.111.0/24"
 #export EXTERNAL_SUBNET_V6="fd2e:6f44:5dd8:c956::/120"
 
+# ENABLE_NAT64 -
+# Enable NAT64/DNS64 to allow IPv6-only clusters (IP_STACK=v6) to run on
+# IPv4-only hosts (HOST_IP_STACK=v4). Uses TAYGA for NAT64 translation and
+# CoreDNS for DNS64 synthesis, enabling cluster VMs to reach external IPv4
+# resources (e.g. container registries) via synthesized IPv6 addresses.
+# Requires: IP_STACK=v6 and HOST_IP_STACK=v4
+# Default: false
+#
+#export ENABLE_NAT64=true
+
+# NAT64_PREFIX -
+# IPv6 prefix used for NAT64 address translation. Packets sent to addresses
+# in this prefix are translated to IPv4 by TAYGA.
+# Default: "64:ff9b::/96"
+# Note: The well-known prefix 64:ff9b::/96 cannot reach RFC1918 addresses.
+# Use a ULA prefix (e.g. "fd00:64::/96") if you need to reach private IPv4.
+#
+#export NAT64_PREFIX="64:ff9b::/96"
+
+# NAT64_V4_POOL -
+# IPv4 address pool used by TAYGA for dynamic NAT64 mappings.
+# Default: "192.168.255.0/24"
+#
+#export NAT64_V4_POOL="192.168.255.0/24"
+
+# NAT64_V4_ADDR -
+# TAYGA's own IPv4 address on the NAT64 TUN interface.
+# Default: "192.168.255.1"
+#
+#export NAT64_V4_ADDR="192.168.255.1"
+
 # ENABLE_BOOTSTRAP_STATIC_IP -
 # Configure a static IP for the bootstrap VM external NIC
 # (Currently this just expects a non-empty value, the IP is fixed to .9)
--- a/host_cleanup.sh
+++ b/host_cleanup.sh
@@ -3,6 +3,8 @@ set -x
 
 source logging.sh
 source common.sh
+source network.sh
+source nat64.sh
 source utils.sh
 source validation.sh
 
@@ -29,10 +31,15 @@ ansible-playbook \
     -e "virthost=$HOSTNAME" \
     -e "manage_baremetal=$MANAGE_BR_BRIDGE" \
     -e "nodes_file=$NODES_FILE" \
-    -i "${VM_SETUP_PATH}/inventory.ini" \
-    -b -vvv "${VM_SETUP_PATH}/teardown-playbook.yml"
+    -i ${VM_SETUP_PATH}/inventory.ini \
+    -b -vvv ${VM_SETUP_PATH}/teardown-playbook.yml
 
-sudo rm -rf "/etc/NetworkManager/dnsmasq.d/openshift-${CLUSTER_NAME}.conf" /etc/yum.repos.d/delorean*
+# Clean up NAT64/DNS64 if enabled
+if [[ "${ENABLE_NAT64:-false}" == "true" ]]; then
+    cleanup_nat64
+fi
+
+sudo rm -rf /etc/NetworkManager/dnsmasq.d/openshift-${CLUSTER_NAME}.conf /etc/yum.repos.d/delorean*
 sudo rm -rf /etc/NetworkManager/conf.d/dnsmasq.conf
 sudo rm -rf /etc/NetworkManager/dnsmasq.d/upstream.conf
 if systemctl is-active --quiet NetworkManager; then
@@ -44,8 +51,8 @@ fi
 # handle upgrade from legacy network scripts
 for interface in ${PROVISIONING_NETWORK_NAME} ${BAREMETAL_NETWORK_NAME} ${PRO_IF} ${INT_IF}; do
   interface_config=/etc/sysconfig/network-scripts/ifcfg-${interface}
-  if [ -e "$interface_config" ]; then
-    sudo rm -f "$interface_config"
+  if [ -e $interface_config ]; then
+    sudo rm -f $interface_config
     IF_FOUND=true
   fi
 done
@@ -57,22 +64,22 @@ fi
 # There was a bug in this file, it may need to be recreated.
 # delete the interface as it can cause issues when not rebooting
 if [ "$MANAGE_PRO_BRIDGE" == "y" ]; then
-    sudo nmcli con del "${PROVISIONING_NETWORK_NAME}" || true
-    sudo ip link delete "${PROVISIONING_NETWORK_NAME}" || true
+    sudo nmcli con del ${PROVISIONING_NETWORK_NAME} || true
+    sudo ip link delete ${PROVISIONING_NETWORK_NAME} || true
     if [[ -d /sys/class/net/pro-ipv6-dummy ]]; then
        sudo ip link delete pro-ipv6-dummy || true
     fi
-    sudo rm -f "/etc/NetworkManager/system-connections/${PROVISIONING_NETWORK_NAME}.nmconnection"
+    sudo rm -f /etc/NetworkManager/system-connections/${PROVISIONING_NETWORK_NAME}.nmconnection
 fi
 # Leaving this around causes issues when the host is rebooted
 # delete the interface as it can cause issues when not rebooting
 if [ "$MANAGE_BR_BRIDGE" == "y" ]; then
-    sudo nmcli con del "${BAREMETAL_NETWORK_NAME}" || true
-    sudo ip link delete "${BAREMETAL_NETWORK_NAME}" || true
+    sudo nmcli con del ${BAREMETAL_NETWORK_NAME} || true
+    sudo ip link delete ${BAREMETAL_NETWORK_NAME} || true
     if [[ -d /sys/class/net/bm-ipv6-dummy ]]; then
        sudo ip link delete bm-ipv6-dummy || true
     fi
-    sudo rm -f "/etc/NetworkManager/system-connections/${BAREMETAL_NETWORK_NAME}.nmconnection"
+    sudo rm -f /etc/NetworkManager/system-connections/${BAREMETAL_NETWORK_NAME}.nmconnection
 fi
 
 # Drop all ebtables rules
--- a/host_cleanup.sh
+++ b/host_cleanup.sh
@@ -31,15 +31,16 @@ ansible-playbook \
     -e "virthost=$HOSTNAME" \
     -e "manage_baremetal=$MANAGE_BR_BRIDGE" \
     -e "nodes_file=$NODES_FILE" \
-    -i ${VM_SETUP_PATH}/inventory.ini \
-    -b -vvv ${VM_SETUP_PATH}/teardown-playbook.yml
+    -i "${VM_SETUP_PATH}/inventory.ini" \
+    -b -vvv "${VM_SETUP_PATH}/teardown-playbook.yml"
 
-# Clean up NAT64/DNS64 if enabled
-if [[ "${ENABLE_NAT64:-false}" == "true" ]]; then
-    cleanup_nat64
-fi
+# Always clean up NAT64/DNS64. cleanup_nat64 is fully idempotent (every step is a
+# no-op when nothing is present), so run it unconditionally rather than gating on
+# the current ENABLE_NAT64 value: otherwise unsetting ENABLE_NAT64 before teardown
+# would strand the unbound service, dnsmasq drop-in, TUN device, routes and rules.
+cleanup_nat64
 
-sudo rm -rf /etc/NetworkManager/dnsmasq.d/openshift-${CLUSTER_NAME}.conf /etc/yum.repos.d/delorean*
+sudo rm -rf "/etc/NetworkManager/dnsmasq.d/openshift-${CLUSTER_NAME}.conf" /etc/yum.repos.d/delorean*
 sudo rm -rf /etc/NetworkManager/conf.d/dnsmasq.conf
 sudo rm -rf /etc/NetworkManager/dnsmasq.d/upstream.conf
 if systemctl is-active --quiet NetworkManager; then
@@ -51,8 +52,8 @@ fi
 # handle upgrade from legacy network scripts
 for interface in ${PROVISIONING_NETWORK_NAME} ${BAREMETAL_NETWORK_NAME} ${PRO_IF} ${INT_IF}; do
   interface_config=/etc/sysconfig/network-scripts/ifcfg-${interface}
-  if [ -e $interface_config ]; then
-    sudo rm -f $interface_config
+  if [ -e "$interface_config" ]; then
+    sudo rm -f "$interface_config"
     IF_FOUND=true
   fi
 done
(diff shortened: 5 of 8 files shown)


mkowalski:
## Deep Review — Multi-Specialist Panel

**Disposition: REQUEST_CHANGES**

Reviewed `28a095e` against merge base `e2790c1` (8 files, +446/-16). Seven specialists reviewed independently; every BLOCKING claim asserting a runtime bug was handed to a separate reproducer agent for verification. **8 of 8 reproducers ran; 8 confirmed** (2 with material corrections that changed severity).

The design is sound and the hard parts — `dns64-synthall` to force all egress through NAT64, pointing the extra networks' dnsmasq at DNS64, the IPv6 SAN on the sushy cert — are the right calls, and the comments explaining *why* are genuinely good. The problems are concentrated in lifecycle: setup is not re-runnable, nothing survives a reboot, and cleanup is gated on config rather than on applied state.

---

### Specialist Findings

<details>
<summary><b>Functional bugs</b> — 2 blocking, 6 suggestions, 4 notes</summary>

- **BLOCKING** `network.sh:158` — `ENABLE_NAT64` leaves `MIRROR_IMAGES` empty, so the installer is still pointed at a registry that is never created or populated. *Reproduced.*
- **BLOCKING** `nat64.sh:49` — `configure_tayga` is not idempotent; `tayga --mktun` is unguarded. *Reproduced.*
- SUGGESTION `nat64.sh:177` — `_nat64_point_libvirt_dns64` destroys the baremetal bridge but never re-attaches `INT_IF` (cf. `02_configure_host.sh:451-453`).
- SUGGESTION `nat64.sh:311` — `tayga --rmtun` runs before the daemon is killed, so the persistent TUN leaks.
- SUGGESTION `nat64.sh:208` — jq aborts on a node lacking `driver_info.address`, leaving `.nat64` temp files behind.
- SUGGESTION `nat64.sh:275` — `_nat64_remove_legacy_coredns` runs `rm -rf /etc/coredns` on every `configure_dns64`.
- SUGGESTION `nat64.sh:68` — no systemd unit for TAYGA, so translation does not survive reboot while DNS64 does.
- SUGGESTION `ocp_cleanup.sh:22` — removes the DNS64 drop-in that `make redeploy` never restores; duplicates the literal instead of using `${NAT64_DNSMASQ_CONF}`.
- NOTE `02_configure_host.sh:487` — `ip6tables` rules appended without `-C`, never removed by cleanup.
- NOTE `network.sh:129` — `NAT64_V6_ADDR` defaults inside `EXTERNAL_SUBNET_V6`, so it is on-link on the bridge rather than reached via the TUN; affects TAYGA-generated ICMPv6 (PMTUD).
- NOTE `02_configure_host.sh:416` — making `EXTERNAL_SUBNET_V6` non-empty on a v4 host silently flips a pre-existing `INT_IF` bridge branch.
- NOTE `host_cleanup.sh:34` — unrelated revert of the METAL-1845 quoting fixes.
</details>

<details>
<summary><b>Adversarial</b> — 5 blocking, 3 suggestions</summary>

- **BLOCKING** `nat64.sh:49` — NAT64 setup is not rerunnable (`Device or resource busy`). *Reproduced.*
- **BLOCKING** `nat64.sh:208` — IPMI endpoints rewritten to IPv6 without an IPv6 listener. *Reproduced.*
- **BLOCKING** `nat64.sh:239` — certificate idempotence check always misses IPv6. *Reproduced.*
- **BLOCKING** `nat64.sh:68` — NAT64 disappears after host reboot. *Reproduced.*
- **BLOCKING** `01_install_requirements.sh:216` — clean EL10 hosts cannot install TAYGA. *Reproduced.*
- SUGGESTION `nat64.sh:311` — cleanup leaves the persistent TUN device.
- SUGGESTION `nat64.sh:185` — `${net}-dmy` can exceed the 15-char interface limit for longer `CLUSTER_NAME` values; the failure is swallowed by `|| true`.
- SUGGESTION `common.sh:437` — agent scenarios bypass stack validation.
</details>

<details>
<summary><b>Security</b> — 6 blocking, 3 suggestions (2 downgraded on verification)</summary>

- **BLOCKING** `nat64.sh:275` — `rm -rf /etc/coredns` can erase an unrelated host CoreDNS deployment.
- **BLOCKING** `nat64.sh:312` — global `/etc/tayga.conf`, fixed interface/data names and a broad `pkill` can overwrite or stop a pre-existing host TAYGA instance.
- **BLOCKING** `01_install_requirements.sh:220` — stock `unbound.service` is stopped and disabled with no record of prior state and never restored.
- BLOCKING → **downgraded to SUGGESTION** `02_configure_host.sh:487` — "IPv6 forwarding bypasses NAT64 isolation". Over-breadth is real, but the isolation-bypass framing was **disproven**: the FORWARD policy is already `accept` on a stock host, `-A` cannot override an earlier DROP, firewalld filters in a separate nft base chain at `priority filter + 10`, and on a real host the rules sit behind `LIBVIRT_FWO` and matched **0 packets**.
- BLOCKING → **downgraded to SUGGESTION** `nat64.sh:210` — "BMC credential permissions widened". The `mv` really does drop the mode (0600 → 0644), but the premise fails: `NODES_FILE` is already 0644 at birth (metal3-dev-env's `template` task sets no `mode:`), so this widens nothing in the default flow.
- BLOCKING → **downgraded to SUGGESTION** `nat64.sh:90` — silent fallback to `8.8.8.8`. Fail-open is worth fixing, but the file's own comment notes NAT64 hosts often cannot reach public resolvers, so this is a broken default more than a leak in a lab tool.
- SUGGESTION `nat64.sh:110` — `module-config: "dns64 iterator"` omits the validator; no trust anchor.
- SUGGESTION `nat64.sh:68` — TAYGA parses packets as root with no user/group/chroot.
- SUGGESTION `host_cleanup.sh:42` — quoting removed from privileged cleanup paths.
</details>

<details>
<summary><b>Architecture</b> — 5 blocking, 1 suggestion</summary>

- **BLOCKING** `nat64.sh:68` — TAYGA has no managed lifecycle; setup, cleanup and reboot all diverge.
- **BLOCKING** `nat64.sh:178` — undefine/redefine drops libvirt `autostart`. *Reproduced.*
- **BLOCKING** `host_cleanup.sh:38` — cleanup keys off current config rather than applied state.
- **BLOCKING** `nat64.sh:239` — certificate check is not idempotent. *Reproduced.*
- BLOCKING → **moved to open question** `network.sh:190` — `PROVISIONING_HOST_EXTERNAL_IP` stays IPv4 under NAT64 (see below).
- SUGGESTION `nat64.sh:279` — host mutations are not symmetrically removed.
</details>

<details>
<summary><b>Consistency</b> — 3 blocking, 3 suggestions, 1 note</summary>

- **BLOCKING** `host_cleanup.sh:38` — NAT64 teardown depends on current config, unlike the unconditional LLDP/BGP cleanup right beside it.
- **BLOCKING** `02_configure_host.sh:487` — unconditional `-A` while the adjacent NAT64 rule at `nat64.sh:64-65` correctly uses `-C ... || -A`.
- **BLOCKING** `01_install_requirements.sh:216` — EL10 branch lacks the repo that ships TAYGA.
- SUGGESTION `nat64.sh:5` — TAYGA state is not instance-scoped.
- SUGGESTION `ocp_cleanup.sh:22` — cluster cleanup partially dismantles host NAT64.
- SUGGESTION `host_cleanup.sh:34` — unrelated quoting regression.
- NOTE `config_example.sh:447` — names the wrong DNS64 implementation.
</details>

<details>
<summary><b>QA</b> — 5 blocking, 1 suggestion, 1 note</summary>

- **BLOCKING** `common.sh:437` — agent scenarios bypass NAT64 validation.
- **BLOCKING** `01_install_requirements.sh:216` — TAYGA unavailable on clean EL10.
- **BLOCKING** `nat64.sh:49` — TAYGA lifecycle is not idempotent; test setup/setup/cleanup/cleanup.
- **BLOCKING** `host_cleanup.sh:38` — disabling the flag skips cleanup; test enable → disable → clean.
- **BLOCKING** `nat64.sh:177` — reboot drops the NAT64 path.
- SUGGESTION `02_configure_host.sh:553` — no job in the Prow matrix sets `ENABLE_NAT64`; nothing exercises this code.
- NOTE `nat64.sh:168` — ShellCheck SC2024 (`sudo` does not apply to the redirect); the Prow shellcheck step passes regardless.
</details>

<details>
<summary><b>Technical writer</b> — 1 blocking, 3 suggestions (1 finding discarded)</summary>

- **BLOCKING** `config_example.sh:444` — enabling NAT64 disables stock unbound, redirects host and libvirt DNS, enables forwarding, adds routes/rules, rewrites BMC data and replaces the sushy cert; cleanup restores none of that and only runs while `ENABLE_NAT64` is still true. Neither the retained changes nor a manual rollback are documented.
- SUGGESTION `config_example.sh:447` — says CoreDNS provides DNS64; the code installs and runs unbound and treats CoreDNS as legacy.
- SUGGESTION `README.md` — NAT64 is undiscoverable; no mention of the libvirt/sushy-only scope (`NODES_PLATFORM=baremetal` exits before all of it), required IPv6 kernel support, or prefix/pool constraints.
- SUGGESTION `config_example.sh:432` — under `ENABLE_NAT64` the documented `EXTERNAL_SUBNET_V6=""` and "IPv6 defaults `MIRROR_IMAGES` to true" statements are both untrue.
- *Discarded:* a reported mismatch for `TAYGA_IPV4_ADDR` / `NAT64_DNS_IPV6` was an artifact of the dispatch prompt, not of this PR. Those names appear nowhere in the diff. `config_example.sh:469` correctly documents the real `NAT64_V4_ADDR`. **Not a defect.**
</details>

---

### Panel Synthesis

Convergence was high: `nat64.sh:49` (idempotence), `host_cleanup.sh:38` (cleanup gating), `01_install_requirements.sh:216` (EL10) and `nat64.sh:68`/`:239` were each raised independently by three or more specialists.

Two BLOCKING security findings were **refuted by their own reproducers** and downgraded — worth stating plainly, because both read as serious until tested. Conversely, the single highest-impact finding (`MIRROR_IMAGES`) came from one specialist and was decisive once replayed against the real sourcing order.

---

### Required Actions

1. **`network.sh:158` — a default `ENABLE_NAT64=true` deploy points the installer at a registry that is never created.**
   `common.sh:672` runs *before* `network.sh` is sourced, and empty `!= "false"`, so `OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE` is still set to the local registry. The new guard then suppresses `MIRROR_IMAGES=true`, so `04_setup_ironic.sh:109` skips mirroring and `use_registry()` (`utils.sh:1144`) skips starting the registry entirely.
   Reproducer xtrace from the real scripts:
   ```
   ++(common.sh:672): [[ '' != \f\a\l\s\e ]]
   ++(common.sh:691): export OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE=virthost.ostest...:5000/localimages/local-release-image:latest
   ++(network.sh:158): [[ true != \t\r\u\e ]]
   ```
   Fix: set `MIRROR_IMAGES=false` explicitly rather than leaving it unset, or move the NAT64 exclusion into `common.sh` above line 672.

2. **`nat64.sh:49` — re-running `02_configure_host.sh` aborts.**
   `sudo tayga --mktun` is unguarded under `set -euxo pipefail`. Run 1 exits 0; run 2 exits 1 with `Unable to attach tun device nat64, aborting: Device or resource busy`, and never reaches `configure_dns64` or `nat64_fixup_sushy_cert`. `cleanup_nat64` only runs from `host_cleanup.sh`, so nothing tears TAYGA down between `make` runs.
   Correction to the panel: `nat64.sh:68` is *not* the abort point — `tayga` daemonizes before `tun_setup()`, so the parent exits 0 while the child dies in the background. That is a silent defect (misleading exit status), not a fatal one. Both need guards; only `:49` breaks the run.

3. **`nat64.sh:239` — the sushy cert is rotated on every run, breaking `bmcVerifyCA`.**
   OpenSSL renders IPv6 SANs uncompressed (`FD2E:6F44:5DD8:C956:0:0:0:1`); `$v6host` is compressed (`fd2e:...::1`), so `grep -qiF` never matches. Case is not the cause — `grep -i` handles that; zero-run compression is. Proven by control: an address with no zero-run matches. The SAN is genuinely valid throughout (`openssl verify -verify_ip` → `OK`), yet the fingerprint changed on all 3 consecutive runs, contradicting the "Idempotent" docstring at `nat64.sh:224`. `ocp_install_env.sh:482-483` embeds this exact file into the install-config trust bundle.
   Fix: use `openssl x509 -checkip "$v6host"`.

4. **`01_install_requirements.sh:216` — clean EL10 cannot install TAYGA.**
   `centos10`/`rhel10` are supported (`:111`), but only the EL9 branch installs EPEL (`:97`, `:106`); `tayga` ships **only** in EPEL. Verified against remote repodata and end-to-end in a clean `quay.io/centos/centos:stream10` container: `Error: Unable to find a match: tayga`, exit 1 — fatal under `set -e`. `unbound` is fine on both.
   Correction to the panel: EPEL 10 **does** ship `tayga` (`0.9.6-...el10_2`); adding `epel-release` to the EL10 branch makes the install succeed.

5. **`nat64.sh:68` + `nat64.sh:177-180` — NAT64 does not survive a host reboot, and fails silently when it doesn't.**
   DNS64 is persistent (unit + `WantedBy=multi-user.target`, `enable --now`, dnsmasq drop-in, network XML). The translator is not: bare forked daemon, `ip route add`, `sysctl -w`, `iptables -A`. After a reset, DNS64 still synthesizes AAAA in the prefix while `ip -6 route get 64:ff9b::102:304` returns `Network is unreachable` — DNS answers successfully and traffic blackholes.
   Also: `virsh net-undefine` clears `autostart` (runtime-confirmed: `yes` → `net-destroy` still `yes` → after undefine/define `no`), and metal3-dev-env explicitly sets it (`network_setup_tasks.yml:65-70`). So the networks stay down after reboot too.
   Note the installed `tayga` RPM already ships `/usr/lib/systemd/system/tayga@.service`, which the PR never enables — and it reads `/etc/tayga/<instance>.conf`, not the `/etc/tayga.conf` written at `nat64.sh:5`.

6. **`nat64.sh:275` — `rm -rf /etc/coredns` runs on every `configure_dns64`, not just cleanup.**
   Invoked from `configure_dns64` (`nat64.sh:78`) as well as `cleanup_nat64` (`:283`). No merged dev-scripts code has ever created `coredns-nat64.service` or `/etc/coredns`, so on a real host the only thing this can delete is an unrelated, host-owned CoreDNS directory. Please drop the helper, or scope it to a dev-scripts-owned path and call it only from cleanup.

7. **`nat64.sh:208` — `ipmi://` BMCs are rewritten to IPv6 but VirtualBMC only listens on IPv4.**
   The rewrite is scheme-agnostic and gated only on `ENABLE_NAT64` (`02_configure_host.sh:300-302`). metal3-dev-env overrides vbmc's own `::` default with an IPv4 literal (`virtbmc/tasks/setup_tasks.yml:41-53`, `:90`), and pyghmi derives the socket family from it. Verified with real `ipmitool` against the real vbmc image: bound to IPv4, `-H ::1` fails; bound to `::`, both succeed. sushy is unaffected because its entrypoint force-adds `--interface ::`.
   Scope: the non-CI default `BMC_DRIVER=redfish` is safe. This bites `BMC_DRIVER=ipmi`, and CI where the default is `mixed` (`common.sh:202`) and roughly one node in three gets ipmi. Either reject ipmi/mixed under NAT64 or rewrite the vbmc address to `::`.

8. **`host_cleanup.sh` + `ocp_cleanup.sh` — unrelated revert of merged commit `e741cb5`.**
   `host_cleanup.sh:34-35, 54-55, 67-68, 72, 77-78, 82` and `ocp_cleanup.sh:21` drop the quoting added by "METAL-1845: shellcheck fixes part 4 (#1906)". Confirmed as a verbatim revert by diffing against `e741cb5` — almost certainly a bad rebase. It reintroduces word splitting and globbing on `${VM_SETUP_PATH}`, `${CLUSTER_NAME}`, `${PROVISIONING_NETWORK_NAME}`, `${BAREMETAL_NETWORK_NAME}` and `$interface_config` inside `rm -f`/`rm -rf`/`ip link delete`. Please restore and keep the diff to NAT64.

9. **`host_cleanup.sh:38` — cleanup is gated on current config, not applied state.**
   Raised independently by architecture, consistency and QA. Unset `ENABLE_NAT64` before `make clean` and the unbound service, dnsmasq drop-in, TUN, routes and rules all survive. Worse, changing `IP_STACK`/`HOST_IP_STACK` first makes the new `common.sh` validation `exit 1` *before* cleanup starts. `cleanup_nat64` is already absence-tolerant, so calling it unconditionally is the smaller change.

10. **`common.sh:436` — agent scenarios bypass the new validation.**
    Validation runs at `436-446`, but `IP_STACK` is derived from `AGENT_E2E_TEST_SCENARIO` at `common.sh:488`. With `IP_STACK` unset, `ENABLE_NAT64=true` + `HA_IPV4` passes on the `:-v6` default, then configures NAT64 and rewrites BMCs for an IPv4 cluster. Move the check after scenario parsing.

---

### Question for the author

`network.sh:190` — `PROVISIONING_HOST_EXTERNAL_IP` keys off `HOST_IP_STACK`, which is `v4` under NAT64, so it resolves to the IPv4 baremetal address. Consumers include `NTP_SERVERS` (`utils.sh:126`), `bootstrapExternalStaticGateway`/`bootstrapExternalStaticDNS` (`ocp_install_env.sh:111,115`) and the agent DHCP option (`agent/05_agent_configure.sh:685`). Is handing IPv6-only nodes an IPv4 address there intentional (reached via NAT64), or should NAT64 introduce a distinct cluster-facing address? **Not reproduced — flagging for your judgement rather than asserting a bug.**

---

### Optional Follow-ups

- `02_configure_host.sh:487-488` — scope to `-i $BAREMETAL -o $NAT64_TUN -s $EXTERNAL_SUBNET_V6 -d $NAT64_PREFIX`, guard with `-C ... || -A` (3 runs currently yield 3 copies), and add matching `-D` to cleanup. Note these mirror the pre-existing IPv4 rules at `:481-482`, and on a real host they sit behind `LIBVIRT_FWO` and match nothing.
- `nat64.sh:311-312` — kill the daemon *then* `--rmtun`; `pkill -f "^tayga"` also misses `/usr/sbin/tayga`.
- `nat64.sh:5-6`, `:312` — namespace `/etc/tayga.conf`, the data dir and the TUN per cluster so cleanup cannot disturb a pre-existing host TAYGA.
- `01_install_requirements.sh:220` — record and restore the stock `unbound.service` state instead of disabling it permanently.
- `nat64.sh:90` — fail closed instead of falling back to `8.8.8.8`; the comment above it already notes NAT64 hosts often cannot reach it.
- `nat64.sh:206-210` — preserve mode/ownership across the `mv` (`install -m`, or `chmod --reference`). Not exploitable today since `NODES_FILE` is already 0644, but it silently relaxes a user-supplied `NODES_FILE` protected at 0600, and fires even when the rewrite matches nothing.
- `nat64.sh:208` — `select(.driver_info.address? != null)` so a node without a BMC address doesn't abort step 02 and strand `.nat64` temp files.
- `nat64.sh:177` — re-attach `INT_IF` after `net-destroy`, mirroring `02_configure_host.sh:451-453`.
- `nat64.sh:185` — `${net}-dmy` can exceed the 15-char interface limit for longer `CLUSTER_NAME` values; the `|| true` hides it.
- `nat64.sh:129`/`network.sh:129` — `NAT64_V6_ADDR` defaults on-link on the bridge rather than behind the TUN; also undocumented in `config_example.sh`.
- `ocp_cleanup.sh:22` — `make redeploy` never re-runs `02_configure_host.sh`, so this permanently removes DNS64 forwarding; it is also ungated, hardcodes the path instead of `${NAT64_DNSMASQ_CONF}`, and skips the NetworkManager reload.
- `config_example.sh:447` — CoreDNS → unbound; document `NAT64_V6_ADDR` and qualify the `EXTERNAL_SUBNET_V6` / `MIRROR_IMAGES` defaults under NAT64.
- Add a CI job setting `ENABLE_NAT64=true`, or record the equivalent manual matrix — nothing currently exercises any of this. Highest-value cases given the above: run 02 twice, reboot, and enable → disable → clean.
- `nat64.sh:110` — `module-config: "dns64 iterator"` omits the validator and no trust anchor is initialized.

---

### Stats

| | |
|---|---|
| Specialists | 7 (bugs, adversarial, security, architecture, consistency, QA, writer) |
| Raw findings | 54 |
| After dedupe / filtering | 10 required, 1 open question, 14 follow-ups |
| Reproducers run | 8 |
| Confirmed | 8 (2 with corrections that lowered severity) |
| Refuted or discarded | 1 security framing, 1 security premise, 1 doc finding |

Reproducers ran read-only against `28a095e` in throwaway namespaces and containers; the checkout was verified unmodified afterwards.

<sub>Generated by [/code-review:deep-review](https://github.com/openshift-eng/ai-helpers/tree/main/plugins/code-review/skills/deep-review). This message was generated using AI. Please verify before acting on it.</sub>

network.sh
```diff
@@ -140,7 +155,8 @@ elif [[ "$IP_STACK" = "v6" ]]; then
   export SERVICE_SUBNET_V4=""
   export SERVICE_SUBNET_V6=${SERVICE_SUBNET_V6:-"fd02::/112"}
   export NETWORK_TYPE=${NETWORK_TYPE:-"OVNKubernetes"}
-  if [[ ${AGENT_E2E_TEST_BOOT_MODE} != "ISO_NO_REGISTRY" ]]; then
+  if [[ "${ENABLE_NAT64}" != "true" ]] && [[ ${AGENT_E2E_TEST_BOOT_MODE} != "ISO_NO_REGISTRY" ]]; then
+    # NAT64 provides external registry access, so mirroring is not required
     export MIRROR_IMAGES=${MIRROR_IMAGES:-true}
```
mkowalski, line 168: **Bug: `ENABLE_NAT64` leaves `MIRROR_IMAGES` empty, so the installer is pointed at a registry that is never created**

This guard stops `export MIRROR_IMAGES=${MIRROR_IMAGES:-true}` from running, leaving `MIRROR_IMAGES` as the empty string set in `common.sh:80`. But `common.sh:672` (`[[ "${MIRROR_IMAGES,,}" != "false" ]] && [[ ... || "$IP_STACK" = "v6" ]]`) is evaluated **before** `network.sh` is sourced — every script does `source common.sh` then `source network.sh` — and empty is not `"false"`, so `OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE` is still exported pointing at the local registry.

Meanwhile `04_setup_ironic.sh:109` (`[[ ! -z "${MIRROR_IMAGES}" && ... ]]`) is false, so `setup_release_mirror` and the `podman push` never run. `use_registry()` (`utils.sh:1144`) uses the same `! -z` gate, so `02_configure_host.sh:139` never even starts the registry container.

Note the guard is only reachable in the broken case: a user who sets `MIRROR_IMAGES=false` explicitly already gets the intended behaviour without this change.

Fix: set the value explicitly — `export MIRROR_IMAGES=${MIRROR_IMAGES:-false}` in the NAT64 branch — or move the NAT64 exclusion into `common.sh` above line 672.

<details>
<summary>Reproducer</summary>

**Steps:** Sourced the real `common.sh` then `network.sh` from this branch in the real `04_setup_ironic.sh` order, with the mirroring gates extracted from the files at runtime via `sed` so they cannot drift. Ran with `ENABLE_NAT64=true/false` as the only difference.

**Expected:** the two decisions agree — either mirroring is off and the override points upstream, or mirroring runs and populates the registry.

**Actual:** xtrace from the real scripts:
```
++(common.sh:672): [[ '' != \f\a\l\s\e ]]
++(common.sh:672): [[ v6 = \v\6 ]]
++(common.sh:691): export OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE=virthost.ostest.test.metalkube.org:5000/localimages/local-release-image:latest
++(network.sh:158): [[ true != \t\r\u\e ]]
```
`MIRROR_IMAGES=[]`, `MIRRORING_BLOCK_RUNS=no`, `USE_REGISTRY=no`, override still `.../local-release-image:latest`. Control run with `ENABLE_NAT64=false` is self-consistent (`MIRROR_IMAGES=true`, mirroring runs, registry started). 7/7 assertions passed.
</details>

nat64.sh
```diff
@@ -0,0 +1,326 @@
...
+tun-device ${NAT64_TUN_INTERFACE}
+ipv4-addr ${NAT64_V4_ADDR}
+ipv6-addr ${NAT64_V6_ADDR}
+prefix ${NAT64_PREFIX}
+dynamic-pool ${NAT64_V4_POOL}
+data-dir ${NAT64_TAYGA_DATA_DIR}
+EOF
+
+    # Create the TUN device via TAYGA
+    sudo tayga --mktun
```
mkowalski: **Bug: `configure_tayga` is not idempotent — re-running `02_configure_host.sh` aborts here**

Every other command in this function is deliberately re-runnable (`mkdir -p`, `ip route add ... || true`, `iptables -C ... || iptables -A ...`), but `sudo tayga --mktun` has no guard. `02_configure_host.sh` runs with `set -euxo pipefail` and calls `configure_tayga` unconditionally at line 556, so any second run — `make` re-run after a later failure, a very common dev-scripts workflow — aborts here. `cleanup_nat64` is only reachable from `host_cleanup.sh`, so nothing tears TAYGA down in between.

Compare `04_setup_ironic.sh:234`, which guards container creation with `if ! is_running sushy-tools`.

Fix: `[[ -d /sys/class/net/${NAT64_TUN_INTERFACE} ]] || sudo tayga --mktun`, or better, use the `tayga@.service` unit the RPM already ships (see the separate comment on line 68).

<details>
<summary>Reproducer</summary>

**Steps:** Extracted `configure_tayga` verbatim, ran it twice under the same `set -euxo pipefail`, inside a throwaway network namespace with the real `network.sh` defaults. `tayga-0.9.2-17.el9` installed from EPEL.

**Expected:** second run reconfigures and continues on to `configure_dns64` and `nat64_fixup_sushy_cert`.

**Actual:** run 1 exit 0; run 2 exit **1**:
```
+ sudo tayga --config .../tayga.conf --mktun
Unable to attach tun device nat64, aborting: Device or resource busy
```
`MARKER_REACHED_configure_dns64` never printed — host setup aborts partway through NAT64 setup. Control proves the fix: `ip link show nat64 >/dev/null 2>&1 || tayga --mktun` exits 0.

Root cause: the daemon started at line 68 holds the TUN queue open, so `TUNSETIFF` returns `EBUSY` for any second attach. Boundary condition: `--mktun` *is* idempotent when the tun exists but no daemon holds it — the failure requires a live daemon, which is exactly the steady state after a successful run.
</details>

nat64.sh
```diff
@@ -0,0 +1,326 @@
...
+    if [[ ! -f "${cert}" || ! -f "${key}" || -z "${v6host}" ]]; then
+        echo "nat64_fixup_sushy_cert: sushy cert/key or IPv6 host address missing, skipping"
+        return 0
+    fi
+
+    # Already valid for the IPv6 address? Nothing to do.
+    if sudo openssl x509 -in "${cert}" -noout -text 2>/dev/null | grep -qiF "${v6host}"; then
+        echo "sushy BMC cert already valid for ${v6host}"
+        return 0
+    fi
```
mkowalski: **Bug: this idempotence check can never match, so the sushy cert is rotated on every run**

OpenSSL renders IPv6 SANs **uncompressed** (`ipaddr_to_asc()` emits each group with `%X` and no `::` compression), while `$v6host` comes from `nth_ip` → `ansible.utils.nthhost` → netaddr, which returns the **compressed** form. So `grep -qiF` never matches even when the SAN is present and valid. Case is not the problem — `grep -i` handles that; zero-run compression is.

This matters because `ocp_install_env.sh:482-483` embeds this exact file into the install-config `bmcVerifyCA`. Rotating it invalidates the trust anchor already baked into an existing install-config or installed cluster, and contradicts the "Idempotent" docstring at line 224.

Fix: `sudo openssl x509 -in "${cert}" -noout -checkip "${v6host}"` — and ideally add the IPv6 SAN where the cert is originally generated rather than post-processing it.

<details>
<summary>Reproducer</summary>

**Steps:** Extracted `nat64_fixup_sushy_cert` (lines 226-266) and `nth_ip` (`network.sh:6-11`) verbatim, seeded the metal3-dev-env starting state (key + IPv4-only-SAN cert), then invoked the function 3× with the real defaults.

**Expected:** run 2 detects the SAN, prints "already valid", leaves the cert byte-identical.

**Actual:** printed "Regenerating..." every time; serial and SHA-256 changed on all 3 runs.
```
X509v3 Subject Alternative Name:
    IP Address:192.168.111.1, IP Address:FD2E:6F44:5DD8:C956:0:0:0:1

v6host searched for : fd2e:6f44:5dd8:c956::1
grep -qiF           => NOMATCH
```
The SAN was valid the whole time: `openssl verify -verify_ip fd2e:6f44:5dd8:c956::1` → `OK`, and `cryptography` parses it as containing that exact address.

Root-cause control (proves compression, not case): subnet `2001:db8:1:2:3:4:5:0/120` → host has no zero-run → renders identically → **MATCH**. The real default → **NOMATCH**.
</details>

01_install_requirements.sh
```diff
@@ -207,6 +207,19 @@ if [[ "${NODES_PLATFORM:-}" == "baremetal" ]] ; then
     sudo dnf -y install ipmitool
 fi
 
+# Install NAT64 dependencies if enabled
+if [[ "${ENABLE_NAT64:-false}" == "true" ]]; then
+    echo "Installing NAT64 dependencies (TAYGA, unbound)..."
+    # TAYGA provides the NAT64 translation; unbound provides DNS64 synthesis.
+    # unbound is used rather than CoreDNS because its built-in dns64 module
+    # reliably synthesizes AAAA records (the CoreDNS dns64 plugin build did not).
+    sudo dnf -y install tayga unbound
```
mkowalski, line 223: **Bug: this fails on a clean EL10 host — `tayga` ships only in EPEL, which the EL10 branch never enables**

`centos10`/`rhel10` are an explicitly supported branch (line 111; the catch-all at 119-121 exits 1 for anything else), but only the **EL9** branch installs EPEL (line 97 `epel-release`, line 106 the EPEL-9 RPM). The EL10 branch (111-118) only enables CRB via `subscription-manager ... || true`. This install sits after the `case`, so it runs on every branch. Under `set -euxo pipefail` (line 2) the failure aborts the whole script.

Nothing else enables EPEL first: the only other reference is `agent/01_agent_requirements.sh:58` (different script, gated on `AGENT_E2E_TEST_BOOT_MODE`), and the pinned metal3-dev-env `packages_installation` role has no `epel`/`tayga` references.

Fix: enable EPEL 10 in the EL10 branch — `dnf -y install epel-release` (CentOS Stream) or the EPEL-10 release RPM (RHEL) — or explicitly reject NAT64 on EL10. `unbound` is fine; it is in BaseOS/AppStream on both.

<details>
<summary>Reproducer</summary>

**Steps:** Queried remote repodata with `dnf repoquery --repofrompath` against `mirror.stream.centos.org` and `dl.fedoraproject.org`, then reproduced end-to-end in a clean `quay.io/centos/centos:stream10` container using the EL10 branch semantics (CRB on, no EPEL).

**Expected:** `dnf -y install tayga unbound` succeeds, as it does on EL9.

**Actual:**
```
Error: Unable to find a match: tayga
DNF_EXIT_CODE=1
tayga: NOT FOUND in enabled repos
unbound: FOUND
```
| repo | tayga | unbound |
|---|---|---|
| cs10-BaseOS / AppStream / CRB | ABSENT | present |
| epel10 | **tayga-0.9.6-...el10_2** | absent |
| cs9-BaseOS / AppStream / CRB | ABSENT | present |
| epel9 | tayga-0.9.2-17.el9 | absent |

Correction to one panel claim: EPEL 10 **does** ship tayga, and `epel-release` + `dnf -y install tayga unbound` on EL10 returns RC=0. So enabling EPEL is a sufficient fix.
</details>

nat64.sh
```diff
@@ -0,0 +1,326 @@
...
+
+    # Enable IPv4 forwarding
+    sudo sysctl -w net.ipv4.ip_forward=1
+
+    # Set up iptables MASQUERADE for translated traffic
+    sudo iptables -t nat -C POSTROUTING -s "${NAT64_V4_POOL}" -j MASQUERADE 2>/dev/null || \
+        sudo iptables -t nat -A POSTROUTING -s "${NAT64_V4_POOL}" -j MASQUERADE
+
+    # Start TAYGA daemon
+    sudo tayga
```
mkowalski: **Bug: TAYGA is started as an unmanaged daemon, so NAT64 does not survive a host reboot — and fails silently when it doesn't**

The DNS64 half is fully persistent: unit file with `WantedBy=multi-user.target` (119-132), `systemctl enable --now` (135), the NetworkManager drop-in (138-141), and the libvirt dnsmasq option persisted via `net-define` (179). The translator half is entirely runtime-only: this bare forked daemon, plus `ip route add` (55), `ip -6 route add` (58), `sysctl -w` (21-22, 61) and `iptables -A` (64-65).

After a reboot unbound comes back and keeps synthesizing AAAA inside `${NAT64_PREFIX}`, and the libvirt dnsmasq keeps forwarding to it, but there is no translator and no route — every synthesized address blackholes behind a **successful** DNS answer. That is the worst failure shape: it looks like DNS works.

Worth knowing: the `tayga` RPM you install already ships `/usr/lib/systemd/system/tayga@.service` (`ExecStart=/usr/sbin/tayga --pidfile ... -d --config /etc/tayga/%i.conf`). It is never enabled here, and it reads `/etc/tayga/<instance>.conf` — not the `/etc/tayga.conf` written at line 5. Using it (or an equivalent unit owning the TUN, routes and firewall state) fixes reboot persistence and the idempotence problem on line 49 at once.

<details>
<summary>Reproducer</summary>

**Steps:** Applied exactly what `configure_tayga` applies inside a throwaway netns, then destroyed and recreated the namespace — routes, tun devices, `net.*` sysctls and iptables are all namespaced, so this resets precisely the state a reboot resets. Separately used `systemctl --root=<tmpdir>` so the host was never touched.

**Expected:** both halves of the NAT64/DNS64 pair have matching lifetimes.

**Actual:**
```
before reset: tun=1 v4route=1 v6route=1 ipv6fwd=1 masq=1
after  reset: tun=0 v4route=0 v6route=0 ipv6fwd=0 masq=0

ip -6 route get 64:ff9b::102:304  ->  RTNETLINK answers: Network is unreachable
ping6 64:ff9b::102:304            ->  connect: Network is unreachable
```
(`64:ff9b::102:304` is exactly what `dns64-synthall` synthesizes for an A record.)

`systemctl --root enable unbound-dns64.service` → creates `multi-user.target.wants/` symlink, `disabled`→`enabled`. `is-enabled tayga.service` → `No such file or directory`. 6/6 checks passed.

The end-to-end post-reboot blackhole was **not** executed (no reboot performed); it follows by composition of the runtime-verified facts above.
</details>

nat64.sh
```diff
@@ -0,0 +1,326 @@
...
+    if ! grep -q "server=127.0.0.1#${NAT64_DNS64_PORT}" "${xml}"; then
+        sudo sed -i "/<\/dnsmasq:options>/i\\
+    <dnsmasq:option value='server=127.0.0.1#${NAT64_DNS64_PORT}'/>\\
+    <dnsmasq:option value='no-resolv'/>" "${xml}"
+    fi
+
+    sudo virsh net-destroy "${net}"
+    sudo virsh net-undefine "${net}"
+    sudo virsh net-define "${xml}"
+    sudo virsh net-start "${net}"
```
mkowalski: **Bug: `net-undefine` clears the libvirt `autostart` flag, so these networks stay down after a reboot**

metal3-dev-env explicitly marks these networks autostart (`vm-setup/roles/libvirt/tasks/network_setup_tasks.yml:65-70`, `autostart: "yes"` — verified at the commit pinned in `01_install_requirements.sh:22`). Undefining and redefining silently drops it, and nothing here restores it.

Combined with the TAYGA persistence gap on line 68, a rebooted host comes up with no baremetal/extra networks **and** no translator.

Fix: `sudo virsh net-autostart "${net}"` after the `net-define`, or update the network in place instead of undefine/define.

<details>
<summary>Reproducer</summary>

**Steps:** Ran this exact sequence against a uniquely-named throwaway libvirt network (no existing network touched).

**Expected:** autostart survives the rewrite.

**Actual:**
```
after net-autostart              = yes
after net-destroy                = yes   <- destroy alone is fine
after net-undefine + net-define  = no    <- cleared here
autostart symlink after undefine: No such file or directory
grep -n 'net-autostart' nat64.sh -> (no matches)
```
Isolated to `net-undefine`. Upstream confirms the mechanism: `networkUndefine()` (`src/network/bridge_driver.c:3329`) → `virNetworkObjDeleteConfig()` (`src/conf/virnetworkobj.c:1131-1133`) does `unlink(autostartLink); obj->autostart = false;`.
</details>

nat64.sh
```diff
@@ -0,0 +1,326 @@
...
+    sudo ip link add name "${net}-dmy" up master "${net}" type dummy 2>/dev/null || true
+    echo 0 | sudo dd of="/proc/sys/net/ipv6/conf/${net}/addr_gen_mode" 2>/dev/null || true
+}
+
+# When running an IPv6-only cluster via NAT64, the in-cluster Ironic runs as
+# IPv6-only pods (hostNetwork on v6-only nodes) and cannot reach the sushy/redfish
+# BMC emulator at its IPv4 baremetal address. Rewrite the generated node BMC
+# addresses to the host's IPv6 baremetal address (which sushy also listens on) so
+# both the bootstrap Ironic and the pivoted in-cluster Ironic can control the nodes.
+function nat64_fixup_bmc_addresses() {
```
mkowalski, line 282: **Bug: `ipmi://` BMCs are rewritten to IPv6, but VirtualBMC only listens on IPv4**

This rewrite is scheme-agnostic (split/join on the `//<v4host>:` token) and is gated only on `ENABLE_NAT64` (`02_configure_host.sh:300-302`) — never on `BMC_DRIVER`. sushy/redfish is fine because the sushy-tools entrypoint force-adds `--interface ::` when `SUSHY_EMULATOR_LISTEN_IP` is absent from `conf.py`, which metal3-dev-env never sets. **vbmc has no equivalent escape hatch**: metal3-dev-env overrides vbmc's own `::` default with an IPv4 literal (`vm-setup/roles/virtbmc/tasks/setup_tasks.yml:41-53`, written to the per-domain config at `:90`), and pyghmi derives the socket family from that address (`serversession.py:275-277` → `session.py:356-362`).

Scope, to be fair: the non-CI default is `BMC_DRIVER=redfish` (`common.sh:200`), which is unaffected. This bites `BMC_DRIVER=ipmi` explicitly, and CI where the default is `mixed` (`common.sh:202`) and `ironic_nodes.json.j2:22-33` assigns ipmi to roughly one node in three.

Fix: reject `ipmi`/`mixed` under `ENABLE_NAT64`, or rewrite the vbmc config `address` to `::` and restart the container — mirroring what `nat64_fixup_sushy_cert` already does for sushy.

<details>
<summary>Reproducer</summary>

**Steps:** Ran the real pyghmi/virtualbmc from `quay.io/metal3-io/vbmc` with a real `ipmitool` client, on loopback, unprivileged port, inside the container netns.

**Expected:** the rewritten `ipmi://[v6]:port` endpoint is reachable by the IPv6-only in-cluster Ironic.

**Actual:**
```
### BMC bound to '127.0.0.1'   (what metal3-dev-env writes)
  /proc/net/udp : 0100007F:3F7A     /proc/net/udp6 : (empty)
  ipmitool -H 127.0.0.1 : Chassis Power is on            [rc=0]
  ipmitool -H ::1       : Unable to establish IPMI v2 / RMCP+ session [rc=1]

### BMC bound to '::'          (control: vbmc's own CLI default)
  ipmitool -H 127.0.0.1 : Chassis Power is on            [rc=0]
  ipmitool -H ::1       : Chassis Power is on            [rc=0]
```
And the rewrite itself, run against a mixed-driver nodes file:
```
ipmi -> ipmi://[fd2e:6f44:5dd8:c956::1]:6230
grep -rn ENABLE_NAT64 *.sh | grep -i BMC_DRIVER  ->  (no matches)
```
</details>

nat64.sh
```diff
@@ -0,0 +1,326 @@
...
+# Remove the legacy CoreDNS DNS64 service (superseded by unbound). No-op if absent.
+function _nat64_remove_legacy_coredns() {
+    if sudo systemctl is-active --quiet coredns-nat64.service 2>/dev/null; then
+        sudo systemctl stop coredns-nat64.service
+    fi
+    sudo systemctl disable coredns-nat64.service 2>/dev/null || true
+    sudo rm -f /etc/systemd/system/coredns-nat64.service
+    sudo rm -rf /etc/coredns
+    sudo systemctl daemon-reload
+}
```
mkowalski, line 400: **Bug: `rm -rf /etc/coredns` runs on every `configure_dns64`, not just on cleanup**

This helper is invoked from `configure_dns64` (line 78) on **every** `02_configure_host.sh` run, not only from `cleanup_nat64` (line 283). No merged dev-scripts code has ever created `coredns-nat64.service` or `/etc/coredns` — they existed only in an earlier revision of this PR — so on any real host the only thing this can delete is an unrelated, host-owned CoreDNS configuration directory. Enabling NAT64 should not be able to wipe someone else's DNS deployment.

Fix: drop the helper entirely, or scope it to a dev-scripts-owned path and invoke it only from `cleanup_nat64`.

Raised independently by the bugs and security specialists.

host_cleanup.sh
```diff
@@ -29,10 +31,15 @@ ansible-playbook \
...
-    -i "${VM_SETUP_PATH}/inventory.ini" \
-    -b -vvv "${VM_SETUP_PATH}/teardown-playbook.yml"
+    -i ${VM_SETUP_PATH}/inventory.ini \
+    -b -vvv ${VM_SETUP_PATH}/teardown-playbook.yml
 
-sudo rm -rf "/etc/NetworkManager/dnsmasq.d/openshift-${CLUSTER_NAME}.conf" /etc/yum.repos.d/delorean*
+# Clean up NAT64/DNS64 if enabled
+if [[ "${ENABLE_NAT64:-false}" == "true" ]]; then
+    cleanup_nat64
+fi
```
mkowalski: **Bug: cleanup is gated on current config rather than on applied state**

Raised independently by the architecture, consistency and QA specialists.

If a deployment was configured with NAT64 and the user later unsets `ENABLE_NAT64` before `make clean`, the unbound service, the dnsmasq drop-in, the TUN, the routes and the firewall rules all survive. Worse: changing `IP_STACK`/`HOST_IP_STACK` first makes the new validation in `common.sh:436-446` `exit 1` **before** cleanup even starts, so a user correcting their config cannot clean up.

Compare the LLDP/BGP cleanup right beside this, which is unconditional and absence-tolerant. `cleanup_nat64` is already written to be a no-op when nothing is present, so calling it unconditionally is the smaller change.

Fix: call `cleanup_nat64` unconditionally (or detect applied state), and keep the deploy-time stack validation out of the cleanup path.

common.sh
```diff
@@ -433,6 +433,18 @@ if [[ "${BOOTSTRAP_IN_PLACE}" == "true" ]]; then
...
+if [[ "${ENABLE_NAT64:-false}" == "true" ]]; then
+  if [[ "${IP_STACK:-v6}" != "v6" ]]; then
+    error "ENABLE_NAT64=true requires IP_STACK=v6 (got IP_STACK=${IP_STACK:-v6})"
+    exit 1
+  fi
+  if [[ "${HOST_IP_STACK:-${IP_STACK:-v6}}" != "v4" ]]; then
+    error "ENABLE_NAT64=true requires HOST_IP_STACK=v4 (got HOST_IP_STACK=${HOST_IP_STACK:-${IP_STACK:-v6}})"
+    exit 1
+  fi
+fi
```
mkowalski: **Bug: agent scenarios bypass this validation entirely**

This runs at lines 436-446, but `IP_STACK` is not derived from `AGENT_E2E_TEST_SCENARIO` until `common.sh:488` (`IP_STACK=$(echo "${arr[1]##*IP}" | tr V v)`). With `IP_STACK` unset at this point, `ENABLE_NAT64=true` together with an IPv4 or dual-stack agent scenario (e.g. `HA_IPV4`) passes on the `:-v6` default — and then NAT64 is configured and BMC addresses are rewritten for a cluster that is not IPv6-only.

Fix: move this block below the scenario parsing at line 481-489, so it validates the values that are actually in effect.

host_cleanup.sh
```diff
@@ -29,10 +31,15 @@ ansible-playbook \
     -e "virthost=$HOSTNAME" \
     -e "manage_baremetal=$MANAGE_BR_BRIDGE" \
     -e "nodes_file=$NODES_FILE" \
-    -i "${VM_SETUP_PATH}/inventory.ini" \
-    -b -vvv "${VM_SETUP_PATH}/teardown-playbook.yml"
+    -i ${VM_SETUP_PATH}/inventory.ini \
+    -b -vvv ${VM_SETUP_PATH}/teardown-playbook.yml
```
mkowalski: **Unrelated revert of merged commit `e741cb5`**

These two lines drop the quoting added by "METAL-1845: shellcheck fixes part 4" (#1906). I confirmed by diffing against `e741cb5` that this is a verbatim revert, almost certainly a bad rebase rather than an intentional change.

It reintroduces word splitting and globbing on `${VM_SETUP_PATH}`. The same revert appears at lines 54-55, 67-68, 72, 77-78, 82 and in `ocp_cleanup.sh:21`.

Fix: restore the quoting and keep this PR's diff to NAT64.

host_cleanup.sh
```diff
@@ -57,22 +64,22 @@ fi
...
 if [ "$MANAGE_PRO_BRIDGE" == "y" ]; then
-    sudo nmcli con del "${PROVISIONING_NETWORK_NAME}" || true
-    sudo ip link delete "${PROVISIONING_NETWORK_NAME}" || true
+    sudo nmcli con del ${PROVISIONING_NETWORK_NAME} || true
+    sudo ip link delete ${PROVISIONING_NETWORK_NAME} || true
     if [[ -d /sys/class/net/pro-ipv6-dummy ]]; then
        sudo ip link delete pro-ipv6-dummy || true
     fi
-    sudo rm -f "/etc/NetworkManager/system-connections/${PROVISIONING_NETWORK_NAME}.nmconnection"
+    sudo rm -f /etc/NetworkManager/system-connections/${PROVISIONING_NETWORK_NAME}.nmconnection
```
mkowalski: **Unrelated revert of merged commit `e741cb5` (continued)**

Same verbatim revert of the METAL-1845 quoting fixes as at lines 34-35. Here it reintroduces word splitting and globbing on `${PROVISIONING_NETWORK_NAME}` and `${BAREMETAL_NETWORK_NAME}` inside privileged `nmcli con del`, `ip link delete` and `rm -f` commands — and again at lines 77-78 and 82 for the baremetal bridge.

ocp_cleanup.sh
```diff
@@ -18,7 +18,8 @@ if [ -d "${OCP_DIR}" ]; then
     sudo rm -rf "${OCP_DIR}"
 fi
 
-sudo rm -rf "/etc/NetworkManager/dnsmasq.d/openshift-${CLUSTER_NAME}.conf"
+sudo rm -rf /etc/NetworkManager/dnsmasq.d/openshift-${CLUSTER_NAME}.conf
```
mkowalski: **Unrelated revert of merged commit `e741cb5`**

Line 21 loses the quoting added by METAL-1845 (#1906), reintroducing globbing/word-splitting on `${CLUSTER_NAME}` in a `rm -rf`. Same bad-rebase pattern as `host_cleanup.sh`.

02_configure_host.sh
```diff
@@ -474,6 +482,12 @@ if [ "$EXT_IF" ]; then
   sudo $IPTABLES -A FORWARD --in-interface "${BAREMETAL_NETWORK_NAME}" -j ACCEPT
 fi
 
+# When NAT64 is enabled, ensure IPv6 forwarding rules are set for the bridge
+if [[ "${ENABLE_NAT64}" == "true" ]]; then
+  sudo ip6tables -A FORWARD --in-interface "${BAREMETAL_NETWORK_NAME}" -j ACCEPT 2>/dev/null || true
+  sudo ip6tables -A FORWARD --out-interface "${BAREMETAL_NETWORK_NAME}" -m state --state RELATED,ESTABLISHED -j ACCEPT 2>/dev/null || true
+fi
```
mkowalski, line 498: **Suggestion: these rules are broader than needed, duplicate on every run, and are never removed**

Three separate issues, in descending order of importance:

1. **Never cleaned up.** `cleanup_nat64` (`nat64.sh:279-326`) contains zero `ip6tables` lines — its only netfilter teardown is the IPv4 MASQUERADE at `nat64.sh:319`. These rules survive `make clean`.
2. **Not idempotent.** Unconditional `-A`, unlike the adjacent NAT64 rule at `nat64.sh:64-65` which correctly does `-C ... || -A ...`. I measured 3 copies of each after 3 runs; combined with (1) they accumulate without bound.
3. **Over-broad.** They accept *all* forwarded IPv6 on the bridge rather than cluster-subnet ↔ NAT64-prefix via the TUN.

I want to be precise about severity, because this was initially filed as a security bypass and that framing did **not** hold up. I tested it four ways: the ip6 `FORWARD` policy is already `accept` on a stock host; `-A` cannot override an earlier DROP; firewalld filters in a *separate* nft base chain at `priority filter + 10` so an `accept` here does not bypass it; and on a real dev-scripts host these rules sit behind `LIBVIRT_FWO` — which already ends with `-s <bm v6 subnet> -i ostestbm -j ACCEPT` — and matched **0 packets**. So this is code hygiene, not a reachability change. Also note lines 481-482 already do the same thing for IPv4, so the breadth follows existing repo convention.

Fix: `ip6tables -C ... || ip6tables -A ...` scoped to `-i ${BAREMETAL_NETWORK_NAME} -o ${NAT64_TUN_INTERFACE} -s ${EXTERNAL_SUBNET_V6} -d ${NAT64_PREFIX}` plus the reverse established rule, with matching `-D` calls in `cleanup_nat64`. Dropping `2>/dev/null || true` would also stop hiding real failures.

nat64.sh
```diff
@@ -0,0 +1,326 @@
...
+
+    # Remove dnsmasq DNS64 forwarding config
+    sudo rm -f "${NAT64_DNSMASQ_CONF}"
+    if systemctl is-active --quiet NetworkManager; then
+        sudo systemctl reload NetworkManager
+    fi
+
+    # Stop TAYGA
+    sudo tayga --rmtun 2>/dev/null || true
+    sudo pkill -f "^tayga" 2>/dev/null || true
```
mkowalski: **Suggestion: `--rmtun` runs before the daemon is killed, so the persistent TUN leaks**

`tayga --rmtun` runs while the daemon still holds the tun, so the detach fails with `Device or resource busy` and the error is swallowed by `|| true`. Only then does line 312 kill the daemon — by which point nothing removes the device, leaving a stale persistent `nat64` interface after `make clean`.

Separately, `pkill -f "^tayga"` only matches when argv[0] is exactly `tayga`; it misses a daemon started as `/usr/sbin/tayga`.

Fix: swap the order — `sudo pkill -x tayga || true`, then `sudo tayga --rmtun`, then `sudo ip link del "${NAT64_TUN_INTERFACE}" 2>/dev/null || true` as a backstop.

nat64.sh
```diff
@@ -0,0 +1,326 @@
+#!/bin/bash
+# NAT64/DNS64 helper functions for enabling IPv6-only clusters
+# on IPv4-only hosts using TAYGA (NAT64) and unbound (DNS64).
+
+export NAT64_TAYGA_CONF="/etc/tayga.conf"
+export NAT64_TAYGA_DATA_DIR="/var/db/tayga"
```
mkowalski, line 6: **Suggestion: `/etc/tayga.conf` and the fixed TUN/data names are not instance-scoped**

This helper owns the global `/etc/tayga.conf` (line 5) and `/var/db/tayga` (line 6), uses a fixed interface name, and `cleanup_nat64` then removes that state and `pkill`s broadly. A host that already runs its own TAYGA instance can have its configuration overwritten and its daemon stopped by a dev-scripts run.

Fix: namespace the config, data dir, TUN and (ideally) systemd instance per cluster, and stop only the instance dev-scripts owns.

nat64.sh
```diff
@@ -0,0 +1,326 @@
...
+    # reach public resolvers (e.g. 8.8.8.8) through their firewall, so forward
+    # DNS64 queries to whatever resolvers the host itself uses. When NetworkManager
+    # runs in dnsmasq mode /etc/resolv.conf points at a loopback stub and the real
+    # upstream servers live in no-stub-resolv.conf.
+    local upstreams
+    upstreams=$(awk '/^nameserver/ && $2 != "127.0.0.1" && $2 != "::1" {print $2}' /run/NetworkManager/no-stub-resolv.conf 2>/dev/null)
+    if [[ -z "${upstreams//[[:space:]]/}" ]]; then
+        upstreams=$(awk '/^nameserver/ && $2 != "127.0.0.1" && $2 != "::1" {print $2}' /etc/resolv.conf 2>/dev/null)
+    fi
+    upstreams=${upstreams:-8.8.8.8}
```
mkowalski: **Suggestion: fail closed instead of silently falling back to `8.8.8.8`**

When resolver discovery finds nothing, every cluster and host query routed through DNS64 is forwarded to a public resolver. Two problems: internal names leak across the configured DNS trust boundary, and — per the comment you wrote directly above at lines 80-84 — NAT64 hosts frequently cannot reach `8.8.8.8` through their firewall anyway, so this fallback is likely non-functional exactly when it triggers.

Fix: require an explicitly configured upstream, or abort with a clear message when none can be discovered.

nat64.sh
```diff
@@ -0,0 +1,326 @@
...
+    port: ${NAT64_DNS64_PORT}
+    do-ip4: yes
+    do-ip6: yes
+    do-udp: yes
+    do-tcp: yes
+    access-control: 127.0.0.0/8 allow
+    access-control: ::1 allow
+    module-config: "dns64 iterator"
+    dns64-prefix: ${NAT64_PREFIX}
+    dns64-synthall: yes
```
mkowalski, line 161: **Note: DNS64 answers are not validated**

`module-config: "dns64 iterator"` omits unbound's `validator`, and the generated unit does not initialize a trust anchor, so answers rely entirely on the upstream resolver — which, per line 90, may be a silent public fallback.

Reasonable to accept deliberately for a lab tool, but worth either using the supported `dns64 validator iterator` ordering with an initialized trust anchor, or stating the trust assumption in a comment so it is an explicit choice.

nat64.sh
```diff
@@ -0,0 +1,326 @@
...
+
+    sudo virsh net-destroy "${net}"
+    sudo virsh net-undefine "${net}"
+    sudo virsh net-define "${xml}"
+    sudo virsh net-start "${net}"
+
+    # net-destroy drops the bridge; restore a dummy for carrier and addr_gen_mode=0
+    # so the network's IPv6 address comes up before the VMs provide carrier (needed
+    # for IPv6 on EL9).
+    sudo ip link add name "${net}-dmy" up master "${net}" type dummy 2>/dev/null || true
```
mkowalski: **Suggestion: `${net}-dmy` can exceed the 15-character interface name limit**

With the default `CLUSTER_NAME=ostest` the name is `ostestbm-dmy` (12 chars) and this is fine. A `CLUSTER_NAME` of 12+ characters pushes `${net}-dmy` past Linux's 15-char `IFNAMSIZ` limit; `|| true` then hides the failure and the recreated IPv6 bridge is left without carrier before the VMs boot — which is the exact condition this line exists to prevent.

Fix: derive a fixed or truncated dummy name bounded to 15 characters, and don't swallow the failure.

nat64.sh
```diff
@@ -0,0 +1,326 @@
...
+    fi
+    echo "Rewriting node BMC addresses from ${v4host} to [${v6host}] for NAT64..."
+    local f tmp
+    for f in "${NODES_FILE}" "${NODES_FILE}.orig" "${EXTRA_NODES_FILE:-}" "${ARM_NODES_FILE:-}"; do
+        [[ -n "${f}" && -f "${f}" ]] || continue
+        tmp="${f}.nat64"
+        # Literal (non-regex) host replacement via split/join on the "//host:" token.
+        jq --arg old "//${v4host}:" --arg new "//[${v6host}]:" \
+            '(.nodes[]?.driver_info.address) |= (. / $old | join($new))' \
+            "${f}" > "${tmp}" && mv "${tmp}" "${f}"
```
mkowalski: **Suggestion: preserve file mode across the `mv`, and don't abort on a node without a BMC address**

Two small robustness issues in this loop:

1. **Mode is not preserved.** The temp file is created with the current umask and `mv` (rename) carries its mode over the destination. I measured 0600 → 0644 under umask 0022, and it fires even when the rewrite matches nothing. To be clear about impact: this is **not** exploitable in the default flow — metal3-dev-env's `template` task sets no `mode:`, so `NODES_FILE` is already 0644 at birth and `WORKING_DIR` is `chmod 755` (`common.sh:355`). It matters only for a user-supplied `NODES_FILE` (`config_example.sh:546`) deliberately protected at 0600 with real BMC credentials. Cheap defensive hygiene: `install -m "$(stat -c %a "$f")"` or `chmod --reference="$f"` before the `mv`.

2. **`jq` aborts on a missing address.** `(.nodes[]?.driver_info.address) |= (. / $old | join($new))` applies to a `null` when the path is absent: `jq: error ... null (null) and string cannot be divided`, exit 5. The `&& mv` means the loop continues but leaves `${f}.nat64` behind, and the non-zero return then aborts `02_configure_host.sh` at line 301. Stock generated files always have the key, so this only bites user-supplied inventories. Fix: `select(.driver_info.address? != null)` and `rm -f` the temp on failure.

ocp_cleanup.sh
```diff
@@ -18,7 +18,8 @@ if [ -d "${OCP_DIR}" ]; then
...
+sudo rm -f /etc/NetworkManager/dnsmasq.d/nat64-dns64.conf
```
mkowalski: **Suggestion: `make redeploy` never restores this, and the constant is now duplicated**

`make redeploy` is `ocp_cleanup ironic_cleanup build_installer ironic install_config ocp_run` — `02_configure_host.sh`, the only writer of `${NAT64_DNSMASQ_CONF}`, is not re-run. So the first redeploy of a NAT64 environment permanently removes host DNS64 forwarding.

Three smaller points: this is ungated (it runs even when `ENABLE_NAT64` is false), it hardcodes the literal instead of using `${NAT64_DNSMASQ_CONF}` (`ocp_cleanup.sh` does not source `nat64.sh`, so the constant now lives in two files and can drift), and unlike `cleanup_nat64` it does not reload NetworkManager, so the running dnsmasq keeps the deleted config until some unrelated later reload.

Fix: drop this line and leave host-level NAT64 state to `cleanup_nat64`, or source `nat64.sh`, gate it, use the variable, and reload NetworkManager.

config_example.sh
```diff
@@ -441,6 +441,37 @@ set -x
 #export EXTERNAL_SUBNET_V4="192.168.111.0/24"
 #export EXTERNAL_SUBNET_V6="fd2e:6f44:5dd8:c956::/120"
 
+# ENABLE_NAT64 -
+# Enable NAT64/DNS64 to allow IPv6-only clusters (IP_STACK=v6) to run on
+# IPv4-only hosts (HOST_IP_STACK=v4). Uses TAYGA for NAT64 translation and
+# CoreDNS for DNS64 synthesis, enabling cluster VMs to reach external IPv4
```
mkowalski: **Suggestion: this names the wrong DNS64 implementation**

The implementation installs and runs **unbound** (`01_install_requirements.sh:216`, `nat64.sh:99-135`) and treats CoreDNS as legacy to be removed (`nat64.sh:269-277`). Anyone following this comment will inspect the wrong service when troubleshooting.

While here, three other doc gaps worth closing in the same block:
- `NAT64_V6_ADDR` (`network.sh:128-129`) is a supported override but is not documented.
- Line 432 says `EXTERNAL_SUBNET_V6` is empty for `HOST_IP_STACK=v4`; under NAT64 it defaults to `fd2e:6f44:5dd8:c956::/120`.
- The later claim that IPv6 defaults `MIRROR_IMAGES` to true is also not true under NAT64.

Most importantly, enabling this disables any existing `unbound.service`, redirects host and libvirt DNS, enables forwarding, adds routes and firewall rules, rewrites BMC data and replaces the sushy certificate — and cleanup restores none of that and only runs while `ENABLE_NAT64` is still true. That is worth stating here explicitly.

network.sh
```diff
@@ -114,6 +125,10 @@ else
   exit 1
 fi
 
+if [[ "${ENABLE_NAT64}" == "true" ]]; then
+  export NAT64_V6_ADDR=${NAT64_V6_ADDR:-$(nth_ip "$EXTERNAL_SUBNET_V6" 3)}
```
mkowalski, line 129: **Note: `NAT64_V6_ADDR` defaults to an address that is on-link on the baremetal bridge**

`nth_ip "$EXTERNAL_SUBNET_V6" 3` is `fd2e:6f44:5dd8:c956::3` with the default subnet — inside the /120 libvirt puts on the baremetal bridge. The host therefore has an on-link route for it via `${BAREMETAL_NETWORK_NAME}` rather than via the `nat64` tun. TAYGA uses this address as the source of the ICMPv6 errors it generates (unreachable, packet-too-big/PMTUD), so those replies and neighbour resolution get directed at the bridge where nothing answers.

The index itself is free (::1 virthost, ::2 DNS VIP, ::4 ingress, ::5 api, ::7/::8 provisioning, ::9 bootstrap, ::20-::60 DHCP), so this is about routing, not collision. Worth a deliberate default outside any locally-attached subnet, plus documenting the knob in `config_example.sh`.

mkowalski (changes requested):
## Deep Review — inline findings

Line-anchored companion to the [full panel verdict](https://github.com/openshift-metal3/dev-scripts/pull/1952#issuecomment-5345318183). **Disposition: REQUEST_CHANGES.**

Reviewed `28a095e` against merge base `e2790c1`. Seven specialists reviewed independently; every BLOCKING claim asserting a runtime bug went to a separate reproducer agent. 8 reproducers ran, 8 confirmed — 2 with corrections that *lowered* severity, noted inline where relevant.

**23 inline comments below**, grouped by what they block:

- **Breaks a default `ENABLE_NAT64=true` deploy** — `network.sh:158` (installer pointed at a registry that is never created)
- **Breaks re-running `02_configure_host.sh`** — `nat64.sh:49`
- **Breaks an existing cluster's BMC trust** — `nat64.sh:239` (cert rotates every run, invalidating `bmcVerifyCA`)
- **Breaks on a supported distro** — `01_install_requirements.sh:216` (EL10 has no EPEL, and `tayga` ships only there)
- **Breaks after reboot, silently** — `nat64.sh:68` + `nat64.sh:177` (DNS64 persists, translator and libvirt autostart do not, so DNS answers succeed while traffic blackholes)
- **Destructive to unrelated host state** — `nat64.sh:275` (`rm -rf /etc/coredns` on every configure run)
- **Conditional on `BMC_DRIVER`** — `nat64.sh:194` (ipmi endpoints rewritten to IPv6; vbmc listens on IPv4 only)
- **Cleanup and scope** — `host_cleanup.sh:37` (teardown gated on current config), `common.sh:436` (validation runs before agent scenarios derive `IP_STACK`), plus an unrelated revert of merged commit `e741cb5` in `host_cleanup.sh` and `ocp_cleanup.sh`

The rest are suggestions and notes: firewall rule scope/idempotence, cleanup ordering, DNS fallback, docs.

### One finding I could not anchor inline

`network.sh:190` is outside the diff, so it has no inline comment. `PROVISIONING_HOST_EXTERNAL_IP` keys off `HOST_IP_STACK`, which is `v4` under NAT64, so it resolves to the IPv4 baremetal address. Consumers include `NTP_SERVERS` (`utils.sh:126`), `bootstrapExternalStaticGateway`/`bootstrapExternalStaticDNS` (`ocp_install_env.sh:111,115`) and the agent DHCP option (`agent/05_agent_configure.sh:685`). Is handing IPv6-only nodes an IPv4 address there intentional — reached via NAT64 — or should NAT64 introduce a distinct cluster-facing address? **Not reproduced; flagging for your judgement rather than asserting a bug.**

### Two claims that did not survive verification

Stating these plainly so they don't cost you time:

- The `ip6tables` rules at `02_configure_host.sh:487-488` were initially filed as a NAT64 **isolation bypass**. That framing is wrong — on a real host the `FORWARD` policy is already `accept`, `-A` cannot override an earlier DROP, firewalld filters in a separate nft base chain at `priority filter + 10`, and these rules sit behind `LIBVIRT_FWO` and matched **0 packets**. Reduced to hygiene (scope, idempotence, missing cleanup).
- The BMC temp-file `mv` was filed as **widening credential permissions**. The mechanism is real (0600 → 0644) but the premise is not: `NODES_FILE` is already 0644 at birth from metal3-dev-env's `template` task. Reduced to defensive hygiene.

Reproducers ran read-only in throwaway namespaces and containers; the checkout was verified unmodified afterwards.

<sub>Generated by [/code-review:deep-review](https://github.com/openshift-eng/ai-helpers/tree/main/plugins/code-review/skills/deep-review). This message was generated using AI. Please verify before acting on it.</sub>

mkowalski:
Re-reviewed commit 29cd165 against the [deep review findings](https://github.com/openshift-metal3/dev-scripts/pull/1952#pullrequestreview-4974876340). All 13 blocking bugs are addressed:

- MIRROR_IMAGES now defaults to false in common.sh before the registry-override decision
- TAYGA runs as a systemd service (idempotent, reboot-persistent)
- Sushy cert idempotence check uses `openssl x509 -checkip` (handles compressed/uncompressed IPv6 SANs)
- EPEL enabled on EL10 so tayga installs
- `virsh net-autostart` restored after network redefine
- Non-redfish BMC_DRIVER rejected at validation (vbmc is IPv4-only)
- `/etc/coredns` no longer removed on every run
- `cleanup_nat64` runs unconditionally in host_cleanup.sh
- NAT64 validation moved after agent scenario parsing
- METAL-1845 quoting reverts in host_cleanup.sh / ocp_cleanup.sh restored

The suggestions are also addressed (ip6tables check-then-add + cleanup, rmtun ordering, `NAT64_DNS64_UPSTREAM` override with loud fallback warning, IFNAMSIZ-safe dummy names, BMC rewrite mode/null-address handling, redeploy no longer strips the DNS64 dnsmasq drop-in, config_example docs corrected). The only item left open is instance-scoping the global `/etc/tayga.conf`/TUN name — minor for a lab tool, fine to leave as-is.

Review addressed — LGTM.

---
This message was generated using AI. Please verify before acting on it.
Assisted-By: Claude Fable 5

mkowalski:
/lgtm

Ideal would be to have e2e CI lane before this merges so we can do pre-submit. But I am fine with going without and fixing whatever arises

emy:
@mkowalski Just did a first version of e2e CI lane over at: https://github.com/openshift/release/pull/83984 This should provide a signal for this PR right?

mkowalski:
The last failure in nat64 job is

```
    Error running oc --namespace=e2e-test-no-egress-firewall-e2e-qpxsl --kubeconfig=/tmp/configfile3574480851 exec dummy -- ping -c 1 8.8.8.8:
    StdOut>
    ping: sendto: Network unreachable
    PING 8.8.8.8 (8.8.8.8): 56 data bytes
    command terminated with exit code 1
    StdErr>
    ping: sendto: Network unreachable
    PING 8.8.8.8 (8.8.8.8): 56 data bytes
    command terminated with exit code 1
    exit status 1
```

what means we have a pod which cannot connect to 8.8.8.8

mkowalski:
/lgtm

We are fine, the cluster stack is IPv6 and we don't promise IPv4 connectivity like this pod expects. It's rather test issue rather than product issue. We may need to figure out why this test is even executing

PR #1957, opened 2026-09-02
Do not source a file from /tmp

dtantsur:
Sourcing files from a world-writeable location as root isn't a great
security practice. We don't expect anyone to run dev-scripts on a
multi-user system, but let's still fix it.


MahnoorAsghar:
/lgtm
Not sure if I have the rights in this org

dtantsur:
/cc @andfasano @bfournie 

PTAL for approval, this change may have broader impact

PR #1958, opened 2026-09-02
Run vbmc and sushy-tools with less privileges

dtantsur:
Neither really need --privileged and only vbmc needs host networking.
Removing --privileged does mean disabling selinux labeling.


dtantsur:
/retest

Ansible galaxy took a break

PR #1960, opened 2026-09-03
Adopt all relevant parts of metal3-dev-env and drop dependency

dtantsur:
The initial idea was that we'd stay pretty close to metal3-dev-env. This
is not the case. OpenShift works very differently from upstream Metal3,
and our dev-scripts have grown a number of deviations from dev-env.
The pinned hash requires regular updates. We use an older Ansible
version. Go installation logic gets in the way all the time because of
unstable mirrors.

On top of this, upstream Metal3 community wants to gradually phase out
metal3-dev-env in favour of leaner development scripts in CAPM3 and BMO.

With this change, I've tasked Claude to transfer all relevant parts of
dev-scripts and drop any hacks around upstream metal3-dev-env. Most of
the code concerns creating testing VMs.

Further commits clean up irrelevant parts of the transferred code, such as
support for Ubuntu/Suse or our hacks.

Metal3-dev-env commit was c963d687d23c6a2e1f024359415680606de157ff


dtantsur:
/hold

This is a proof of concept

PR #1961, opened 2026-09-04
Hack: apply metal3-dev-env/pull/1725 until that merges

tdomnesc:
To be reverted


iurygregory (commented):
/lgtm

tdomnesc:
This is technically incorrect since we call the playbook with `-e "go_custom_mirror=$GO_CUSTOM_MIRROR"` not go_mirror, but we don't change it from https://go.dev/dl so it works anyway. Will let it merge like this since we plan to revert it.
The retries are the important part for now.

PR #1964, opened 2026-09-11
Do not break DNS if nameservers are configured via resolv.conf

dtantsur:
DevScripts forces the host to use NetworkManager's dnsmasq instance for
DNS. After this point, only DNS discovered via DHCP or provided
explicitly via ADDN_DNS remains. This is a problem for machines that had
their DNS settings in resolv.conf: NetworkManager changes it to refer
to 127.0.0.1, while its dnsmasq is not aware of any upstream servers.

Then in host_cleanup dev-scripts removes all NetworkManager DNS
configuration. At this point, resolv.conf refers to 127.0.0.1,
upstream.conf does not exists, dnsmasq is still only aware of DNS
servers received via DHCP. This results in no DNS on the machine.

Even worse, dev-scripts uses DNS before ADDN_DNS is applied, so only
manual recovery is possible.

This patch does two modifications:

1) ADDN_DNS now defaults to the first non-loopback address from
   /etc/resolv.conf (same logic as already used in the CI).
   This solves DNS configuration on the first dev-scripts run.

2) On cleanup, upstream.conf and dnsmasq.conf are preserved, i.e.
   NetworkManager remains the DNS provider with the same configuration.
   This ensures that subsequent dev-scripts runs don't break DNS.

ADDN_DNS is now documented in config_example.sh.


dtantsur:
/hold cancel

WFM

common.sh
```diff
@@ -393,6 +393,25 @@ if [ ! -f "$IRONIC_IMAGES_DIR/.permissions" ]; then
   touch "$IRONIC_IMAGES_DIR/.permissions"
 fi
 
+# NOTE(dtantsur): NetworkManager is very keen on nuking any existing DNS
+# configuration, even if it does not have any configuration itself. Detect and
+# cache ADDN_DNS for later use.
+if [[ -z "${ADDN_DNS}" ]]; then
+    # Current information from resolv.conf takes priority.
+    for addr in $(awk '/nameserver/ { print $2; }' /etc/resolv.conf); do
+        if ! [[ "$(ipcalc --addrspace ${addr})" =~ Loopback ]]; then
```
tdomnesc, line 410: ipcalc is not guaranteed to be installed
Either add it to install requirements or something like:
case "$addr" in 127.*|::1) continue ;; esac

PR #1967, opened 2026-09-17
For ABI, handle ssh key field in UI when using prebuilt ISO

bfournie:
If the pre-built ISO contains an ssh key, for example when creating the ISO with the SaaS UI and the optional ssh key is added, then the interactive UI will hang since it waits unconditionally for the ssh key text area. This fixes that case. It does not affect the more common case in which the ISO does not have an embedded ssh key.


agent/isobuilder/ui_driven_cluster_installation/main.go
```diff
@@ -622,6 +628,47 @@ func networkingDetails(page *rod.Page, path string) error {
...
+// key is injected at fetch time - it shows a checked "Use the same host
+// discovery SSH key" checkbox instead and never renders the text area. Wait for
+// whichever variant the UI puts on the page, and only type when there is
+// somewhere to type. Match on the field ids: the checkbox label differs between
+// single-cluster (OVE) mode and the regular one, see
+// https://github.com/openshift-assisted/assisted-installer-ui/pull/4038.
+func enterSSHPublicKey(page *rod.Page) error {
+	const (
+		sshFieldSelector = "#form-input-sshPublicKey-field"
+		shareKeySelector = "#form-checkbox-shareDiscoverySshKey-field"
```
pawanpinjarkar, line 646: As the dev-script is only for below the sea level, I am not sure how this scenario could be replicated into our testing without an integration from SaaS UI. Are we heading towards a testing scenario where we first create OVE ISO via dev-scripts then update ISO ignition to put the optional configurations a user could enter from SaaS UI ( rendezvous IP, pull secret, ssh key etc.) and finally consider the dev-scripts generated ISO like the one downloaded from SaaS UI in real environment? 

Without this additional step, I think there won't a way to test it. Am I missing something?

bfournie: This is for manual testing only when downloading the ISO from console.redhat.com and adding in the ssh key. It was never intended for automated testing, see https://github.com/openshift-metal3/dev-scripts/pull/1951.

